PR #3, opened 2023-01-02
fix: error when installing

ofrades:
Just a small change to make it work when cloning for the first time.

Feel free to disregard the fix if the idea is to add the first param into the options in the future.


ofrades:
Was just trying it out :smile:

PR #6, opened 2023-01-03
Use 'mini.starter' as start screen

echasnovski:
Hi!

This is a PR to give you options for a default start screen. Nothing against 'goolord/alpha.nvim', will totally understand rejecting this. 

Functionally it is almost the same:
- Press highlighted letter and it will execute corresponding command.
- User can also navigate to item with `<C-n>/<C-p>` or `<Down>/<Up>`.

Visually there are more differences:
- Items are automatically organized in sections.
- No easy way of adding different icons for different items. Instead there is a single bullet for all items serving as grouping symbol for sections.
- Padding for bullets, section names, and footer are picked manually to achieve some kind of alignment based on the header.

Here is a preview:
lazyvim_mini-starter

### DIFF
--- a/lua/lazyvim/plugins/ui.lua
+++ b/lua/lazyvim/plugins/ui.lua
@@ -137,6 +137,71 @@ return {
     },
   },
 
+  -- start screen
+  {
+    "echasnovski/mini.starter",
+    enabled = false,
+    event = "VimEnter",
+    config = function()
+      local logo = table.concat({
+        "██╗      █████╗ ███████╗██╗   ██╗██╗   ██╗██╗███╗   ███╗          Z",
+        "██║     ██╔══██╗╚══███╔╝╚██╗ ██╔╝██║   ██║██║████╗ ████║      Z",
+        "██║     ███████║  ███╔╝  ╚████╔╝ ██║   ██║██║██╔████╔██║   z",
+        "██║     ██╔══██║ ███╔╝    ╚██╔╝  ╚██╗ ██╔╝██║██║╚██╔╝██║ z",
+        "███████╗██║  ██║███████╗   ██║    ╚████╔╝ ██║██║ ╚═╝ ██║",
+        "╚══════╝╚═╝  ╚═╝╚══════╝   ╚═╝     ╚═══╝  ╚═╝╚═╝     ╚═╝",
+      }, "\n")
+      local pad = string.rep(" ", 22)
+      local new_section = function(name, action, section)
+        return { name = name, action = action, section = pad .. section }
+      end
+
+      local starter = require("mini.starter")
+      --stylua: ignore
+      local config = {
+        evaluate_single = true,
+        header = logo,
+        items = {
+          new_section("Find file",    "Telescope find_files", "Telescope"),
+          new_section("Recent files", "Telescope oldfiles",   "Telescope"),
+          new_section("Grep text",    "Telescope live_grep",  "Telescope"),
+          new_section("init.lua",     "e $MYVIMRC",           "Config"),
+          new_section("Lazy",         "Lazy",                 "Config"),
+          new_section("New file",     "ene | startinsert",    "Built-in"),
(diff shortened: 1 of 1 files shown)


folke:
Hi! This repo is still a big work in progress :)

Will defintely look into mini.starter. Would be great if you could update the PR (rebase) and keep alpha in there too. And for now set `enabled=false` for mini.starter.

Will play around with it.

echasnovski:
> Hi! This repo is still a big work in progress :)
> 
> Will defintely look into mini.starter. Would be great if you could update the PR (rebase) and keep alpha in there too. And for now set `enabled=false` for mini.starter.
> 
> Will play around with it.

Will do (around tomorrow, once I get to my main machine).

echasnovski:
Rebased, added `enabled = false`, and slightly tweaked.

Some observations about plugin updating with 'lazy.nvim' I had during play-testing it:
- I very often got `fatal: unable to access 'https://github.com/...': Could not resolve host: github.com` errors. Every time with different set of plugins.
- It sometimes seems to think that first commit is the latest commit, as seen from 'lazy-lock.json'. In particular in 'mini.starter' and 'mini.indentscope'. I noticed that because the latter got update around a month ago that makes it work better with 'indent-blankline.nvim' (does not cover it with blank space)

folke:
Those Github errors are because of a spotty internet connection. Or a router that can't handle too many simultaneous requests. You could configure the `concurrency` setting for lazy to work around this.

As for the other issue, that's really odd. `LazyVim` does default to installing the latest version if a plugin has versions. For `mini.indentscope` it would install version `0.6.0`:
image

Is that what you mean?

That's configured by `config.defaults.version = "*"`

folke:
I could set `verion=false` for those two plugins, or you could make a new release with the fixes? :)

echasnovski:
> Those Github errors are because of a spotty internet connection. Or a router that can't handle too many simultaneous requests. You could configure the concurrency setting for lazy to work around this.

Yeah, I guessed something like this. Good you know about this.

> As for the other issue, that's really odd. LazyVim does default to installing the latest version if a plugin has versions. For mini.indentscope it would install version 0.6.0:

Oh, I see. In my mind, seeing "branch: main" in 'lazy-lock.json' means that it would update to latest `main`, which is not the same as updating to latest tag. But with this logic behavior is correct: `v0.6.0` is the latest available version which just happens to be the first commit in standalone repositories.

> I could set verion=false for those two plugins, or you could make a new release with the fixes? :)

If everything goes well, I plan to release 'mini.basics' some time soon, wait for about a week or two, and then make 0.7.0 release (for 'mini.nvim' and all standalone versions).

If you want to track only stable releases, you could track `stable` branch. This is ought to be the same as tracking latest tag. Otherwise, I really try not to break things even in `main`.

folke:
lazy has full semver support. You can use semver ranges etc. `version="*"` automatically picks the last stable version, excluding pre-release versions.

The lock-file only needs to know the commit and the branch where to find that commit.

A tag (version) is simply a reference to a commit on a certain branch, so that's why you see `main` in there.

folke:
`mini.basics` a new module? Looking forward to that :)

echasnovski:
Ok, sorry for the noise, then.

> lazy has full semver support. You can use semver ranges etc. `version="*"` automatically picks the last stable version, excluding pre-release versions.

So much functionality... Great plugin, as always👍

echasnovski:
> `mini.basics` a new module? Looking forward to that :)

I'm sorry, but you most certainly won't benefit from it :) It will be a collection of options and basic mappings presets, (like `nnoremap j gj`, 'tpope/vim-unimpaired' stuff, etc.) following a Neovim options survey I did. You know, the *basics* :)

folke:
Still looking forward to it. Thank you for the PR!

People overriding this config, can enable it by adding the following to their specs. Will add docs soon:

```lua
  { "echasnovski/mini.starter", enabled = true },
  { "goolord/alpha-nvim", enabled = false },
```

PR #15, opened 2023-01-05
better lualine with git status, diagnostics and navic lsp information

appelgriebsch:
add some more context information to lualine:
- git information
- diagnostics
- navic (current lsp context)

Looks like this:

image


folke:
Will change it a bit more, but merged already. Thank you!

PR #21, opened 2023-01-07
fix: provide LSP server name in setup function

appelgriebsch:


folke:
right, still had to test that. thanks! :)

appelgriebsch:
You're faster with merging than I'm with writing a proper comment ;)....

folke:
haha :)

PR #24, opened 2023-01-07
Use rust-tools + potential bug fix

amaanq:
This PR adds rust-tools and is used when the detected client is named rust_analyzer. The reason for this is because rust-tools adds a lot of features and niceties similar to what vscode does. I would find beginners that write in Rust will greatly appreciate this QoL improvement.

On another note a small bug with c/c++ lsps is that there can be a flurry of notifs with "warning: multiple different client offset_encodings detected for buffer, this is not supported yet". The fix is to just make capabilities.offsetEncoding utf-16, this can really be discarded if you would think it could cause potential issues with something else..

Thanks again so much for your work on neovim plugin development!
Amaan

### DIFF
--- a/doc/LazyVim.txt
+++ b/doc/LazyVim.txt
@@ -1,4 +1,4 @@
-*LazyVim.txt*         For Neovim >= 0.8.0         Last change: 2023 January 09
+*LazyVim.txt*         For Neovim >= 0.8.0         Last change: 2023 January 10
 
 ==============================================================================
 Table of Contents                                  *LazyVim-table-of-contents*
--- a/lua/lazyvim/plugins/extras/lang/clangd.lua
+++ b/lua/lazyvim/plugins/extras/lang/clangd.lua
@@ -1,0 +1,58 @@
+return {
+
+  -- add c/c++ to treesitter
+  {
+    "nvim-treesitter/nvim-treesitter",
+    opts = function(_, opts)
+      vim.list_extend(opts.ensure_installed, { "c", "cpp" })
+    end,
+  },
+
+  -- correctly setup lspconfig
+  {
+    "neovim/nvim-lspconfig",
+    dependencies = {
+      "p00f/clangd_extensions.nvim",
+      dependencies = {
+        "hrsh7th/cmp-nvim-lsp",
+      },
+    },
+    opts = {
+      -- make sure mason installs the server
+      servers = {
+        clangd = {},
+      },
+      setup = {
+        clangd = function(_, opts)
+          require("clangd_extensions").setup({
+            server = opts,
+            extensions = {
+              ast = {
+                --These require codicons (https://github.com/microsoft/vscode-codicons)
+                role_icons = {
+                  type = "",
+                  declaration = "",
+                  expression = "",
+                  specifier = "",
+                  statement = "",
+                  ["template argument"] = "",
+                },
+
+                kind_icons = {
+                  Compound = "",
+                  Recovery = "",
+                  TranslationUnit = "",
+                  PackExpansion = "",
+                  TemplateTypeParm = "",
+                  TemplateTemplateParm = "",
+                  TemplateParamObject = "",
+                },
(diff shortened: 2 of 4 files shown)


folke:
I'd rather not include rust-tools in LazyVim by default. I'm working on a way where users can easily import additional layers on top of the core set of LazyVim plugins. It would be a good fit to add rust-tools to those extra layers.

I'll keep this PR open for now, but eventually it will have to be structured differently.

You probably know, but you can easily add rust-tools to your own lspconfig, by only overriding the `setup_server` part in your personal specs. It has been added specifically for this purpose.

amaanq:
I actually did know I could override the key, but completely forgot while doing this..oops. I do think it should be a bit easier to add/import additional layers for things like rust-tools and typescript (and whatever else someone wants)...looking forward to that!

appelgriebsch:
You may want to have a look at my override here https://github.com/appelgriebsch/Nv/blob/main/lua/plugins/lsp.lua, which checks for a module in plugins.langs.<lsp-server-name> and calls it’s setup function if found…

folke:
I'm pretty sure I broke your configs, so sorry about that :)

But doing so, will now make it a lot easier to configure LazyVim.

For more info, see here:
https://github.com/LazyVim/LazyVim/discussions/22#discussioncomment-4625576

folke:
I just added an extra for typescript. It will add typescirpt to treesitter, properly configure lspconfig and make sure its installed with mason.

See [here](https://github.com/LazyVim/LazyVim/blob/main/lua/lazyvim/plugins/extras/lang/typescript.lua)

Users can enable this with:

```lua
![image](https://user-images.githubusercontent.com/292349/211203043-12bf827c-9cc3-4fd3-bfe7-3781ad7eba14.png)
```

That's how I would like any optional plugins to be structured.

amaanq:
Hi again @folke, I love the change for optional lang/ui extensions! 

Now in light of that change, I've added rust-tools and clangd_extensions inside extras/lang, and have made it super easy to add to anyone's config...how does it look?

zolrath:
From what I can tell by pulling your `rust.lua` file into my config the `<leader>co` and `<leader>cR` hotkeys you've defined don't seem to be registering? They don't do anything while in a rust file. `<C-k>` brings up the standard lsp hover for rust so the language server is functioning.

zolrath:
I've copied the `rust.lua` file into my files and the `<leader>co` and `<leader>cR` commands you've bound simply paste in the text "RustHoverActions" and "CodeActionGroups" into the buffer.

amaanq:
That config is relatively old @zolrath, can you try this [here](https://github.com/amaanq/LazyVim/blob/5a464f6db5d007bcf95f144cc18d441fc56ca908/lua/lazyvim/plugins/extras/lang/rust.lua)? And do you use copilot? I had an issue where copilot was loading after null-ls/rust_analyzer and so the keymaps were being reset back to the default lsp ones in the on_attach function, which you can fix with something like [this](https://github.com/amaanq/LazyVim/blob/5a464f6db5d007bcf95f144cc18d441fc56ca908/lua/lazyvim/plugins/lsp/init.lua#L74-76)

zolrath:
Hah perfect! Thats actually the rust config I was using before testing this and it worked great, didn't realize it was also you! 

Wanted to make sure this lang file worked properly in case it ended up getting merged into LazyVim, but if you're the author of the other file as well I'm sure you'll get this PR to a good state before it (potentially) gets merged in.

amaanq:
That crates source addition is awesome, I had never known about that..thanks! I am probably not gonna update this PR unless folke explicitly states interest in adding rust-tools. You were likely using https://github.com/appelgriebsch/Nv, I made some tweaks to my rust conf based on his haha :slightly_smiling_face:

acehinnnqru:
I wonder why the "rust-tools" plugin will load even if I am not opening a rust project.

Do you also meet this problem?

PR #36, opened 2023-01-11
fix(bufferline): load bufferline on VeryLazy event (#35)

### LINKED ISSUE #35: bug: Bufferline does not load if multiple files are open on init

- [X] I have searched the exsiting issues of plugins related to this issue

### Neovim version (nvim -v)

NVIM v0.8.2

### Operating system/version

6.0.12-arch1-1

### Describe the bug

Opening two or more files at init with LazyVim's default config results in a missing bufferline. This is because the BufAdd event [only fires for buffers added during runtime](https://neovim.io/doc/user/autocmd.html#BufAdd), not at init. Relatedly, the bufferline still does not open when navigating these buffers, as this does not fire BufAdd.

To fix this, you'd probably have to use the VeryLazy event instead of BufAdd.

- The other events (BufEnter, BufLeave) don't guarantee that two or more buffers are open, so we can't catch this on the rebound if the user were to switch among their open buffers in this state.
- There's no event to tell if two or more files are open at init
- You can check the number of active buffers and load the plugin yourself, but this is very not lazy.

### Steps To Reproduce

1. `nvim file1 file2`
2. Observe bufferline does not open

For comparison:
1. `nvim file`
2. `:e file2`
3. Observe opened bufferline

### Expected Behavior

Bufferline should load if buffers are opened at init

### Repro

```Lua
-- set stdpaths to use .repro
end

-- bootstrap lazy
end

-- install plugins
local plugins = {
  "folke/tokyonight.nvim",
  "folke/LazyVim",
  -- add any other plugins here
}
  root = root .. "/plugins",
})

-- add anything else here
```

gregdan3:
fixes #35

### DIFF
--- a/lua/lazyvim/plugins/ui.lua
+++ b/lua/lazyvim/plugins/ui.lua
@@ -42,7 +42,7 @@ return {
   -- bufferline
   {
     "akinsho/nvim-bufferline.lua",
-    event = "BufAdd",
+    event = "VeryLazy",
     opts = {
       options = {
         diagnostics = "nvim_lsp",


folke:
thank you! I will set `always_show_bufferline=false` though, so it doesn't show when only one buffer is active

folke:
Nevermind, seems like I already set that :)

PR #48, opened 2023-01-11
fix: add tsx treesitter to typescript lang module

appelgriebsch:
After cleaning up my default treesitter modules (and move most into language specific plugins), I noticed that the typescript module is missing on the tsx parser (for typescript based jsx files)...

### DIFF
--- a/lua/lazyvim/plugins/extras/lang/typescript.lua
+++ b/lua/lazyvim/plugins/extras/lang/typescript.lua
@@ -4,7 +4,7 @@ return {
   {
     "nvim-treesitter/nvim-treesitter",
     opts = function(_, opts)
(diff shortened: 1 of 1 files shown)


lua/lazyvim/plugins/extras/lang/typescript.lua
```diff
@@ -4,7 +4,7 @@ return {
   {
     "nvim-treesitter/nvim-treesitter",
     opts = function(_, opts)
-      table.insert(opts.ensure_installed, "typescript")
+      table.insert(opts.ensure_installed, { "typescript", "tsx" })
```
folke, line 7: should be `vim.list_extend`

folke (changes requested):

folke:
Thanks!

PR #59, opened 2023-01-14
docs: add mini.surround keybindings to readme

sebffischer:


folke:
Thank you for this PR! However, those keymaps in the readme are auto-generated. I've updated the plugin spec so that they are now included in the readme.

PR #65, opened 2023-01-16
adding autoformat option

xpressivecode:


folke:
Thank you for the suggestion! Since I'd like to keep most configs with their plugins, I've added this as an option for `lspconfig` instead, so you can now do:

image

Documentation about custom plugin options is coming soon.

PR #77, opened 2023-01-17
`<` and `>` indentation mappings are not better

ueaner:
It should be better to use `.` for repeated indentation blocks.


folke:
Sorry, but I prefer the mappings I've added.

You can remove them in your `lua/config/keymaps.lua` with `vim.keymap.del`

PR #82, opened 2023-01-20
Add context to tab related keymaps

elijahmanor:
The description for the tab keymaps are currently a bit generic. This gives extra information so they are more helpful in which-key and also in the documentation.

Question, are you manually updating the docs in the other repo or do you have a script that is abstracting the keymaps to generate the keymaps table?


folke:
Thanks!

No need to updated the docs, I have indeed a script that automtates the whole process :)

Btw, let me know what you think about the new website! https://lazyvim.github.io/

Especially, about what might be missing for new users.

folke:
I just updated the docs as well. They should appear in a couple of minutes...

PR #120, opened 2023-01-24
fix(config): don't set options to their defaults

### LINKED ISSUE #119: A few vim.opt options are already nvim defaults
Some of the options in [options.lua](https://github.com/LazyVim/LazyVim/blob/main/lua/lazyvim/config/options.lua) are being set to their default values and could be omitted.

`opt.shortmess` can be appended to for brevity when it's a _superset_ of the defaults, e.g.:

```lua
opt.shortmess:append { c = true, W = true}
```

Just pointing that out, you may have your own reasons/preferences.

adrian5:
Closes #119.

Note: I've set the viable `opt.shortmess` options outside of the `has("nvim-0.9.0")` feature gate as well, assuming they weren't left out intentionally.

### DIFF
--- a/lua/lazyvim/config/options.lua
+++ b/lua/lazyvim/config/options.lua
@@ -5,24 +5,21 @@ vim.g.maplocalleader = " "
 
 local opt = vim.opt
 
-opt.autowrite = true -- enable auto write
-opt.clipboard = "unnamedplus" -- sync with system clipboard
-opt.cmdheight = 1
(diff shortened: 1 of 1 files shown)


adrian5:
Also, I left `spelllang` in (it's `en` by default according to the manual) because I wasn't sure if it interacts with the system locale or something. Could be removed as well otherwise. Or set to `en_us` since that's probably applicable to more users.

folke:
Thanks!

Issue #150, opened 2023-01-29
feature: Make included autocommands revertable/optional


TamasBarta:
### Did you check the docs?

- [X] I have read all the LazyVim docs

### Is your feature request related to a problem? Please describe.

Whenever I resize the window, it switches to the last tab, so I cannot continue working, I have to switch back. I had the same issue with LunarVim, and it was tracked down, that the issue causing it was this autocommand: https://github.com/LazyVim/LazyVim/blob/main/lua/lazyvim/config/autocmds.lua#L16 (the link shows its occurrence in LazyVim)

At the moment I found no solution to disable individual autocommands in LazyVim, because they are not in groups, and the Neovim API I found to remove autocommands take an integer, which my understanding is coming from registering the autocommand, that's not in my code, but in LazyVim's.

### Describe the solution you'd like

In LunarVim I could control individual autocommands, as they were in groups, and groups can be removed by name. If the autocommands were in groups, I could nullify the ones I don't like from LazyVim.

### Describe alternatives you've considered

I was thinking about ways to prevent setting the autocommands altogether, but that just looks unnecessary work, and as far as I can tell it doesn't align with the philosophy of the plugin either, but you tell me.

### Additional context

I even checked the sources, but the lua files tell me there's no conditionality anywhere near loading the autocommands (or the options, but those are easy to change afterwards).

TamasBarta:
I forked "LazyVim/LazyVim" to experiment, and I found that if I wrap "VimResized" in a group, and try to delete a group, I cannot do that, the group is not found.

I thought maybe my autocmds.lua is loaded earlier than LazyVim's, but https://github.com/LazyVim/LazyVim/blob/main/lua/lazyvim/config/init.lua#L123 suggests otherwise.

After some amount of research I found that iteration over tables is not deterministic (my source: https://stackoverflow.com/questions/66329522/what-makes-lua-tables-key-order-be-undeterministic sorry, I'm a Lua noob).

Running the plugin's and the user's configs in non-deterministic order itself is I believe a bug, but even if you accounted for that in the original design, this improvement would probably need that to change, if this idea is accepted.

TamasBarta:
I got it all working in my fork. Have a look at my proposal in #151. It also fixes the indeterministic config load (that I could reproduce with the current foreach solution).

folke:
I'll have a look at the PR later today.

[...]

PR #151, opened 2023-01-29
fix(config): add groups to autocommands

TamasBarta:
I draft proposed solution for the problems described in #150


dpetka2001:
Wouldn't it be better to have them all under the same group? So, if we wanted to disable them we just disable a single group instead of every group for every single autocmd? Depends on the needs of course, but i wouldn't imagine me having to disable every single augroup. I would prefer if i could disable them all under a single group.

dpetka2001:
Oh, i just re-read your issue and you're interested in individual autocmds. Your solution makes more sense when you consider this. Sorry for that.

folke:
I've cleaned it up and merged it. Thanks!

PR #161, opened 2023-01-31
make argc if statement more readable

syIvii:
I was just skimming the code until I saw the vim.fn.argc snippet, I didn't know what was going on at first and did a quick google search
just declared a variable so that the reader knows whats going on


folke:
I prefer the original code. Sorry

syIvii:
> I prefer the original code. Sorry

That's reasonable, all it took for me was a quick google search.
No need to apologize!

PR #171, opened 2023-02-04
feat: use trouble quicklist / loclist if available

appelgriebsch:
If `trouble.nvim` is available: use the quicklist / loclist feature of it, otherwise fallback to the buildin ones.


folke:
Thank you for this suggestion! But I think a lot of people would still want a keybind for the regular loclist and qflist. Maybe add new keymaps with capital letters instead that open Trouble?

So `<leader>xL` and `<leader>xQ`?

Issue #172, opened 2023-02-04
bug: can not overwrite mini.surround keybinds


pprotas:
### Did you check docs and existing issues?

- [X] I have read all the LazyVim docs
- [X] I have searched the existing issues of LazyVim
- [X] I have searched the exsiting issues of plugins related to this issue

### Neovim version (nvim -v)

v0.9.0-dev-2057+g8376486e8-dirty

### Operating system/version

MacOS 13.2

### Describe the bug

You can not overwrite the keybinds for surround objects with `mini.surround`. Here is an example config:

```lua
return {
  {
    "echasnovski/mini.surround",
    opts = {
      mappings = {
        add = "sa",
        delete = "sd",
        find = "sf",
        find_left = "sF",
        highlight = "sh",
        replace = "sr",
        update_n_lines = "sn",
      },
    },
  },
}
```

This doesn't work, see how the keybinds look like in Telescope:

![image](https://user-images.githubusercontent.com/30273418/216764768-760715a9-2952-476b-9f9c-be2bd92de2d7.png)

I tried replicating [the full spec from the docs](https://www.lazyvim.org/plugins/coding#minisurround), but now I have duplicate keybinds:

```lua
return {
  {
    "echasnovski/mini.surround",
    keys = function(plugin, keys)
      local opts = require("lazy.core.plugin").values(plugin, "opts", false)
      local mappings = {
        { opts.mappings.add, desc = "Add surrounding", mode = { "n", "v" } },
        { opts.mappings.delete, desc = "Delete surrounding" },
        { opts.mappings.find, desc = "Find right surrounding" },
        { opts.mappings.find_left, desc = "Find left surrounding" },
        { opts.mappings.highlight, desc = "Highlight surrounding" },
        { opts.mappings.replace, desc = "Replace surrounding" },
        { opts.mappings.update_n_lines, desc = "Update `MiniSurround.config.n_lines`" },
      }
      return vim.list_extend(mappings, keys)
    end,
    opts = {
      mappings = {
        add = "sa",
        delete = "sd",
        find = "sf",
        find_left = "sF",
        highlight = "sh",
        replace = "sr",
        update_n_lines = "sn",
      },
    },
  },
}
```

![image](https://user-images.githubusercontent.com/30273418/216764898-307d16b7-7c7d-431d-b832-723a4bc07eae.png)

Flit and Leap are both **disabled**.

### Steps To Reproduce

1. Use one of the example configs above
[...]

PR #177, opened 2023-02-05
fix(mini.surround): user can change keybinds again

pprotas:
This is related to this issue: https://github.com/LazyVim/LazyVim/issues/172

This `keys` function is preventing the user from changing `mini.surround` keybinds. I removed this and now the user can change the keybinds again.

It looks like this code was here so the keybinds correctly show up in WhichKey, but after removing them they still show up there like expected. Here is a screenshot:

image

The reason that it didn't work before is because the `plugin` parameter contains the `opt` **before** the user's custom keybinds. So it was still using the `gz...` mappings, instead of the user's custom mappings. This means that no matter what the user provided in `opts`, it would always use the old `opts.mappings`.

### DIFF
--- a/lua/lazyvim/plugins/coding.lua
+++ b/lua/lazyvim/plugins/coding.lua
@@ -92,20 +92,6 @@ return {
   -- surround
   {
     "echasnovski/mini.surround",
-    keys = function(plugin, keys)
-      -- Populate the keys based on the user's options
-      local opts = require("lazy.core.plugin").values(plugin, "opts", false)
-      local mappings = {
-        { opts.mappings.add, desc = "Add surrounding", mode = { "n", "v" } },
-        { opts.mappings.delete, desc = "Delete surrounding" },
-        { opts.mappings.find, desc = "Find right surrounding" },
-        { opts.mappings.find_left, desc = "Find left surrounding" },
-        { opts.mappings.highlight, desc = "Highlight surrounding" },
-        { opts.mappings.replace, desc = "Replace surrounding" },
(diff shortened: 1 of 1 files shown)


dpetka2001:
I think the original implementation with `keys` was meant so that the plugin could be lazy loaded only when those keys were pressed. Your implementation removes this option and plugin is always loaded. Correct me if i'm wrong cuz i'm not that adept with Lua and the intrisics of Lazyvim(lazy.nvim) yet. I believe the developer will have a more accurate opinion.

pprotas:
But doesn't the plugin set the keymaps anyway?

Does the `keys` property cause the plugin to lazy load? Is there documentation on this? Seems counterintuitive to me

dpetka2001:
It's a feature of `[lazy.nvim](https://github.com/folke/lazy.nvim)`. If you read at the section `Plugin Spec` you should see the keywords available and what they do. That is just my assumption though(and maybe i'm wrong), as I don't have a full understanding of how it works. The dev would have a more accurate opinion, so you should wait for his input.

dpetka2001:
I tried your change and the plugin does indeed load at the start of Neovim. With the original configuration the 
plugin does not load at startup. It only lazy loads when the `keys` are called. My opinion is that it should stay like it
is and users should just tinker with the configuration according to their preferences. From my understanding this is 
the purpose of LazyVim. But that is just my opinion. It's up to the dev to decide about your "fix".

pprotas:
Interesting! I missed this from the docs.

What if we just make it lazy load on a different event instead of `keys`? Seems like there are some options here.

folke:
This should now be fixed

PR #178, opened 2023-02-05
fix: add missing border

### LINKED ISSUE #173: LSP windows don't have border
I'm loving lazyvim!

The only thing do fat that's bugging me is I don't know how to make LSP windows like hover, diagnostic etc to have a border.

Any idea?
Thanks!

loichyan:
This PR sets rounded border for some plugins.

Fixes #173

### DIFF
--- a/lua/lazyvim/plugins/coding.lua
+++ b/lua/lazyvim/plugins/coding.lua
@@ -40,6 +40,7 @@ return {
     },
     opts = function()
       local cmp = require("cmp")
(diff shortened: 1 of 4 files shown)


folke:
Thank you for this PR, however I personally prefer not having borders :)

But as you found out yourself, it's pretty easy to add borders to your config.

loichyan:
How about using an extra layer instead (like `lazyvim.plugins.extras.ui.border`)? This provides a shortcut for users who want to add a border.

PR #189, opened 2023-02-06
Improve Telescope mappings

Cretezy:
- Adds `<C-j>`/`<C-k>` in insert mode to scroll in results list
- Adds `<C-n>`/`<C-p>` in insert mode to scroll in preview
  - Maybe `<C-b>`/`<C-f>` or `<C-u>`/`<C-d>` could be used?
- Adds `q` in normal mode to quit

> Inspired by AstroNvim


folke:
The `<c-n>` and `<c-p>` mappings you added got back and forth in history like the mappings we already have. That's probably not intentional?

folke:
Fixed by #231

PR #195, opened 2023-02-07
feat: add copilot icon

joshmedeski:
While setting up [copilot-cmp](https://github.com/zbirenbaum/copilot-cmp) I figured adding to the icons in LavyVim is the quickest way for me to get the icon support.

Screenshot 2023-02-06 at 9 32 53 PM

Let me know if there's a simple way to extend the kinds icons if you don't think this is the right place to add it.


folke:
thanks!

PR #196, opened 2023-02-07
fix #191 Release unused `<c-k>` `<c-b>` `<c-f>` keymaps in `insert mode`

### LINKED ISSUE #191: Release unused `<c-k>` `<c-b>` `<c-f>` keymaps in `insert mode`
I'm not sure what events generate floating windows in insert mode.

https://github.com/LazyVim/LazyVim/blob/2ff213126ceb7c85619a084f9bde4c92199248f3/lua/lazyvim/plugins/ui.lua#L206-L207

Seeing `<c-k>` triggers `Signature Help`.

https://github.com/LazyVim/LazyVim/blob/2ff213126ceb7c85619a084f9bde4c92199248f3/lua/lazyvim/plugins/lsp/keymaps.lua#L21

It should be rare to move to a function in insert mode to view its `Signature Help`, and the `Signature Help` is often not large and does not need to scroll.

So I think in insert mode, we should release `<c-k>` `<c-b>` `<c-f>` these keys to do other things.

What do you think of my humble opinion?

ueaner:
If I overlooked something, please do let me know.


folke:
Those mappings were added for a reason. I'm not just going to remove them.

PR #231, opened 2023-02-10
Improved Telescope mappings in normal and insert mode

sourproton:
I saw @Cretezy 's pull request and realized I made the same modifications in my own config, but with `move_selection_next` and `move_selection_previous` instead of `cycle_history`. So, this PR
- In `insert` mode
  - Adds `<C-j>`/`<C-k>` to scroll in results list
  - Adds `<C-n>`/`<C-p>`  to scroll in preview
- In `normal` mode
  - Adds `q` to quit

### DIFF
--- a/lua/lazyvim/plugins/editor.lua
+++ b/lua/lazyvim/plugins/editor.lua
@@ -133,6 +133,23 @@ return {
             ["<C-Up>"] = function(...)
               return require("telescope.actions").cycle_history_prev(...)
             end,
+            ["<C-j>"] = function(...)
+              return require("telescope.actions").move_selection_next(...)
+            end,
+            ["<C-k>"] = function(...)
+              return require("telescope.actions").move_selection_previous(...)
+            end,
+            ["<C-n>"] = function(...)
+              return require("telescope.actions").preview_scrolling_down(...)
+            end,
+            ["<C-p>"] = function(...)
+              return require("telescope.actions").preview_scrolling_up(...)
+            end,
+          },
+          n = {
+            ["q"] = function(...)
+              return require("telescope.actions").close(...)
+            end,
           },
         },
       },


ditsuke:
Made the same mod in my config, thanks for adding here!

sourproton:
Thank you @folke for accepting the PR. On the Telescope subject, in my config I also have `nvim-telescope/telescope-file-browser.nvim` setup in the following way
```lua
{
    "nvim-telescope/telescope-file-browser.nvim",
    dependencies = { "nvim-telescope/telescope.nvim" },
    config = function()
        require("telescope").load_extension "file_browser"
    end,
    keys = {
        { "<leader>fd", "<cmd>Telescope file_browser<cr>", desc = "File Browser (cwd)" },
    },
}
```
Would you be interested in having this merged too ?

folke:
I don't think it's all that useful, given that we already have `neo-tree` and `Telescope find_files`.

sourproton:
I agree. I just though it would be worth it asking.

Finally, I also have vim-slime setup to work with the Julia and Python REPLs with tmux. I think this is even less worth it merging, but I think others would like having this template to build their configs upon (there aren't that many examples on how to nicely work with REPLs and NeoVim).

Where do you think would be the ideal place to showcase it?

folke:
You can use https://github.com/LazyVim/LazyVim/discussions/categories/show-and-tell for that.

folke:
Hi! I just wanted to let you know that I changed the Telescope keybindings to make them more consistent with cmp / signature help and hover.

`<c-f>` and `<c-b>` should now be used for scrolling the preview. This is the same as used for the cmp doc window, signature help and for scrolling the hover window.

I've also removed the new `<c-i>` and `<c-j>` mappings since in other places we also use the default way of navigating a list with `<c-n>` and `<c-p>`. Again this is to make all mappings more consistent.

You can of course still override this in your own mappings if that's what you prefer.

PR #232, opened 2023-02-10
chore: better dependency declaration for leap/flit

ditsuke:
## What

Decouples plugin table entries for leap.nvim and flit.nvim

## Why

As `flit` depends on `leap` and `leap` should be a component on its own (and works as such), declarations are decoupled with a more apt  flit -> leap dependency declaration to make the intent and effect clearer.

### DIFF
--- a/lua/lazyvim/plugins/editor.lua
+++ b/lua/lazyvim/plugins/editor.lua
@@ -139,11 +139,10 @@ return {
     },
   },
 
-  -- easily jump to any location and enhanced f/t motions for Leap
+  -- enhanced 2-char search motions
   {
     "ggandor/leap.nvim",
     event = "VeryLazy",
-    dependencies = { { "ggandor/flit.nvim", opts = { labeled_modes = "nv" } } },
     config = function(_, opts)
       local leap = require("leap")
       for k, v in pairs(opts) do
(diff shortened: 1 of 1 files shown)


ditsuke:
Opinionated take: we should disable flit's `multiline` option by default. As someone used to vanilla vim `f`/`t`, the multiline felt more unnatural than useful. I have it disabled in my configuration.

folke:
What does this solve?

folke:
In your personal config you can have these as two different specs.
So what does this solve for you?

ditsuke:
To me it makes it clear LazyVim includes `leap.nvim`

folke:
Sorry, but this is not making any sense. Everyting works as expected, no?

A dependency is an extension on its own as well obviously.

The docs show them separately:
image

Closing this.

folke:
> Opinionated take: we should disable flit's `multiline` option by default. As someone used to vanilla vim `f`/`t`, the multiline felt more unnatural than useful. I have it disabled in my configuration.

If I'd have to chose how vanilla vim works for everything, then I'd just not include any plugins.

The goal with LazyVim is not to make it work like vanlla Vim. It's quite the opposite. I want to have an IDE that is an improvement over vanilla vim

ditsuke:
> If I'd have to chose how vanilla vim works for everything, then I'd just not include any plugins.
> 
> The goal with LazyVim is not to make it work like vanlla Vim. It's quite the opposite. I want to have an IDE that is an improvement over vanilla vim

Thanks for considering, it was an opinionated take after all.

While we're at it I wanted to ask if you would be open to a PR adding `<C-j>` `C-k>` bindings for `nvim-cmp` results, in spirit of the mappings from #231?

sourproton:
Would you mind sharing your `nvim-cmp` bindings here, even if it doesn't get merged ?

ditsuke:
> Would you mind sharing your `nvim-cmp` bindings here, even if it doesn't get merged ?

PR #240 is open for review and discussion.

PR #240, opened 2023-02-11
feat: improved keybinds to navigate completions

ditsuke:
## What
`<C-j>`, `<C-k>` to navigate completions, both for `nvim-cmp` and for the command mode.

## Why
Widely used mappings, already used in LazyVim's Telescope config after #189.

### DIFF
--- a/lua/lazyvim/config/keymaps.lua
+++ b/lua/lazyvim/config/keymaps.lua
@@ -139,3 +139,7 @@ map("n", "<leader><tab><tab>", "<cmd>tabnew<cr>", { desc = "New Tab" })
 map("n", "<leader><tab>]", "<cmd>tabnext<cr>", { desc = "Next Tab" })
 map("n", "<leader><tab>d", "<cmd>tabclose<cr>", { desc = "Close Tab" })
 map("n", "<leader><tab>[", "<cmd>tabprevious<cr>", { desc = "Previous Tab" })
+
+-- navigate completions in command-mode with <C-j>, <C-k>
+vim.keymap.set("c", "<C-j>", 'pumvisible() ? "\\<C-n>" : "\\<C-j>"', { expr = true, noremap = true })
+vim.keymap.set("c", "<C-k>", 'pumvisible() ? "\\<C-p>" : "\\<C-k>"', { expr = true, noremap = true })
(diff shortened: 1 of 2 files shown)


dpetka2001:
I would prefer if the `command-line` changes didn't go through. I already use `C-j(k)` as `Up(Down)` for command line history previous(next) tokens. This has been the default in Neovim (I mean `C-n(p)`) and some of us have other personal keybindings that conflict with these. Please keep the defaults as is and let people tinker the configuration according to their preferences.

ditsuke:
> I would prefer if the `command-line` changes didn't go through. I already use `C-j(k)` as `Up(Down)` for command line history previous(next) tokens. This has been the default in Neovim (I mean `C-n(p)` for selection) and some of us have other personal keybindings that conflict with these. Please keep the defaults as is and let people tinker the configuration according to their preferences. Just my humble opinion. Up to the maintainer to decide.

I think your existing  `<C-j(k)`  bindings will override those set by LazyVim. These mappings should only only take effect if you don't have your own.

dpetka2001:
> > I would prefer if the `command-line` changes didn't go through. I already use `C-j(k)` as `Up(Down)` for command line history previous(next) tokens. This has been the default in Neovim (I mean `C-n(p)` for selection) and some of us have other personal keybindings that conflict with these. Please keep the defaults as is and let people tinker the configuration according to their preferences. Just my humble opinion. Up to the maintainer to decide.
> 
> I think your existing `<C-j(k)` bindings will override those set by LazyVim. These mappings should only only take effect if you don't have your own.
Then it's my mistake. I didn't think this through. If that's the case please disregard my previous comment. My bad

folke:
Seeing this PR, actually made me realize that the Telescope PR I merged did not add the same consistent keymaps as is used in other places.

So I just changed the Telescope keybindings to make them more consistent with cmp / signature help and hover.

`<c-f>` and `<c-b>` should now be used for scrolling the preview. This is the same as used for the cmp doc window, signature help and for scrolling the hover window.

I've also removed the new `<c-i>` and `<c-j>` mappings since in other places we also use the default way of navigating a list with `<c-n>` and `<c-p>`. Again this is to make all mappings more consistent.

You can of course still override this in your own mappings if that's what you prefer.

ditsuke:
> Seeing this PR, actually made me realize that the Telescope PR I merged did not add the same consistent keymaps as is used in other places.
> 
> So I just changed the Telescope keybindings to make them more consistent with cmp / signature help and hover.

Thank you for considering the PR and making the existing bindings consistent. Cheers!

PR #284, opened 2023-02-16
feat(telescope): add `<leader>fR` for `:Telescope resume`

cbochs:
I've recently switched to LazyVim and really liked many of the out-of-the-box defaults. One keybind I would like though is a way to quickly call `Telescope resume`. I thought `<leader>fR` might be a good place for it. Let me know what you think.

Thanks!


alezunino:
I use <leader>sR for :Telescope resume because there are are many Telescope keys under "<leader>s"

cbochs:
Fair. I'm happy to move it under the `<leader>s` (search) category instead.

alezunino:
I have this in keymaps.lua:
`
vim.keymap.set("n", "<leader>sR", "<cmd>Telescope resume<cr>", { desc = "Resume telescope search" })
`

folke:
`sR` probably makes more sense indeed.

folke:
Thank you!

PR #300, opened 2023-02-18
feat(project): add project management

Isti115:
I think that this is a generally useful functionality that's present in many editors / configs, so many users would expect it by default. Also, it's very simple and easy to ignore, so I don't think that it would bother / bloat the experience for those who don't use it.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -16,7 +16,7 @@ body:
           required: true
         - label: I have searched the existing issues of LazyVim
           required: true
-        - label: I have searched the exsiting issues of plugins related to this issue
+        - label: I have searched the existing issues of plugins related to this issue
           required: true
   - type: input
     attributes:
@@ -59,6 +59,7 @@ body:
       value: |
         -- DO NOT change the paths and don't remove the colorscheme
         local root = vim.fn.fnamemodify("./.repro", ":p")
+
         -- set stdpaths to use .repro
         for _, name in ipairs({ "config", "data", "state", "cache" }) do
           vim.env[("XDG_%s_HOME"):format(name:upper())] = root .. "/" .. name
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,8 +15,12 @@ jobs:
       - name: Install Neovim
         shell: bash
         run: |
-            wget -q https://github.com/neovim/neovim/releases/download/nightly/nvim-linux64.deb -O /tmp/nvim.deb
-            sudo dpkg -i /tmp/nvim.deb
+          mkdir -p /tmp/nvim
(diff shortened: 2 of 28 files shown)


folke:
I'm not using projects myself, so want to check this fully first before merging. It might also make more sense to put this under extras.

I'm leaving on vacation tomorrow, so won't have the time to test this, but will do so next week. 

Thank you!

alexander-born:
Just want to bring the nvim-telescope https://github.com/nvim-telescope/telescope-project.nvim plugin into discussion.

lucasrabiec:
> I'm not using projects myself, so want to check this fully first before merging. It might also make more sense to put this under extras.
> 
> I'm leaving on vacation tomorrow, so won't have the time to test this, but will do so next week.
> 
> Thank you!

What do you use (if any) to switch between local projects (e.g. git repos) within one nvim instance? I am just wondering what are the other solutions to achieve this.

briandipalma:
> > I'm not using projects myself, so want to check this fully first before merging. It might also make more sense to put this under extras.
> > I'm leaving on vacation tomorrow, so won't have the time to test this, but will do so next week.
> > Thank you!
> 
> What do you use (if any) to switch between local projects (e.g. git repos) within one nvim instance? I am just wondering what are the other solutions to achieve this.

Why not just have multiple tabs open in your terminal (kitty in my case) and have a Neovim in each? Very straightforward and quick too using keymappings/fzf.

folke:
Yeah, that's what I do...

lucasrabiec:
Ofc this is an option. I only asked for other plugins than (telescope)project.nvim, but that's okay and no problem at all. I will stick with tmux.

briandipalma:
I'm not saying that there shouldn't be project management in LazyVim, I'm agnostic TBH. I'm just saying that a terminal with built-in tabs/window management like kitty or wezterm seems to make such functionality redundant. I used to use Alacritty and tmux, in the end I had too many issues with tmux and ended up moving. Couldn't be happier and have no desire to go back to tmux/Alacritty. I may one day move to wezterm as I'd like to configure my terminal with Lua but I'm not too bothered and I'm happy with kitty.

Isti115:
For me the use case is not at all about switching _between_ projects, but rather opening a blank instance of neovim and choosing a recent project to jump back into.
I mostly run LazyVim as a separate app (no shell, just the executable running in a terminal window), so I don't set the working directory before opening it.
(This breaks session restoration from the dashboard as well, since that relies on the working directory. 😐)
This is why a project manager to choose what I'd like to work in after opening a _"blank"_ LazyVim is very useful to me.

ibeex:
Can this go to the documentation about extending? So anybody can add it to the local config or leave it as is (usage to terminal or multiplexer?)

folke:
I've moved the project stuff to extras. Use it by importing this in your lazy confg:

```lua
      { import = "lazyvim.plugins.extras.util.project" },
```

PR #304, opened 2023-02-19
feat: persist pinned tabs

ditsuke:
## What does this PR do?
Enables persistence of pinned buffers across sessions.

## Summary of changes
Adds `globals` to session opts


folke:
thank you!

PR #309, opened 2023-02-20
feat(telescope): add `<leader>sS` for `:Telescope lsp_workspace_symbols`

NyCodeGHG:
I found myself often wanting to search trough the workspace symbols and i think this keymap makes the most sense, since there is already the same thing for document symbols with `<leader>ss`.


folke:
Makes sense!

PR #331, opened 2023-02-24
fix(icons): replace obsolete Nerd icons

xzsk2:
same as https://github.com/folke/lazy.nvim/pull/529

these obsoleted icons will be rendered as CJK characters for CJK font users, replace them with correct new icons


folke:
Thank you!

PR #348, opened 2023-02-28
Only map lsp go definition when client has definitionProvider

alexander-born:
When using null-ls formatter my custom "gd" mapping gets overwritten even thouh null-ls formatters don't have a `definitionProvider".


folke:
thank you!

PR #383, opened 2023-03-04
feat: add luap to nvim-treesitter ensure_installed

amaanq:
Needs https://github.com/nvim-treesitter/nvim-treesitter/pull/4433 but I think it'll benefit anyone using Lua (similar to regex highlights in js)


folke:
works really great. thanks!

amaanq:
np :)

PR #421, opened 2023-03-12
mini-starter: add Session restore to starter menu

willnorris:
This mirrors the menu item added in alpha-nvim. I named this "Session restore" rather than "Restore session", since mini.starter uses the first letter of the item as the shortcut key.  This way it doesn't conflict with "Recent files".

image


folke:
thanks!

PR #433, opened 2023-03-14
fix/indent-blankline-setup

lessthanseventy:
The setup function for indent-blankline is ```require("indent_blankline").setup(opts)```, so it was not being called for me with the default configuration.


folke:
It does get called. This PR is not needed

PR #453, opened 2023-03-18
fix(copilot): use full name of nvim-cmp plugin

willnorris:
Otherwise, `:checkhealth` complains of:

> lazy.nvim
> - WARNING {nvim-cmp}: overriding <1>


folke:
That's actually ok and a bug in lazy.nvim's **health**. Just fixed it there. Will make a new release soon. Thanks for letting me know!

PR #490, opened 2023-03-24
feat(treesitter): use Python indents as they've improved greatly

amaanq:
Thanks to https://github.com/nvim-treesitter/nvim-treesitter/pull/4524 being merged, Python indents are quite good now


folke:
Sweet!

bluss:
I still have some blocking problems with this, like typing `_logger.info("something:` will indent the current line while typing, which is wrong because the `:` is going to be inside a string.

PR #512, opened 2023-03-31
feat(autocmds): add checkhealth filetype to close_with_q list

shashanktomar:
Add `checkhealth` filetype to `close_with_q` autocmd list.


folke:
thanks!

PR #552, opened 2023-04-08
feat(catppuccin): better look

loichyan:
This PR enables catppuccin integrations for used plugins and changes the highlight groups of alpha-nvim to `Dashboard*` (which look the same as `Alpha*` in tokyonight and are supported by many colorschemes).


amaanq:
This is entirely personal preference with quite a bit of customization and doesn't belong here imo, not sure what the dashboard hl changes are all about as well as alpha isn't dashboard.

folke:
I agree with @amaanq on this one.

PR #557, opened 2023-04-10
feat(keymaps): add some telescope keymaps

loichyan:
Adds document/workspace diagnostics and todo/fixme telescope keymaps.


folke:
thanks!

Issue #576, opened 2023-04-13
bug: workspace symbols returns an error


aymanbagabas:
### Did you check docs and existing issues?

- [X] I have read all the LazyVim docs
- [X] I have searched the existing issues of LazyVim
- [X] I have searched the existing issues of plugins related to this issue

### Neovim version (nvim -v)

NVIM v0.10.0-dev-1409+g23e9d625c-Homebrew

### Operating system/version

MacOS 13.3.1

### Describe the bug

Whenever I try to find workspace symbols for example using the keybinding <kbd>\<leader>sS</kbd>, I get this error:

```
[telescope.filter_symbols]: class, function, method, constructor, interface, module, struct, trait, field, property symbol(s) were not part of the query results
```

![image](https://user-images.githubusercontent.com/3187948/231817955-f7b565ae-dd9d-4ecf-9ae8-586614639b5d.png)

### Steps To Reproduce

1. Start LazyVim in a LSP aware env
2. Press <kbd>\<leader>sS</kbd>
3. 

### Expected Behavior

Telescope should show up with workspace symbols.

PR #577, opened 2023-04-13
fix(telescope): use dynamic workspace symbols

aymanbagabas:
`lsp_workspace_symbols` may fail or return nothing when query is empty. Use the new `lsp_dynamic_workspace_symbols` to dynamically query symbols in the workspace.

Related: https://github.com/nvim-telescope/telescope.nvim/issues/964
Fixes: https://github.com/LazyVim/LazyVim/issues/576


folke:
thank you!

PR #579, opened 2023-04-14
feat(keymaps): Get rename command from inc_rename configuration

aarondill:
Gets the rename command from the inc_rename module (when installed), rather than assuming the default `IncRename`, allowing the user to modify the inc_rename options without having to create a new keymap.


folke:
Thanks!

PR #612, opened 2023-04-18
feat(dap): add more keybindings and lualine component

lkhphuc:
I am not sure what the criteria for adding default keybindings to LazyVim, just want to share a few missing keys from my config.
Plus a lualine components showing debug adapter status.


folke:
I updated some of the keymaps and merged it. Thanks!

PR #629, opened 2023-04-19
feat(treesitter): use `auto_install` to install missing parser

qazxcdswe123:
even though it recommend "set to false if you don't have `tree-sitter` CLI installed locally", it actually works perfectly

by doing this we can also potentially reduce the number of `ensure_installed`


folke:
Thank you for the PR, but I'm not a fan of `auto_install`. But you can of course still add it to your own config.

PR #640, opened 2023-04-20
feat(project): add to mini.starter

Jomik:
I am unsure where to put this, as it sort of belongs to two extras.
What I added here will cause mini.starter to always be loaded if you include the projects extra, right?

### DIFF
--- a/lua/lazyvim/plugins/extras/util/project.lua
+++ b/lua/lazyvim/plugins/extras/util/project.lua
@@ -27,4 +27,17 @@ return {
       table.insert(dashboard.section.buttons.val, 4, button)
     end,
   },
+  {
+    "echasnovski/mini.starter",
+    opts = function(_, opts)
+      local items = {
+        {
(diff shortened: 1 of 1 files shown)


nunestiago:
Gave a check and couldn't really understand what else it brings that the already installed [goolord/alpha-nvim](https://github.com/goolord/alpha-nvim) is not able to do.

Jomik:
> Gave a check and couldn't really understand what else it brings that the already installed [goolord/alpha-nvim](https://github.com/goolord/alpha-nvim) is not able to do.
It is to support https://github.com/LazyVim/LazyVim/blob/81ab5bed7a81c3005f862d6508dacc26c4ed553a/lua/lazyvim/plugins/extras/ui/mini-starter.lua

Jomik:
That is awesome! Should be ready now :)

folke:
Just tested and it all works as expected :)

Thanks!

PR #642, opened 2023-04-20
feat(tests): adds test management

Jomik:
Adds [neotest](https://github.com/nvim-neotest/neotest) setup based on my very early understanding of neotest.

We can then add test adapters in the following way.
```lua
return {
  {
    "nvim-neotest/neotest",
    dependencies = {
      "nvim-neotest/neotest-go",
    },
    opts = function(_, opts)
      -- Would like "cleaner" way than this..
      opts.adapters = vim.list_extend(opts.adapters or {}, {
        require("neotest-go")({ args = { "-tags=integration" } }),
      })
    end,
  },
}
```

Can we somehow only add the "Debug Nearest" keybind if `nvim-dap` exists?


loqusion:
`FixCursorHold` may need to be added as a dependency even though part of its functionality has already been merged into `neovim`:

> Neotest uses the `CursorHold` event. This uses the `updatetime` setting which is by default very high, and lowering this can lead to excessive writes to disk. It's recommended to use https://github.com/antoinemadec/FixCursorHold.nvim which allows detaching `updatetime` from the frequency of the `CursorHold` event. The repo claims it is no longer needed but it is still recommended (See [this issue](https://github.com/antoinemadec/FixCursorHold.nvim/issues/13))

Jomik:
> `antoinemadec/FixCursorHold.nvim` may need to be added as a dependency even though part of its functionality has already been merged into `neovim`:
> 
> > Neotest uses the `CursorHold` event. This uses the `updatetime` setting which is by default very high, and lowering this can lead to excessive writes to disk. It's recommended to use https://github.com/antoinemadec/FixCursorHold.nvim which allows detaching `updatetime` from the frequency of the `CursorHold` event. The repo claims it is no longer needed but it is still recommended (See [this issue](https://github.com/antoinemadec/FixCursorHold.nvim/issues/13))

I am unsure of the impact - decided to not add it as I do not understand it fully.
It should be easy to add from userland. But I am definitely not against adding it either.
Just need to be told what is "correct" 😄

Jomik:
@folke What do you think about this? Do you prefer me opening an Idea in discussions?

kevinrobayna:
I would love to have this tbh! Something else i would add to the default configuration is neotest as it's a best effort plugin to run test automatically. see my config for neotest https://github.com/kevinrobayna/dotfiles/blob/c699b32f7493e0caf44dc94db738447fd41b03f9/config/.config/nvim/lua/plugins/neotest.lua

```
				require("neotest-vim-test")({
					ignore_file_types = { "ruby", "go" },
				})
```

Jomik:
> I would love to have this tbh! Something else i would add to the default configuration is neotest as it's a best effort plugin to run test automatically. see my config for neotest https://github.com/kevinrobayna/dotfiles/blob/c699b32f7493e0caf44dc94db738447fd41b03f9/config/.config/nvim/lua/plugins/neotest.lua
> 
> ```
> require("neotest-vim-test")({
>   ignore_file_types = { "ruby", "go" },
> })
> ```
> 
> Although i think is also makes sense to leave it to the consumer so that they do something like this
> 
> ```
>   {
>     "nvim-neotest/neotest",
>     dependencies = {
>       "nvim-neotest/neotest-vim-test",
>     },
>     opts = function(_, opts)
>       opts.adapters = vim.list_extend(opts.adapters or {}, {
>         require("neotest-vim-test")({
>           ignore_file_types = { "ruby", "go" },
>         }),
>       })
>     end,
>   },
> ```

I would personally prefer not to have `neotest-vim-test` as a dependency, it seems like that is only there because there is not a specific adapter for all languages yet, so this work as backwards compatible bridge to `vim-test`.

I am not sure how this relates to auto running tests - unless `vim-test` does that?
But if I understand correctly, you can auto run tests using neotest with an auto command.
Not sure if we want to include that in a default configuration either though?

folke:
Sorry for the delay. Will have a look into this tomorrow!

kevinrobayna:
> > I would love to have this tbh! Something else i would add to the default configuration is neotest as it's a best effort plugin to run test automatically. see my config for neotest https://github.com/kevinrobayna/dotfiles/blob/c699b32f7493e0caf44dc94db738447fd41b03f9/config/.config/nvim/lua/plugins/neotest.lua
> 
> > 
> 
> > ```
> 
> > require("neotest-vim-test")({
> 
> >   ignore_file_types = { "ruby", "go" },
> 
> > })
> 
> > ```
> 
> > 
> 
> > Although i think is also makes sense to leave it to the consumer so that they do something like this
> 
> > 
> 
> > ```
> 
> >   {
> 
> >     "nvim-neotest/neotest",
> 
> >     dependencies = {
> 
> >       "nvim-neotest/neotest-vim-test",
> 
> >     },
> 
> >     opts = function(_, opts)
> 
> >       opts.adapters = vim.list_extend(opts.adapters or {}, {
> 
> >         require("neotest-vim-test")({
> 
> >           ignore_file_types = { "ruby", "go" },
> 
> >         }),
> 
> >       })
> 
> >     end,
> 
> >   },
> 
> > ```
> 
> 
> 
> I would personally prefer not to have `neotest-vim-test` as a dependency, it seems like that is only there because there is not a specific adapter for all languages yet, so this work as backwards compatible bridge to `vim-test`.
> 
> 
> 
> I am not sure how this relates to auto running tests - unless `vim-test` does that?
> 
> But if I understand correctly, you can auto run tests using neotest with an auto command.
> 
> Not sure if we want to include that in a default configuration either though?

Yeah I think the default configuration should be as minimal as possible to be honest

lua/lazyvim/plugins/extras/test/core.lua
```diff
@@ -0,0 +1,51 @@
...
+        callback = function(event)
+          vim.bo[event.buf].buflisted = false
+          vim.keymap.set("n", "q", "<cmd>close<cr>", { buffer = event.buf, silent = true })
+        end,
+      })
+
+      require("neotest").setup(opts)
+    end,
+    -- stylua: ignore
+    keys = {
```
kevinrobayna, line 41: Some of these are clashing with other LazyVim mappings we should probably avoid that? 😅 (leader t t, is for sure one of those)

Jomik: I do not see `<leader>t` anywhere in the keymaps?
There is no `<leader>t+` group anywhere https://github.com/LazyVim/LazyVim/blob/25d37a2cdd6202f1d767595454a7f84f77bbd83e/lua/lazyvim/plugins/editor.lua#L240-L250
Is this something you have in your config maybe?

I see them defined here in your dotfiles 😄 
https://github.com/kevinrobayna/dotfiles/blob/373cc8669849bb77874c9a9b50629e1de089ad91/config/.config/nvim/lua/config/keymaps.lua#L24-L27 https://github.com/kevinrobayna/dotfiles/blob/373cc8669849bb77874c9a9b50629e1de089ad91/config/.config/nvim/lua/plugins/overseer.lua#L5

Jomik:
> I recently expedited some effort in setting up neotest with language plugins broken up into language layers:
> 
> 1. Core config with ships with `neotest-vim-test`: https://github.com/ditsuke/nvim-config/blob/main/lua/ditsuke/extras/code/neotest.lua

Why do you want `neotest-vim-test`? It seems like a "catch-all" / "bloated" package.
We want to stay minimal, so we should really only have language specific adapters. `neotest-vim-test` seems easy to add in userland anyway?

ditsuke:
> > I recently expedited some effort in setting up neotest with language plugins broken up into language layers:
> > 
> > 1. Core config with ships with `neotest-vim-test`: https://github.com/ditsuke/nvim-config/blob/main/lua/ditsuke/extras/code/neotest.lua
> 
> Why do you want `neotest-vim-test`? It seems like a "catch-all" / "bloated" package.
> We want to stay minimal, so we should really only have language specific adapters. `neotest-vim-test` seems easy to add in userland anyway?
> 
> 

I'm fine either way wrt vim-test. My concern is having a clean way to enable neotest adaptors for imported language layers.

Jomik:
@folke Did you have time to give feedback here? 😄

lua/lazyvim/plugins/extras/test/core.lua
```diff
@@ -0,0 +1,51 @@
+return {
+  {
+    "nvim-neotest/neotest",
+    dependencies = {
+      -- which key integration
+      {
```
folke, line 6: nitpick, but you can remove this as a dep and just add it separately avove the neotest spec

lua/lazyvim/plugins/extras/test/core.lua
```diff
@@ -0,0 +1,51 @@
...
+      local neotest_ns = vim.api.nvim_create_namespace("neotest")
+      vim.diagnostic.config({
+        virtual_text = {
+          format = function(diagnostic)
+            local message = diagnostic.message:gsub("\n", " "):gsub("\t", " "):gsub("%s+", " "):gsub("^%s+", "")
+            return message
+          end,
+        },
+      }, neotest_ns)
+      local group = vim.api.nvim_create_augroup("lazyvim_neotest_close_with_q", { clear = true })
```
folke, line 26: no need to create a separate autocmd here. It's ok to add `neotest-output` to the default close_with_q autocmd

lua/lazyvim/plugins/extras/test/core.lua
Jomik: Done. Added this fancy new `optional` keyword too 😄

lua/lazyvim/plugins/extras/test/core.lua
```diff
@@ -0,0 +1,38 @@
...
+        },
+      }, neotest_ns)
+
+      require("neotest").setup(opts)
+    end,
+    -- stylua: ignore
+    keys = {
+      { "<leader>tr", function() require("neotest").run.run() end, desc = "Run Nearest" },
+      { "<leader>tR", function() require("neotest").run.run(vim.fn.expand("%")) end, desc = "Run File" },
+      { "<leader>td", function() require("neotest").run.run({strategy = "dap"}) end, desc = "Debug Nearest" },
```
Jomik, line 31: @folke How can we handle this? It should probably only be enabled when `nvim-dap` is present.

folke: Take it outside of the neotest spec and add a new spec for `nvim-dap`. Add the keys there. And make that spec optional.

`keys` are always merged even as a list, so no need to use a keys function for that

Jomik: Perfect solution! This `optional` property is a great addition 👍

folke: Yeah, should have added that a while ago :) Useful for other LazyVim plugin specs too. Making some changes there as well

folke:
Let me know when it's ready to merge

Jomik:
> Let me know when it's ready to merge

Just copied into my personal config and it works as expected.
Unsure if there is a good way to test other than that.. 😄 
Do we need any documentation / examples on how to configure a test adapter/plugin?
I only have `neotest-go` as a working example currently.
```lua
  {
    "nvim-neotest/neotest",
    dependencies = {
      "nvim-neotest/neotest-go",
    },
    opts = function(_, opts)
      opts.adapters = vim.list_extend(opts.adapters or {}, {
        require("neotest-go")({
          args = { "-tags=integration" },
        }),
      })
    end,
  },
```

folke:
Looks good. I don't really use neotest myself, but may give it a try :)

folke:
Merged. Thanks!

PR #651, opened 2023-04-21
fix: remove unintentional `<lt>nop>` mapping

loqusion:
There is quite a noticable amount of lag for the visual mode `<lt>` mapping because of the `<lt>nop>` mapping made by treesitter. (`<lt>` is escaped `<`)

Before:

Kapture 2023-04-21 at 02 10 54

After:

Kapture 2023-04-21 at 02 13 18

[Relevant mapping code is here](https://github.com/nvim-treesitter/nvim-treesitter/blob/b0338a056525e83d88aaa76e8cf396af07410bed/lua/nvim-treesitter/incremental_selection.lua)


folke:
thanks!

PR #653, opened 2023-04-21
fix(telescope): mixed up cwd and root dir

butschi:
The `cwd` and `root dir` were mixed up in the description of a few telescope commands.

This fixes the description while keeping the functionality like before.


folke:
no, the descriptions were correct. your changes make them wrong

PR #670, opened 2023-04-24
feat(bufferline): use `mini.bufremove` to close a buffer

loichyan:
Use `mini.bufremove` to close buffers without breaking window layout in commands such as `BufferLineCloseLeft`, `BufferLineCloseRight`.


folke:
Merged, but changed it to not force remove buffers. Thank you!

PR #675, opened 2023-04-25
Improve description of notify.dismiss() binding

albertfgu:
I got confused by the description and thought it deleted the history of notifications so I never tried it. Went to add a keymap and it turned out this one did what I wanted already, but I think calling it "Dismiss" is more clear.


folke:
thanks!

PR #719, opened 2023-05-06
Update project.lua

FBamx:
When directly inputting "Telescope projects" in the command mode, an error will be reported, and the error message is "[telescope.run_command]: Unknown command". If the character 's' in word "projects" is removed, the command will be executed successfully and will not affect the execution of < leader >fp

### DIFF
--- a/lua/lazyvim/plugins/extras/util/project.lua
+++ b/lua/lazyvim/plugins/extras/util/project.lua
@@ -12,7 +12,7 @@ return {
           require("telescope").load_extension("projects")
         end,
         keys = {
-          { "<leader>fp", "<Cmd>Telescope projects<CR>", desc = "Projects" },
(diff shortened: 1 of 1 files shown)


dpetka2001:
As far as I know `Telescope project` is used for the `telescope-project.nvim` whereas `Telescope projects` is used for `ahmedkhalf/project.nvim` that Lazyvim uses. Are you sure there's nothing wrong with your configuration? Did you import the spec or did you make one yourself and maybe misspelled something?

FBamx:
> As far as I know `Telescope project` is used for the `telescope-project.nvim` whereas `Telescope projects` is used for `ahmedkhalf/project.nvim` that Lazyvim uses. Are you sure there's nothing wrong with your configuration? Did you import the spec or did you make one yourself and maybe misspelled something?

I checked and found you were right, thanks for the reminder

PR #752, opened 2023-05-12
Latest copilot-cmp requires passing opts to _on_insert_enter

ryansch:
Upstream changes happened here: https://github.com/zbirenbaum/copilot-cmp/pull/67


amaanq:
I don't think that's needed though

folke:
thanks!

PR #779, opened 2023-05-19
Fix comments on autocmds and keymaps loading

jpaquim:
Very minor, but it seems that `autocmds.lua` and `keymaps.lua` are both loaded by the `config/init.lua` module, maybe the comments are out of date?


folke:
Thanks

PR #781, opened 2023-05-19
Fix the mason-nvim-dap setting name

jyuan0:
The `mason-nvim-dap` setting name is `automatic_installation`, rather than `automatic_setup`.

https://github.com/jay-babu/mason-nvim-dap.nvim#default-configuration


folke:
Thanks!

PR #784, opened 2023-05-20
feat(lsp): add NullLsInfo binding

utkarshgupta137:
I know it is not possible to include every key binds that every user wants, but I found no way so far to hook into this function.


folke:
```lua
  {
    "jose-elias-alvarez/null-ls.nvim",
    keys = {
      { "<leader>cn", "<cmd>NullLsInfo<cr>", desc = "NullLs Info" },
    },
  },
```

utkarshgupta137:
Right, I already have that in my config, but this should ideally only be added for buffers with LSP.

folke:
Not every buffer with an lsp attached has null-ls either.

You could add an `init()` with an autocmd listening for `LspAttach` that register a buffer local mapping when null-ls attaches

PR #811, opened 2023-05-24
fix(go): actually set semantic tokens

Jomik:
Just realised that we set the settings in the wrong place, apologies..

### DIFF
--- a/lua/lazyvim/plugins/extras/lang/go.lua
+++ b/lua/lazyvim/plugins/extras/lang/go.lua
@@ -15,7 +15,11 @@ return {
     opts = {
       servers = {
         gopls = {
-          semanticTokens = true,
+          settings = {
+            gopls = {
+              semanticTokens = true,
+            },
+          },
         },
       },
     },


folke:
thanks!

kcaashish:
I don't think this is enough for getting semantic token highlighting in go. Gopls needs some workaround.

Jomik:
> I don't think this is enough for getting semantic token highlighting in go. Gopls needs some workaround.

Really? Can you elaborate? Did you get it working some other way? I am actually a bit unsure what this is supposed to show, added it based on @amaanq's request : https://github.com/LazyVim/LazyVim/pull/795#issuecomment-1558053279

It is listed here https://github.com/golang/tools/blob/master/gopls/doc/settings.md#semantictokens-bool
https://github.com/golang/tools/blob/master/gopls/doc/semantictokens.md

Just realised that it is experimental, should we just remove it from the settings here and leave it to the user @folke ?
Though, if it requires some involved workaround, I would probably do it here..

kcaashish:
I have a working setup, let me send a pr with it. It can be kept until Gopls starts supporting it out of the box.

PR #815, opened 2023-05-24
fix(go): add workaround to get semantic token highlighting

kcaashish:
Add a work around based on this comment: https://github.com/golang/go/issues/54531#issuecomment-1464982242
Related to #811

### DIFF
--- a/lua/lazyvim/plugins/extras/lang/go.lua
+++ b/lua/lazyvim/plugins/extras/lang/go.lua
@@ -22,6 +22,28 @@ return {
           },
         },
       },
+      setup = {
+        gopls = function()
+          -- workaround for gopls not supporting semantictokensprovider
+          -- https://github.com/golang/go/issues/54531#issuecomment-1464982242
+          require("lazyvim.util").on_attach(function(client, _)
+            if client.name == "gopls" then
+              if not client.server_capabilities.semanticTokensProvider then
+                local semantic = client.config.capabilities.textDocument.semanticTokens
+                client.server_capabilities.semanticTokensProvider = {
+                  full = true,
+                  legend = {
+                    tokenTypes = semantic.tokenTypes,
+                    tokenModifiers = semantic.tokenModifiers,
+                  },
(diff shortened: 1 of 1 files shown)


lua/lazyvim/plugins/extras/lang/go.lua
```diff
@@ -14,13 +14,31 @@ return {
     "neovim/nvim-lspconfig",
     opts = {
       servers = {
-        gopls = {
-          settings = {
+        gopls = function(_, opts)
```
folke, line 17: This should always be a table. Did you mean to put it under `opts.setup`?

kcaashish: Oh, yes! It should be insider setup. Fixed that.

folke:
thanks!

fredrikaverpil:
I was hoping to enable inline SQL but it seems I have to set `semanticTokens = false` in the gopls settings to make that work. 😞 

```lua
  {
    "hexdigest/go-enhanced-treesitter.nvim",
    dependencies = {
      "derekstride/tree-sitter-sql",
    },
    build = ":TSInstall go sql",
    ft = "go",
  },
```

hojjatabdollahi:
@fredrikaverpil did you ever figure out how to get the inline sql syntax highlighting to work? I tried to look at your dot files by they're too complex for my simple Go dev setup.

fredrikaverpil:
@hojjatabdollahi yes, I got it working. You need to:

1. set gopls to not use semantic tokens: https://github.com/fredrikaverpil/dotfiles/blob/f55cfa07b53efed740b27ace830a1edf439bfc9f/nvim-fredrik/lua/fredrik/plugins/lang/go.lua#L221
2. add injections: https://github.com/fredrikaverpil/dotfiles/blob/f55cfa07b53efed740b27ace830a1edf439bfc9f/nvim-fredrik/after/queries/go/injections.scm#L10-L17

PR #859, opened 2023-05-30
feat(dap): add extra for dap js

ofrades:


folke:
Did you check the existing dap integration?
https://github.com/LazyVim/LazyVim/blob/main/lua/lazyvim/plugins/extras/lang/typescript.lua#L66

folke:
nvim-dap-vscode-js is actually no longer needed with the latest microsoft-jsdebug. IT's now a full-blown dap server.

ofrades:
Ok so I will close this PR and if I find a nice way to add the test config to the existing integration I will then open a new PR. Thanks

PR #870, opened 2023-06-01
fix: fix(indent-blankline): don't load indent-blankline for nvim-notify

rofrol:
fixes https://github.com/LazyVim/LazyVim/commit/a94ef5953a6f14731d9aea447e769c280db140ab


folke:
Right :) But you can keep it in indentscope as well. That's actually the one I wanted to fix. The commit message was incorrect :)

PR #881, opened 2023-06-04
Add ruby as extras for languages

kevinrobayna:
Following the example of go I wanted to go ahead and try to add the setup for ruby.

This is the configuration I currently use locally so it does work pretty well.

### DIFF
--- a/lua/lazyvim/plugins/extras/lang/ruby.lua
+++ b/lua/lazyvim/plugins/extras/lang/ruby.lua
@@ -4,6 +4,13 @@ return {
     opts = function(_, opts)
       vim.list_extend(opts.ensure_installed, {
         "ruby",
+      })
+    end,
+  },
+  {
+    "williamboman/mason.nvim",
+    opts = function(_, opts)
+      vim.list_extend(opts.ensure_installed, {
         "solargraph",
       })
     end,
--- a/lua/lazyvim/plugins/extras/lang/ruby.lua
+++ b/lua/lazyvim/plugins/extras/lang/ruby.lua
@@ -4,13 +4,6 @@ return {
     opts = function(_, opts)
       vim.list_extend(opts.ensure_installed, {
         "ruby",
-      })
-    end,
-  },
-  {
-    "williamboman/mason.nvim",
-    opts = function(_, opts)
-      vim.list_extend(opts.ensure_installed, {
         "solargraph",
       })
     end,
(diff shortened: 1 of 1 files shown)


kevinrobayna:
👋🏻 @folke would you be interested in this pr? 

Let me know if you'd be interested in having several languages in the extras folder, I guess we don't really want all existing languages in the extras folders as otherwise it will be a massive folder 😅

cabaleirog:
I am unsure if this might be a misconfiguration on my side or simple lack of understanding, as I just began using Vim, but l noticed that my custom ruby dap configuration no longer appears when I try to run the debugger, and it shows only the default configuration.
After running a git bisect, I found this was the commit which causes that to happen, and in particular, line 32 seems to be the reason (`require("dap-ruby").setup()`), removing that single line restores the custom scenarios in my case at least, is this something that should be done here instead of in someone's personal config?
cc: @kevinrobayna @folke 

In case I am just doing something wrong, this is my config.

```lua
local dap = require('dap')

dap.adapters.ruby = function(callback, config)
  local function str_to_list (inputstr)
    local t={}
    for str in string.gmatch(inputstr, "[^%s]+") do
      table.insert(t, str)
    end
    return t
  end
  callback {
    type = "server",
    host = "127.0.0.1",
    port = "${port}",
    executable = {
      command = "bundle",
      args = { "exec", "rdbg", "-n", "--open", "--port", "${port}",
        "-c", "--", "bundle", "exec", config.command, unpack(str_to_list(config.script))
      },
    },
  }
end

dap.configurations.ruby = {
  {
    type = "ruby",
    name = "Debug current file",
    request = "attach",
    localfs = true,
    command = "ruby",
    script = "${file}",
  },
  {
    type = "ruby",
    name = "Run current spec file",
    request = "attach",
    localfs = true,
    command = "rspec",
    script = "${file}",
  },
  {
    type = "ruby",
    name = "Run all spec files",
    request = "attach",
    localfs = true,
    command = "rspec",
    script = ".",
  },
  {
    type = "ruby",
    name = "Test - Redacted",
    request = "attach",
    localfs = true,
    command = "/opt/redacted/bin/redacted",
    script = "redacted"
  },
  {
    type = "ruby",
    name = "Redacted - Run application",
    request = "attach",
    localfs = true,
    command = "/opt/redacted/bin/redacted",
    script = "--fg"
  },
}

return {
  {
    'mfussenegger/nvim-dap',
    opt = {}
  },
  {
    'suketa/nvim-dap-ruby',
    opt = {}
  }
}
```

PR #897, opened 2023-06-07
fix(format): eslint should respect autoformat

mikeduminy:
The eslint extra will autoformat if there are any diagnostics associated with eslint, but it does not currently respect the stateful autoformat setting (toggled via `<leader>uf`).


folke:
thanks!

PR #944, opened 2023-06-19
feat(treesitter): allow run TSUpdateSync from command line

robsonpeixoto:
Use a script to keep the neovim updated is really useful, but is not
possible by default because TSUpdateSync is not on `cmd` commands.

```sh
nvim --headless "+Lazy! sync" "+TSUpdateSync" +qa
```


folke:
thank you!

PR #964, opened 2023-06-23
feat(mini.files): "g." to toggle hidden files

lkhphuc:
Very useful features, copy straight from the doc of mini.files

### DIFF
--- a/lua/lazyvim/plugins/extras/editor/mini-files.lua
+++ b/lua/lazyvim/plugins/extras/editor/mini-files.lua
@@ -26,4 +26,26 @@ return {
       desc = "Open mini.files (cwd)",
     },
   },
+  config = function(_, opts)
+    require("mini.files").setup(opts)
+
+    local show_dotfiles = true
+    local filter_show = function(fs_entry) return true end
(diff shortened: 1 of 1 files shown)


lua/lazyvim/plugins/extras/editor/mini-files.lua
```diff
@@ -26,4 +26,24 @@ return {
       desc = "Open mini.files (cwd)",
     },
   },
+  init = function()
```
folke, line 29: `init` is not ideal for this, since this will always be loaded.

Can you change this to the config function below?

```lua
  config = function(_, opts)
    require("mini.files").setup(opts)
    -- add that toggle code here 
  end,
```

folke (changes requested):

lua/lazyvim/plugins/extras/editor/mini-files.lua
lkhphuc: Done. I found that the setup call is not needed, but added anyway.

folke:
Thank you!

PR #977, opened 2023-06-25
fix(keymaps): remove redundant ':' on lazy keymap

jyuan0:
Super minor. Just removing `:` in the Lazy keymap since it has `<cmd>`:
```
map("n", "<leader>l", "<cmd>:Lazy<cr>", { desc = "Lazy" })
```
I'm assuming removing `:` is preferable.


folke:
Thanks!

PR #993, opened 2023-06-27
Add Elixir support and config

georgeguimaraes:
Hi folke, adding support for Elixir in extras.
Not sure if you'd want this but here it is :)

Thanks for all the work in the neovim community.

### DIFF
--- a/lua/lazyvim/plugins/extras/lang/elixir.lua
+++ b/lua/lazyvim/plugins/extras/lang/elixir.lua
@@ -1,0 +1,32 @@
+return {
+  {
+    "nvim-treesitter/nvim-treesitter",
+    opts = function(_, opts)
+      vim.list_extend(opts.ensure_installed, {
+        "elixir",
+        "heex",
+        "eex",
+      })
+    end,
+  },
+  {
+    "williamboman/mason.nvim",
+    opts = function(_, opts)
+      vim.list_extend(opts.ensure_installed, {
+        "elixir-ls",
+      })
+    end,
+  },
+  {
+    "nvim-neotest/neotest",
+    optional = true,
+    dependencies = {
+      "jfpedroza/neotest-elixir",
+    },
+    opts = {
+      adapters = {
+        ["neotest-elixir"] = {},
+      },
+    },
+  },
+}


dvic:
Thanks for this! I just want to mention also https://github.com/elixir-tools/elixir-tools.nvim, which I'm currently using and it's working great 👍 The difference is that the elixir-ls is installed on the fly depending on the elixir version. Also, it includes the credo language server. I can share my config if people are interested.

robsonpeixoto:
Please @dvic , share your config ❤️

dvic:
This is my setup:

```lua
return {
  "elixir-tools/elixir-tools.nvim",
  lazy = true,
  ft = { "elixir", "heex", "eelixir" },
  config = function()
    local elixir = require("elixir")
    local elixirls = require("elixir.elixirls")
    local wk = require("which-key")

    elixir.setup({
      elixirls = {
        tag = "v0.15.0",
        enabled = true,
        settings = elixirls.settings({
          dialyzerEnabled = true,
          enableTestLenses = false,
        }),
        on_attach = require("lazyvim.util").on_attach(function(_, buffer)
          wk.register({
            e = {
              name = "elixir",
              p = { ":ElixirToPipe<cr>", "Convert to pipe", noremap = true, buffer = buffer },
              P = { ":ElixirFromPipe<cr>", "Convert from pipe", noremap = true, buffer = buffer },
              m = { ":ElixirExpandMacro<cr>", "Expand macro", noremap = true, buffer = buffer },
            },
          }, { prefix = "<leader>c" })
        end),
      },
    })
  end,
  dependencies = {
    "folke/which-key.nvim",
    "nvim-lua/plenary.nvim",
  },
```

As you can see, I also configured some key mappings for the pipe conversion commands from `elixir-tools`, which are very handy :)

georgeguimaraes:
Interesting. I didn't know about `elixir-tools.nvim` before. Although, since we have elixir-ls in Mason and credo in null-ls, it doesn't seem to be very useful for my workflow. I can see it sets some integration with mix and some keymaps (It could be a much small nvim plugin just for that), but it seems to be overlapping with Mason too much.

folke:
thanks!

PR #1012, opened 2023-06-30
feat(lang): Add Rust config

amaanq:


folke:
thanks!

dvic:
@amaanq Thanks for this! Quick question: when you goto definition to the standard library, do you also get this problem? https://github.com/simrat39/rust-tools.nvim/issues/394  

It seems like the issue here is that a new instance of the LSP is boot up and that there are duplicate definitions.

amaanq:
> @amaanq Thanks for this! Quick question: when you goto definition to the standard library, do you also get this problem? https://github.com/simrat39/rust-tools.nvim/issues/394  
> 
> It seems like the issue here is that a new instance of the LSP is boot up and that there are duplicate definitions.

I do not, just tried going to definition on `Result` and I couldn't replicate that

lua/lazyvim/plugins/extras/lang/rust.lua
```diff
@@ -0,0 +1,132 @@
...
+          -- rust tools configuration for debugging support
+          local codelldb = mason_registry.get_package("codelldb")
+          local extension_path = codelldb:get_install_path() .. "/extension/"
+          local codelldb_path = extension_path .. "adapter/codelldb"
+          local liblldb_path = vim.fn.has("mac") == 1 and extension_path .. "lldb/lib/liblldb.dylib"
+            or extension_path .. "lldb/lib/liblldb.so"
+          local rust_tools_opts = vim.tbl_deep_extend("force", opts, {
+            dap = {
+              adapter = require("rust-tools.dap").get_codelldb_adapter(codelldb_path, liblldb_path),
+            },
```
shanesveller, line 72: Pretty novice with Lua here, and I had trouble figuring out how to successfully run with a fork/branch of LazyVim, in order to test my own attempts at addressing the concern below. Pointers welcome!

This code has an implicit hard requirement on `mason`, which I keep disabled because I manage CLIs separately. Any chance this module's implementation could be refactored slightly to fail gracefully is the `mason-registry` require fails and still set up any of the options that can be determined without it? It looks like it would just be the DAP piece that depends on it?

lua/lazyvim/plugins/extras/lang/rust.lua
```diff
@@ -0,0 +1,132 @@
...
+                    extraArgs = { "--no-deps" },
+                  },
+                  procMacro = {
+                    enable = true,
+                    ignored = {
+                      ["async-trait"] = { "async_trait" },
+                      ["napi-derive"] = { "napi" },
+                      ["async-recursion"] = { "async_recursion" },
+                    },
+                  },
```
shanesveller, line 105: There's some subjective choices included here, and I'm not clear on if/how I could inject different choices if one of my projects needs it. Is that possible, or would it require further code changes in this module? My beginner understanding for `vim.tbl_deep_extend`'s behavior on conflicts is last-one-wins when the first argument is `force`, is that accurate?

lua/lazyvim/plugins/extras/lang/rust.lua
folke: I refactored the whole rust extra. It can now work without mason and you can easily override the rust-tools options.

lua/lazyvim/plugins/extras/lang/rust.lua
folke: This is also easily configurable now.

Check the updated docs here: https://www.lazyvim.org/plugins/extras/lang.rust

shanesveller: Great improvements all around, thank you so much ❤️

PR #1015, opened 2023-06-30
Add rust improvements

loqusion:
- Fixes indentation
- Moves lspconfig opts to correct location in rust-tools config
- Adds optional neotest-rust
- Allows user to extend rust-tools with `opts` field, without calling rust-tools's `setup()` more than once


loqusion:
135c342 allows the user to add a `rust-tools.nvim` spec to their config with an `opts` field that is merged into the options passed to rust-tools's `setup()` function. Setting `config = false` doesn't prevent `setup()` from being called multiple times; thus, an empty function is used.

Is there a better way to do this?

folke:
Sorry, I just merged another PR that fixed the indent. Can you rebase?

PR #1024, opened 2023-06-30
feat(go): add fine-tuned LSP options, add some null-ls sources

amaanq:
The null-ls sources *could* be controversial, but they provide niceties most people would appreciate I'd think

### DIFF
--- a/lua/lazyvim/plugins/extras/lang/go.lua
+++ b/lua/lazyvim/plugins/extras/lang/go.lua
@@ -23,8 +23,8 @@ return {
         },
       },
       setup = {
-        gopls = function()
-          -- workaround for gopls not supporting semantictokensprovider
+        gopls = function(_, opts)
+          -- workaround for gopls not supporting semanticTokensProvider
           -- https://github.com/golang/go/issues/54531#issuecomment-1464982242
           require("lazyvim.util").on_attach(function(client, _)
             if client.name == "gopls" then
@@ -42,9 +42,60 @@ return {
             end
           end)
           -- end workaround
+          opts.settings = {
+            gopls = {
+              gofumpt = true,
+              codelenses = {
+                gc_details = false,
+                generate = true,
+                regenerate_cgo = true,
+                run_govulncheck = true,
+                test = true,
+                tidy = true,
+                upgrade_dependency = true,
+                vendor = true,
+              },
+              hints = {
+                assignVariableTypes = true,
+                compositeLiteralFields = true,
+                compositeLiteralTypes = true,
+                constantValues = true,
+                functionTypeParameters = true,
+                parameterNames = true,
+                rangeVariableTypes = true,
+              },
+              analyses = {
+                fieldalignment = true,
+                nilness = true,
+                unusedparams = true,
+                unusedwrite = true,
+                useany = true,
+              },
+              usePlaceholders = true,
+              completeUnimported = true,
+              staticcheck = true,
+              directoryFilters = { "-.git", "-.vscode", "-.idea", "-.vscode-test", "-node_modules" },
+              semanticTokens = true,
+            },
+          }
         end,
       },
     },
+  },
+  -- Ensure Go tools are installed
+  {
+    "jose-elias-alvarez/null-ls.nvim",
+    opts = function(_, opts)
+      if type(opts.sources) == "table" then
+        local nls = require("null-ls")
+        vim.list_extend(opts.sources, {
+          nls.builtins.code_actions.gomodifytags,
+          nls.builtins.code_actions.impl,
+          nls.builtins.formatting.gofumpt,
+          nls.builtins.formatting.goimports_reviser,
+        })
+      end
+    end,
   },
   {
     "mfussenegger/nvim-dap",


folke:
thanks! Let's see if anyone has any objections :)

aofei:
Sorry for commenting on a closed PR. This PR has caused a lot of surprise. My nvim is now flooded with warnings, including ones that should be ignored by using thins like `//lint:ignore ...`. IMHO, as a core plugin, LazyVim's configuration of LSP should be kept as minimal as possible. Such LSP options as introduced in this PR should be left to users to manually enable according to their own preferences, just like before (and also like the other `lazyvim.plugins.extras.lang.*`). No offense, but enabling so many options by default feels like overkill to me. Defaulting to `false` is always a good thing, it doesn't make people think "Is my editor broken or what?"

Just my personal opinion, though. Of course, they can always be disabled.

folke:
I don't agree with that. It's better for LazyVim to enable sensible defaults especially in terms of diagnostics and such.

You can easily disable some or all in your own config. Check  https://www.lazyvim.org/plugins/extras/lang.go

aofei:
Drawing the line for "sensible defaults" is always tough, so sticking to the LSP's defaults is a good idea, just like the defaults provided by vscode-go for gopls: https://github.com/golang/vscode-go/wiki/settings#settings-for-gopls. Since vscode-go is maintained by the Go team, so why not use it as a reference in this case?

amaanq:
Well, semantic tokens are also off by default - that is explicitly enabled here despite it being not perfect

I'm not sure which diagnostics are excessive to you - I find many of them helpful, like reorganizing struct fields to reduce its size in memory, though it'd be nice to know which you find excessive

aofei:
Hi @amaanq,

Don't get me wrong, I've enabled some of the options you provided myself, so I'm not denying your interpretation of "sensible defaults".

What I'm trying to say is that, in my opinion, most of these options come down to personal preference. For instance, as you mentioned, I understand that you find `fieldalignment` and `unusedparams` useful, but to me, they're mostly just annoying. It's important to acknowledge that not all projects people work on are written according to their own personal taste. Enabling such "harsh" options by default means that most of the time people are flooded with all kinds of warnings. Well, that's not very good for mental health, is it? I believe vscode-go leaves them `false` by default for a reason. I hope you understand my perspective, and please know that I don't mean any offense.

aofei:
Given that it's hard to draw the line for "sensible defaults", I suggest that we refer to vscode-go's defaults for gopls to fine-tune these options: https://github.com/golang/vscode-go/wiki/settings#settings-for-gopls.

Let's just leave the line drawing thing to the Go team, as they have likely put sufficient effort into evaluating whether these options should be enabled by default when they were added to gopls.

folke:
It's not because it's vscode that the defaults make sense.
As mentioned before, you can just change the settings you don't like in your personal config

aofei:
> It's not because it's vscode that the defaults make sense.
> As mentioned before, you can just change the settings you don't like in your personal config

As I said in my first comment https://github.com/LazyVim/LazyVim/pull/1024#issuecomment-1619824440, yes, I know I can disable them. All I'm doing here at this moment is expressing an opinion.

I'm not saying that vscode-go's defaults are the best, even though it's maintained by the Go team. I'm just saying that people debated a lot about the defaults for these options in the Go issue tracker before they were finalized. Disregarding them doesn't feel right.

Well, I thought I was doing this for the good of the plugin. I'm just a user anyway, and this is a great plugin, so I'll eventually live with what this plugin has to offer.

amaanq:
I take no offense - if I come across as harsh then I apologize. 

Does commenting out the staticcheck line remove what you'd find annoying? that's the only one that could be more on the controversial to add compared to the other opotions.

aofei:
> Does commenting out the staticcheck line remove what you'd find annoying?

I've always had `staticcheck` enabled myself.

> that's the only one that could be more on the controversial end compared to the other options.

That's not the point. It doesn't make much sense to get into an argument about drawing the line for the defaults of LSP options. I think I have made my opinion clear in my previous comments. I tried it, and the author objected, so let it be as it is. These options are configurable and harmless anyway. 👌

---

This is my current setup for gopls, as you can see, it's nothing special but just personal preference (I'm not the kind of person who likes a screen full of hints, they don't do anything but distract me):

<details><summary>Click to expand</summary>

```lua
gopls = {
  gofumpt = true,
  codelenses = {
    run_govulncheck = false,
    test = false,
    tidy = true,
    upgrade_dependency = true,
    vendor = true,
  },
  hints = {
    assignVariableTypes = false,
    compositeLiteralFields = false,
    compositeLiteralTypes = false,
    constantValues = false,
    functionTypeParameters = false,
    parameterNames = false,
    rangeVariableTypes = false,
  },
  analyses = {
    fieldalignment = false,
    nilness = true,
    unusedparams = false,
    unusedwrite = true,
    useany = true,
  },
  usePlaceholders = true,
  completeUnimported = true,
  staticcheck = true,
  semanticTokens = true,
}
```
</details>

fredrikaverpil:
@aofei quick question, I was looking at the [gopls settings docs](https://github.com/golang/tools/blob/master/gopls/doc/settings.md) and I can't see a `completeUnimported` setting (which was part of the config [you provided](https://github.com/LazyVim/LazyVim/pull/1024#issuecomment-1620051172)). Do you know what this is?

aofei:
@fredrikaverpil 

It's [here](https://github.com/golang/tools/blob/451716b51585d61afb394f6d6f0d604192705aa8/gopls/internal/lsp/source/options.go#L611-L613), and it's [enabled by default](https://github.com/golang/tools/blob/451716b51585d61afb394f6d6f0d604192705aa8/gopls/internal/lsp/source/options.go#L174) now, so it's ok to remove it from your settings.

PR #1025, opened 2023-06-30
feat(lang): add C/C++ support

amaanq:
Not sure if the file should be called c or clangd (lsp name, supports c/c++/objc/objc++)

### DIFF
--- a/lua/lazyvim/plugins/extras/lang/clangd.lua
+++ b/lua/lazyvim/plugins/extras/lang/clangd.lua
@@ -1,0 +1,152 @@
+return {
+
+  -- Add C/C++ to treesitter
+  {
+    "nvim-treesitter/nvim-treesitter",
+    opts = function(_, opts)
+      if type(opts.ensure_installed) == "table" then
+        vim.list_extend(opts.ensure_installed, { "c", "cpp" })
+      end
+    end,
+  },
+
+  {
+    "p00f/clangd_extensions.nvim",
+    lazy = true,
+    config = function() end,
+    opts = {
+      extensions = {
+        inlay_hints = {
+          inline = false,
+        },
+        ast = {
+          --These require codicons (https://github.com/microsoft/vscode-codicons)
+          role_icons = {
+            type = "",
+            declaration = "",
+            expression = "",
+            specifier = "",
+            statement = "",
+            ["template argument"] = "",
+          },
+          kind_icons = {
+            Compound = "",
+            Recovery = "",
+            TranslationUnit = "",
+            PackExpansion = "",
+            TemplateTypeParm = "",
+            TemplateTemplateParm = "",
+            TemplateParamObject = "",
+          },
+        },
+      },
+    },
+  },
+
+  -- Correctly setup lspconfig for clangd 🚀
+  {
+    "neovim/nvim-lspconfig",
+    opts = {
+      servers = {
+        -- Ensure mason installs the server
+        clangd = {
+          keys = {
+            { "<leader>cR", "<cmd>ClangdSwitchSourceHeader<cr>", desc = "Switch Source/Header (C/C++)" },
+          },
+          root_dir = function(...)
+            -- using a root .clang-format or .clang-tidy file messes up projects, so remove them
+            return require("lspconfig.util").root_pattern(
+              "compile_commands.json",
+              "compile_flags.txt",
+              "configure.ac",
+              ".git"
+            )(...)
+          end,
+          capabilities = {
+            offsetEncoding = { "utf-16" },
+          },
+          cmd = {
+            "clangd",
+            "--background-index",
+            "--clang-tidy",
+            "--header-insertion=iwyu",
+            "--completion-style=detailed",
+            "--function-arg-placeholders",
+            "--fallback-style=llvm",
+          },
+          init_options = {
+            usePlaceholders = true,
+            completeUnimported = true,
+            clangdFileStatus = true,
+          },
+        },
+      },
+      setup = {
+        clangd = function(_, opts)
+          local clangd_ext_opts = require("lazyvim.util").opts("clangd_extensions.nvim")
+          require("clangd_extensions").setup(vim.tbl_deep_extend("force", clangd_ext_opts or {}, { server = opts }))
+          return true
+        end,
+      },
+    },
+  },
+
+  {
+    "nvim-cmp",
+    opts = function(_, opts)
+      table.insert(opts.sorting.comparators, 1, require("clangd_extensions.cmp_scores"))
+    end,
+  },
+
+  {
+    "mfussenegger/nvim-dap",
+    optional = true,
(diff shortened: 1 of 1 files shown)


Meigumikato:
Maybe should add a keymap for ClangdSwitchSourceHeader ?

amaanq:
> Maybe should add a keymap for ClangdSwitchSourceHeader ?

I set it to `<leader>cR`, better suggestions welcome

lua/lazyvim/plugins/extras/lang/c.lua
```diff
@@ -0,0 +1,97 @@
...
+        vim.list_extend(opts.ensure_installed, { "c", "cpp" })
+      end
+    end,
+  },
+
+  -- Correctly setup lspconfig for clangd 🚀
+  {
+    "neovim/nvim-lspconfig",
+    dependencies = {
+      "p00f/clangd_extensions.nvim",
```
folke, line 17: Can you try another structure to make sure clangd_extensions is configurable to the user?

I think it should work :)

Move this as a seperate spec to line 12 for example.

There, add the opts as you specify them below. Also add `lazy=true`.

lua/lazyvim/plugins/extras/lang/c.lua
```diff
@@ -0,0 +1,97 @@
...
+            "--completion-style=detailed",
+            "--function-arg-placeholders",
+            "--fallback-style=llvm",
+          }
+          opts.init_options = {
+            usePlaceholders = true,
+            completeUnimported = true,
+            clangdFileStatus = true,
+          }
+          require("clangd_extensions").setup({
```
folke, line 51: so move these options to `opts` on line 12.

Then here, you just do `require("clangd_extensions")`
This will properly load and setup `clangd_extensions` and will use any custom options the user may have set.

folke (changes requested):

folke:
`clangd` is probably a better name here indeed for the extra

lua/lazyvim/plugins/extras/lang/c.lua
amaanq: Hmm, maybe I'm overthinking this but clangd_extensions sets up the clangd lsp (hence the return true), so don't we need to call setup here (and if so - then how can we get the opts that the user has set for clangd inside here?)

folke: Have a spec outside of lspconfig, so not in its dep  like:
```lua
{".../extenstions", opts= {the options you currently pass to seup}, lazy = true}
```

Having that spec, users can easily override its `opts`.

Then where you currently call `extenstions.setup`, just do `require("clangd_extensions")` instead.
By doing this, lazy will see the plugin has not been loaded yet, and will automatically call `require("clangd_extensions").setup(ops)` where opts would be the mergded opts for the plugin.

We should do the same for rust-tools btw. Currently there it's done by using an internal lazy API, but the approach I suggest is much easier.

amaanq: Yeah that makes sense - thanks for the clarification. My only lingering question is how we can set the server key to be the clangd lspconfig opts `server = opts`. It's epecially important that offsetEncoding is utf-16 otherwise the user is bombarded with warning messages about multiple client encodings)

folke: Right! Didn't look at it that well it seems :)

Then the approach from https://github.com/LazyVim/LazyVim/blob/fca09b12e8188fb0a9b7bda040890c096e6cc9be/lua/lazyvim/plugins/extras/lang/rust.lua#L73 should be used instead.

But I'd still move all the opts to the opts of the plugin as I mentioned, but then get it's opts and call setup where you already did it.

folke: And make sure to move it outside of lspconfig deps with lazy=false.
Also add a `config = function()end` to the extensions spec to prevent it being loaded twice

amaanq: done..but should lazy be true or false? (you mentioned true at first, but false here)

folke: Sorry, lazy should be true indeed, to prevent it loading during startup

amaanq: done

lua/lazyvim/plugins/extras/lang/clangd.lua
```diff
@@ -0,0 +1,105 @@
...
+          },
+        },
+      },
+    },
+  },
+
+  -- Correctly setup lspconfig for clangd 🚀
+  {
+    "neovim/nvim-lspconfig",
+    dependencies = {
```
folke, line 48: This dependency should be removed, no longer needed.
Move the `config=...` to the spec at line 16

lua/lazyvim/plugins/extras/lang/clangd.lua
```diff
@@ -0,0 +1,105 @@
...
+            "--fallback-style=llvm",
+          }
+          opts.init_options = {
+            usePlaceholders = true,
+            completeUnimported = true,
+            clangdFileStatus = true,
+          }
+          require("lazyvim.util").on_attach(function(client, buffer)
+            if client.name == "clangd" then
+              -- stylua: ignore
```
folke, line 87: move this to the new lsp keys. See rust for an example

lua/lazyvim/plugins/extras/lang/clangd.lua
```diff
@@ -0,0 +1,105 @@
...
+              "compile_flags.txt",
+              "configure.ac",
+              ".git"
+            )(...)
+          end,
+        },
+      },
+      setup = {
+        clangd = function(_, opts)
+          opts.capabilities.offsetEncoding = { "utf-16" }
```
folke, line 70: move all these options to lspconfig.opts.servers.clangd, so that users can easily override them.

Check the reworked rust extra for an example

folke (changes requested):

lua/lazyvim/plugins/extras/lang/clangd.lua
amaanq: Just saw - thanks for the rework esp w/ the keys!

amaanq:
Should be good now

amaanq:
added dap support

folke:
thanks!

dpetka2001:
I would like to ask a question to whomever can answer. Why do we add `mason` as `optional` in the `nvim-dap`? I understand for `nvim-dap`, since someone might not have a spec configured for `nvim-dap` if he doesn't want to use debugging. But for `mason` is it really necessary here? Since there is already a spec for mason, doesn't it make obsolete for `mason` to be `optional` here? Since `nvim-dap` is optional and `mason` configured as a dependency, I'm pretty sure `codelldb` will not be installed correct?

PR #1030, opened 2023-07-01
feat(plugins): add terraform extra lang

azdanov:


robsonpeixoto:
I'm using exactly this config for.a long time and works like a charm.

folke:
thanks!

Mjb141:
This doesn't work for me @folke - I _always_ get `terraformls: timeout`. Is there something I'm missing? Terraform-ls is installed via homebrew.

folke:
@Mjb141 I have no idea. I don't use terraform.

robsonpeixoto:
@folke what's the best way to change the `mini.comment`'s `custom_commentstring` options?

For some reason `mini.comment` is detecting the `commentstring`, [maybe because has 3 differents options](https://github.com/MichaHoffmann/tree-sitter-hcl/blob/becebebd3509c02e871c99be55556269be1def1b/make_grammar.js#L347-L354), and IMHO a good hack is change the  `custom_commentstring`.

Do you have a suggestion of how to allow the extra plugins extends it?
https://github.com/LazyVim/LazyVim/blob/main/lua/lazyvim/plugins/coding.lua#L138-L148

folke:
Assuming Treesitter doesn't properly get the commentstring, the easiest would be to create an autocmd for the hcl filetype that sets the correct `commentstring` option for the buffer

reegnz:
https://github.com/JoosepAlviste/nvim-ts-context-commentstring/pull/94 is trying to fix the commentstring situation.

reegnz:
My nvim-ts-context-commentstring change got merged, cutting a PR to drop the custom handing in lazyvim.

PR #1031, opened 2023-07-01
feat(lang): add python support

jakub-kozlowicz:


lua/lazyvim/plugins/extras/lang/python.lua
```diff
@@ -0,0 +1,116 @@
...
+  },
+  {
+    "jose-elias-alvarez/null-ls.nvim",
+    opts = function(_, opts)
+      local nls = require("null-ls")
+      opts.sources = opts.sources or {}
+      vim.list_extend(opts.sources, {
+        -- Order of formatters matters. They are used in order of appearance.
+        nls.builtins.formatting.ruff.with({ filetypes = { "python" }, prefer_local = ".venv/bin" }),
+        nls.builtins.formatting.black.with({ filetypes = { "python" }, prefer_local = ".venv/bin" }),
```
amaanq, line 18: I'm not sure why both of these are installed, imo ruff is far superior than black

lua/lazyvim/plugins/extras/lang/python.lua
```diff
@@ -0,0 +1,116 @@
...
+                autoImportCompletions = true,
+                autoSearchPaths = true,
+                diagnosticMode = "workspace",
+                typeCheckingMode = "basic",
+                useLibraryCodeForTypes = true,
+              },
+            },
+          },
+        },
+        ruff_lsp = {
```
amaanq, line 50: If it's added in null-ls, why add it here? (or vice versa)

lua/lazyvim/plugins/extras/lang/python.lua
```diff
@@ -0,0 +1,116 @@
+return {
+  {
+    "nvim-treesitter/nvim-treesitter",
+    opts = function(_, opts)
+      if type(opts.ensure_installed) == "table" then
+        vim.list_extend(opts.ensure_installed, { "ninja", "python", "rst", "toml" })
```
amaanq, line 6: with rust, we should be avoiding duplicates (toml), not saying it's a problem for you to fix but it can be one for users down the road

lua/lazyvim/plugins/extras/lang/python.lua
```diff
@@ -0,0 +1,116 @@
...
+    },
+  },
+  {
+    "mfussenegger/nvim-dap",
+    dependencies = {
+      "mfussenegger/nvim-dap-python",
+      -- stylua: ignore
+      keys = {
+        { "<leader>dPt", function() require('dap-python').test_method() end, desc = "Debug test method" },
+        { "<leader>dPc", function() require('dap-python').test_class() end, desc = "Debug class" },
```
amaanq, line 103:
```suggestion
        { "<leader>dPt", function() require('dap-python').test_method() end, desc = "Debug Method" },
        { "<leader>dPc", function() require('dap-python').test_class() end, desc = "Debug Class" },
```

lua/lazyvim/plugins/extras/lang/python.lua
robsonpeixoto: But ruff does not have all black features. Ruff is a replacement for isort.

lua/lazyvim/plugins/extras/lang/python.lua
robsonpeixoto: LazyVim already handle duplicated config.

But maybe makes sense extract toml to another plugin

lua/lazyvim/plugins/extras/lang/python.lua
jakub-kozlowicz: Ruff is used here instead od `isort`. I do not include diagnostics here from `ruff` because of `ruff_lsp`.

lua/lazyvim/plugins/extras/lang/python.lua
jakub-kozlowicz: `ruff` is used here as diagnostics and formatting with ruff is done via null-ls

lua/lazyvim/plugins/extras/lang/python.lua
amaanq: ah, I thought ruff also had code formatting capabilities as well

lua/lazyvim/plugins/extras/lang/python.lua
amaanq: that's fine then..didn't know that actually

lua/lazyvim/plugins/extras/lang/python.lua
amaanq: null-ls can also use ruff for diagnostic, is there any difference in moving it there?

lua/lazyvim/plugins/extras/lang/python.lua
```diff
@@ -0,0 +1,116 @@
...
+        ["neotest-python"] = {
+          -- Extra arguments for nvim-dap configuration
+          -- See https://github.com/microsoft/debugpy/wiki/Debug-configuration-settings for values
+          dap = {
+            justMyCode = false,
+            console = "integratedTerminal",
+          },
+          args = { "--log-level", "DEBUG", "--quiet" },
+          runner = "pytest",
+          python = ".venv/bin/python",
```
jyuan0, line 91: I'm not sure, but it might be better to take the defaults for these, and by setting `runner` and `python`, I think you lose the "dynamic" fallback behavior.

https://github.com/nvim-neotest/neotest-python:
>         -- Runner to use. Will use pytest if available by default.
>         -- Can be a function to return dynamic value.
>         runner = "pytest",
>         -- Custom python path for the runner.
>         -- Can be a string or a list of strings.
>         -- Can also be a function to return dynamic value.
>         -- If not provided, the path will be inferred by checking for 
>         -- virtual envs in the local directory and for Pipenev/Poetry configs
>         python = ".venv/bin/python",

lua/lazyvim/plugins/extras/lang/python.lua
jyuan0: > `ruff` is used here as diagnostics and formatting with ruff is done via null-ls

It's really `ruff_lsp` doing the diagnostics. They're separate.

lua/lazyvim/plugins/extras/lang/python.lua
```diff
@@ -0,0 +1,116 @@
...
+            python = {
+              analysis = {
+                autoImportCompletions = true,
+                autoSearchPaths = true,
+                diagnosticMode = "workspace",
+                typeCheckingMode = "basic",
+                useLibraryCodeForTypes = true,
+              },
+            },
+          },
```
jyuan0, line 48: Just wondering, what's the benefit of explicitly setting the defaults for these?

jakub-kozlowicz: Not every setting is set by default. The neovim-lspconfig only set these settings:

```lua
{
  python = {
    analysis = {
      autoSearchPaths = true,
      diagnosticMode = "workspace",
      useLibraryCodeForTypes = true
    }
  }
}
```

I often set some values explicitly because of better control. If you thinking it would be better to leave the decision to lower level package I will change this. I am new to neovim ecosystem and I would really appreciate the feedback.

lua/lazyvim/plugins/extras/lang/python.lua
jakub-kozlowicz: Yea you are right here. I changed the code to use default options to the neotest-python adapter.

lua/lazyvim/plugins/extras/lang/python.lua
jyuan0: > Not every setting is set by default. The neovim-lspconfig only set these settings:

I think the rest are the Pyright server defaults, just not set by nvim-lspconfig.

> I often set some values explicitly because of better control. If you thinking it would be better to leave the decision to lower level package I will change this.

I think this is fine either way. I'm just wary of explicitly setting defaults in general because I think it's a double-edged sword. Although it increases visibility and control, it can also obscure upstream default changes.

lua/lazyvim/plugins/extras/lang/python.lua
```diff
@@ -0,0 +1,110 @@
...
+          -- runner = "pytest",
+          -- python = ".venv/bin/python",
+        },
+      },
+    },
+  },
+  {
+    "mfussenegger/nvim-dap",
+    dependencies = {
+      "mfussenegger/nvim-dap-python",
```
jyuan0, line 93: Just a note: things might be different with this setup, but when I installed `nvim-dap-python` previously, the "Attach remote" configuration didn't work because the `DAP` adapter added by `mason-nvim-dap` doesn't check the request type.

lua/lazyvim/plugins/extras/lang/python.lua
jakub-kozlowicz: > null-ls can also use ruff for diagnostic, is there any difference in moving it there?

`null-ls` cannot use `ruff` as `code_actions`. `ruff_lsp` solves this issue and therefore `ruff_lsp` is used as diagnostics and code_actions server and formatting with `ruff` is done via `null-ls`

laermannjan (commented):
Thanks for submitting this! 

As the python ecosystem offers a bunch of options to choose from for your linter, formatter, type-checker, venv manager, even language-server, I would argue these should not all be set in a singe `extra` in LazyVim. 

I think it makes sense to have treesitter, neotest, dap and potentially pyright setup in here.
But ruff and black are not universally used in all python projects and might actually interfere  with people's needs. 
They could move into the linting, formatting extras, but then again I'm not completely sure whether their setup warrants an "extra" here.

lua/lazyvim/plugins/extras/lang/python.lua
```diff
@@ -0,0 +1,110 @@
+return {
```
laermannjan, line 1: The virtual env doesn't have to live under `.venv/`.
`Pipenv` and `conda` for example put them in designated directories outside of your repo.

I've been using something like this to start my language server with the correct python bin.
```suggestion
local lsp_path = require("lspconfig/util").path

local function is_empty(s)
  return s == nil or s == ""
end

local function get_pipenv_venv_path(workspace)
  workspace = workspace or vim.fn.getcwd()

  -- Find and use virtualenv from pipenv in workspace directory.
  local match = vim.fn.glob(lsp_path.join(workspace, "Pipfile"))
  if not is_empty(match) then
    local venv = vim.fn.trim(vim.fn.system("PIPENV_PIPFILE=" .. match .. " pipenv --venv 2>/dev/null"))
    vim.notify("Found pipenv venv", "debug", { title = "Python" })
    return venv
  end
  return ""
end

local function get_peotry_venv_path(workspace)
  workspace = workspace or vim.fn.getcwd()

  -- Find and use virtualenv via poetry in workspace directory.
  local match = vim.fn.glob(lsp_path.join(workspace, "poetry.lock"))
  if not is_empty(match) then
    local venv = vim.fn.trim(vim.fn.system("poetry env info -p 2>/dev/null"))
    vim.notify("Found poetry venv", "debug", { title = "Python" })
    return venv
  end
  return ""
end

local function get_local_venv_path(workspace)
  local path = lsp_path.join(workspace, ".venv")
  if not is_empty(vim.fn.glob(path)) then
    vim.notify("Found local venv", "debug", { title = "Python" })
    return lsp_path.dirname(path)
  end

  return ""
end

local function get_venv_path(workspace)
  workspace = workspace or vim.fn.getcwd()
  local venv = vim.VIRTUAL_ENV
    or get_pipenv_venv_path(workspace)
    or get_peotry_venv_path(workspace)
    or get_local_venv_path(workspace)

  vim.notify("Using  venv: " .. venv, "debug", { title = "Python" })
  return venv
end

return {
```

Use it later as:
```python
      vim.list_extend(opts.sources, {
        -- Order of formatters matters. They are used in order of appearance.
        nls.builtins.formatting.ruff.with({
          filetypes = { "python" },
          prefer_local = lsp_path.join(get_venv_path(), "bin"),
        }),
        nls.builtins.formatting.black.with({
          filetypes = { "python" },
          prefer_local = lsp_path.join(get_venv_path(), "bin"),
        }),
      })
-- ...
    opts = {
      servers = {
        pyright = {
          before_init = function(_, config)
            local path = lsp_path.join(get_venv_path(), "bin", "python")
            config.settings.python.pythonPath = path
          end,
          settings= {...}

```

lua/lazyvim/plugins/extras/lang/python.lua
fredrikaverpil: Would the entry of ruff here automatically apply any formatting other than the import ordering?

fredrikaverpil:
@jakub-kozlowicz this is a super nice addition. I just would like to add I cannot run the full pytest test suite using the LazyVim-provided `<leader>tT`. Does this work for you?

I had to override this for neotest-python:

```lua
    keys = {
      {
        "<leader>tT",
        ":lua require('neotest').run.run({ suite = true })<CR>",
        desc = "Run all tests (override LazyVim)",
      },
```

lua/lazyvim/plugins/extras/lang/python.lua
```diff
@@ -0,0 +1,110 @@
...
+        -- Order of formatters matters. They are used in order of appearance.
+        nls.builtins.formatting.ruff.with({ filetypes = { "python" }, prefer_local = ".venv/bin" }),
+        nls.builtins.formatting.black.with({ filetypes = { "python" }, prefer_local = ".venv/bin" }),
+      })
+    end,
+    dependencies = {
+      "mason.nvim",
+      opts = function(_, opts)
+        opts.ensure_installed = opts.ensure_installed or {}
+        vim.list_extend(opts.ensure_installed, { "black", "ruff" })
```
fredrikaverpil, line 25: I don't personally load `black` from Mason unless it isn't found on `$PATH` (from my activated `$VIRTUAL_ENV`). 

```lua
formatting.black.with({
  command = function()
    -- if executable is on $PATH, use it
    if vim.fn.executable("black") == 1 then
      return "black"
    end
    -- otherwise, use the executable from Mason
    return mason_registry.get_package("black"):get_install_path()
  end,
}),
```

And I ensure `ruff-lsp` is installed via Mason rather than ensuring that `ruff` is installed. You're also not including `mypy` which is very common, so you're here providing "sane defaults" that might not apply to everyone.

I'm starting to think that maybe as a first step, you should omit ruff and black entirely from this PR and call it a day. It'll be really hard to cater to everyone's need here. I would even go as far as putting in a comment about NOT adding in linters/formatters here.

lua/lazyvim/plugins/extras/lang/python.lua
folke: yes, please don't explicitely set defaults

lua/lazyvim/plugins/extras/lang/python.lua
folke: yep, this is fine. It used to be an issue, but I fixed that in LazyVim

folke (changes requested):
I agree that it's probably best to remove the null-ls setup for black and ruff for now.

ruffls is fine.

jakub-kozlowicz:
> @jakub-kozlowicz this is a super nice addition. I just would like to add I cannot run the full pytest test suite using the LazyVim-provided `<leader>tT`. Does this work for you?
> 
> I had to override this for neotest-python:
> 
> ```lua
>     keys = {
>       {
>         "<leader>tT",
>         ":lua require('neotest').run.run({ suite = true })<CR>",
>         desc = "Run all tests (override LazyVim)",
>       },
> ```

@fredrikaverpil For me the config is working just fine. I can run the whole suite of test using the `<leader>tT`. I do not have to add anything to the `neotest-python` configuration.

lua/lazyvim/plugins/extras/lang/python.lua
jakub-kozlowicz: That's why I used the `prefer_local` option. But for now I removed the `null-ls` package entirely from this extra. There is to many differences between everyones configurations.

lua/lazyvim/plugins/extras/lang/python.lua
jakub-kozlowicz: `pyright` respects the `pyproject.toml` or `pyrightconfig.json` configuration files. You can also set the python path in these files. I do not plan to add automatic detection to this extra as it will complicate a lot.

lua/lazyvim/plugins/extras/lang/python.lua
jakub-kozlowicz: > Would the entry of ruff here automatically apply any formatting other than the import ordering?

No. `ruff` can only organise imports exactly like `isort`. That's why black was also added to this configuration initially.

lua/lazyvim/plugins/extras/lang/python.lua
```diff
@@ -0,0 +1,65 @@
...
+    dependencies = {
+      "mfussenegger/nvim-dap-python",
+      -- stylua: ignore
+      keys = {
+        { "<leader>dPt", function() require('dap-python').test_method() end, desc = "Debug Method" },
+        { "<leader>dPc", function() require('dap-python').test_class() end, desc = "Debug Class" },
+      },
+      config = function()
+        local path = require("mason-registry").get_package("debugpy"):get_install_path()
+        require("dap-python").setup(path .. "/venv/bin/python")
```
fredrikaverpil, line 56: I've tried this setup too, but it doesn't work for me. All I get is that "Python exited with 1".
Does this work for you?

If you get this to work, you might want to have Mason install debugpy too as part of this config.

This kind of setup is a little devious, as you might have Python 3.11 install debugpy with Mason but you're working on a project with activated virtual environment in e.g. 3.7. Then this won't work either as the interpreter versions must match.

I've simply resorted to installing debugpy in each project of mine. Not great, but the only way I got this to work reliably.

fredrikaverpil: Hmm, okay so it might have something to do with that I've got a conflicting setup, [here](https://github.com/fredrikaverpil/dotfiles/blob/main/nvim-lazyvim/lua/plugins/debug.lua#L24-L48) as described [here](https://github.com/mfussenegger/nvim-dap/wiki/Debug-Adapter-installation#python).

jakub-kozlowicz: > Does this work for you?

Yes. The `debugpy` is automatically installed.

fredrikaverpil: I think you can do away with this completely, and instead make sure Mason installs debugpy. Then you'll get debugpy from mason automatically.

robsonpeixoto:
This config is simple and is working very well.

lua/lazyvim/plugins/extras/lang/python.lua
```diff
@@ -0,0 +1,65 @@
...
+        ["neotest-python"] = {
+          -- Here you can specify the settings for the adapter, i.e.
+          -- runner = "pytest",
+          -- python = ".venv/bin/python",
+        },
+      },
+    },
+  },
+  {
+    "mfussenegger/nvim-dap",
```
folke, line 46: Can you add `optional=true` here?

lua/lazyvim/plugins/extras/lang/python.lua
```diff
@@ -0,0 +1,65 @@
...
+      },
+      config = function()
+        local path = require("mason-registry").get_package("debugpy"):get_install_path()
+        require("dap-python").setup(path .. "/venv/bin/python")
+      end,
+    },
+  },
+  {
+    "linux-cultist/venv-selector.nvim",
+    opts = {},
```
folke, line 62: add cmd=VenvSelect

folke (changes requested):

folke:
Merged. Thank you!

PR #1032, opened 2023-07-01
feat(lang): add cmake support

jakub-kozlowicz:


lua/lazyvim/plugins/extras/lang/cmake.lua
```diff
@@ -0,0 +1,39 @@
...
+  {
+    "nvim-treesitter/nvim-treesitter",
+    opts = function(_, opts)
+      if not opts.ensure_installed then
+        opts.ensure_installed = {}
+      elseif opts.ensure_installed == "all" then
+        return
+      end
+
+      vim.list_extend(opts.ensure_installed, { "cmake" })
```
amaanq, line 11:
```suggestion
    if type(opts.ensure_installed) == "table"
      vim.list_extend(opts.ensure_installed, { "cmake" })
    end
```

amaanq:
I don't use CMake much but when I do I am a fan of the utilities https://github.com/Civitasv/cmake-tools.nvim provide, could be nice to add w/ this spec:

```lua
  -- Add CMake Tools
  {
    "Civitasv/cmake-tools.nvim",
    dependencies = { "nvim-lua/plenary.nvim" },
    event = "BufRead",
  }
  ```

**Comment by jakub-kozlowicz** [comment]
> I don't use CMake much but when I do I am a fan of the utilities https://github.com/Civitasv/cmake-tools.nvim provide, could be nice to add w/ this spec:
> 
> ```lua
> -- Add CMake Tools
> {
>   "Civitasv/cmake-tools.nvim",
>   dependencies = { "nvim-lua/plenary.nvim" },
>   event = "BufRead",
> }
> ```

This looks like very nice addition. I have not known about this plugin before :)

File: lua/lazyvim/plugins/extras/lang/cmake.lua [hunk 2]
```diff
@@ -0,0 +1,40 @@
...
+  {
+    "jose-elias-alvarez/null-ls.nvim",
+    opts = function(_, opts)
+      local nls = require("null-ls")
+      opts.sources = opts.sources or {}
+      vim.list_extend(opts.sources, {
+        nls.builtins.diagnostics.cmake_lint,
+      })
+    end,
+    dependencies = {
```
dpetka2001, line 19: I noticed this in your other PRs as well. Out of curiosity, why do you put `mason` as a dependency to `null-ls` and not on its own? In my personal config I've been putting it on its own so far. Is there a particular reason? (Something i should know to also take notice in my configuration perhaps)?

jakub-kozlowicz: The rationale behind this decision is not based on any particular reason. It came from my observation of configurations on GitHub that I found to be a clever approach to specifying the desired configuration. Specifically, the `null-ls` plugin requires the presence of a binary, which in turn relied on the `mason` plugin for installation of that binary.

folke: Yes, please don't add this as a dependency. That's not needed

folke (changes requested):

folke:
thanks!

PR #1048, opened 2023-07-03
feat(catppuccin): enable mason integration

errantepiphany:
Since LazyVim pre-bundles mason.nvim, it is safe (and makes sense) for the mason integration of catppuccin to be enabled. Otherwise, the colors in the mason popup window clash with everything else.


folke:
thanks!

errantepiphany:
Thanks for the merge! :)

PR #1053, opened 2023-07-04
Disable `document_symbols` in Neo-tree

georgeguimaraes:
When we enable document_symbols, the Neo-tree window "jumps" around when opening files and when the Neo-tree window has a lot of expanded directories and files.

I've created a bug in upstream with a config.lua that reproduces the problem:
https://github.com/nvim-neo-tree/neo-tree.nvim/issues/1025


folke:
I'd rather wait til this bug is fixed upstream. 
You can of course make this adjustment in your own config

PR #1061, opened 2023-07-05
fix(autocmds): jump to last loc in buffer

SameAsMuli:
vim.api.nvim_win_set_cursor is expecting a window handle but was being passed the current buffer's handle instead


folke:
thanks! I actually wonder why that did work

SameAsMuli:
No problem! And yeah, not sure - it wasn't working on the couple of environments I investigated it on, which is how I found it in the first place

PR #1076, opened 2023-07-08
decouple mini.ai and which-key.nvim

abeldekat:
At the moment, LazyVim "requires" which-key in mini.ai's config in coding.lua. As a consequence which-key cannot be loaded on demand using a key when mini.ai is also enabled.

I would like to load which-key using a key as I mostly do not need which-keys UI.
As an example:

` {
    "folke/which-key.nvim",
    enabled = true,
    event = function()
      return "RemoteReply"
    end,
    keys = {
      {
        "<leader>j",
        function()
          require("which-key")
        end,
        desc = "Which-key activation",
      },
    },
`

**Note:** I added event "RemoteReply" as a dummy event because I was unable to undo the configured "VeryLazy" event in editor.lua

The changes in this pull request follow the same approach as seen in extra/test and extra/dap: Add which-key to coding.lua using the optional keyword, and provide the options regarding mini.ai, to be applied in which-key's main definition in editor.lua

### DIFF
--- a/lua/lazyvim/plugins/coding.lua
+++ b/lua/lazyvim/plugins/coding.lua
@@ -1,3 +1,5 @@
+local Util = require("lazyvim.util")
+
 return {
 
   -- snippets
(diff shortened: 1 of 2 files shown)


folke:
I've added a new `LazyLoad` event to `lazy.nvim` and used that in LazyVim to only execute the which-key code when which-key actually loads (or immediately when it's already loaded)

abeldekat:
Thanks Folke! The LazyLoad event is a big improvement in keeping mini ai's key descriptions in the plugin config itself. I think the difference in using the "optional" keyword as seen in extras(dap and test) is understandable. These configs merely configure a which-key specific group.

I wonder whether it would be possible to override which-keys "verylazy" event. Adding a key is  of no use when which-key still loads on that event. My "RemoteReply" event is a hack... Maybe something like "event = false", the same way preconfigured keys can be deactivated?

PR #1078, opened 2023-07-08
feat(lang): add docker support

jakub-kozlowicz:


lua/lazyvim/plugins/extras/lang/docker.lua
```diff
@@ -0,0 +1,36 @@
+return {
+  {
+    "nvim-treesitter/nvim-treesitter",
+    opts = function(_, opts)
+      if type(opts.ensure_installed) == "table" then
+        vim.list_extend(opts.ensure_installed, { "dockerfile" })
+      end
```
Jomik, line 7: Shouldn't we use the same pattern as below here? `opts.ensure_installed = opts.ensure_installed or {}` 😄

dpetka2001: I read in some other reply by @amaanq that `treesitter` also accepts `all` to install everything. I believe that is why this check about `ensure_installed` whether it being a table or not is being done here. It's a `treesitter` specific thing and I've seen in other PRs also endorsing this check as far as `ensure_installed` is concerned.

folke:
thanks!

PR #1096, opened 2023-07-11
feat(telescope): allow word and selection searches

SameAsMuli:
Changes the current normal-mode mappings `<leader>sw` / `<leader>sW` to only display matches surrounded by word boundaries. Also adds the equivalent visual-mode mappings to allow searching for the current selection.

Selection searching does not consider word boundaries, so highlighting a single word and searching for it will produce the same results as the word-under-cursor searching did before this change. This allows the user to easily choose whether to include word boundaries in their search.

### DIFF
--- a/lua/lazyvim/plugins/editor.lua
+++ b/lua/lazyvim/plugins/editor.lua
@@ -128,8 +128,10 @@ return {
       { "<leader>sm", "<cmd>Telescope marks<cr>", desc = "Jump to Mark" },
       { "<leader>so", "<cmd>Telescope vim_options<cr>", desc = "Options" },
       { "<leader>sR", "<cmd>Telescope resume<cr>", desc = "Resume" },
-      { "<leader>sw", Util.telescope("grep_string"), desc = "Word (root dir)" },
-      { "<leader>sW", Util.telescope("grep_string", { cwd = false }), desc = "Word (cwd)" },
+      { "<leader>sw", Util.telescope("grep_string", { word_match = "-w" }), desc = "Word (root dir)" },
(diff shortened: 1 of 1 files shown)


fredrikaverpil:
@SameAsMuli Not sure if this is helpful, but I replaced `<leader>/` with `nvim-telescope/telescope-live-grep-args.nvim`, which makes use of ripgrep, so to gain greater control: https://github.com/fredrikaverpil/dotfiles/blob/main/nvim-lazyvim/lua/plugins/telescope.lua#L17

You can then search for `-w something` to search with word boundary.

Just figured I'd mention it 😄

SameAsMuli:
That is helpful, I think adding a search option with more fine-grained control would definitely be worth-while! That wouldn't address the original intention behind this change, though, so I'll leave that for a separate PR.

The purpose of this change was that the current "Search Word" functionality chooses its word based on where the cursor is, i.e. it uses the word boundaries around the cursor to determine the search term. But it doesn't then use those word boundaries as part of the search - which felt wrong to me. So if my cursor is in the word vim, and I use the "Search Word" functionality, it'll currently match "vim", "nvim", "neovim", etc.

This change adds word boundaries to the "Search Word" mappings so that, in the previous example, it will now only match "vim". Whilst I was at it, though, I thought I might as well add visual selection searching, given that Telescope supports it. That also had the added bonus that if someone really _did_ want to search for a word without the boundaries surrounding it, they could still do so by typing `viw<leader>sw`.

These are just handy mappings to allow searches for text immediately around the cursor. As you've suggested, a more configurable search that allows the user to freely type the search term and pass additional arguments to ripgrep would be useful. Although, I believe that would be a complimentary change to this one, rather than a replacement for it.

folke:
thanks!

PR #1115, opened 2023-07-13
feat(lang): add debug adapter plugin for go

fredrikaverpil:
### Why this change?

I'm getting an error with `<leader>td`, when I want to debug a go test. See https://github.com/LazyVim/LazyVim/discussions/1107 for details.

It seems there is no debug adapter defined in the Go extras. I can work around this locally like so:

```lua
return {
  -- extend Go extras setup from lazy.lua, with DAP capability
  {
    "leoluz/nvim-dap-go",
    dependencies = {
      { "mfussenegger/nvim-dap" },
    },
    ft = { "go" },
    config = true,
    keys = {
      -- workaround, as nvim-dap-go does not have a DAP strategy set up for neotest
      -- see https://github.com/nvim-neotest/neotest-go/issues/12
      { "<leader>tg", "<cmd>lua require('dap-go').debug_test()<CR>", desc = "Debug Nearest (Go)" },
    },
  },
}
```

...but I figured it might be better to extend the go extras with this capability, hence this PR.

### What was changed?

- Add `nvim-dap-go` plugin.
- Add keymap override for `<leader>td` in gopls lspconfig.

I can now successfully debug go tests with `<leader>td`.

Screenshot 2023-07-13 at 15 36 59

### Concerns, side-effects, notes etc

Even with the addition of `nvim-dap-go`, you cannot use `<leader>td` or `require("neotest").run.run({strategy = "dap"})` to debug a test, as neotest-go apparently does not implement a DAP strategy. Instead you have to invoke `require('dap-go').debug_test()`. It might be good to mention this in the [extras docs](https://github.com/LazyVim/lazyvim.github.io/blob/main/docs/plugins/extras/lang.go.md?plain=1#L5). Or can we reliably override the command executed by `<leader>td` but only for go buffers, perhaps?

### DIFF
--- a/lua/lazyvim/plugins/extras/lang/go.lua
+++ b/lua/lazyvim/plugins/extras/lang/go.lua
@@ -15,6 +15,10 @@ return {
     opts = {
       servers = {
         gopls = {
+          keys = {
+            -- Workaround for the lack of a DAP strategy in neotest-go: https://github.com/nvim-neotest/neotest-go/issues/12
+            { "<leader>td", "<cmd>lua require('dap-go').debug_test()<CR>", desc = "Debug Nearest (Go)" },
+          },
           settings = {
             gopls = {
               gofumpt = true,
@@ -102,6 +106,10 @@ return {
           opts.ensure_installed = opts.ensure_installed or {}
           table.insert(opts.ensure_installed, "delve")
         end,
+      },
+      {
+        "leoluz/nvim-dap-go",
+        config = true,
       },
     },
   },


fredrikaverpil:
@folke let me know if you want me to squash together the commits into just one commit.

jyuan0:
I thought mason-nvim-dap should have added a Go DAP adapter and configurations automatically without the need for this plugin. Is that not working or are the adapters/configs not working correctly?

fredrikaverpil:
@jyuan0 Hm, not sure. You can see the error I get when I just enable the [go extras](https://www.lazyvim.org/plugins/extras/lang.go) here:https://github.com/LazyVim/LazyVim/discussions/1107

I haven't looked into whether mason-nvim-dap is supposed set both a DAP adapter and a configuration up. Here's the go extras implementation in case you want to take a look: https://github.com/LazyVim/LazyVim/blob/main/lua/lazyvim/plugins/extras/lang/go.lua

dpetka2001:
`mason-nvim-dap` provides configuration for the adapters that are configured in `nvim-dap`. However, this is a case where `neotest-go` does not provide dap configuration support, so that one can use the `strategy=dap` provided by `neotest`. `neotest-go` is a `neotest` adapter not a `nvim-dap` adapter.

dpetka2001:
Or are you talking about `dap-go` and I misunderstood? In this case `mason-nvim-dap` uses the `handlers` field to automatically set up adapters. But the default `handlers` is empty in the extras, so nothing gets configured automatically in LazyVim as far as I understood from the `mason-nvim-dap` README.

jyuan0:
From my understanding of mason-nvim-dap, it should automatically setup a DAP adapter and configs without specifying anything in `handlers` (`handlers` is for customization).

I'm not saying this solves the issue with `neotest`, I just wanted to point out that I think there should already be a Go DAP adapter and config setup (although it might not be working correctly, and the adapter/config provided by `nvim-dap-go` might be better).

dpetka2001:
@jyuan0 Pls see this [issue](https://github.com/jay-babu/mason-nvim-dap.nvim/discussions/68). I think it has to be done via `handlers`.

jyuan0:
> @jyuan0 Pls see this https://github.com/jay-babu/mason-nvim-dap.nvim/discussions/68. I think it has to be done via handlers.

Ah ok, I guess `delve` is special.

jyuan0:
So, from my testing, if I just add the Go extra, delve is automatically installed and mason-nvim-dap adds an DAP adapter and configurations without needing to specify anything in `handlers`. I guess adding `nvim-dap-go` is just about getting the `require('dap-go').debug_test()` functionality.

On another note, the null-ls sources (`gomodifytags`, `impl`, etc.) should probably be added to mason so they get installed automatically.

dpetka2001:
@jyuan0 Thanks for letting us know. I hadn't experimented myself since I don't use Go.

lua/lazyvim/plugins/extras/lang/go.lua
```diff
@@ -15,6 +15,10 @@ return {
     opts = {
       servers = {
         gopls = {
+          keys = {
+            -- Workaround for the lack of a DAP strategy in neotest-go: https://github.com/nvim-neotest/neotest-go/issues/12
+            { "<leader>td", "<cmd>lua require('dap-go').debug_test()<CR>", desc = "Debug Nearest (Go)" },
```
Jomik, line 20: Are we sure this will always overwrite the neotest configuration?
I would suspect that it does not, if you load the go lang extra before the neotest extra?

Also, this will unconditionally bind `<leader>td` even if we do not have the neotest or dap extras.
I think this keybind probably has to be done under an optional neotest dependency, probably under the `nvim-dap` dependency too 😄

Jomik (changes requested):

lua/lazyvim/plugins/extras/lang/go.lua
fredrikaverpil: I'm kind of new to Neovim/LazyVim and can't say I know for sure what's the best way to do this, or what caveats/side-effects there may be. Would be great to have @folke pitch in here (but I see he's on vacation so it can wait until he's back).

But since I think you had an interesting observation, I just now tried to load the go extras _before_ the neotest/dap extras, just to see if the keymap would still be overridden for go tests. I initially did this:

```lua
--  lazy.lua
...

require("lazy").setup({
  spec = {

    -- import any extras modules here
    { import = "lazyvim.plugins.extras.lang.go" },
    { import = "lazyvim.plugins.extras.test.core" },
    { import = "lazyvim.plugins.extras.dap.core" },

    { import = "plugins" },
  },

...
```

I also tried switching them around, and in _all_ cases, I were able to debug a go test with `<leader>td` - which means the keymap seems like it's always overridden regardless of this order.

However, I'm not fully understood with how the Lazy.nvim `dependencies =` works, and perhaps my local config somehow triggers loading of plugins in some order that doesn't show by this rudimental test of mine.

folke: yes, should be fine

folke:
Thanks!

PR #1131, opened 2023-07-14
feat(lsp): add 'reuse_win' for reuse window in lsp definition keymap

leisurelicht:
jump to existing window if buffer is already opened when the builtin.lsp_definitions, builtin.lsp_type_definitions and builtin.lsp_implementations only return one result.

https://github.com/nvim-telescope/telescope.nvim/pull/2454#event-9325221452

### DIFF
--- a/lua/lazyvim/plugins/lsp/keymaps.lua
+++ b/lua/lazyvim/plugins/lsp/keymaps.lua
@@ -5,8 +5,6 @@ M._keys = nil
 
 ---@return (LazyKeys|{has?:string})[]
 function M.get()
-  local telescope = require("telescope.builtin")
-
   local format = function()
     require("lazyvim.plugins.lsp.format").format({ force = true })
   end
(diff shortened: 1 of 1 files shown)


lua/lazyvim/plugins/lsp/keymaps.lua
```diff
@@ -5,6 +5,8 @@ M._keys = nil
 
 ---@return (LazyKeys|{has?:string})[]
 function M.get()
+  local telescope = require("telescope.builtin")
```
folke, line 8: Please inline the require, otherwise Telescope will always be loaded whenever an lsp attaches

folke (changes requested):

lua/lazyvim/plugins/lsp/keymaps.lua
leisurelicht: 👌

lua/lazyvim/plugins/lsp/keymaps.lua
```diff
@@ -16,11 +14,11 @@ function M.get()
     M._keys =  {
       { "<leader>cd", vim.diagnostic.open_float, desc = "Line Diagnostics" },
       { "<leader>cl", "<cmd>LspInfo<cr>", desc = "Lsp Info" },
-      { "gd", function() telescope.lsp_definitions({ reuse_win = true }) end, desc = "Goto Definition", has = "definition" },
+      { "gd", function() require("telescope.builtin").lsp_definitions({ reuse_win = true }) end, desc = "Goto Definition", has = "definition" },
       { "gr", "<cmd>Telescope lsp_references<cr>", desc = "References" },
```
folke, line 18: Does this also wok for lsp references?

leisurelicht: No, lsp references does not support the 'reuse_win' option.

PR #1132, opened 2023-07-14
feat(lang): add java

agriffis:
Main thing here is to use the more powerful https://github.com/mfussenegger/nvim-jdtls instead of the default setup provided by nvim-lspconfig.

Looking for early feedback like "yeah not interested go away"

Otherwise, it works but should do more to merge opts, prolly update some docs, that sort of thing.

### DIFF
--- a/lua/lazyvim/plugins/extras/lang/java.lua
+++ b/lua/lazyvim/plugins/extras/lang/java.lua
@@ -1,0 +1,74 @@
+return {
+  -- Add java to treesitter.
+  {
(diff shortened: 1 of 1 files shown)


folke:
thanks!

appelgriebsch:
Hmmm no test and debugger support yet?!

folke:
Feel free to make a PR! I don't use java myself...

agriffis:
> Hmmm no test and debugger support yet?!

I was hoping someone would follow up. I don't use those parts personally. 😂

PR #1133, opened 2023-07-14
feat(lang): add null-ls tools to mason's ensure_installed

fredrikaverpil:
@jyuan0 [noticed](https://github.com/LazyVim/LazyVim/pull/1115#issuecomment-1634975850) the null-ls sources are not installed by Mason. Figured I'd fix this in a separate PR.

### DIFF
--- a/lua/lazyvim/plugins/extras/lang/go.lua
+++ b/lua/lazyvim/plugins/extras/lang/go.lua
@@ -100,7 +100,7 @@ return {
         "mason.nvim",
         opts = function(_, opts)
           opts.ensure_installed = opts.ensure_installed or {}
-          table.insert(opts.ensure_installed, "delve")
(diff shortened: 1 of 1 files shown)


lua/lazyvim/plugins/extras/lang/go.lua
```diff
@@ -100,7 +100,10 @@ return {
         "mason.nvim",
         opts = function(_, opts)
           opts.ensure_installed = opts.ensure_installed or {}
-          table.insert(opts.ensure_installed, "delve")
+          local tools = { "gomodifytags", "impl", "gofumpt", "goimports-reviser", "delve" }
+          for _, tool in ipairs(tools) do
+            table.insert(opts.ensure_installed, tool)
+          end
```
amaanq, line 106: use `vim.list_extend` instead

fredrikaverpil: Ah yes, fixed. Thanks! 👍

folke:
thanks!

PR #1143, opened 2023-07-15
feat(neo-tree): update to v3.x and change `follow_current_file` to a …

dpetka2001:
…table and enable it

### DIFF
--- a/lua/lazyvim/plugins/editor.lua
+++ b/lua/lazyvim/plugins/editor.lua
@@ -57,16 +57,6 @@ return {
           expander_expanded = "",
           expander_highlight = "NeoTreeExpander",
         },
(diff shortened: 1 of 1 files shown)


Mayrixon:
In the `init` section, the line `vim.g.neo_tree_remove_legacy_commands = 1` could also be removed because Neo-tree drops support for old commands.

Mayrixon (changes requested):
In the init section, the line vim.g.neo_tree_remove_legacy_commands = 1 could also be removed because Neo-tree drops support for old commands.

folke:
I've also removed the legacy commands and nerd font v3 icons

folke:
thanks!

PR #1149, opened 2023-07-16
feat(python): add semantic highlighting

lkhphuc:
While cleaning up my config to use the new extras/lang/python, there is this one plugin that I think might be generally useful to many.

Since no python's lsp support semantic highlighting now or in the near future (pyright definitely won't), this is the best we can have for python.

The default color are hardcoded by semshi, so I either link it to matching semantic @lsp.... highlight group, or just add style and inherit the color defined by colorscheme.

Without semshi:
Screenshot 2023-07-16 at 17 07 38

With semshi:
Screenshot 2023-07-16 at 17 06 46


folke:
I renamed the extra, otherwise it was always loaded when adding the python extra

PR #1156, opened 2023-07-18
feat(lang): add tex support

Mayrixon:
Here it is the support for `bib` and `tex` files mainly based on plugin [vimtex](https://github.com/lervag/vimtex) and LSP server [texlab](https://github.com/latex-lsp/texlab).

Some noticeable points are listed here:
1. Although treesitter provides highlights for `tex` files, vimtex renders better using conceal function. So I disabled the treesitter highlight for `tex` files. It will not affect the text-object function provided by treesitter.
2. For a better rendering result, vimtex requires to set `conceallevel` to 2 as written in line 34. It works fine without touching `<leader>uc`, which is defined in `config/keymaps.lua` as follows.
```lua
local conceallevel = vim.o.conceallevel > 0 and vim.o.conceallevel or 3
map("n", "<leader>uc", function() Util.toggle("conceallevel", false, {0, conceallevel}) end, { desc = "Toggle Conceal" })
```
So I am wondering if it is possible to set the default value of `conceallevel` to 2.

### DIFF
--- a/lua/lazyvim/plugins/extras/lang/tex.lua
+++ b/lua/lazyvim/plugins/extras/lang/tex.lua
@@ -36,6 +36,14 @@ return {
         end,
       })
 
+      vim.api.nvim_create_autocmd({ "FileType" }, {
+        group = vim.api.nvim_create_augroup("lazyvim_vimtex_keymap", { clear = true }),
+        pattern = { "bib", "tex" },
+        callback = function()
+          vim.keymap.set("n", "<LocalLeader>ld", "<plug>(vimtex-doc-package)", { silent = true })
+        end,
+      })
+
       vim.g.vimtex_complete_enabled = 0 -- use texlab for completion
       vim.g.vimtex_mappings_disable = { ["n"] = { "K" } } -- disable `K` as it conflicts with LSP hover
       vim.g.vimtex_quickfix_method = vim.fn.executable("pplatex") == 1 and "pplatex" or "latexlog"
--- a/lua/lazyvim/plugins/extras/lang/tex.lua
+++ b/lua/lazyvim/plugins/extras/lang/tex.lua
@@ -41,15 +41,7 @@ return {
     optional = true,
     opts = {
       servers = {
-        texlab = {
-          settings = {
-            texlab = {
-              chktex = {
-                onOpenAndSave = true,
-              },
-            },
-          },
-        },
+        texlab = {},
       },
     },
   },
--- a/lua/lazyvim/plugins/extras/lang/tex.lua
+++ b/lua/lazyvim/plugins/extras/lang/tex.lua
@@ -26,10 +26,7 @@ return {
   {
     "lervag/vimtex",
     version = "*",
-    ft = { "bib", "tex" },
-    keys = {
-      { "<LocalLeader>ld", "<Cmd>VimtexDocPackage<CR>", desc = "Open Documentation" },
-    },
+    lazy = false, -- lazy-loading will disable inverse search
     config = function()
       vim.opt.conceallevel = 2
       vim.g.vimtex_complete_enabled = 0 -- use texlab for completion
--- a/lua/lazyvim/plugins/extras/lang/tex.lua
+++ b/lua/lazyvim/plugins/extras/lang/tex.lua
@@ -44,7 +44,6 @@ return {
         end,
       })
 
-      vim.g.vimtex_complete_enabled = 0 -- use texlab for completion
       vim.g.vimtex_mappings_disable = { ["n"] = { "K" } } -- disable `K` as it conflicts with LSP hover
       vim.g.vimtex_quickfix_method = vim.fn.executable("pplatex") == 1 and "pplatex" or "latexlog"
     end,
--- a/lua/lazyvim/plugins/extras/lang/tex.lua
+++ b/lua/lazyvim/plugins/extras/lang/tex.lua
@@ -1,9 +1,10 @@
 return {
   {
     "folke/which-key.nvim",
+    optional = true,
     opts = {
       defaults = {
-        ["<LocalLeader>l"] = { name = "+vimtex" },
+        ["<localLeader>l"] = { name = "+vimtex" },
       },
     },
   },
@@ -25,7 +26,6 @@ return {
 
   {
     "lervag/vimtex",
-    version = "*",
     lazy = false, -- lazy-loading will disable inverse search
     config = function()
       vim.api.nvim_create_autocmd({ "FileType" }, {
@@ -33,14 +33,6 @@ return {
         pattern = { "bib", "tex" },
         callback = function()
           vim.wo.conceallevel = 2
-        end,
-      })
-
-      vim.api.nvim_create_autocmd({ "FileType" }, {
-        group = vim.api.nvim_create_augroup("lazyvim_vimtex_keymap", { clear = true }),
-        pattern = { "bib", "tex" },
-        callback = function()
-          vim.keymap.set("n", "<Leader>K", "<plug>(vimtex-doc-package)", { desc = "Vimtex Docs", silent = true })
         end,
       })
 
@@ -55,7 +47,11 @@ return {
     optional = true,
     opts = {
       servers = {
-        texlab = {},
+        texlab = {
+          keys = {
+            { "<Leader>K", "<plug>(vimtex-doc-package)", desc = "Vimtex Docs", silent = true },
+          },
+        },
       },
     },
   },
--- a/lua/lazyvim/plugins/extras/lang/tex.lua
+++ b/lua/lazyvim/plugins/extras/lang/tex.lua
@@ -1,0 +1,61 @@
+return {
+  {
+    "folke/which-key.nvim",
+    opts = {
+      defaults = {
+        ["<LocalLeader>l"] = { name = "+vimtex" },
+      },
+    },
+  },
+
+  -- Add BibTeX/LaTeX to treesitter
+  {
+    "nvim-treesitter/nvim-treesitter",
+    opts = function(_, opts)
+      if type(opts.ensure_installed) == "table" then
+        vim.list_extend(opts.ensure_installed, { "bibtex", "latex" })
+      end
+      if type(opts.highlight.disable) == "table" then
+        vim.list_extend(opts.highlight.disable, { "latex" })
+      else
+        opts.highlight.disable = { "latex" }
+      end
+    end,
+  },
+
+  {
+    "lervag/vimtex",
+    version = "*",
+    ft = { "bib", "tex" },
+    keys = {
+      { "<LocalLeader>ld", "<Cmd>VimtexDocPackage<CR>", desc = "Open Documentation" },
+    },
+    config = function()
+      vim.opt.conceallevel = 2
+      vim.g.vimtex_complete_enabled = 0 -- use texlab for completion
+      vim.g.vimtex_mappings_disable = { ["n"] = { "K" } } -- disable `K` as the confilict to LSP hover
+      vim.g.vimtex_quickfix_method = vim.fn.executable("pplatex") == 1 and "pplatex" or "latexlog"
+
+      vim.g.vimtex_view_method = vim.fn.has("mac") == 1 and "skim" or "zathura"
+    end,
+  },
+
+  -- Correctly setup lspconfig for LaTeX 🚀
+  {
+    "neovim/nvim-lspconfig",
+    optional = true,
+    opts = {
+      servers = {
+        texlab = {
+          settings = {
+            texlab = {
+              chktex = {
+                onOpenAndSave = true,
+              },
+            },
+          },
+        },
+      },
+    },
+  },
+}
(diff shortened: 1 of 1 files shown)


lua/lazyvim/plugins/extras/lang/tex.lua
```diff
@@ -0,0 +1,61 @@
...
+    "nvim-treesitter/nvim-treesitter",
+    opts = function(_, opts)
+      if type(opts.ensure_installed) == "table" then
+        vim.list_extend(opts.ensure_installed, { "bibtex", "latex" })
+      end
+      if type(opts.highlight.disable) == "table" then
+        vim.list_extend(opts.highlight.disable, { "latex" })
+      else
+        opts.highlight.disable = { "latex" }
+      end
```
amaanq, line 22: Why?

lua/lazyvim/plugins/extras/lang/tex.lua
```diff
@@ -0,0 +1,61 @@
...
+    "lervag/vimtex",
+    version = "*",
+    ft = { "bib", "tex" },
+    keys = {
+      { "<LocalLeader>ld", "<Cmd>VimtexDocPackage<CR>", desc = "Open Documentation" },
+    },
+    config = function()
+      vim.opt.conceallevel = 2
+      vim.g.vimtex_complete_enabled = 0 -- use texlab for completion
+      vim.g.vimtex_mappings_disable = { ["n"] = { "K" } } -- disable `K` as the confilict to LSP hover
```
amaanq, line 36:
```suggestion
      vim.g.vimtex_mappings_disable = { ["n"] = { "K" } } -- disable `K` as it conflicts with LSP hover
```

lua/lazyvim/plugins/extras/lang/tex.lua
```diff
@@ -0,0 +1,61 @@
...
+      vim.g.vimtex_quickfix_method = vim.fn.executable("pplatex") == 1 and "pplatex" or "latexlog"
+
+      vim.g.vimtex_view_method = vim.fn.has("mac") == 1 and "skim" or "zathura"
```
amaanq, line 39: These are opinionated choices and might not be installed by everyone, best to leave it to the user probs

lua/lazyvim/plugins/extras/lang/tex.lua
Mayrixon: I have mentioned above in the point 1. I don't want to touch `lazyvim/plugins/treesitter.lua` so there is an else clause.

amaanq: I know, but I'm not sure if conceals alone are a good case to disable highlights...

lua/lazyvim/plugins/extras/lang/tex.lua
Mayrixon: I agree with line 37-39, but still want to keep line 36.

By default, vimtex maps `K` to `<Cmd>VimtexDocPackage<CR>` and will not override existing keymaps. Usually LSP should be initialised before vimtex, which means it's impossible to have an empty `K` binding. However, I think line 36 could be a mark for people who read vimtex's manual about this function.

amaanq: 36 is totally fine, I was only talking about 39

lua/lazyvim/plugins/extras/lang/tex.lua
Mayrixon: Here are images containing the same paragraphs.

截屏2023-07-18 01 44 00

截屏2023-07-18 01 47 34
 
The first image is using vimtex's syntax highlights and conceals without treesitter highlights. The second image is using treesitter highlights and vimtex's highlights/conceals are disabled. Vimtex's highlights (image 1) render citations naturally. For equations containing Greek alphabets, it's easier to find without searching.

lua/lazyvim/plugins/extras/lang/tex.lua
Mayrixon: Already removed

Mayrixon: BTW, in line 37, `latexlog` is provided by vimtex.

Mayrixon:
@amaanq I'm sorry if my language makes you feel uncomfortable. My English training was all focused on academic purpose, so it looks like I'm defending for my own pieces of code. I'm really welcome for your review. Thanks!

amaanq:
No, don't apologize at all - I totally understand where you're coming from and just want to ensure specs can be applicable to everyone :)

lua/lazyvim/plugins/extras/lang/tex.lua
amaanq: I see, I agree the first one does seem a lot nicer, that's fine then

clason:
Do not enable tree-sitter for TeX (and bib) files, and do not set `conceal` by default.

lua/lazyvim/plugins/extras/lang/tex.lua
```diff
@@ -0,0 +1,61 @@
...
+      else
+        opts.highlight.disable = { "latex" }
+      end
+    end,
+  },
+
+  {
+    "lervag/vimtex",
+    version = "*",
+    ft = { "bib", "tex" },
```
clason, line 29: Do not lazy-load VimTeX. That is not supported. (It needs to set things up _before_ you set the filetype.)

lua/lazyvim/plugins/extras/lang/tex.lua
```diff
@@ -0,0 +1,61 @@
...
+    keys = {
+      { "<LocalLeader>ld", "<Cmd>VimtexDocPackage<CR>", desc = "Open Documentation" },
+    },
+    config = function()
+      vim.opt.conceallevel = 2
+      vim.g.vimtex_complete_enabled = 0 -- use texlab for completion
```
clason, line 35: VimTeX's omni-completion is slower but more complete, as it makes full use of the Vim complete API (e.g., can search on `menu`). It's better to allow both to coexist: autocomplete via texlab, and manual omnicomplete via VimTeX.

lua/lazyvim/plugins/extras/lang/tex.lua
clason, line 36: Again, LSP hover and VimTeX `keywordprg` are different and complementary.

lua/lazyvim/plugins/extras/lang/tex.lua
clason, line 39: Same. Viewer is personal preference.

lua/lazyvim/plugins/extras/lang/tex.lua
```diff
@@ -0,0 +1,61 @@
...
+  {
+    "neovim/nvim-lspconfig",
+    optional = true,
+    opts = {
+      servers = {
+        texlab = {
+          settings = {
+            texlab = {
+              chktex = {
+                onOpenAndSave = true,
```
clason, line 53: `chktex` is very spammy; I'd counsel against it.

lua/lazyvim/plugins/extras/lang/tex.lua
```diff
@@ -0,0 +1,61 @@
...
+
+  {
+    "lervag/vimtex",
+    version = "*",
+    ft = { "bib", "tex" },
+    keys = {
+      { "<LocalLeader>ld", "<Cmd>VimtexDocPackage<CR>", desc = "Open Documentation" },
+    },
+    config = function()
+      vim.opt.conceallevel = 2
```
folke, line 34: Probably best to do this in an autocmd:

```lua
vim.api.nvim_create_autocmd({ "FileType" }, {
  group = vim.api.nvim_create_augroup("lazyvim_vimtex", {clear = true}),
  pattern = { "bib", "tex" },
  callback = function()
    vim.wo.conceallevel = 2
  end,
})
```

If users don't want this, they can then simply remove the `augroup`

folke (changes requested):

lua/lazyvim/plugins/extras/lang/tex.lua
folke: lazy-loading works fine though last time I checked, but definitely something OP should verify/test to be 100% sure

lua/lazyvim/plugins/extras/lang/tex.lua
folke: We should probably have both available.

I'm fine with the texlab lsp hover as `K` by default, unless the opposite makes more sense.

You can add the other mapping to:

```lua
nvim-slpconfig.opts.servers.texlab.keys = {
  {"<c-k>", "<Plug>(vimtex-doc-package)", desc = "Vimtex Docs" }
}
```

lua/lazyvim/plugins/extras/lang/tex.lua
clason: Did you test inverse search (which is a core feature)?

From the docs: 
> Note: Many plugin managers provide mechanisms to lazy load plugins. There is
      no need to use such a mechanism for VimTeX, and in fact, doing it will
      prevent Vim/neovim from loading the :VimtexInverseSearch command.

I'd respect the dev's intent here.

lua/lazyvim/plugins/extras/lang/tex.lua
folke: Sorry, `<c-k>` is not a keymap we should use since that's to go to the window above :)
Any suggestions? `<leader>k`?

lua/lazyvim/plugins/extras/lang/tex.lua
clason: Also, VimTeX comes with improved `ftdetect` scripts, which again are broken by lazy-loading.

folke: ok, great

lua/lazyvim/plugins/extras/lang/tex.lua
Mayrixon: So your suggestion is remove line 35, am I right?

I tried to add omni-completion into nvim-cmp according to the VimTeX's document. However, the solution requires a change in `opts.formatting.format`, which is a function. Maybe I can use `require("cmp").setup.buffer()` to only config specific filetypes, but that need either `after/ftplugin` or `autocmd FileType`.

clason: I don't use `cmp`, sorry. But I believe there's a custom source for VimTeX?

lua/lazyvim/plugins/extras/lang/tex.lua
Mayrixon: I think I had already mapped this function to `<localleader>ld`, where `<localleader>l` is VimTeX's default "leader" key and `<localleader>lk` is occupied. I removed this keymap temporarily in the lazy-loading related commit because the keybinding will always be available without lazy-loading.

lua/lazyvim/plugins/extras/lang/tex.lua
```diff
@@ -0,0 +1,61 @@
...
+      vim.g.vimtex_complete_enabled = 0 -- use texlab for completion
+      vim.g.vimtex_mappings_disable = { ["n"] = { "K" } } -- disable `K` as the confilict to LSP hover
+      vim.g.vimtex_quickfix_method = vim.fn.executable("pplatex") == 1 and "pplatex" or "latexlog"
```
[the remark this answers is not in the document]
Mayrixon: You mean `pplatex`? VimTeX's default is `latexlog`

lua/lazyvim/plugins/extras/lang/tex.lua
Mayrixon: No, the source VimTeX using is omni. For a better result, the omni suggestions should be formatted in a function.

If leave VimTeX's completion enabled, users can still find suggestions by `<C-X><C-O>`, but I'm not sure if this is a good behaviour.

Mayrixon: @folke would you like to provide some advise about enable omni-completion?

lua/lazyvim/plugins/extras/lang/tex.lua
Mayrixon: I agreed with you and updated the code.

folke:
I made some small adjustments to the code.

I do get an error by Vimtex saying that syntax is controlled by Treesitter, however latex is disabled in treesitter.

Any ideas on why that's not working?

folke:
I'm referring to this error: https://github.com/lervag/vimtex/blob/08371bf97a9c4ce3dbdf996161faf04ff8b32aa0/autoload/vimtex/nvim.vim#L13

Looking at `:TSConfigInfo`, `latex` is correctly added to `opts.highlight.disable`.

@clason do you have any ideas why that error still triggers?

clason:
Probably an overzealous check in VimTeX. I'll see about that.

folke:
> Probably an overzealous check in VimTeX. I'll see about that.

Looks like we posted a comment at the same time :) See my comment above with the actual check

folke:
Seems that the vimtex error is not correct. It only shows when restoring a session. But even then treesitter doesn't seem to attach.

All looks good

folke:
Thank you for the PR!

PR #1159, opened 2023-07-18
Improve summary documentation of deps

tmcw:
This adds more detailed documentation strings to LazyVim's dependencies. As far as I can tell, the documentation for these on the website is generated from these docstrings.

This is just trying to add enough description so that people know what more of the plugins do without needing to click through to their individual sites.


folke:
thanks!

PR #1165, opened 2023-07-20
refactor: update clangd root detection

amaanq:
It'd be nice if there was a precedence associated w/ the ordering of arguments, but I don't think there is and it seems to match the most nested dir. That's not ideal for a lot of projects, like [tree-sitter](https://github.com/tree-sitter/tree-sitter) where compile_flags.txt is in `lib/`

Suggestions welcome!

### DIFF
--- a/lua/lazyvim/plugins/extras/lang/clangd.lua
+++ b/lua/lazyvim/plugins/extras/lang/clangd.lua
@@ -53,14 +53,19 @@ return {
           keys = {
             { "<leader>cR", "<cmd>ClangdSwitchSourceHeader<cr>", desc = "Switch Source/Header (C/C++)" },
           },
-          root_dir = function(...)
-            -- using a root .clang-format or .clang-tidy file messes up projects, so remove them
+          root_dir = function(fname)
             return require("lspconfig.util").root_pattern(
-              "compile_commands.json",
-              "compile_flags.txt",
+              "Makefile",
+              "CMakeLists.txt",
               "configure.ac",
-              ".git"
-            )(...)
+              "configure.in",
+              "config.h.in",
+              "meson.build",
+              "meson_options.txt",
+              "build.ninja"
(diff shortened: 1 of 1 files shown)


folke:
If you want to have precedence, you can do something like the below:

```lua
return function(...)
  return require("lspconfig.util").root_pattern("Makefile")(...) or require("lspconfig.util").root_pattern("another pattern")(...) or --the others
end
```

amaanq:
that's smart..thanks!

amaanq:
updated, that should perform much better

folke:
thanks!

lucobellic:
Now that "compile_commands.json" and "compile_flags.txt" are no longer the first default, I was surprised to see several clangd processes running in single file mode on a project with one CMakeLists.txt per folder.

It's easy enough to restore the previous behavior, but @folke, @amaanq, was the default root path search order change expected?

amaanq:
that's interesting, I guess there's no easy solution for this after all :sweat_smile:, for me the original behavior was usually fine for projects purely in C, but add anything else like Rust and it can get wonky.. (which I usually work with)

lucobellic:
My concern is that projects with multiple CMakeLists.txt are common when following "modern" cmake practice. and IMO this change doesn't seem like a good default configuration for C++ development.

Also by default, browsing throw such a code base will start spawning lot of clangd processes and probably gradually increase CPU usage, which may not be noticeable at fisrt since lsp/completion work as expected.

Do you thing removing the search of CMakeLists.txt could provide better default for most C++ users without introducing issue with your way of working?

amaanq:
we can remove CMakeLists.txt, I don't mind that

folke:
maybe a good idea to remove indeed.
for reference, this is the default root detection in lspconfig: https://github.com/neovim/nvim-lspconfig/blob/master/lua/lspconfig/server_configurations/clangd.lua#L24

ReKylee:
Ok but shouldn't you add back CMakeLists.txt for people who use the extras.cmake ???

PR #1173, opened 2023-07-20
Fix line number toggling

agriffis:
This PR consists of a fix to `Util.toggle()` followed by a separate commit to fix how `<leader>ul` works.

See the individual commit messages for details.


agriffis:
Hmm, I should mention that the old code would silence the toggling of `relativenumber` so there would only be a single message from pressing `<leader>ul`

I opted to let both messages display, since it clarifies what's happening with the two separate settings.

folke:
I've just pushed as fix for this. Thank you for bringing this to my attention!

PR #1184, opened 2023-07-22
fix(go): add redundancy to null-ls config

Mayrixon:


folke:
That's not needed. You should always import LazyVim first, followed by extras, followed by your own plugins

PR #1192, opened 2023-07-22
feat: add test and debug support for Java extension

appelgriebsch:
as title says: this PR adds `java-test` and `java-debug-adapter` support to the language package; as there is no official `neotest` plugin yet, it uses the features from `nvim-jdtls` for running tests


folke:
thanks!

PR #1195, opened 2023-07-23
fix(lsp): supports_method for checking inlay hint support

MariaSolOs:
Updates the check for inlay hints support to use `supports_method` instead of `client.server_capabilities`, given that with dynamic registration checking for static server capabilities isn't enough:
image


folke:
thanks!

PR #1196, opened 2023-07-23
fix: rust dap installation

appelgriebsch:
on a clean system `codelldb` was not installed automatically when importing the `rust.lua` extension. this PR provides a fix

### DIFF
--- a/lua/lazyvim/plugins/extras/lang/rust.lua
+++ b/lua/lazyvim/plugins/extras/lang/rust.lua
@@ -31,13 +31,17 @@ return {
 
   -- Ensure Rust debugger is installed
   {
-    "williamboman/mason.nvim",
+    "mfussenegger/nvim-dap",
     optional = true,
-    opts = function(_, opts)
-      if type(opts.ensure_installed) == "table" then
-        vim.list_extend(opts.ensure_installed, { "codelldb" })
-      end
-    end,
+    dependencies = {
+      {
+        "williamboman/mason.nvim",
+        opts = function(_, opts)
+          opts.ensure_installed = opts.ensure_installed or {}
+          vim.list_extend(opts.ensure_installed, { "codelldb" })
+        end,
+      },
+    },
   },
 
   {


dpetka2001:
Strange, with the current rust extra it got installed just fine for me also on a clean installation (that I keep aside for testing purposes). How does making `mason` a dependency of `nvim-dap` solve this?

appelgriebsch:
Not sure how the sequence of loading is build, but I found a similar setup in the `go.lua` (https://github.com/LazyVim/LazyVim/blob/142e6bec209704210db23b3424b4d51896bb206a/lua/lazyvim/plugins/extras/lang/go.lua#L100) and tried it. For me it was more reliably installing `codelldb`. Also when configured this way I expect it to not install the dap adapter if the whole dap-module is not installed…

dpetka2001:
I can understand the latter, you wanting it to be installed only when `nvim-dap` is configured. I'm just baffled with the automatic installation. Cuz I tried it a 2nd time on my clean installation and it still installs without problems. Just wondering what might be causing the different behavior.

appelgriebsch:
One important differences between both versions is this line https://github.com/LazyVim/LazyVim/blob/142e6bec209704210db23b3424b4d51896bb206a/lua/lazyvim/plugins/extras/lang/go.lua#L106C12-L106C62.

folke:
@appelgriebsch probably because you don't import specs in the correct order?

Order should always be:
1. LazyVim
2. all extras
3. personal plugins

appelgriebsch:
Hmm actually have it in the same order, and can really say that on a clean machine codelldb is not installed automatically: https://github.com/appelgriebsch/Nv/blob/main/lua/config/lazy.lua#L54

abeldekat:
@dpetka2001

> ...  can understand the latter, you wanting it to be installed only when nvim-dap is configure ....

This is not how lazy.nvim operates currently. Even without nvim-dap, its dependencies (ie mason) will still apply, and
codelldb will be installed.
See discussion: [Given a disabled plugin, don't consider it's dependencies](https://github.com/folke/lazy.nvim/discussions/949)

@appelgriebsch 

Given the original code, your opts.ensure-installed is not a table when entering the function. That's really strange because lsp.lua in stock LazyVim does create a table. I had a peek at your dotfiles and noticed a "deep-extend" in your config/lazy.lua. Would it be possible that during the extend the list of your dependencies changes order?

dpetka2001:
@abeldekat 
> This is not how lazy.nvim operates currently. Even without nvim-dap, its dependencies (ie mason) will still apply, and
codelldb will be installed.
See discussion: https://github.com/folke/lazy.nvim/discussions/949

Thank you for clarifying this. I was under the impression that since `nvim-dap` is configured as `optional` the whole spec would only apply if `nvim-dap` was already enabled and configured somewhere. But I just tried enabling rust with the changes applied from @appelgriebsch (`mason` as a dependency of `nvim-dap`) and indeed it did install `codelldb` without me enabling `nvim-dap` extras. So, yes the dependencies did get apply. I guess `optional` is only for the plugin's own spec that is configured as optional and not its dependencies included?

abeldekat:
Hello @appelgriebsch,

After loading all extras, your top-level mason-config overwrites ensure-installed.
The problem is caused by this fragment in your repository, lua/plugins/lang.lua:
```
{
    "williamboman/mason.nvim",
    opts = {
      ensure_installed = {
        "lua-language-server",
        "marksman"
      },
      ui = {
        icons = {
          package_installed = "✓",
          package_pending = "",
          package_uninstalled = "✗"
        }
      }
    },
  }
```

This could be the fix:
```
  {
    "williamboman/mason.nvim",
    opts = function(_, opts)
      vim.list_extend(opts.ensure_installed, { "lua-language-server", "marksman" })
      opts.ui = {
        icons = {
          package_installed = "✓",
          package_pending = "",
          package_uninstalled = "✗",
        },
      }
    end,
  },

```

LazyVim's examples mention this situation in the treesitter configuration:
> -- since `vim.tbl_deep_extend`, can only merge tables and not lists, the code above
>  -- would overwrite `ensure_installed` with the new value.
>  -- If you'd rather extend the default config, use the code below instead:

PR #1197, opened 2023-07-23
fix: #1194 use cmakelang instead of cmakelint

### LINKED ISSUE #1194: bug: the command cmake-lint is not executable for cmake_lint

### Neovim version (nvim -v)

v0.9.1

### Operating system/version

Fedora 38

### Describe the bug

`cmake-lint` command is not an executable as the the `cmakelint`  linter specified  in mason for `cmake_lint` diagnostics (null-ls) is incorrect.

### Steps To Reproduce

1. Enable `plugins.extras.lang.cmake` in `init.lua`
2. Launch nvim in any project that starts cmake linter
3. Run `:checkhealth`
4. Under the `null-ls` section you will find the error message

### Expected Behavior

A correct cmake linter executable should be installed.

According to https://github.com/jose-elias-alvarez/null-ls.nvim/blob/main/doc/BUILTINS.md, that `cmake_lint`  is now provided by https://github.com/cheshirekow/cmake_format which is given the alias of  `cmakelang`  in mason.

### Repro

_No response_

appelgriebsch:
use `cmakelang` instead of  `cmakelint`, closes #1194

### DIFF
--- a/lua/lazyvim/plugins/extras/lang/cmake.lua
+++ b/lua/lazyvim/plugins/extras/lang/cmake.lua
@@ -21,7 +21,7 @@ return {
     "mason.nvim",
     opts = function(_, opts)
       opts.ensure_installed = opts.ensure_installed or {}
-      vim.list_extend(opts.ensure_installed, { "cmakelint" })
+      vim.list_extend(opts.ensure_installed, { "cmakelang" })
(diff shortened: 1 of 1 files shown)


folke:
thanks!

PR #1206, opened 2023-07-24
feat: Enable flash.nvim highlight on catppuccin integration

radvil:
Since flash.nvim has already been the default one, and catppuccin already support it


skoch13:
This might be not needed as catppuccin [activates](https://github.com/catppuccin/nvim/commit/a84ee1848bfac4601771805396552bdbaa0a0e91) flash support themselves by default

PR #1229, opened 2023-07-26
lang: yaml: Fix TypeError undefined length

jetm:
In YAML files LSP completion is not happening because yaml-language-server is failing with error:

    TypeError: Cannot read properties of undefined (reading 'length')

from `out/server/src/languageserver/handlers/settingsHandlers.js:78:51`, which it's expecting to have `yaml.schemaStore.url.length` and it's doesn't exist.

Signed-off-by: Javier Tia <javier.tia@gmail.com>


folke:
thanks

PR #1230, opened 2023-07-26
refactor(clangd): remove CMakelists from clangd root_dir search

lucobellic:
Hello,

I would like to remove the search for CMakelisits.txt in the clangd root directory, as discussed in #1165

The reason for this change is that projects following "modern" cmake convention contain multiple CMakelists.txt and are therefore not suitable for root directory detection.

Feel free to edit or comment


folke:
makes sense. thanks!

PR #1232, opened 2023-07-26
feat(lang): add R support

m-muecke:


lua/lazyvim/plugins/extras/lang/r.lua
```diff
@@ -0,0 +1,30 @@
...
+      if type(opts.ensure_installed) == "table" then
+        vim.list_extend(opts.ensure_installed, { "r" })
+      end
+    end,
+  },
+  {
+    "neovim/nvim-lspconfig",
+    opts = {
+      servers = {
+        r_language_server = {},
```
jolars, line 14:
```suggestion
        r_language_server = {
          root_dir = function(fname)
            return require("lspconfig.util").root_pattern("DESCRIPTION", "NAMESPACE", ".Rbuildignore")(fname)
              or require("lspconfig.util").find_git_ancestor(fname)
          end,
        },

```

m-muecke: Is this the desired effect, that it won't run unless the file is in an R project structure or a git repo? What is the general benefit of specifying the root_dir?

jolars: Sorry, you're right of course. We should fall back to the home dir:

        r_language_server = {
          root_dir = function(fname)
            return require("lspconfig.util").root_pattern("DESCRIPTION", "NAMESPACE", ".Rbuildignore")(fname)
              or require("lspconfig.util").find_git_ancestor(fname) or vim.loop.os_homedir()
          end,
        },

The default is to look for `.git` and otherwise fall back to the home directory. But the point of having more granular root detection is that if you have an R package in a subdirectory of your git repo, then the root will be set to that directory, which I believe is the correct behavior.

m-muecke: Sounds good!

jolars (approved):

PR #1240, opened 2023-07-27
fix(lsp): keymaps when telescope is disabled

agriffis:
I'm using fzf instead of telescope, but lsp keymaps were borked


agriffis:
Actually this seems like only a partial fix. A more complete fix would be an extras plugin to enable fzf as a full replacement for how telescope is used in LazyVim. I think it wouldn't be too hard, so I'm going to look into that instead.

PR #1245, opened 2023-07-28
feat(black): added an extra for black with null-ls

markis:
Just like [prettier format](https://github.com/LazyVim/LazyVim/blob/main/lua/lazyvim/plugins/extras/formatting/prettier.lua), it would be nice if there was a similar one for black.

### DIFF
--- a/lua/lazyvim/plugins/extras/formatting/black.lua
+++ b/lua/lazyvim/plugins/extras/formatting/black.lua
@@ -1,0 +1,15 @@
+return {
+  {
+    "williamboman/mason.nvim",
+    opts = function(_, opts)
+      table.insert(opts.ensure_installed, "black")
+    end,
+  },
+  {
+    "jose-elias-alvarez/null-ls.nvim",
+    opts = function(_, opts)
+      local nls = require("null-ls")
+      table.insert(opts.sources, nls.builtins.formatting.black)
+    end,
+  },
+}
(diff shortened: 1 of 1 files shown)


loqusion:
I'd definitely be interested in seeing this in LazyVim.

Perhaps we should use [blackd](https://github.com/jose-elias-alvarez/null-ls.nvim/blob/main/doc/BUILTINS.md#blackd) instead, there's already precedent with LazyVim using [prettierd](https://github.com/LazyVim/LazyVim/blob/566049aa4a26a86219dd1ad1624f9a1bf18831b6/lua/lazyvim/plugins/extras/formatting/prettier.lua#L12), plus by my own benchmarks it's much faster than black.

However, a major problem is that blackd doesn't start automatically like prettierd does, so the user would have to already have it running on their system with a systemd service or likewise.

IndianBoy42:
Can you check if it's already running and either connect or start the server?

rafo:
But shouldn't it be called `none-ls` instead of `null-ls` now? null has been archived by the owner.

Whether black or blacked, together with ruff, it would be my dream combination for Python.

folke:
Can you change the refs to `none-ls` and also add an optional spec for conform.nvim?

folke:
The `none-ls` spec should also be `optional=true`

markis:
> Can you change the refs to `none-ls` and also add an optional spec for conform.nvim?

> The none-ls spec should also be optional=true

Done and done.

folke:
Thanks!

PR #1251, opened 2023-07-29
Fix yaml validate and enable line folding

Jomik:
`yamlls` settings is a bit different from `jsonls`. The `validate` property is just a boolean.
see https://github.com/redhat-developer/yaml-language-server#language-server-settings


folke:
thanks!

Issue #1253, opened 2023-07-29
bug: Hit <space> on float term has wait time 1 sec to react (depends on timeoutlen)


radvil:
### Did you check docs and existing issues?

- [X] I have read all the LazyVim docs
- [X] I have searched the existing issues of LazyVim
- [X] I have searched the existing issues of plugins related to this issue

### Neovim version (nvim -v)

NVIM v0.10.0-dev-616+g203181234

### Operating system/version

Arch Linux KDE

### Describe the bug

[current]
[Peek 2023-07-30 01-41.webm](https://github.com/LazyVim/LazyVim/assets/36059968/64ded230-104e-4d0f-afc3-b515aa00d062)

### Steps To Reproduce

1. open float terminal
2. Hit <space>
3. wait 1 sec to react

### Expected Behavior

[expected]
[Peek 2023-07-30 01-40.webm](https://github.com/LazyVim/LazyVim/assets/36059968/9c94656f-fd5d-41ac-8de4-4b01e59e3b3f)

### Repro

```Lua
-- DO NOT change the paths and don't remove the colorscheme
local root = vim.fn.fnamemodify("./.repro", ":p")

-- set stdpaths to use .repro
for _, name in ipairs({ "config", "data", "state", "cache" }) do
  vim.env[("XDG_%s_HOME"):format(name:upper())] = root .. "/" .. name
end

-- bootstrap lazy
local lazypath = root .. "/plugins/lazy.nvim"
if not vim.loop.fs_stat(lazypath) then
  vim.fn.system({ "git", "clone", "--filter=blob:none", "https://github.com/folke/lazy.nvim.git", lazypath, })
end
vim.opt.runtimepath:prepend(lazypath)

-- install plugins
local plugins = {
  "folke/tokyonight.nvim",
  "folke/LazyVim",
  -- add any other plugins here
}
require("lazy").setup(plugins, {
  root = root .. "/plugins",
})

vim.cmd.colorscheme("tokyonight")
-- add anything else here
```

radvil:
wrong tag, that [current] recording should be [expected] and vise versa

folke:
This must be caused because of a terminal keymap you have on leader.
Default LazyVim does not have this issue.

You can debug your own config by checking `:tmap <leader>`

PR #1254, opened 2023-07-29
fix: type <space> on float terminal should not wait 1 sec to react #1253

radvil:
Seems like the mapleader <space> has effected this behavior


Jomik:
This is the fix for #1253  right?
PR title/commit message is a bit weird ? Isn't it "should not wait" ? 😄

radvil:
@Jomik Ah yeah, I think so, I typo so many time. Did you also encounter this issue ? or is it just me ?

Jomik:
I haven't really felt this issue.

PR #1262, opened 2023-07-30
Fix(keymaps): change `v` mode to `x` mode for `save file`

dpetka2001:
While trying to disable `<C-s>` entirely by `vim.keymap.del({ "i", "v","n", "s" }, "<C-s>")` in my own `keymaps.lua`, I got an error on reopening Neovim `No such mapping`. After tinkering around a bit, I changed `v` to `x`, since `v` incorporates both `x` and `s` and the error stopped occuring.

### DIFF
--- a/lua/lazyvim/config/keymaps.lua
+++ b/lua/lazyvim/config/keymaps.lua
@@ -82,7 +82,7 @@ map("i", ".", ".<c-g>u")
 map("i", ";", ";<c-g>u")
 
 -- save file
-map({ "i", "v", "n", "s" }, "<C-s>", "<cmd>w<cr><esc>", { desc = "Save file" })
+map({ "i", "x", "n", "s" }, "<C-s>", "<cmd>w<cr><esc>", { desc = "Save file" })
 
 --keywordprg
 map("n", "<leader>K", "<cmd>norm! K<cr>", { desc = "Keywordprg" })


jyuan0:
Strange, `vim.keymap.del({ "i", "v", "n", "s" }, "<c-s>")` also gives me a `No such mapping` error, but `vim.keymap.del({ "i", "n", "s", "v" }, "<c-s>")` does not.

dpetka2001:
Yes, it gives the error as long as `s` comes after `v`. If `s` comes beforehand, then I also don't notice such error. The first time i noticed it because I tried with the order the keymap was set in lazyvim keymaps. Anyway, I'm not too confident about Neovim's internals, but also from the help it seems that `v` supersedes `x` and `s`. So, I thought maybe it would be good to distinguish them when setting the keymap?

abeldekat:
I also noticed the error. 

As an alternative, maybe *removing* the `s` from LazyVim's definition is a solution as well, as `v` implies both `x` and `s`. 
In that case, when removing `v`, `s` is also removed. Thus, when removing `s` after `v`, the `no mapping error` occurs. 

I do hope I interpret `:h map-overview` correctly.

In my own keymaps, I currently don't include the `s` specified by LazyVim:
```
-- save file with C-s in all modes
-- "s", -- stacktrace ...
for _, mode in pairs({ "i", "v", "n" }) do
  vim.keymap.del(mode, "<C-s>")
end
```

The result is as expected.

dpetka2001:
Yes, of course it could be done that way as well. I just left it there, because I interpreted it like folke would like to distinguish between them for some reason from the initial code. It's up to him to decide which way to go. Either change `v` to `x` or remove `s` all together. The same result will be achieved.

abeldekat:
Hi @dpetka2001,

I described an alternative and its reasoning, on a problem I recognized, without disqualifying your changes. Just to make sure:

> Yes, of course it could be done that way as well.

I am uncertain on how to interpret that quoted line. I hope my comment did not offend you.

dpetka2001:
No no, absolutely no offense at all taken. I totally agree with your point as well. It's one way or the other. I'm fine with either to be honest.

abeldekat:
> No no, absolutely no offense at all taken

That's a relief, as I am a big fan of your contributions to this project! It's not always easy to know how written text is interpreted by the receiver. Thanks!

PR #1266, opened 2023-08-01
fix(Editor): Set keybindings to correct telescope `find_files` expected behaviour

bamdadsabbagh:
Hello,

## Description

This PR sets the expected behaviour according to the keybinding descriptions for telescope `find_files` calls.

The current implementation seems to call `find_files` within current working directory when pressing `<leader>ff` and pressing `<leader>fF` invokes `find_files` in root directory.

## Impacted lines in current implementation

https://github.com/LazyVim/LazyVim/blob/566049aa4a26a86219dd1ad1624f9a1bf18831b6/lua/lazyvim/plugins/editor.lua#L98
https://github.com/LazyVim/LazyVim/blob/566049aa4a26a86219dd1ad1624f9a1bf18831b6/lua/lazyvim/plugins/editor.lua#L99

## Example in my configuration

https://github.com/bamdadsabbagh/config-nvim/blob/1ee03ce6a818c05c2640e3ffc712653f40fe8959/lua/plugins/telescope.lua#L9
https://github.com/bamdadsabbagh/config-nvim/blob/1ee03ce6a818c05c2640e3ffc712653f40fe8959/lua/plugins/telescope.lua#L10

### DIFF
--- a/lua/lazyvim/plugins/editor.lua
+++ b/lua/lazyvim/plugins/editor.lua
@@ -95,8 +95,8 @@ return {
       { "<leader><space>", Util.telescope("files"), desc = "Find Files (root dir)" },
       -- find
       { "<leader>fb", "<cmd>Telescope buffers<cr>", desc = "Buffers" },
-      { "<leader>ff", Util.telescope("files"), desc = "Find Files (root dir)" },
-      { "<leader>fF", Util.telescope("files", { cwd = false }), desc = "Find Files (cwd)" },
+      { "<leader>ff", Util.telescope("files", { cwd = false }), desc = "Find Files (root dir)" },
+      { "<leader>fF", Util.telescope("files"), desc = "Find Files (cwd)" },
       { "<leader>fr", "<cmd>Telescope oldfiles<cr>", desc = "Recent" },
       { "<leader>fR", Util.telescope("oldfiles", { cwd = vim.loop.cwd() }), desc = "Recent (cwd)" },
       -- git


dpetka2001:
That is the intended behavior as far as I can tell. From other issues raised, any change should be done in your personal config if you'd like. Take a look at #952. It also suggests a solution if you don't want this behavior.

bamdadsabbagh:
Hi @dpetka2001,

Thanks for your answer and the link to the related discussion.

My custom configuration already satisfies my needs.

I actually opened this PR because the annotations / descriptions seemed confusing as they did not seem to match, especially when you compare with the descriptions and behaviours of recent/old files.

For a more precise explanation, I find misleading that, given searches within current working directory, one has to press `<leader>fR` for finding recent files and not `<leader>fF` for finding all files. (Current implementation seems to be `<leader>ff` as I read the code and observed behaviour. If not, I guess it's time for me to get some sleep and apologize for this unwanted PR 👀)

dpetka2001:
I just expressed my opinion based on other issues that have already been opened for this behavior. It was stated by the maintainer that it is expected behavior and how to disable it if you don't want it. I'm not deciding anything, just expressing my opinion. You'll have to wait for a month until maintainer comes back from vacation and he will decide on this PR. Please don't take it as if I'm trying to discourage you or anything.

bamdadsabbagh:
Sorry if I gave you this feeling, it was definitely not intended and I felt no discouragement or anything from you for even a second! Apologies for the tone of my message if it did offend you.

Actually, this PR is also opinion based after all ^^ I guess this is more about interpretation of the english language which I believe is not our mother tongue.

It is totally fine and expected to wait for proper reviews. And I am sure there are more important topics than this one ;)

dpetka2001:
No offense at all, so no worries :smile:

Jomik:
Well, from your change it does indeed look like the description does not match the action being run 🤔 
E.g. the one that says "Find Files (cwd)" has `cwd = false`.
I think the PR is a good change, that may be why I get confused at times, never thought more about it though 😅

ejazahm3d:
This issue exists for other shortcuts as well like `<leader><leader>` , `<leader>sg` and other commands that rely on root path.

dpetka2001:
You can try putting in `lazy.lua`
```lua
{
      "LazyVim/LazyVim",
      init = function()
        require("lazyvim.util").get_root = function()
          return vim.loop.cwd()
        end
      end,
      import = "lazyvim.plugins",
    },
```
This way it won't use the `get_root()` function and default to `vim.loop.cwd()`. Again this is intended behavior to return the `root_dir` of the currently attached LSP clients on the buffer(this depends on how you specify the `root_pattern` in your projects for the LSP to attach, so it might be the case that `root_dir` and `cwd` are the same). If you don't like this behavior you can try the above snippet or change the keymap for telescope in your personal configuration to not use `Util.telescope()` and instead use the command `Telescope` directly.

folke:
This is indeed intended behavior. I'm planning to document this soon and also make it configurable.

Jomik:
> This is indeed intended behavior. I'm planning to document this soon and also make it configurable.

@folke, can you explain this line, just because it confuses me :)
https://github.com/LazyVim/LazyVim/blob/566049aa4a26a86219dd1ad1624f9a1bf18831b6/lua/lazyvim/plugins/editor.lua#L99
It looks like we are saying that it should find files, not using `cwd`, since we set the option to false, but the description indicates that we do use `cwd`?

folke:
When `cwd=false`, the root detection will be used instead.

PR #1281, opened 2023-08-03
feat(neoconf): Move to separate plugin spec

mertzt89:
Moves neoconf to its own plugin spec so that neoconf can be used in an environment where LSP is not necessary.


lua/lazyvim/plugins/editor.lua
```diff
@@ -461,4 +461,16 @@ If you rather use leap/flit instead, you can add the leap extra:
       { "<leader>sT", "<cmd>TodoTelescope keywords=TODO,FIX,FIXME<cr>", desc = "Todo/Fix/Fixme" },
     },
   },
+
```
mertzt89, line 464: I wasn't sure where to stick this. If it belongs somewhere else i can move it.

PR #1294, opened 2023-08-06
fix(telescope): replace anonymous functions in mappings by named functions

MariaSolOs:
When adding mappings to Telescope with anonymous functions, the following notification is shown when invoking Telescope's which-key:
image

The fix is to replace the anonymous functions by references to named ones, as described in [this comment](https://github.com/nvim-telescope/telescope.nvim/issues/2237#issuecomment-1322642896).

### DIFF
--- a/lua/lazyvim/plugins/editor.lua
+++ b/lua/lazyvim/plugins/editor.lua
@@ -161,48 +161,47 @@ return {
         desc = "Goto Symbol (Workspace)",
       },
     },
-    opts = {
-      defaults = {
-        prompt_prefix = " ",
-        selection_caret = " ",
-        mappings = {
-          i = {
-            ["<c-t>"] = function(...)
-              return require("trouble.providers.telescope").open_with_trouble(...)
-            end,
-            ["<a-t>"] = function(...)
-              return require("trouble.providers.telescope").open_selected_with_trouble(...)
-            end,
-            ["<a-i>"] = function()
-              local action_state = require("telescope.actions.state")
-              local line = action_state.get_current_line()
-              Util.telescope("find_files", { no_ignore = true, default_text = line })()
-            end,
-            ["<a-h>"] = function()
-              local action_state = require("telescope.actions.state")
-              local line = action_state.get_current_line()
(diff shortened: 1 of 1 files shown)


lua/lazyvim/plugins/editor.lua
```diff
@@ -161,48 +161,47 @@ return {
...
-            end,
+    opts = function()
+      local actions = require("telescope.actions")
+
+      local open_with_trouble = function(...)
+        return require("trouble.providers.telescope").open_with_trouble(...)
+      end
+      local open_selected_with_trouble = function(...)
+        return require("trouble.providers.telescope").open_selected_with_trouble(...)
+      end
```
jyuan0, line 172: You may want to leave these and the insert mode trouble.nvim maps as anonymous functions to keep the lazy-loading.

MariaSolOs: Lazy loading shouldn't be affected though, since I'm requiring the modules inside the function bodies.

MariaSolOs: I've also changed `opts` into its function form, which `lazy` will invoke until loading Telescope, so this shouldn't do any premature loading.

jyuan0: > Lazy loading shouldn't be affected though, since I'm requiring the modules inside the function bodies.

I guess I'm mistaken, when I made the same changes locally, it seemed like Trouble with being loaded by Telescope.

jyuan0: I gosh, yeah I'm sorry, I misread the code.

MariaSolOs: @jyuan0 oh my don't sweat it, it happens to me all the time <3
Thank you for the review!

dpetka2001: Yes, this is the same code essentially, expect that it assigns a function name. Should have no effect on lazy loading. Looking forward to this PR. I always looked at those anonymous functions, but got used to them (by reading about the keymaps from the source code) and never bothered with them. Thank you for this.

PR #1297, opened 2023-08-06
[feat] add codeium extension

appelgriebsch:
add new extension for [Codeium](https://codeium.com/) based on `cmp-codeium` plugin - similar to `copilot`

### DIFF
--- a/lua/lazyvim/plugins/extras/coding/codeium.lua
+++ b/lua/lazyvim/plugins/extras/coding/codeium.lua
@@ -1,0 +1,22 @@
+return {
+
+  -- codeium cmp source
+  {
+    "nvim-cmp",
+    dependencies = {
+      {
+        "jcdickinson/codeium.nvim",
+        opts = {},
+        config = function(_, opts)
+          require("codeium").setup(opts)
+        end,
+      },
+    },
+    ---@param opts cmp.ConfigSchema
+    opts = function(_, opts)
+      table.insert(opts.sources, 1, { name = "codeium", group_index = 2 })
+      opts.sorting = opts.sorting or require("cmp.config.default")().sorting
+    end,
+  },
+
+}
--- a/lua/lazyvim/config/init.lua
+++ b/lua/lazyvim/config/init.lua
@@ -41,6 +41,7 @@ local defaults = {
       Array = " ",
       Boolean = " ",
       Class = " ",
+      Codeium = " ",
       Color = " ",
       Constant = " ",
       Constructor = " ",


lua/lazyvim/plugins/extras/coding/codeium.lua
```diff
@@ -0,0 +1,22 @@
...
+        "jcdickinson/codeium.nvim",
+        opts = {},
+        config = function(_, opts)
+          require("codeium").setup(opts)
+        end,
+      },
+    },
+    ---@param opts cmp.ConfigSchema
+    opts = function(_, opts)
+      table.insert(opts.sources, 1, { name = "codeium", group_index = 2 })
```
dpetka2001, line 17: If you put it in `group_index = 2`, doesn't that mean it will only give suggestions if there are no suggestions from `nvim_lsp` and rest which are by default `group_index = 1`. Wouldn't it be better for the `codeium` source to work alongside `nvim_lsp` source? How does it work for you with this configuration?

jyuan0: > If you put it in group_index = 2, doesn't that mean it will only give suggestions if there are no suggestions from nvim_lsp and rest which are by default group_index = 1

From my understanding of this setting, you're right, but the Copilot extra plugin is also using `group_index = 2`, so maybe that's why.

lua/lazyvim/plugins/extras/coding/codeium.lua
```diff
@@ -0,0 +1,23 @@
+return {
+
+  -- codeium cmp source
+  {
+    "nvim-cmp",
+    commit = "6c84bc75c64f778e9f1dcb798ed41c7fcb93b639", -- lock update (break codeium)
```
dpetka2001, line 6: @appelgriebsch This codeium [commit](https://github.com/jcdickinson/codeium.nvim/pull/84) seems to fix the problem with `nvim-cmp`, so you might want to remove the pinning of `nvim-cmp` to the specific commit. I tested it in my personal config and seems to work, but you might also want to test it out in case I missed something.

appelgriebsch: Thanks for notifying me. Just tested and updated PR accordingly.

lua/lazyvim/plugins/extras/coding/codeium.lua
jjohnston-aims: [jcdickinson/codeium.nvim](https://github.com/jcdickinson/codeium.nvim) has just been archived

appelgriebsch:
Closed due to depreciation of plug-in

PR #1298, opened 2023-08-07
let keymap n and N opens folds

doitian:
It's strange that mappings to n and N do not open folds like its original behavior, so I append `zv` to manually open folds.


lkhphuc:
Gently ping @folke for your opinion.

With the new fold column indicator and treesitter foldtext, I've been using fold more often and this is really helpful.
Just to note, expand fold to view search line is the default behaviors of Vim as well, so this feels very natural.

PR #1300, opened 2023-08-07
feat(java): better java lang support

aqav:
Better Java support:

1. The existing `java.lua` put `nvim-jdtls` in the dependency of `nvim-lspconfig`. `nvim-jdtls` should not be the dependency of `nvim-lspconfig`, if so, every time I open a file need `nvim-lspconfig`, the `nvim-jdtls` would also be loaded, even if it's not a Java file. The dependency should be reversed.
existing:

3. Enable cmp doc preview(which the existing `java.lua` not)
existing:
Snipaste_2023-08-07_22-40-15
mine:
Snipaste_2023-08-07_22-42-20
4. Use which-key to register all the keymap to avoid the unamed(+prefix) group(which the existing `java.lua` not)
existing:
Snipaste_2023-08-07_22-40-55
mine:
Snipaste_2023-08-07_22-43-08
5. The existing `java.lua` use `require("lazyvim.util").on_attach`([95](https://github.com/LazyVim/LazyVim/blob/main/lua/lazyvim/plugins/extras/lang/java.lua#L95C20-L95C20)) to create `on_attach` would cause the endless function invoke loop(try to put a `vim.notify` into `on_attach`, you will see. Everytime user enter a new buffer, the invoke times will add one. For example, when you enter the fifth buffer, the `on_attach` would be invoked 5 times)
6. Register keymaps and setup dap after the LSP fully attached
7. More simple and straight config, help user to understand what the config did, and how to create a own one according this one
8. Reasonable keymap


agriffis:
Hi @aqav! Thanks for this PR. I contributed the original java support to LazyVim, which clearly needs some fixes as you mentioned.

I appreciate your changes overall, but some of the changes seem unexplained so I asked some questions in code review.

amaanq:
If you started a review, you need to "submit" these changes in the files tab (bad ui choice, I know), otherwise they'll be stuck in a "pending" state that you can see at the top of any comment

agriffis:
> If you started a review, you need to "submit" these changes in the files tab (bad ui choice, I know), otherwise they'll be stuck in a "pending" state that you can see at the top of any comment

Thanks!

lua/lazyvim/plugins/extras/lang/java.lua
```diff
@@ -1,125 +1,143 @@
...
+      "mfussenegger/nvim-dap",
+      "hrsh7th/nvim-cmp",
+      "folke/which-key.nvim",
+    },
+    ft = { "java" },
+    config = function()
+      local project_name = vim.fn.fnamemodify(vim.fn.getcwd(), ":p:h:t")
+      local workspace_dir = vim.fn.stdpath("data") .. "/lsp/java/" .. project_name
+      local config = {
+        cmd = {
```
appelgriebsch, line 55: providing the full cmd is usually no longer required as most of the default parameters are given by the jdtls.sh script now, so can you evaluate/explain if there is something in addition to those default parameters needed? I would rather rely on the jdtls bundled startup script instead of override everything…

lua/lazyvim/plugins/extras/lang/java.lua
```diff
@@ -1,125 +1,143 @@
...
+          "settings.gradle",
+          "settings.gradle.kts",
+          "build.xml",
+          ".project",
+        }),
+        settings = {
+          java = {},
+        },
+        handlers = {
+          ["$/progress"] = function() end,
```
appelgriebsch, line 91: Any explanation why you override those handlers with empty functions?

lua/lazyvim/plugins/extras/lang/java.lua
```diff
@@ -1,125 +1,143 @@
...
+          ["language/status"] = function() end,
+        },
+        init_options = {
+          bundles = {
+            vim.fn.expand("$MASON/share/java-debug-adapter/com.microsoft.java.debug.plugin.jar"),
```
appelgriebsch, line 96: The original version checks if the user has installed the debug adapter and test extension… you assume here it’s always there…

appelgriebsch:
> 4\. The existing `java.lua` use `require("lazyvim.util").on_attach`([95](https://github.com/LazyVim/LazyVim/blob/main/lua/lazyvim/plugins/extras/lang/java.lua#L95C20-L95C20)) to create `on_attach` would cause the endless function invoke loop(try to put a `vim.notify` into `on_attach`, you will see. Everytime user enter a new buffer, the invoke times will add one. For example, when you enter the fifth buffer, the `on_attach` would be invoked 5 times)

But if you check the on_attach code you will notice that it will set up the keybindings for the current buffer, so having it invoked per buffer shouldn’t be an issue but rather the expected behavior… also calling jdtls.start_or_attach multiple times is not an issue but the preferred way of doing (check nvim-jdtls documentation)

lua/lazyvim/plugins/extras/lang/java.lua
```diff
@@ -1,125 +1,143 @@
...
-      if type(opts.ensure_installed) == "table" then
-        vim.list_extend(opts.ensure_installed, { "java" })
-      end
-    end,
+    opts = {
+      ensure_installed = {
+        "java",
+      },
+    },
+  },
```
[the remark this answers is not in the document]
aqav: You're right, my bad, I have been recommitted, please review 5e25b7c9ab7b217e26fc267e8bc7ed39a42fe88f

lua/lazyvim/plugins/extras/lang/java.lua
```diff
@@ -1,125 +1,143 @@
...
+  {
+    "williamboman/mason.nvim",
+    opts = {
+      ensure_installed = {
+        "jdtls",
+      },
+    },
```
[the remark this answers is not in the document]
aqav: As I explained above, please review 5e25b7c9ab7b217e26fc267e8bc7ed39a42fe88f

lua/lazyvim/plugins/extras/lang/java.lua
```diff
@@ -1,125 +1,143 @@
...
   },
-
-  -- Ensure java debugger and test packages are installed
```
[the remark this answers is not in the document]
aqav: have add some comment for explain why I write this way and some links to the doc. 5e25b7c9ab7b217e26fc267e8bc7ed39a42fe88f

lua/lazyvim/plugins/extras/lang/java.lua
```diff
@@ -1,125 +1,143 @@
...
+    dependencies = {
+      "neovim/nvim-lspconfig",
+      "mfussenegger/nvim-dap",
+      "hrsh7th/nvim-cmp",
+      "folke/which-key.nvim",
+    },
+    ft = { "java" },
+    config = function()
+      local project_name = vim.fn.fnamemodify(vim.fn.getcwd(), ":p:h:t")
+      local workspace_dir = vim.fn.stdpath("data") .. "/lsp/java/" .. project_name
```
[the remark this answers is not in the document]
aqav: The way to get the path is just my coding convention, the main diff is I don't think the LSP data is a "cache"

lua/lazyvim/plugins/extras/lang/java.lua
```diff
@@ -1,125 +1,143 @@
...
+          "java.base/java.util=ALL-UNNAMED",
+          "--add-opens",
+          "java.base/java.lang=ALL-UNNAMED",
+          "-jar",
+          vim.fn.expand("$MASON/share/jdtls/plugins/org.eclipse.equinox.launcher.jar"),
+          "-configuration",
+          vim.fn.expand("$MASON/share/jdtls/config"),
+          "-data",
+          workspace_dir,
+        },
```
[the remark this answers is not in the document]
aqav: According to the jdtls doc(https://github.com/eclipse-jdtls/eclipse.jdt.ls#running-from-command-line-with-wrapper-script), the command "jdtls" would work if the user has Python 3.9 installed, can't guarantee that.

The more secure way I think is to use the "java -De...", because if someone enable the LazyVim Java support, the requirement of "java" existing in PATH is reasonable.

And more, exposing more detail can help user to understand what the config do, and help them to create their own config if they have more requirement(for example, "-xmx1g")

lua/lazyvim/plugins/extras/lang/java.lua
```diff
@@ -1,125 +1,143 @@
...
+          "mvnw",
+          "gradlew",
+          "pom.xml",
+          "build.gradle",
+          "build.gradle.kts",
+          "settings.gradle",
+          "settings.gradle.kts",
+          "build.xml",
+          ".project",
+        }),
```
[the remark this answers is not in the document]
aqav: I just put every filetype in `require("lspconfig.server_configurations.jdtls").default_config` into the `root_dir`, 

The reason is the same as I explained above, exposing more detail can help user to understand what the config do, and help them to create their own config if they have more requirement

Add some more filetype that most user should have in the project root(such as `pom.xml`, `.git`)

lua/lazyvim/plugins/extras/lang/java.lua
aqav: Hi, plz ref this https://github.com/LazyVim/LazyVim/pull/1300#discussion_r1286696050

lua/lazyvim/plugins/extras/lang/java.lua
aqav: Because if I don't, every time I enter a word when I coding, the process of LSP will show in the right corner, This may be a bit annoying？

lua/lazyvim/plugins/extras/lang/java.lua
aqav: I want to provide a full feature config (not only include LSP, but also DAP, keymaps, and Test support), and I have added the debug adapter and test extension in the `mason-nvim-dap.nvim` ensure_installed, So I think they should always be there...

For instead of referring $MASON, sorry for that, I didn't know this before, my bad

aqav:
Hi @appelgriebsch 

Sorry I have not explained clearly, I agree with you that the `on_attach` should be invoked per buffer, But should be invoked only once per buffer

If you open buffer A(1st buffer), the on_attach would be invoked once, then you open the buffer B(2nd buffer), and on_attach would be called twice instead of being called a second time, then the total times the on_attach would sum to 3(1+2), image what would happen if user opens the 10th buffer, the on_attach would be invoked 1+2+3+4+5+6+7+8+9+10=55 total!

Try to put `vim.notify("------------")` between the 95 and 108
Snipaste_2023-08-08_15-40-06
I just close the buffer and reopen it multiple times, after repeating a couple of times, every buffer opened would be involved a number of times of on_attach...
Snipaste_2023-08-08_15-40-35

aqav:
The buffer would be invoked the "∑j=1+2+3+…+n" times when you open the nth buffer

appelgriebsch:
> The on_attach would be invoked the "∑n=1+2+3+…+n" times when you open the nth buffer.
> 
> Maybe is the bug only in my PC? Can you reproduce that?
> 
> And the way I register keymap and setup dap would be invoked just one time per buffer, after the LSP fully attaches the buffer

will give it a try, but one difference I notice: you set the test keymaps globally.... not sure if that breaks test integration with other languages if they happen to appear in same project (e.g. javascript/typescript via jest test extension)

lua/lazyvim/plugins/extras/lang/java.lua
```diff
@@ -1,125 +1,151 @@
...
+            ["<leader>cxv"] = { [[<ESC><CMD>lua require('jdtls').extract_variable_all(true)<CR>]], "Extract Variable" },
+            ["<leader>cxc"] = { [[<ESC><CMD>lua require('jdtls').extract_constant(true)<CR>]], "Extract Constant" },
+          }, { mode = "v", buffer = args.buf })
+
+          require("jdtls.dap").setup_dap()
+          require("jdtls.dap").setup_dap_main_class_configs()
+
+          require("which-key").register({
+            ["<leader>t"] = { name = "+test" },
+            ["<leader>tT"] = { require("jdtls.dap").test_class, "Run All Test" },
```
appelgriebsch, line 143: not sure if this only registers the keys for *.java buffers, or in general. If the latter it might break test integrations done via neotest for other languages in the same project (e.g. javascript/typescript via neotest-jest)

lua/lazyvim/plugins/extras/lang/java.lua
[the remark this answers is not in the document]
agriffis: Hmm, if you're going to change this, I think it would be better to expose it as a standalone option than to change it unilaterally for everybody.

This code also calculates the project name from the cwd, which might not be the top level. (Something I also mentioned below but seems to have been lost.) Previously the code would find the project root before calculating the project name.

lua/lazyvim/plugins/extras/lang/java.lua
agriffis: Then say that in a comment! :smile: 

It would also be helpful to say in the comment whether disabling the progress handler will potentially cause any problems for the user, or if it's entirely benign.

lua/lazyvim/plugins/extras/lang/java.lua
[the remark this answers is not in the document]
agriffis: Hmm, it's a shame that the script depends on Python (even if most people will have it installed)

I'm still uneasy about this change. The purpose of the script is to abstract some details so that a plugin such as this doesn't need to reiterate them. The args you used are slightly different from those produced by the Python script, for reasons that aren't explained. Additionally there is no way for the user to augment these arguments without completely overriding the config function, so the idea that the user can create their own config is somewhat lacking.

I don't mean to be completely discouraging. It is possible that this is still the proper way forward. But you've just made a lot of changes without explaining them in comments or genuinely making things easier for the user to tune, which leads to asking lots of questions in code review.

lua/lazyvim/plugins/extras/lang/java.lua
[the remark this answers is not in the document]
agriffis: I'll just reiterate that finding the project root needs to happen before calculating the project name.

About the root detection method, `pom.xml` is already in the default list, see https://github.com/neovim/nvim-lspconfig/blob/3fe1e8de80b98c7a6b16f730711b5eafe84212e1/lua/lspconfig/server_configurations/jdtls.lua#L79-L89

For java projects specifically, `.git` is a false indicator because it can trip on submodules.

I think you should stick with calling the upstream-provided functions as this code did previously. It's counter-productive to reinvent things. If changes deserve to be made, consider proposing them to the lspconfig project.

agriffis:
@aqav I'm not the owner of this project, so you can take my thoughts here with a grain of salt. Nonetheless, I contributed this plugin originally, and I'm concerned by the trajectory of this PR, and I think we need to figure out how to move forward together.

This PR consists of two major fixes (reversing deps and avoiding accumulation of `on_attach`), a few smaller fixes ("Extract" and improved completion), and a number of other changes that primarily reflect your preferences.

I think it would be good to make a separate PR consisting of only the two major fixes, preferably as separate commits for the sake of review. That PR should include comments so that future contributors understand how things work. Internal documentation is critical for a project like this where various individuals will contribute over time without all of the context you have personally.

Then we should look at the smaller fixes and opinionated changes case-by-case. Maybe they make sense, maybe they should be adjustable in opts, or maybe they should stay as-is. But at the moment, all the extra baggage is holding up the important fixes.

What do you think?

aqav:
Hi @agriffis @appelgriebsch 

I've also been thinking for the past few days that maybe I have made too many changes this time, and maybe some of the changes were unnecessary

Java is one of my main development languages, and I prefer simple, clean, fewer configurations (If LazyVim support it, I wouldn't need to create my own), so when LazyVim started to support it, I tried it as soon as I could, but issues 1 and 4 made me switch back to my own configuration.

My main goal is still to "take from the community and contribute to the community", and I could just keep my own configurations private, and this PR could have been unnecessary. But I decided to do something for LazyVim

How about this? For now, this PR I will change based on the existing configuration, and only changes 1 and 2. I think the other parts need more discussion, and may be presented in other PRs in the future.

I think we all agree on 1 and 4 need to change, right?

appelgriebsch:
> I think we all agree on 1 and 4 need to change, right? :)

hey @aqav I would even suggest to look into 2, 3 and 5 as well, which seems to be bugs in the current configuration. I think the most discussion is around the configuration of the java lsp and how to make that better understandable for users.... (jdtls.sh vs. java cmd with a bunch of cli params, workdir finding)... on the rest of your changes we might already be aligned mostly....

aqav:
> > I think we all agree on 1 and 4 need to change, right? :)
> 
> hey @aqav thanks for taking these changes up. I would even suggest to look into 2, 3 and 5 as well, which seems to be bugs in the current configuration. I think the most discussion is around the configuration of the java lsp and how to make that better understandable for users.... (jdtls.sh vs. java cmd with a bunch of cli params, workdir finding)... on the rest of your changes we might already be aligned mostly...

2, 3 and 5 is my suggestion too. 3 I think it may be a bug of `which-key`, but I can't figure out where the problem is, I just find a way to walk around it(register all keymaps and groups together)

For the jdtls vs java cmd, I still prefer the java cmd, here is my reason:
1. We can't assume that the user has installed Python 3.9, and there is no way to ensure that(Can `:checkhealth` help to check ? I don't know), the requirement of Python 3.9 installed for Java developer is not reasonable 
2. Expose more detail for advanced users. If the user is a Java developer who just want play around, enable the extra Java support in config/lazy.lua is the only thing he must do, In this scenario, jdtls or java cmd is no diff; If the user is a Java advanced developer, change the options of Java is always happened, for example, change the `-xmx -xms` to test the GC, so expose these options to the user can help them understand what the default config do and make their own(Maybe support a way to allow user define cmd is better?)

For workdir finding, I'm not so adamant about my opinion, Is LSP data `data`? or `cache`? How to get the root_dir? I am willing to follow the existing standard

agriffis:
I agree with 1 and 4 definitely; and 2, 3, 5 are good, too.

I think the plugin should continue to use the jdtls wrapper script by default. The vast majority of users will have Python 3.9 or later installed (it was released in 2020 and is the default on all Linux distros as well as macOS) so I think there isn't a strong reason to avoid it.

However I'm not opposed to this PR supporting a custom command specified in opts. Then the advanced user has the option to call java directly via their local config. The plugin could even include a commented example.

aqav:
Hi @agriffis @appelgriebsch 

Please review the commit cc3dfa496ea53d0476b51ea8370f67695937620a

All change base the already existing `java.lua` for compatible, solve the 1, 2, 3, 4, 5, add comment if need, and do as more as I can to preserve the original way of implementation

Please try it out, welcome all advice, thank you all

lua/lazyvim/plugins/extras/lang/java.lua
```diff
@@ -24,102 +24,131 @@ return {
...
+          require("jdtls").start_or_attach(vim.tbl_deep_extend("force", jdtls_opts or {}, jdtls_base_config))
+          -- not need to require("jdtls.setup").add_commands(), start automatically adds commands
+        end,
+      })
+
+      -- Setup keymap and dap after the lsp is fully attached
+      -- https://github.com/mfussenegger/nvim-jdtls#nvim-dap-configuration
+      -- https://neovim.io/doc/user/lsp.html#LspAttach
+      vim.api.nvim_create_autocmd("LspAttach", {
+        pattern = "*.java",
```
appelgriebsch, line 113: use fileTypes here again

appelgriebsch (approved):

lua/lazyvim/plugins/extras/lang/java.lua
aqav: `vim.api.nvim_create_autocmd("LspAttach"xx` only work when the pattern is "*.java"(I guess it require a regex), the filetype ref to the `lsp_config.filetypes` is `{"java"}`

If I use fileTypes there, the `autocmd` would work, try it out

agriffis (changes requested):

aqav:
@agriffis Thank you for the advice, I have tested it and it works. Please review 69f5a30af82d95f775409439e0665020b24f7cee

agriffis (approved):

agriffis:
@folke After a lot of back and forth, we finally have consensus on this PR. Feel free to merge when you're ready. :+1:

agriffis:
> @folke After a lot of back and forth, we finally have consensus on this PR. Feel free to merge when you're ready. 👍

Unfortunately I have to take this back. After some testing, there are still some problems. I'll start another review.

lua/lazyvim/plugins/extras/lang/java.lua
```diff
@@ -24,102 +24,134 @@ return {
...
         end,
       },
     },
   },
+
+  -- Set up nvim-jdtls
+  {
+    "mfussenegger/nvim-jdtls",
+    dependencies = { "folke/which-key.nvim" },
+    ft = { "java" },
```
agriffis, line 47: This `ft = { "java" }` needs to be omitted, because the config function also sets up filetype-based autocmds, and the new ones don't fire properly if loading depends on the filetype.

lua/lazyvim/plugins/extras/lang/java.lua
```diff
@@ -24,102 +24,134 @@ return {
...
+        capabilities = require("cmp_nvim_lsp").default_capabilities(),
+      }
+
+      -- Attach jdtls for the proper filetypes (i.e. java).
+      -- Existing server will be reused if the root_dir matches.
+      vim.api.nvim_create_autocmd("FileType", {
+        pattern = filetypes,
+        callback = function()
+          local jdtls_opts = require("lazyvim.util").opts("nvim-jdtls")
+          require("jdtls").start_or_attach(vim.tbl_deep_extend("force", jdtls_opts or {}, jdtls_base_config))
```
agriffis, line 104: This is from the original code, not something you changed. Nonetheless, it should be

```lua
require("jdtls").start_or_attach(vim.tbl_deep_extend("force", jdtls_base_config, jdtls_opts or {})
```

so that local opts can override the base config, rather than the other way around.

agriffis (changes requested):
I tested this with a java file and the jdtls didn't load. Needs some fixes still.

aqav:
@agriffis Please review ebc537540165a31404d05872ab8e0c7c6cde2169

I have tested it and everything seems to be okay.

If remove the `ft = { "java" }`, the `nvim-jdtls` will be loaded by Lazy.nvim even if there is not a Java file, I think that the plugin would not do many things before the user enters a real Java file, but I am not sure if it's reasonable
image

lua/lazyvim/plugins/extras/lang/java.lua
dpetka2001: Plz take what I'm about to say with a grain of salt, since I don't know anything about Java, I'm just commenting based on my experience with `lazy.nvim`. Couldn't you maybe keep the `ft = { "java" }` and then load the autocmds in the config function based on some event other than `FileType`? For example something like `BufReadPre(Post)`, since `jdtls` will load on `java` filetypes either way? Have you tested such a possibility or is this not going to work? If it works this way, then you will also not have the problem of `jdtls` loading when not in Java files.

aqav:
So.. What I should do next now :)

agriffis:
@aqav I'm sorry I haven't replied! We had a major thunderstorm and lost electricity for a number of days. I have an idea for this that I'll test and then follow up shortly.

agriffis:
@aqav @appelgriebsch Could you look at #1388 and let me know what you think.

lua/lazyvim/plugins/extras/lang/java.lua
agriffis: @dpetka2001 Thanks for your comment, and my apologies for not replying earlier.

We'll keep `ft = { "java" }` to preserve lazy loading, but then handle the first-load case specially. See #1388

PR #1343, opened 2023-08-17
Do not jump on invocation of *

knpwrs:
I had this in my configuration before switching to lazyvim and pulled it in to my lazyvim configuration. I figure it might be useful for other folks as well.


Jomik:
This is probably _very_ subjective, hm? Should be paired with `#` right? But then `*` and `#` has the same behaviour 😁 

Curious - what do you use this behaviour for?

folke:
Thank you for this PR, but this is not something I want in LazyVim.

PR #1365, opened 2023-08-22
fix: disable ruff hover correctly

fecet:
image

The hover ability of ruff seems not be correctly disabled.

### DIFF
--- a/lua/lazyvim/plugins/extras/lang/python.lua
+++ b/lua/lazyvim/plugins/extras/lang/python.lua
@@ -9,17 +9,22 @@ return {
   },
   {
     "neovim/nvim-lspconfig",
-    opts = function(_, opts)
-      opts.servers["ruff_lsp"] = {}
-      opts.servers["pyright"] = {}
-
-      require("lazyvim.util").on_attach(function(client, _)
-        if client.name == "ruff_lsp" then
-          -- Disable hover in favor of Pyright
-          client.server_capabilities.hoverProvider = false
-        end
-      end)
-    end,
+    opts = {
+      servers = {
+        pyright = {},
+        ruff_lsp = {},
+      },
+      setup = {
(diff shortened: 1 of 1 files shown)


lkhphuc:
A more general way if using noice would be
```
 { "folke/noice.nvim",
    opts = {
      routes = {
        {
          filter = {
            event = "notify",
            any = {  { find = "No information available"} }
          }
        }
      },
    },
  }
```

fecet:
This approach is better if we actually don't want to disable hover of ruff completely, is that the case here?

dpetka2001:
The `setup` in the original code is outside of `opts`. See [here](https://github.com/LazyVim/LazyVim/blob/566049aa4a26a86219dd1ad1624f9a1bf18831b6/lua/lazyvim/plugins/extras/lang/python.lua#L12-L26). So it's not a problem with the `setup` itself not working. I have it like the original code configured and it works correctly with the `setup`. Would you mind changing your PR to the original code, but put `setup` inside `opts` so that it does take effect?

fecet:
Hi @dpetka2001. I'm new to lazyvim, did you mean change it to 
```
opts = {
      servers = {
        pyright = {},
        ruff_lsp = {},
      },
setup = {
      ruff_lsp = function()
        require("lazyvim.util").on_attach(function(client, _)
          if client.name == "ruff_lsp" then
            -- Disable hover in favor of Pyright
            client.server_capabilities.hoverProvider = false
          end
        end)
      end,
}
    },
    
```

dpetka2001:
Yes exactly. Does that work for you in your local environment? It should in my opinion.

fecet:
> Yes exactly. Does that work for you in your local environment? It should in my opinion.

yeah this works perfexrly

folke:
thanks!

PR #1381, opened 2023-08-26
feat(python): add more venv names

jakub-kozlowicz:
By default `venv-selector` only set virtual environment name to the `venv`. To be more versatile I've added more name possibilities.


robsonpeixoto:
Just for curiosity, why do you need to configure the `venv` that you are using?

PR #1388, opened 2023-08-28
fix(java): don't accumulate on_attach, and make more configurable

agriffis:
This is a replacement for #1300. It keeps the fixes from that PR:

1. Reverse dependency between `nvim-jdtls` and `nvim-lspconfig` so the former isn't loaded unnecessarily.
2. Enable cmp doc preview.
3. Use which-key to register all the keymap to avoid the unamed(+prefix) group.
4. Don't accumulate calls to `on_attach` for each java buffer opened.
5. Register keymaps and setup dap after the LSP fully attached.

and adds the following fixes:

1. Lazy load the plugin with `ft = { "java" }` but additionally handle the race condition internally by calling `attach_jdtls()` directly when the plugin loads.
2. Properly check whether debug is enabled with `Util.has("nvim-dap") and mason_registry.is_installed("java-debug-adapter")` (the original code in #1300 called `mason_registry.has_package()` which isn't the right check)

and improves configuration with the following optional in `opts`:

- `opts.root_dir` function to find root dir for given filename
- `opts.project_name` function to find project name for given root dir
- `opts.jdtls_config_dir` function to find config dir for given project name
- `opts.jdtls_workspace_dir` function to find workspace dir for given project name
- `opts.cmd` base command, so you can use `{ "java", ... }` instead of the script
- `opts.full_cmd` function to compute the full command from the base command
- `opts.jdtls` table (or function modifying/returning table) of configuration for `require("jdtls").start_or_attach(config)`
- `opts.dap` can be false to disable, or table (or function modifying/returning table) of configuration for `require("jdtls").setup_dap(config)`
- `opts.test` can be false to disable
- `opts.on_attach` can augment the existing `LspAttach` autocmd for additional keybindings

For example you can configure a replacement `jdtls` command to avoid the script:

```lua
return {
  "mfussenegger/nvim-jdtls",
  opts = {
    cmd = { "java", ... }
  },
}
```

### DIFF
--- a/lua/lazyvim/plugins/extras/lang/java.lua
+++ b/lua/lazyvim/plugins/extras/lang/java.lua
@@ -64,6 +64,49 @@ return {
     "mfussenegger/nvim-jdtls",
     dependencies = { "folke/which-key.nvim" },
     ft = java_filetypes,
+    opts = function()
+      return {
+        -- How to find the root dir for a given filename. The default comes from
+        -- lspconfig which provides a function specifically for java projects.
+        root_dir = require("lspconfig.server_configurations.jdtls").default_config.root_dir,
+
+        -- How to find the project name for a given root dir.
+        project_name = function(root_dir)
+          return root_dir and vim.fs.basename(root_dir)
+        end,
+
+        -- Where are the config and workspace dirs for a project?
+        jdtls_config_dir = function(project_name)
+          return vim.fn.stdpath("cache") .. "/jdtls/" .. project_name .. "/config"
+        end,
+        jdtls_workspace_dir = function(project_name)
+          return vim.fn.stdpath("cache") .. "/jdtls/" .. project_name .. "/workspace"
+        end,
+
+        -- How to run jdtls. This can be overridden to a full java command-line
+        -- if the Python wrapper script doesn't suffice.
+        cmd = { "jdtls" },
+        full_cmd = function(opts)
+          local fname = vim.api.nvim_buf_get_name(0)
+          local root_dir = opts.root_dir(fname)
+          local project_name = opts.project_name(root_dir)
+          local cmd = vim.deepcopy(opts.cmd)
+          if project_name then
+            vim.list_extend(cmd, {
+              "-configuration",
+              opts.jdtls_config_dir(project_name),
+              "-data",
+              opts.jdtls_workspace_dir(project_name),
+            })
+          end
+          return cmd
+        end,
+
+        -- These depend on nvim-dap, but can additionally be disabled by setting false here.
+        dap = { hotcodereplace = "auto", config_overrides = {} },
+        test = true,
+      }
+    end,
     config = function()
       local opts = Util.opts("nvim-jdtls") or {}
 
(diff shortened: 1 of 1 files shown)


agriffis:
In particular, this means that the user can configure a replacement jdtls command:

```lua
return {
  "mfussenegger/nvim-jdtls",
  opts = {
    jdtls = function(config)
      config.cmd = { "java", ... }
    end,
  },
}
```

agriffis:
@aqav @appelgriebsch I made another pass to move defaults into opts. Let me know if you can find the time to review!

aqav:
Good work, I have tested the fixes from https://github.com/LazyVim/LazyVim/pull/1300 and added, they all work as expected

I really like that you allow users to override the config, like `cmd`, excellent move

There are little questions I meet when I test:

1. I am not really sure if the user needs to  invoke the `attach_jdtls` in row 224, I deleted this row and tested, and everything is still fine, so how to reproduce the "race condition and autocmd won't file' you mentioned
2. When I enter a Java file in a project, the file type autocmd is invoked multiple times(If delete 224, it's invoked 2 times, if not, it's invoked 3 times), but only occurs in the first buffer, subsequent buffers do not have this problem
3. When I enter a single java file(like a Hello.java) that is not in a project, I notice there is a warning message from lsp "msg_show Could not resolve java executable: Index 1 out of bounds for length 1"

The 2 and 3 also occur in my version https://github.com/LazyVim/LazyVim/pull/1300.

agriffis:
>     1. I am not really sure if the user needs to  invoke the `attach_jdtls` in row 224, I deleted this row and tested, and everything is still fine, so how to reproduce the "race condition and autocmd won't file' you mentioned

This does not happen for me.

>     2. When I enter a Java file in a project, the file type autocmd is invoked multiple times(If delete 224, it's invoked 2 times, if not, it's invoked 3 times), but only occurs in the first buffer, subsequent buffers do not have this problem

Your filetype auto-commands are running twice for some reason. This bug is why item 1 appears to work for you.

>     3. When I enter a single java file(like a Hello.java) that is not in a project, I notice there is a warning message from lsp "msg_show Could not resolve java executable: Index 1 out of bounds for length 1"

This doesn't happen for me.

I should mention that I am testing this personally by installing it as `.config/nvim/lua/plugins/java.lua` instead of running a fork of LazyVim. I doubt that matters, but for the sake of comparing our setups.

Could you share your full neovim configuration and version?

agriffis:
If you'd like to look at mine, it's here:

https://github.com/agriffis/skel/tree/master/neovim/.config/nvim

aqav:
My neovim version is v0.9.1, I test it when installing it in `.config/nvim/lua/plugins/java.lua` or `.local/share/nvim/lazy/LazyVim/lua/lazyvim/plugins/extras/lang/java.lua`, and test it in a pure new install Lazyvim, got same issue

I think it's very likely that this is due to the related issues you mentioned, but it is not your fault 

Anyway, please go on with your PR process

agriffis:
@folke I think you could merge this now. It's a pretty big improvement over the original java extra.

folke:
Thanks!

PR #1411, opened 2023-09-03
feat(ui): add dashboard animation

MattiasMTS:
# 🧑‍🏭 What has changed 🧑‍🏭

Just added a tweak to the awesome logo by having it as an animation that updates the sleeping part. 

I have this as my default with other animations and just felt like it would be cool to share it here. If it is not preferable to have it in the default setup, then I'm happy to share it in the Wiki/Discussions or so for inspo/knowledge-sharing! 😋 

--- 

# 🎥  Preview of how it looks like 🎥 :

https://github.com/LazyVim/LazyVim/assets/86059470/4fc42e9b-1e44-4f1a-a97e-1f8c0e2453fa

### DIFF
--- a/lua/lazyvim/plugins/ui.lua
+++ b/lua/lazyvim/plugins/ui.lua
@@ -279,16 +279,67 @@ return {
     event = "VimEnter",
     opts = function()
       local dashboard = require("alpha.themes.dashboard")
-      local logo = [[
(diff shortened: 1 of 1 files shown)


folke:
This looks nice, but not really something I want to add to LazyVim.

Feel free to post it in the show and tell discussion forum instead!

PR #1424, opened 2023-09-07
chore: fix template for repro.lua

MoetaYuko:


dpetka2001:
Why do you want this change? It doesn't make sense to me, as the template for `repro.lua` is supposed to be minimal and only test for potential bugs against Lazyvim, not the plugins that it incorporates.

MoetaYuko:
oh ok, closing then

dpetka2001:
I'm under no circumstances trying to tell you what to do. I'm just a regular user like you and not the maintainer of this project. I just asked a question so that you could clarify your position. You could let this open and wait for the maintainer to decide what is the proper decision (after he comes back from vacation at the end of September) or close it yourself like you did, but only if you think this is the right decision because you yourself think so, not because of me. I meant no offense with my previous comment, just expressed my opinion based on the logic of my understanding.

MoetaYuko:
> Why do you want this change? It doesn't make sense to me, as the template for `repro.lua` is supposed to be minimal and only test for potential bugs against Lazyvim, not the plugins that it incorporates.

I simply don't know the scope of `repro.lua`, since the template was written *before* https://github.com/LazyVim/LazyVim/commit/aa6ba297722b20fd47f713d1885f7902157ddd36 that disables "direct" use of this repo.

dpetka2001:
It's just a minimal configuration to test for bugs against Lazyvim itself. If you can reproduce something with this template, then it's likely a Lazyvim bug. The plugins are left out of scope of `repro.lua` because if the bug you're experiencing is caused by some plugin, then it would be best to make an issue to the corresponding plugin's repo. That is my understanding at least.

Issue #1413, opened 2023-09-04
bug: textobject-based tag block selection breaks with mini.ai enabled


idnsunset:
### Did you check docs and existing issues?

- [X] I have read all the LazyVim docs
- [X] I have searched the existing issues of LazyVim
- [X] I have searched the existing issues of plugins related to this issue

### Neovim version (nvim -v)

v0.9.0

### Operating system/version

Ubuntu 20.04

### Describe the bug

With `mini.ai` enabled, textobject-based tag selection in XML file selects unexpected area but not the text between the pair of tags **that contain underscore**.

### Steps To Reproduce

1. Open a XML file
2. Move the cursor to any text between a pair of tags that contain underscore, e.g. `<first_name>First Name</first_name>` 
3. Press `vit`

### Expected Behavior

The text between the pair of tags gets selected.

madjxatw:
Hi, @abeldekat, looks pretty similar. If it is confirmed a bug of mini.ai, this plugin would better be disabled by default.

abeldekat:
I also think the issue should be registered in `mini.ai`. 

I don't think the plugin should be disabled by default, as @madjxatw suggested. There are several ways to disable the plugin on a per buffer/ per filetype basis, described [here](https://github.com/echasnovski/mini.nvim/blob/3a3e19d7b7850e711485b8adde73ba44789f2cf6/doc/mini.txt#L270).

For example, a user can create file `after/ftplugin/xml.lua`, containing:

``` lua
vim.b.miniai_disable = true
```

abeldekat:
@madjxatw 

In my opinion, this is not a buggy plugin. Its limitations are clearly defined in the docs. See also this [comment](https://github.com/echasnovski/mini.nvim/issues/156#issuecomment-1308429031).

dpetka2001:
As @abeldekat noted, the maintainer of the plugin himself has clarified his position as to expect certain false positives since Lua pattern regex engine is quite limited. This is due to the language not being able to offer the more granual control as other languages offer in regards to regex engines. You are welcome to open an issue at the plugin's repo to ask the maintainer if something more can be done in order to solve this issue. Asking to disable this plugin is also a valid opinion of yours, but it's up to the maintainer of Lazyvim to decide on this since he has more experience in regards to Neovim and its ecosystem programming environment. You can easily just disable the plugin in your own configuration if it doesn't meet your needs. Simple as that.

madjxatw:
[...]

Issue #1217, opened 2023-07-25
bug: vat/vit yit/yat etc don't work as intended 


StitiFatah:
### Did you check docs and existing issues?

- [X] I have read all the LazyVim docs
- [X] I have searched the existing issues of LazyVim
- [x] I have searched the existing issues of plugins related to this issue

### Neovim version (nvim -v)

0.9.1

### Operating system/version

Archlinux

### Describe the bug

 vat/vit  and other at/it combination  don't work like normally in neovim, if the tags are spread into multiple lines they seem to select every parent nodes and not just what's inside the tags  

### Steps To Reproduce

try the `vat` motion on a fresh lazyvim config

### Expected Behavior

Normal `it/at` behavior

### Repro

Here are two screen recording with the `vat` motion 

With stock lazyvim :  https://github.com/LazyVim/LazyVim/assets/68331131/653a4849-4a4b-469c-a12a-a86ad7677d65

With stock neovim :  https://github.com/LazyVim/LazyVim/assets/68331131/20886e4c-6501-47e2-939b-71b52730baea

loqusion:
I use the following overrides in my `mini.ai` config, they have worked well for me in the past:

```lua
{
  custom_textobjects = {
    t = { "<(%w-)%f[^<%w][^<>]->.-</%1>", "^<.->%s*().-()%s*</[^/]->$" },
    T = { "<(%w-)%f[^<%w][^<>]->.-</%1>", "^<.->().*()</[^/]->$" },
  },
}
```

folke:
right, report to mini.ai if you think this is a bug

dpetka2001:
This is a known limitation of `mini.ai`, as the maintainer himself has also admitted (because `mini.ai` uses lua patterns and some false positives are expected to exist). He recommends the [answer])https://github.com/echasnovski/mini.nvim/issues/110#issuecomment-1212277863) in this thread. You could take a look and see if it works for you. The answer is for `.html` files, but you could customize it for other filetypes you work with as well I imagine.

folke:
LazyVim already has that treesitter mapping, so instead of vit, one could do vif I think

StitiFatah:
[...]

PR #1434, opened 2023-09-10
feat(coding)!: Move mini.ai to extras. Add default textobjects to treesitter

abeldekat:
### Resources:

[bug: textobject-based tag block selection breaks with mini.ai enabled #1413](https://github.com/LazyVim/LazyVim/issues/1413)
[bug: vat/vit yit/yat etc don't work as intended #1217](https://github.com/LazyVim/LazyVim/issues/1217)
[mini.ai: tag selection breaks when tag contains an underscore #474](https://github.com/echasnovski/mini.nvim/issues/474)

### Motivation:

All thoughts leading to this PR can be found in #1413.
An alternative for this PR is also provided: [mini.ai disable tag textobject](https://github.com/LazyVim/LazyVim/pull/1444)

Quoting myself:

> Quoting my reply to @madjxatw:
>
> > I don't think the plugin should be disabled by default, as @madjxatw suggested
>
> For completeness: I changed my mind. My preference would be to have treesitter-textobjects fully configured in the core of LazyVim, and _opt-in_ to mini.ai in extras.
>
> Mini.ai is a high quality, powerful plugin. However, "it enhances some builtin textobjects". There are benefits to this approach, but also some disadvantages, as [stated](https://github.com/echasnovski/mini.nvim/issues/156#issuecomment-1308429031) by the author.
>
> I was considering a pull request for `LazyVim`, preventing mini.ai from taking over the `tag` textobject:
>
> ```lua
> {
>     "echasnovski/mini.ai",
>     opts = function(_, opts)
>       opts.custom_textobjects = {
>         t = false, -- fallback to neovim for tags
>       }
>     end,
> }
> ```
>
> That would fix this issue, and also the issue referenced [here](https://github.com/LazyVim/LazyVim/issues/1413#issuecomment-1705238732)
>
> However, there might be more enhancements up for debate. As an example, consider:
>
> `local msg = "hello world"`
>
> Position the cursor on the "w", and press `vi"`. Now, the cursor is positioned on the first word, the letter "h". Without mini.ai, the cursor would be positioned on the last word, the letter "d".

Also:
> In the end, the main motivation is the enhanced builtin textobjects, because of the changes to the core behavior of neovim's textobjects. In my personal opinion, for a setup like LazyVim, that should be opt-in.

### Approach

I used the same approach as chosen in the change from `leap.nvim` to `flash.nvim`, in this [commit](https://github.com/LazyVim/LazyVim/commit/ae759b947b1ef16d9814fcddfcafe2cdd767bc6a).
This PR is also breaking.
Users should add the `extra` to their config, in order to opt-in to `mini.ai`.  In that case, all newly added `textobjects` in `lazyvim.plugins.treesitter` are overwritten.

In case this PR is accepted, I can also write a PR for the website

### Additional changes

I disabled the `tag` textobject of `mini.ai`:

```lua
opts.custom_textobjects = {
  -- opt-out: using neovim's text object selection for tags:
  t = false,
  -- opt-in: use the following override in the "opts" function of your spec:
  -- opts.custom_textobjects.t = nil -- opt-in to mini.ai's tag textobject
```

This fixes issues #1413 and #1217

### Testing

I tested the following scenario's, using this [setup](https://github.com/abeldekat/starter/tree/miniai_to_extras):

1. No mini.ai in config. The new textobjects are working.
2. mini.ai in config, having enabled = false. Warning does not appear, as expected
3. mini.ai in config, having enabled = true. Warning does not appear!
4. mini.ai in config, but not installed. Warning appears as expected
5. mini.ai in config, and installed. Warning appears, the UI of `lazy.nvim` offers to clean.
6. mini.ai in config, added extra.
7. mini.ai in config, added extra, modified config to use `mini.ai`'s `tag` `textobject`.
8. mini.ai in config, added extra, added `swap` textobjects to treesitter

The third test case fails, because the user has set the enabled property explicitly to true.
As a consequence, his `mini.ai` spec will lack `LazyVim`'s `mini.ai`. I think this scenario would also have failed when switching from `leap` to `flash`. 

### Related discussions:
[broken text object in tsx files](https://github.com/LazyVim/LazyVim/discussions/1341)
[Deleting inside Tag](https://github.com/LazyVim/LazyVim/discussions/778)
[visually select jsx html tags](https://github.com/LazyVim/LazyVim/discussions/500)

### DIFF
--- a/lua/lazyvim/plugins/extras/coding/mini-ai.lua
+++ b/lua/lazyvim/plugins/extras/coding/mini-ai.lua
@@ -4,7 +4,7 @@ local load_textobjects = false
 local function source_runtime_textobjects(ts_opts)
   -- PERF: no need to load the plugin, if we only need its queries for mini.ai
 
-  if ts_opts.textobjects then -- the user defines textobjects
+  if ts_opts.textobjects then -- the user defined textobjects
     for _, mod in ipairs({ "move", "select", "swap", "lsp_interop" }) do
       if ts_opts.textobjects[mod] and ts_opts.textobjects[mod].enable then
         local Loader = require("lazy.core.loader")
@@ -17,20 +17,15 @@ local function source_runtime_textobjects(ts_opts)
   end
 end
 
-local function spec_decorate_config_property()
-  -- avoid code duplication:
-  -- decorate config function defined in lazyvim.plugins.treesitter
+---@param treesitter_plugin LazyPlugin
+local function decorate_config_function(treesitter_plugin)
+  -- the config function is defined in lazyvim.plugins.treesitter
+  local config_to_decorate = treesitter_plugin.config
 
-  local Config = require("lazy.core.config")
-
-  ---@type LazyPlugin
-  local spec_treesitter = Config.spec.plugins["nvim-treesitter"]
-  local config_to_decorate = spec_treesitter.config
-  spec_treesitter.config = function(_, opts)
+  treesitter_plugin.config = function(_, opts)
     if config_to_decorate then
-      config_to_decorate(spec_treesitter, opts)
+      config_to_decorate(treesitter_plugin, opts)
     end
-
     if load_textobjects then
       source_runtime_textobjects(opts)
     end
@@ -40,8 +35,8 @@ end
 return {
   {
     "nvim-treesitter/nvim-treesitter",
-    init = function()
-      spec_decorate_config_property()
+    init = function(plugin)
+      decorate_config_function(plugin) -- avoiding code duplication
     end,
     dependencies = {
       {
@@ -66,7 +61,7 @@ return {
   -- Better text-objects
   {
     "echasnovski/mini.ai",
-    enabled = true,
+    enabled = true, -- overrides the function defined in lazyvim.plugins.coding
     -- keys = {
     --   { "a", mode = { "x", "o" } },
     --   { "i", mode = { "x", "o" } },
--- a/lua/lazyvim/plugins/extras/coding/mini-ai.lua
+++ b/lua/lazyvim/plugins/extras/coding/mini-ai.lua
@@ -1,24 +1,48 @@
 local load_textobjects = false
 
----@param langs table<string>
----@return table<string>
-local function dedup(langs)
-  -- TODO: move this utility function to a better location
-  ---@type table<string, boolean>
-  local added = {}
-  return vim.tbl_filter(function(lang)
-    if added[lang] then
-      return false
+---@param ts_opts PluginOpts
+local function source_runtime_textobjects(ts_opts)
+  -- PERF: no need to load the plugin, if we only need its queries for mini.ai
+
+  if ts_opts.textobjects then -- the user defines textobjects
+    for _, mod in ipairs({ "move", "select", "swap", "lsp_interop" }) do
+      if ts_opts.textobjects[mod] and ts_opts.textobjects[mod].enable then
+        local Loader = require("lazy.core.loader")
+        Loader.disabled_rtp_plugins["nvim-treesitter-textobjects"] = nil
+        local plugin = require("lazy.core.config").plugins["nvim-treesitter-textobjects"]
+        require("lazy.core.loader").source_runtime(plugin.dir, "plugin")
+        break
+      end
     end
-    added[lang] = true
-    return true
-  end, langs)
+  end
+end
+
+local function spec_decorate_config_property()
+  -- avoid code duplication:
+  -- decorate config function defined in lazyvim.plugins.treesitter
+
+  local Config = require("lazy.core.config")
+
+  ---@type LazyPlugin
+  local spec_treesitter = Config.spec.plugins["nvim-treesitter"]
+  local config_to_decorate = spec_treesitter.config
+  spec_treesitter.config = function(_, opts)
+    if config_to_decorate then
+      config_to_decorate(spec_treesitter, opts)
+    end
+
+    if load_textobjects then
+      source_runtime_textobjects(opts)
+    end
+  end
 end
 
 return {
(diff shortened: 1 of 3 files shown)


lua/lazyvim/plugins/extras/coding/mini-ai.lua
```diff
@@ -0,0 +1,139 @@
...
+    opts = function(_, opts)
+      -- override the default textobjects, set in plugins/treesitter.lua
+      -- when using this extra, the user is expected to add textobjects
+      -- on a need-to-have basis
+      opts.textobjects = nil
+    end,
+    ---@param opts TSConfig
+    config = function(_, opts) -- override config in plugins/treesitter.lua
+      if type(opts.ensure_installed) == "table" then
+        opts.ensure_installed = dedup(opts.ensure_installed)
```
dpetka2001, line 43: @abeldekat I don't believe you need the `dedup` function here as it's already handled internally by [Lazyvim](https://github.com/LazyVim/LazyVim/blob/a72a84972d85e5bbc6b9d60a0983b37efef21b8a/lua/lazyvim/plugins/treesitter.lua#L64-L75)?

abeldekat: @dpetka2001,

I appreciate your input!

The link you used points to the old code in `treesitter.lua`, now present in the `dedup` function in both `treesitter.lua` and the `extra`. This explicitly marks the code duplication.
See [here](https://github.com/LazyVim/LazyVim/blob/0c12cc332a7e6f260b8dec361b5474742b432a3a/lua/lazyvim/plugins/treesitter.lua#L3)

In the `extra`, I have to *override* the config function defined in core `treesitter.lua`. Still thinking about a better way...

Best regards!

dpetka2001: Oh, you're right. Didn't pay attention that you're also overriding the treesitter config in the extras. Essentially you have 2 different treesitter configs depending on if `mini.ai` is enabled or not. I see. Tnx for clarifying.

dpetka2001: @abeldekat I may be talking rubbish, but here is an idea. What if you left the original code in core `treesitter.lua` like it was, but removing the `init` function from `nvim-treesitter-textobjects`? That ways the code for PERF would never get loaded since `load_textobjects` would always be false. Then in the extras you change in `init` function `load_textobjects` to `true`. Would that help getting rid of having to override the treesitter `config` in the extras as well?

abeldekat: @dpetka2001,

Thanks for the suggestion. Inside the `extra`, I would need to access the `load_textobjects` variable defined in `lazyvim.plugins.treesitter`. That is a spec, thus the variable is not within reach, as far as I can see.

However, your comment was very helpful. I found a way to decorate the `config` function in `extra`, thus avoiding any code duplication. Maybe you could inspect the new [commit](https://github.com/LazyVim/LazyVim/pull/1434/commits/cdf8036e7fd2f1cada5c488e5319a26cc71fb33c#diff-597497ffb8a9f32f2e151a7d3ad8b479b0f3999c442d02a55368778282f3f3a7)?

dpetka2001: @abeldekat 

Thank you for asking my opinion, however I'm going to disappoint you due to lack of programming knowledge. I have read through your code. I believe I understand what it's doing at a pretty good level (for my standards at least). From my understanding it seems good, but you definitely should defer your review request to someone experienced like Folke. I shouldn't be the one reviewing. I can only offer some ideas based on how I can logically deduce some things, but even then I do mistakes like the local variable not being able to be accessed across specs. I really admire the work you put in regardless of the outcome of this PR. And I honestly appreciate it you wanting to include me in something like this, even though you absolutely don't have to. I really hope I had the knowledge to offer you more concrete and valuable feedback, but I'm not at that level. I can only read and understand some code due to my effort to be helpful to this project because I like it. I tried to inspect your PR as best as I could, but I can under no circumstances offer a review opinion on this. If it was something simpler, I could do so, but I know my limits and don't want to offer information on something I can barely understand.
Wish you luck with this :smile:

abeldekat: I liked your idea. Thanks, @dpetka2001!

cStralpt:
i just update my config by adding the configuration:
```
return {
  "echasnovski/mini.ai",
  opts = function(_, opts)
    opts.custom_textobjects = {
      t = false, -- fallback to neovim for tags
    }
  end,
}
```

now i can use vat/vit etc normally, but after that i can't use vaf/vif, dif/daf and yif/yif
`
(mini.ai) No textobject "af" found covering region within 500 lines and `search_method = 'cover_or_next'`.
`

dpetka2001:
@cStralpt I believe you should use `vim.tbl_deep_extend` to extend the default `custom_textobjects` table, because this way I believe you're overwriting it.

abeldekat:
@cStralpt, 

In order to *add* the `t` to `LazyVim`'s `opts.custom_textobjects`, you can use:
You can use:
```lua
    {
      "echasnovski/mini.ai",
      opts = function(_, opts)
        opts.custom_textobjects["t"] = false
      end,
    },
```

dpetka2001:
@abeldekat So just defining the key/value like this will add it to the table? I thought `vim.tbl_deep_extend` would be needed to extend the table. Good to know this way is also possible. Tnx a lot :smile:

folke:
`mini.ai` remains a core plugin.

Closing in favor of https://github.com/LazyVim/LazyVim/commit/7c79cfe07314245decd73ec67e7247bba22abaa9

PR #1446, opened 2023-09-13
fix(python): ensure debugpy installed if using nvim-dap

imroc:


dpetka2001:
There is no need for that. `debugpy` already gets automatically installed. It's just that `mason-nvim-dap` is lazy loaded, so if you set a breakpoint inside a python file, it will load and start installing `debugpy`.

imroc:
Oh I see, because I am using a containerized development environment, every time I pull up a new environment and open nvim for the first time, I will get the error that nvim-dap-python cannot find debugy, but it will not happen the second time.

PR #1450, opened 2023-09-14
feat(lang): added php support

hacktivista:
Configures tree-sitter, phpactor as LSP server and xdebug/vscode-php-debug as DAP server.

### DIFF
--- a/lua/lazyvim/plugins/extras/lang/php.lua
+++ b/lua/lazyvim/plugins/extras/lang/php.lua
@@ -1,0 +1,47 @@
+return {
+  {
+    "nvim-treesitter/nvim-treesitter",
+    opts = function(_, opts)
+      vim.list_extend(opts.ensure_installed, {
+        "php",
+      })
+    end,
+  },
+  {
+    "williamboman/mason.nvim",
+    opts = function(_, opts)
+      vim.list_extend(opts.ensure_installed, {
+        "phpactor",
+      })
+    end,
+  },
+  {
+    "neovim/nvim-lspconfig",
+    opts = {
+      servers = {
+        phpactor = {},
+      },
+    },
+  },
+  {
+    "mfussenegger/nvim-dap",
+    optional = true,
+    dependencies = {
+      "williamboman/mason.nvim",
+      opts = function(_, opts)
+        if type(opts.ensure_installed) == "table" then
+          table.insert(opts.ensure_installed, "php-debug-adapter")
+        end
+      end,
+    },
+    opts = function()
+      local dap = require("dap")
+      local path = require("mason-registry").get_package("php-debug-adapter"):get_install_path()
+      dap.adapters.php = {
+        type = "executable",
+        command = "node",
+        args = { path .. "/extension/out/phpDebug.js" },
+      }
+    end,
+  },
+}
--- a/lua/lazyvim/plugins/extras/lang/php.lua
+++ b/lua/lazyvim/plugins/extras/lang/php.lua
@@ -1,20 +1,16 @@
 return {
+  recommended = {
+    ft = "php",
+    root = { "composer.json", ".git", ".phpactor.json", ".phpactor.yml" },
+  },
+
   {
     "nvim-treesitter/nvim-treesitter",
     opts = function(_, opts)
-      vim.list_extend(opts.ensure_installed, {
-        "php",
-      })
+      vim.list_extend(opts.ensure_installed, { "php" })
     end,
   },
-  {
-    "williamboman/mason.nvim",
-    opts = function(_, opts)
-      vim.list_extend(opts.ensure_installed, {
-        "phpactor",
-      })
-    end,
-  },
+
   {
     "neovim/nvim-lspconfig",
     opts = {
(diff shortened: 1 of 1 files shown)


hacktivista:
I never wrote Lua before, nor ever configured NeoVim nor lazy.nvim. I just looked at other language configs and tried to do the same, it worked surprisingly well, but probably there are some things that could be improved, so please review it thoroughly.

Thanks for this project!

willopez:
It would be great to have this feature!

madflow:
This is one flavour of PHP suppport, which would be highly appreciated. However: The more mature LSP server is in my opinion "intelephense". phpactor is great, too. Maybe this can made configurable? or optional? It is a matter of taste, I suppose.

For testing there is also a neotest adapter, which would fit into the Lazyvim ecosystem:

* https://github.com/olimorris/neotest-phpunit

willopez:
I have created a custom plugin using the config on this PR and it's working great! I prefer to use Phpactor as it's free and has many features. For anybody interested, here is my config. https://github.com/willopez/LazyVim/blob/main/lua/plugins/php-lsp.lua

gmnz:
I work as a web dev (using php a lot) and found this combination to be working pretty well:

- lsp: intelephense (it has limited features if you don't buy a license, but I've tried the alternatives and they didn't seem to work properly after installed by Mason)
- linter: phpcs (first one I got to work and was able to set a coding standard using args)
- formatter: pretty-php (first one I got to work and was able to set a coding standard args)

I've set both formatter and linter to follow psr12 so that the linter doesn't complain about how the formatter is formatting.

Just wanted to share my experience and maybe get some inspiration for improving my config from other php devs. :-)

Tmeister:
@gmnz 

You may want to share your configuration, so everyone interested can take a look.

gmnz:
@Tmeister 

```lua
return {
  "mfussenegger/nvim-lint",
  opts = function(_, opts)
    local phpcs = require("lint").linters.phpcs
    phpcs.args = {
      "-q",
      "--standard=PSR12",
      "--report=json",
      "-",
    }
    --
    opts.linters_by_ft = vim.tbl_extend("force", opts.linters_by_ft, {
      php = { "phpcs" },
    })

  end,
}
```

```lua
return {
  {
    "stevearc/conform.nvim",
    opts = function(_, opts)
      opts.formatters_by_ft = vim.tbl_extend("force", opts.formatters_by_ft, {
        php = { { "pretty-php", }, },
      })

      opts.formatters = vim.tbl_extend("force", opts.formatters, {
        ["pretty-php"] = {
          prepend_args = {
            "--psr12",
          },
        },
      })
    end,
  },
}
```

Any feedback on this is welcome, I'm not an expert on lua nor linters, nor formatters. :-)

MystikReasons:
Are there any updates regarding PHP support?

willopez:
I use a lot of PHP, WordPress and for other projects, it would be great to have PHP support! @folke thanks a lot for your work!

folke:
Updated and merged. Thanks!

esetnik:
@folke what's the recommended approach for language extras that have multiple good variants for lsp? For example, the lang extra now merged in this PR uses `phpactor` but a lot of php devs use `intelephense` instead. I tested it out today and the two seem to be able to coexist peacefully (e.g. they can both be enabled simultaneously) although i'm not sure about whether this is actually a good idea. 

It feels a bit weird to me that the LazyVim community would need to be opinionated about which lsp is the best for a particular language. Is there a value to having language variants in extras, e.g. `lang.php-phpactor` and `lang.php-intelephense`? Are there other languages which have multiple good lsp implementations that should be considered?

dpetka2001:
You can take a look at [Python Extra](https://github.com/LazyVim/LazyVim/blob/main/lua/lazyvim/plugins/extras/lang/python.lua) about taking into considerations different variants of which LSP server will be used. Of course the users can add whatever they want to their personal configurations as well. 

The Extras are mostly best effort setups in my opinion and it depends on the PR creator's experience with the language as well. It's not a be-all/end-all kind of thing. 

If you think you can improve some Extra, then just make a PR and it will either be accepted or not.

folke:
Like @dpetka2001 says.

Feel free to submit a PR similar to how it's done in the pyhthon extra.

Ideally the php extra should default to the best LSP available, but since I don't do php anymore, I'm not the best person to decide on that.

folke:
@esetnik if intelephense is better or more widely used, then feel free to make it the default

muneebusmani:
[LSP] Format request failed, no matching language servers.
it says it all the time

dpetka2001:
That's because no formatting has been set up for this Extra. Feel free to improve it.

hacktivista:
Many people use intelephense and phpactor. I prefer phpactor because it uses a libre license, and does not require buying one for unlocking features. IMHO proprietary software should never be a default on a FLOSS project.

muneebusmani:
It seems like there are 2 thing more to do in the this extra, configure formatting, Lin ing with the way python extra does, for formatting there could be 3 options ppl go for, php cs fixer, pint or pretty-php, and about linting there seems to be options in .phpactor.json for choosing phpstan and/or psalm, but there are other linters too like php codesniffer, tlint, etc

PR #1456, opened 2023-09-15
feat(mini-pairs): add toggle function

kevintraver:
Add function and keymap to toggle mini pairs.

### DIFF
--- a/lua/lazyvim/plugins/coding.lua
+++ b/lua/lazyvim/plugins/coding.lua
@@ -97,6 +97,21 @@ return {
     "echasnovski/mini.pairs",
     event = "VeryLazy",
     opts = {},
+    keys = {
+      {
+        "<leader>up",
+        function()
+          local Util = require("lazy.core.util")
+          vim.g.minipairs_disable = not vim.g.minipairs_disable
+          if vim.g.minipairs_disable then
+            Util.warn("Disabled auto pairs", { title = "Option" })
+          else
+            Util.info("Enabled auto pairs", { title = "Option" })
+          end
+        end,
+        desc = "Toggle auto pairs",
+      },
+    },
   },
 
   -- Fast and feature-rich surround actions. For text that includes


lua/lazyvim/config/keymaps.lua
```diff
@@ -110,6 +110,7 @@ map("n", "<leader>uf", require("lazyvim.plugins.lsp.format").toggle, { desc = "T
 map("n", "<leader>us", function() Util.toggle("spell") end, { desc = "Toggle Spelling" })
 map("n", "<leader>uw", function() Util.toggle("wrap") end, { desc = "Toggle Word Wrap" })
 map("n", "<leader>ul", function() Util.toggle_number() end, { desc = "Toggle Line Numbers" })
+map("n", "<leader>up", function() Util.toggle_autopairs() end, { desc = "Toggle Auto Pairs" })
```
jyuan0, line 113:
```lua
if Util.has("mini.pairs") then
  map("n", "<leader>up", function() Util.toggle_autopairs() end, { desc = "Toggle Auto Pairs" })
end
```

Think it would be nice to add a check in case the plugin is disabled.

lua/lazyvim/util/init.lua
```diff
@@ -200,6 +200,18 @@ function M.toggle_number()
   end
 end
 
+function M.toggle_autopairs()
+  if package.loaded["mini.pairs"] then
```
jyuan0, line 204: But this way, the key/mapping is still set even if the plugin is disabled. I figured by guarding the mapping, the mapping wouldn't be created and it's unlikely the toggle function is called, unless explicitly called.

kevintraver: Ah yes, I will move it to the keymap.

Would it be better to use `package.loaded["mini.pairs"]` or `Util.has("mini.pairs") `?

lua/lazyvim/util/init.lua
```diff
@@ -200,6 +200,18 @@ function M.toggle_number()
...
+    if vim.g.minipairs_disable then
+      vim.g.minipairs_disable = false
+      Util.info("Enabled auto pairs", { title = "Option" })
+    else
+      vim.g.minipairs_disable = true
+      Util.warn("Disabled auto pairs", { title = "Option" })
```
jyuan0, line 210: Maybe `INFO` level would be better here since this isn't unexpected or an error? The user is manually toggling the state/functionality.

lua/lazyvim/util/init.lua
jyuan0: I think in `keymaps.lua`, better to use `Util.has` like the other plugin checks in that file.

lua/lazyvim/util/init.lua
kevintraver: I was just following the same pattern as `toggle_number` and `toggle_diagnostics`. They both use `warn` when disabling.

jyuan0: Ah ok, didn't realize that was the precedent.

lua/lazyvim/config/keymaps.lua
```diff
@@ -110,6 +110,11 @@ map("n", "<leader>uf", require("lazyvim.plugins.lsp.format").toggle, { desc = "T
 map("n", "<leader>us", function() Util.toggle("spell") end, { desc = "Toggle Spelling" })
 map("n", "<leader>uw", function() Util.toggle("wrap") end, { desc = "Toggle Word Wrap" })
 map("n", "<leader>ul", function() Util.toggle_number() end, { desc = "Toggle Line Numbers" })
+map("n", "<leader>up", function()
```
jyuan0, line 113: Think it should be:
```lua
if Util.has("mini.pairs") then
  map("n", "<leader>up", function() Util.toggle_autopairs() end, { desc = "Toggle Auto Pairs" })
end
```
No need/use to map the key if the plugin is disabled.

lua/lazyvim/config/keymaps.lua
```diff
@@ -110,6 +110,9 @@ map("n", "<leader>uf", require("lazyvim.plugins.lsp.format").toggle, { desc = "T
 map("n", "<leader>us", function() Util.toggle("spell") end, { desc = "Toggle Spelling" })
 map("n", "<leader>uw", function() Util.toggle("wrap") end, { desc = "Toggle Word Wrap" })
 map("n", "<leader>ul", function() Util.toggle_number() end, { desc = "Toggle Line Numbers" })
+if Util.has("mini.pairs") then
+  map("n", "<leader>up", function() Util.toggle_autopairs() end, { desc = "Toggle Auto Pairs" })
+end
```
jyuan0, line 115: Another possibility is adding this mapping and toggle function to the `mini.pairs` plugin spec, e.g. in a `config` function. It would remove the need for the plugin enabled/disabled check, but I'm not sure it's preferable/better. Just a thought.

kevintraver: Keeping them in once place probably makes sense. I will add it in the `keys` section of mini.pairs to make sure its shows up in which-key.

dpetka2001:
@kevintraver 

Couldn't we just do the following
```lua
return {
  {
    "echasnovski/mini.pairs",
    keys = {
      {
        "<leader>up",
        function()
          local Util = require("lazy.core.util")
          if vim.g.minipairs_disable then
            vim.g.minipairs_disable = false
            Util.info("Enabled auto pairs", { title = "Option" })
          else
            vim.g.minipairs_disable = true
            Util.warn("Disabled auto pairs", { title = "Option" })
          end
        end,
        desc = "Toggle mini.pairs",
      },
    },
  },
}
```
Why introduce new code of other plugins in `lazyvim.util`? This way the change will be entirely dependent on the plugin itself.

jyuan0:
> @kevintraver
> 
> Couldn't we just do the following
> 
> ```lua
> return {
>   {
>     "echasnovski/mini.pairs",
>     keys = {
>       {
>         "<leader>up",
>         function()
>           local Util = require("lazy.core.util")
>           vim.g.minipairs_disable = not vim.g.minipairs_disable
>           if not vim.g.minipairs_disable then
>             Util.info("Enabled auto pairs", { title = "Option" })
>           else
>             Util.warn("Disabled auto pairs", { title = "Option" })
>           end
>         end,
>         desc = "Toggle mini.pairs",
>       },
>     },
>   },
> }
> ```
> 
> Why introduce new code of other plugins in `lazyvim.util`? This way the change will be entirely dependent on the plugin itself.

This is pretty much what I was thinking of too. The mapping and function can all be part of the plugin spec.

kevintraver:
Ok, I updated the keymap to be in the spec. I also switched the enabled/disabled logic and remove the `not` to make the logic more clear.

jyuan0 (commented):
:+1:

Only other improvement I can think of is a more descriptive title than `"Option"`, but it's minor.

jyuan0 (commented):
I think this looks fine. In case you're not aware, you'll have to wait until folke (the maintainer of LazyVim) returns from vacation for any additional feedback, and to get this merged.

folke:
thansk!

PR #1457, opened 2023-09-15
Update python.lua

haininhhoang94:
dap_enabled = true, -- Ensure that the venv selector affect PythonPath in nvim-dap as well! (in venv-selector.nvim)


folke:
thanks!

PR #1459, opened 2023-09-16
Add labels to telescope buffer keymap

latif-k:
Hello,

Right now when you list the keys in the telescope buffer (with "C_/"), some of the keys are mapped to anonymous functions, which makes it hard to see what those keys do. This PR fixes that.


latif-k:
This is already covered by https://github.com/LazyVim/LazyVim/pull/1294#issue-1838004376, my mistake.

PR #1484, opened 2023-09-24
feat(clojure): add clojure support

itsjoeoui:
Add a minimal Clojure support. 

The config is taken from:
https://github.com/Olical/conjure

Not a Clojure expert here haha. 

The only change I made is I removed the Python support from the default configuration, such that it will only work for Clojure and won't interfere with any Python functionalities. Has been working well on my end.


folke:
Closing in favor of https://github.com/LazyVim/LazyVim/pull/2179

PR #1487, opened 2023-09-24
fix(test): add missing neotest dependencies

ChristianMoesl:
There are some dependencies defined in the [installation section of the Neotest documentation](https://github.com/nvim-neotest/neotest#installation), which are missing here in the test extra. On my machine, the test extra wasn't working without these dependencies defined, so I went ahead and thought I could create a PR to fix this. 

I just started learning LazyVim today, so I am not totally sure if my reasoning is right here 😄


folke:
Must have been something else that didtn work, since plenary and treesitter are already included and fixcursorhold is no longer needed

PR #1498, opened 2023-09-26
chore: fix LazyVim url in bug_report.yml

abeldekat:
The repro specifies:
```lua
local plugins = {
  "folke/tokyonight.nvim",
  "folke/LazyVim",
  -- add any other plugins here
}
```

Adding `import="lazyvim.plugins"` leads to the following error:
> lazy.nvim Two plugins with the same name and different url:

This PR changes "folke/LazyVim" into "LazyVim/LazyVim"

On a sidenote, specifying just "LazyVim/LazyVim" merely loads the `LazyVim` plugin itself. The options, autocommands, keymaps and plugins are not loaded.
In the bug reports, I am guessing that 90% contains an unused `repro` fragment.

I created a more flexible [repro](https://github.com/LazyVim/LazyVim/discussions/1493), but I don't think it is suitable for the end user.

Perhaps, adding `import=lazyvim.plugins` is worth considering.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -75,7 +75,7 @@ body:
         -- install plugins
         local plugins = {
           "folke/tokyonight.nvim",
-          "folke/LazyVim",
+          "LazyVim/LazyVim",
           -- add any other plugins here
         }
         require("lazy").setup(plugins, {


folke:
Not home right now, but will have a look later today!

The reason I didn't merge this yet, is because all my plugins' CI/build files are autogenerated across all my plugin repos.

If I'd merge this, it would be overwritten again.

So i need to change that code a bit first 🙂

abeldekat:
In `lazyflex` v2.0 the example repro I mentioned in this [comment](https://github.com/LazyVim/LazyVim/pull/1498#issuecomment-1750157132) has been changed.

example repro:
```lua
-- Minimal `init.lua` to reproduce an issue. Save as `repro.lua` and run with `nvim -u repro.lua`

-- sets std paths to use .repro and bootstraps lazy
local function bootstrap(root) -- DO NOT change the paths
  end
    vim.fn.system({ "git", "clone", "--filter=blob:none", "https://github.com/folke/lazy.nvim.git", lazypath })
  end
  vim.opt.rtp:prepend(vim.env.LAZY or lazypath)
end
bootstrap(root)

-- install plugins
local plugins = {
  -- optional: reduce the number of plugins needed to reproduce the problem
  { "abeldekat/lazyflex.nvim", enabled = false, version = "*", import = "lazyflex.hook" },

  "folke/tokyonight.nvim",
  -- add any other plugins here
}
  root = root .. "/plugins",
})
-- add anything else here

```

PR #1512, opened 2023-09-28
fix(telescope): add enums and consts to search symbols

abusch:
I've been wondering for a while why I could never find my enums when writing Rust, so this should fix this. Also included constants for good measure.


folke:
Thanks!

PR #1517, opened 2023-09-28
feat: use none-ls instead of null-ls

### LINKED ISSUE #1516: feature:  use none-ls instead of null-ls
### Did you check the docs?

With null-ls being archived, none-ls is the community fork of it and should be used.

It's straightforward to use it:

```lua
	{
		'nvimtools/none-ls.nvim',
		opts = function(_, opts)
			local nls = require('null-ls')
			opts.sources = vim.list_extend(opts.sources or {}, {
				nls.builtins.code_actions.gitsigns,
				-- go
				nls.builtins.code_actions.gomodifytags,
				nls.builtins.code_actions.impl,
				nls.builtins.diagnostics.golangci_lint,
				-- ts
				nls.builtins.formatting.biome,
				require('typescript.extensions.null-ls.code-actions'),
				-- other
				nls.builtins.formatting.stylua,
				nls.builtins.formatting.shfmt,
			})
			return opts
		end,
	},

```

N/A

### Additional context

_No response_

OneOfOne:
Fixes #1516

### DIFF
--- a/lua/lazyvim/config/init.lua
+++ b/lua/lazyvim/config/init.lua
@@ -183,6 +183,14 @@ function M.init()
     local add = Plugin.Spec.add
     Plugin.Spec.add = function(self, plugin, ...)
       if type(plugin) == "table" and M.renames[plugin[1]] then
+        require("lazy.core.util").warn(
+          ("Plugin `%s` was renamed to `%s`.\nPlease update your config for `%s`"):format(
+            plugin[1],
+            M.renames[plugin[1]],
+            self.importing or "LazyVim"
+          ),
+          { title = "LazyVim" }
+        )
         plugin[1] = M.renames[plugin[1]]
       end
       return add(self, plugin, ...)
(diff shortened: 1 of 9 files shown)


lua/lazyvim/config/init.lua
```diff
@@ -79,6 +79,8 @@ local defaults = {
 
 M.renames = {
   ["windwp/nvim-spectre"] = "nvim-pack/nvim-spectre",
+  ["jose-elias-alvarez/null-ls.nvim"] = "nvimtools/none-ls.nvim",
+  ["null-ls.nvim"] = "none-ls.nvim",
```
folke, line 83: Added these renames, so that users don't end up with both `null-ls` and `none-ls`, in case they reference `null-ls` in their own setup.

OneOfOne: Perfect

folke:
Merged, thanks!

PR #1535, opened 2023-09-30
fix(extras.go): remove unneeded formatters since gopls handles them

### LINKED ISSUE #1533: feature: extras.go disable gofumpt
### Did you check the docs?

gopls already handles gofumpt, having gofumpt enabled in none-ls/conform.nvim does nothing except adding overhead and sometimes none-ls will get confused and leave extra temp files.

Remove gofumpt from none-ls/conform's settings in extras.go.

Override none-ls's sources in my personal config to disable it.

### Additional context

_No response_

OneOfOne:
Fixes #1533

### DIFF
--- a/lua/lazyvim/plugins/extras/lang/go.lua
+++ b/lua/lazyvim/plugins/extras/lang/go.lua
@@ -84,27 +84,15 @@ return {
   -- Ensure Go tools are installed
   {
     "nvimtools/none-ls.nvim",
-    optional = true,
     opts = function(_, opts)
       if type(opts.sources) == "table" then
         local nls = require("null-ls")
         vim.list_extend(opts.sources, {
           nls.builtins.code_actions.gomodifytags,
           nls.builtins.code_actions.impl,
-          nls.builtins.formatting.gofumpt,
-          nls.builtins.formatting.goimports_reviser,
         })
       end
     end,
-  },
-  {
-    "stevearc/conform.nvim",
-    optional = true,
-    opts = {
-      formatters_by_ft = {
-        go = { "goimports", "gofumpt" },
-      },
-    },
   },
   {
     "mfussenegger/nvim-dap",
@@ -114,7 +102,7 @@ return {
         "mason.nvim",
         opts = function(_, opts)
           opts.ensure_installed = opts.ensure_installed or {}
-          vim.list_extend(opts.ensure_installed, { "gomodifytags", "impl", "gofumpt", "goimports-reviser", "delve" })
+          vim.list_extend(opts.ensure_installed, { "gomodifytags", "impl", "delve" })
         end,
       },
       {
--- a/lua/lazyvim/plugins/extras/lang/go.lua
+++ b/lua/lazyvim/plugins/extras/lang/go.lua
@@ -84,6 +84,7 @@ return {
   -- Ensure Go tools are installed
   {
     "nvimtools/none-ls.nvim",
+    optional = true,
     opts = function(_, opts)
       if type(opts.sources) == "table" then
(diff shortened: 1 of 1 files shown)


lua/lazyvim/plugins/extras/lang/go.lua
```diff
@@ -84,28 +84,16 @@ return {
   -- Ensure Go tools are installed
   {
     "nvimtools/none-ls.nvim",
-    optional = true,
```
folke, line 87: `optional=true` needs to stay.

folke (changes requested):

folke:
Thanks!

delatorrejuanchi:
Either it's not working on my machine, or `gopls` does not handle `goimports` or `goimports_reviser` automatically.

To verify this, you can save a file which is missing imports and it should add imports for them. `gopls` will only add the imports if you accept the autocomplete prompt for the module while writing, but it will not do so on save. This means copy-pasted code will not get imports added on save, for example. `goimports` handles this.

`goimports_reviser` sorts imports and splits them in groups, but it doesn't seem to be handled by gopls either.

fredrikaverpil:
@delatorrejuanchi @OneOfOne Yeah, I don't think `gopls` handles this either and I wonder if this change was made in error...

I actually want `goimports` for adding missing imports on save. As `gopls` doesn't do this for you, I'm adding this back into [my own null-ls config](https://github.com/fredrikaverpil/dotfiles/blob/0bfc017c6db44c08c16dc2ec18226a3cc1b8c9e2/nvim-lazyvim/lua/plugins/lsp.lua#L133).

However, I also _don't_ want `goimports-reviser` (which was removed here). [I wasn't able to disable it](https://github.com/LazyVim/LazyVim/discussions/1538) from my own null-ls config (not sure why) but since it is now removed here, maybe I don't have to worry about that anymore...

Bonus: I also think the gopls lsp server setting [`completeUnimported`](https://github.com/LazyVim/LazyVim/blob/964dd6c4b2135399758c87afed7e1f3c8e30b487/lua/lazyvim/plugins/extras/lang/go.lua#L52) is not a valid setting given how it is not documented [here](https://github.com/golang/tools/blob/master/gopls/doc/settings.md).

folke:
I don't use go myself. @OneOfOne you said this was inlcuded in gopls?
@amaanq do you know?

amaanq:
I don't believe gopls provides this, but I'm not 100% sure, I can confirm later today in a few hrs

delatorrejuanchi:
Assuming this is confirmed later and it turns out it's not included in gopls, I agree with @fredrikaverpil in that the default should be to only include `goimports` and let the user opt into the stricter `goimports_reviser` if they so desire.

OneOfOne:
The goimports removal was in error

OneOfOne:
https://github.com/LazyVim/LazyVim/pull/1549/files fixes this, my bad.

cehoffman:
`gofumpt` appears to have made its way back into the current settings. I just disabled this locally myself and started searching for why it was enabled in the first place. https://github.com/LazyVim/LazyVim/blob/91126b9896bebcea9a21bce43be4e613e7607164/lua/lazyvim/plugins/extras/lang/go.lua#L115-L121

fredrikaverpil:
I stopped using this extras and I'm just [cooking my own](https://github.com/fredrikaverpil/dotfiles/blob/main/nvim-lazyvim/lua/plugins/lang_go.lua), since I don't agree with some of the opinions - and I feel I need a more specialized setup.

Same thing goes for the python extras. But for most other languages that I rarely touch, the provided extras is a great base.

PR #1546, opened 2023-10-01
feat(elixir): Add credo if none-ls is used

georgeguimaraes:
(sorry, sent the PR without the description by mistake)

Credo is the de-facto linter for Elixir. I was not aware that null-ls already had Credo as a builtin before, so that's an easy addition for folks already using null-ls.

btw, LazyVim was recently mentioned in ElixirForum:
https://elixirforum.com/t/neovim-elixir-setup-configuration-from-scratch-guide/46310/35

### DIFF
--- a/lua/lazyvim/plugins/extras/lang/elixir.lua
+++ b/lua/lazyvim/plugins/extras/lang/elixir.lua
@@ -29,4 +29,24 @@ return {
       },
     },
   },
+  {
+    "nvimtools/none-ls.nvim",
+    optional = true,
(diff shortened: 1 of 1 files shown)


folke:
Thanks! Can you add a spec for `nvim-lint` as well similar to how we do this for other languages? It has credo as well

georgeguimaraes:
Sure, let me take a look and will update

georgeguimaraes:
@folke nvim-lint added 👍

folke:
merged. Thanks!

PR #1549, opened 2023-10-01
add goimports

OneOfOne:


folke:
Thanks!

Jomik:
Why isn't this handled through `gopls` as well? 😅 
https://github.com/golang/tools/blob/master/gopls/doc/settings.md#local-string

OneOfOne:
@Jomik for whatever reason gopls doesn't handle automatic imports in nvim, where it does in vscode, I didn't look much into it but I use the exact same config for gopls in both vscode and nvim and it works in vscode.

PR #1552, opened 2023-10-02
fix(indent-blankline): set `config.indent.tab_char`

loqusion:
See lukas-reineke/indent-blankline.nvim#665.

Without patch:

image

With patch:

image


folke:
Thanks!

PR #1558, opened 2023-10-02
fix(which-key): remove duplicate which-key install

pmoieni:
I noticed `which-key` plugin has been installed two times in separate files ([here](https://github.com/LazyVim/LazyVim/blob/7c5a458761fe7002c6603d602e8d130b9a62dd68/lua/lazyvim/plugins/ui.lua#L229) and [here](https://github.com/LazyVim/LazyVim/blob/7c5a458761fe7002c6603d602e8d130b9a62dd68/lua/lazyvim/plugins/editor.lua#L278)) so I merged those two into one config.

### DIFF
--- a/lua/lazyvim/plugins/editor.lua
+++ b/lua/lazyvim/plugins/editor.lua
@@ -277,27 +277,33 @@ return {
   {
     "folke/which-key.nvim",
     event = "VeryLazy",
-    opts = {
-      plugins = { spelling = true },
-      defaults = {
-        mode = { "n", "v" },
-        ["g"] = { name = "+goto" },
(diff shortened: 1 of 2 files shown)


dpetka2001:
This is not `which-key` being installed 2 times. It's just 2 different sub-specs for the same plugin defined in 2 different files that will get merged in the end. It's just a personal preference of how you structure your configuration (I also prefer having the sub-specs for a plugin under different files i.e different LSP configs for each language instead of defining everything in the same file). You don't have to configure everything in one file under one combined spec.

folke:
That's on purpose indeed :)

PR #1598, opened 2023-10-06
fix(which-key): change surround group key

latipun7:
Fix missing changes from https://github.com/LazyVim/LazyVim/commit/d28c69e49eab62b45f7d1562b3bc00afee5fbb65


folke:
Thanks!

PR #1620, opened 2023-10-07
perf: enable lazy_update_context option

radvil:
to update context on "CursorHold" instead of of using default "CursorMove"


folke:
Thanks!

PR #1629, opened 2023-10-08
feat(dap): add clangd debuggee launch arguments

genmzy:
Clang debuggee launch should be able to pass arguments, now add that

### DIFF
--- a/lua/lazyvim/plugins/extras/lang/clangd.lua
+++ b/lua/lazyvim/plugins/extras/lang/clangd.lua
@@ -141,11 +141,7 @@ return {
             cwd = "${workspaceFolder}",
             args = function()
               local s = vim.fn.input("Args to pass: ")
-              local t = {}
-              _ = string.gsub(s, "[^ ]+", function(w)
-                table.insert(t, w)
-              end)
-              return t
+              return vim.split(s, " ")
             end,
           },
           {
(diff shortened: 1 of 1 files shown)


folke:
I added a new keymap `<leader>da` that you can use with any dap configuration to alter the args

fredrikaverpil:
@folke it also looks like that args are being asked about even if I don't pick the option that says `(Arguments)` in the debug window. Not sure if this is a go/delve specific issue?



https://github.com/LazyVim/LazyVim/assets/994357/4648dcfb-1beb-4413-93bb-2d1f799b1c9b

folke:
probably a go thing indeed. The new functionality is only for the `<leader>da` keymap

genmzy:
> I added a new keymap `<leader>da` that you can use with any dap configuration to alter the args

Good job😄

fredrikaverpil:
> probably a go thing indeed. The new functionality is only for the `<leader>da` keymap

Yeah, but [this](https://github.com/LazyVim/LazyVim/pull/1629#issuecomment-1757678475) is not related to go, right?

folke:
right, will move those old dap adapter mappings I made at some point. They're conflicting

folke:
it works though, just which-key that acts up :)

folke:
Also, I just checked and once you did the with args thing, the normal mapping also asks args. Will look into this.

PR #1639, opened 2023-10-08
fix(cmp): properly set cmp group_index

folke:
**nvim-cmp** has a `group_index` property that we used incorrectly in LazyVim, resulting in only showing entries from the first group that had entries.

This lead to several issues with different sources like copilot.

Probably fixes a number of issues at `copilot-cmp`:
- [ ] https://github.com/zbirenbaum/copilot-cmp/issues/93
- [ ] https://github.com/zbirenbaum/copilot-cmp/issues/75
- [ ] https://github.com/zbirenbaum/copilot-cmp/issues/86

# `group_index` docs

For instance, you can set the `buffer`'s source `group_index` to a larger number
  if you don't want to see `buffer` source items while `nvim-lsp` source is available:
```lua
    cmp.setup {
      sources = {
        { name = 'nvim_lsp', group_index = 1 },
        { name = 'buffer', group_index = 2 },
      }
    }
```

  You can also achieve this by using the built-in configuration helper like this:
```lua
    cmp.setup {
      sources = cmp.config.sources({
        { name = 'nvim_lsp' },
      }, {
        { name = 'buffer' },
      })
    }
```

### DIFF
--- a/lua/lazyvim/plugins/coding.lua
+++ b/lua/lazyvim/plugins/coding.lua
@@ -67,12 +67,12 @@ return {
             select = true,
           }), -- Accept currently selected item. Set `select` to `false` to only confirm explicitly selected items.
         }),
-        sources = cmp.config.sources({
-          { name = "nvim_lsp" },
-          { name = "luasnip" },
-          { name = "buffer" },
-          { name = "path" },
-        }),
+        sources = {
+          { name = "nvim_lsp", group_index = 1 },
+          { name = "luasnip", group_index = 1 },
+          { name = "buffer", group_index = 2 },
+          { name = "path", group_index = 2 },
+        },
         formatting = {
           format = function(_, item)
             local icons = require("lazyvim.config").icons.kinds
--- a/lua/lazyvim/plugins/extras/coding/codeium.lua
+++ b/lua/lazyvim/plugins/extras/coding/codeium.lua
@@ -17,9 +17,8 @@ return {
     },
     ---@param opts cmp.ConfigSchema
     opts = function(_, opts)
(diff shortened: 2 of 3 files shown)


dpetka2001:
I had questioned the choice of `group_index = 2` in a previous [PR](https://github.com/LazyVim/LazyVim/pull/1297#discussion_r1285265423) (that was closed when `codeium.nvim` was originally archived), but never got an answer about the choice made, so I brushed it off as personal preference. In my personal spec I had it in the same group as `nvim_lsp`. Not using it any more in favor of [codeium.vim](https://github.com/Exafunction/codeium.vim), since I found that `nvim-cmp` had become quite sluggish for my old pc with the addition of `codeium.nvim`.

folke:
My bad, I blindly copied the `group_index` from https://github.com/zbirenbaum/copilot-cmp#source-definition

And we used `cmp.sources()` that sets a different `group_index` for each source, which is obviously never something someone actually wants.

dpetka2001:
From my understanding `cmp.config.sources()` doesn't set a different `group_index` for each source. It sets a different `group_index` for each subtable it contains. For example in this snippet
```lua
 sources = cmp.config.sources({
                    { name = 'nvim_lsp' },
                    { name = 'luasnip' },
                }, {
                    { name = 'buffer' },
                }),
```
`nvim_lsp` and `luasnip` will have `group_index = 1` and `buffer` `group_index = 2`. But I also prefer using the first method with just `table_insert` and specifically declaring `group_index`.

folke:
oh, I see, that does make sense :)

PR #1640, opened 2023-10-09
fix(black): remove unknown blackd from conform

jyuan0:
conform.nvim doesn't actually have a config for `blackd`.

Selection_001_03:04:04_361x77_20231009


folke:
thanks!

jyuan0:
Wow, that was fast.

PR #1619, opened 2023-10-07
[feat] add codeium extension

appelgriebsch:
re-activate `codeium`extension (free alternative to GH Copilot), which is maintained by Exafunction now

PR #1651, opened 2023-10-10
feat: add Tabnine for LazyVim

jellydn:
# What
Add TabNine plugin for hrsh7th/nvim-cmp

# Why
LazyVim has Codeium by this PR https://github.com/LazyVim/LazyVim/pull/1619 so it would be good if we have Tabnine :)


dpetka2001:
I believe it would be good to leave the options in the `opts` instead of `config` to give the user the opportunity to change them, instead of changing the `config` function.

lua/lazyvim/plugins/extras/coding/tabnine.lua
```diff
@@ -8,14 +11,14 @@ return {
...
         build = "./install.sh",
         dependencies = "hrsh7th/nvim-cmp",
-        -- Only limit top 3 suggestions from Tabnine
-        config = function()
+        opts = {
+          max_lines = 1000,
+          max_num_results = 3,
+          sort = true,
+        },
+        config = function(_, opts)
```
dpetka2001, line 19: Is there a reason for the `config` function, since you don't seem to alter the default `lazy.nvim` config function behavior? I believe this is redundant? Pls correct me if I'm wrong.

folke:
Refactored a bit and merged. Thanks!

PR #1683, opened 2023-10-12
feat(go): add gofumpt formatter with conform/none-ls

delatorrejuanchi:
After https://github.com/LazyVim/LazyVim/pull/1535 and https://github.com/LazyVim/LazyVim/pull/1549, some formatters (i.e.: `gofumpt`) were removed from the `conform.nvim` and `none-ls.nvim` configurations in `extras.lang.go` because they were already already handled by the LSP `gopls`.

With the release of LazyVim 10.0.0, only one formatter can be active at a time so `gopls` is no longer taking care of running `gofumpt` when `conform` or `none-ls` are active.

This PR adds it back to the `conform.nvim` and `none-ls.nvim` configurations so people using any of the two can use `gofumpt`. This PR also makes sure packages are only installed when strictly necessary: for example, `delve` is only needed when `nvim-dap` is available and `gomodifytags` and `impl` are not supported by conform since they are code actions, so they are only installed when using `none-ls`.

### DIFF
--- a/lua/lazyvim/plugins/extras/lang/go.lua
+++ b/lua/lazyvim/plugins/extras/lang/go.lua
@@ -92,12 +92,14 @@ return {
   {
     "nvimtools/none-ls.nvim",
     optional = true,
-    {
-      "williamboman/mason.nvim",
-      opts = function(_, opts)
-        opts.ensure_installed = opts.ensure_installed or {}
-        vim.list_extend(opts.ensure_installed, { "gomodifytags", "impl" })
-      end,
+    dependencies = {
+      {
+        "williamboman/mason.nvim",
+        opts = function(_, opts)
+          opts.ensure_installed = opts.ensure_installed or {}
+          vim.list_extend(opts.ensure_installed, { "gomodifytags", "impl" })
+        end,
+      },
     },
     opts = function(_, opts)
       local nls = require("null-ls")
@@ -123,7 +125,7 @@ return {
     optional = true,
     dependencies = {
       {
-        "mason.nvim",
+        "williamboman/mason.nvim",
         opts = function(_, opts)
           opts.ensure_installed = opts.ensure_installed or {}
           vim.list_extend(opts.ensure_installed, { "delve" })
--- a/lua/lazyvim/plugins/extras/lang/go.lua
+++ b/lua/lazyvim/plugins/extras/lang/go.lua
@@ -105,6 +105,7 @@ return {
         nls.builtins.code_actions.gomodifytags,
         nls.builtins.code_actions.impl,
         nls.builtins.formatting.goimports,
+        nls.builtins.formatting.gofumpt,
       })
     end,
   },
@@ -113,7 +114,7 @@ return {
     optional = true,
     opts = {
       formatters_by_ft = {
-        go = { "goimports" },
+        go = { "goimports", "gofumpt" },
       },
     },
   },
--- a/lua/lazyvim/plugins/extras/lang/go.lua
+++ b/lua/lazyvim/plugins/extras/lang/go.lua
@@ -83,8 +83,22 @@ return {
   },
   -- Ensure Go tools are installed
   {
+    "williamboman/mason.nvim",
+    opts = function(_, opts)
+      opts.ensure_installed = opts.ensure_installed or {}
+      vim.list_extend(opts.ensure_installed, { "goimports", "gofumpt" })
+    end,
+  },
+  {
     "nvimtools/none-ls.nvim",
     optional = true,
+    {
+      "williamboman/mason.nvim",
+      opts = function(_, opts)
+        opts.ensure_installed = opts.ensure_installed or {}
+        vim.list_extend(opts.ensure_installed, { "gomodifytags", "impl" })
+      end,
+    },
     opts = function(_, opts)
       local nls = require("null-ls")
       opts.sources = vim.list_extend(opts.sources or {}, {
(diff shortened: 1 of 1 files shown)


delatorrejuanchi:
Also, this is my first time contributing to the project. Please do let me know if in the future you'd like me to first create an issue or similar.

Thanks :heart:

willnorris:
I'm still catching up myself on the new interaction between conform, none-ls, and the primary/secondary formatters.  But would this now *enforce* the use of gofumpt if none-ls is enabled?  That's certainly doesn't seem desirable, since gofumpt is far from standard in the Go community.

delatorrejuanchi:
> But would this now enforce the use of gofumpt if none-ls is enabled?
With this PR, `gofumpt` would be enabled if `none-ls` or `conform` are installed. Before `10.0.0` this was the default, since it was managed by `gopls`.

Of course, if someone does not wish to use `gofumpt` they could always disable it (possibly in favor of the less restrictive `gofmt`?)

willnorris:
I would argue that having gofumpt enabled by default pre-10.0.0 was a mistake.  If someone wants to use it, they could always add it, could they not?

delatorrejuanchi:
> I would argue that having gofumpt enabled by default pre-10.0.0 was a mistake. If someone wants to use it, they could always add it, could they not?

That's a good argument, I guess it ends up as a matter of preference. I'm happy with closing this if other people agree.

Out of curiosity, do you regularly use a different formatter yourself?

willnorris:
Yeah, it's always tricky to pick the "right" set of defaults for a project like this.  It definitely often comes down to individual preference.  The fact that `gopls` supports gofumpt but has it disabled by default, seems like reasonable behavior to follow.

I just the default `goimports` behavior.

folke:
If I understand correctly, the ideal case is golsp being the primary formatter and goimports being a secundary?

If so I can change it like that.

Jomik:
@willnorris 
> ...since gofumpt is far from standard in the Go community.

I sort of disagree with this. `gopls` ships `gofumpt`, it really does become a standard.
I am aware that it is opt-in, but still, I would argue that it is better to enable it here in LazyVim, as most people would be satisfied by it. It is compatible with `gofmt`.
Of course it may make sense to add docs on how to use `gofmt` instead of `gofumpt`.


@folke It does seem to work for me, but yes `gopls` should be primary, and `goimports` secondary.
I do have times where `goimports` times out, and is unable to actually fix my imports, but it is rarely. I also rarely have to manually modify my imports due to that.
My setup should be default: https://github.com/Jomik/dotfiles/blob/e71bdf109896ccb12a336c3212199c4326fe56b6/.config/nvim/lua/plugins/lang/go.lua

amaanq:
> @willnorris 
> > ...since gofumpt is far from standard in the Go community.
> 
> I sort of disagree with this. `gopls` ships `gofumpt`, it really does become a standard.
> I am aware that it is opt-in, but still, I would argue that it is better to enable it here in LazyVim, as most people would be satisfied by it. It is compatible with `gofmt`.
> Of course it may make sense to add docs on how to use `gofmt` instead of `gofumpt`.
> 
> 
> @folke It does seem to work for me, but yes `gopls` should be primary, and `goimports` secondary. I am wondering if I missed something though 😄 
> I do have times where `goimports` times out, and is unable to actually fix my imports, but it is rarely. I also rarely have to manually modify my imports due to that.
> My setup should be default: https://github.com/Jomik/dotfiles/blob/e71bdf109896ccb12a336c3212199c4326fe56b6/.config/nvim/lua/plugins/lang/go.lua

Yeah it would be nice to have gofumpt as the default, gofumpt got (unnecessarily imo) removed when someone updated the spec a while back

Jomik:
@amaanq I don't think `gofumpt` is removed, it is just handled by `gopls`, see https://github.com/LazyVim/LazyVim/blob/ea4174d46016dd191de48c03bdf23462483f9f6a/lua/lazyvim/plugins/extras/lang/go.lua#L24

amaanq:
> @amaanq I don't think `gofumpt` is removed, it is just handled by `gopls`, see
> 
> https://github.com/LazyVim/LazyVim/blob/ea4174d46016dd191de48c03bdf23462483f9f6a/lua/lazyvim/plugins/extras/lang/go.lua#L24

ooh yeah, forgot about that :sweat_smile:, I just noticed it being removed here but then saw why - https://github.com/LazyVim/LazyVim/commit/964dd6c4b2135399758c87afed7e1f3c8e30b487

folke:
* Initially we had `gofumpt` and `goimports`. That means that `gopls` was not used for formatting.
* Both got removed in favor of `gopls`
* apparently removing `goimports` was in error, so that was moved back
* because of that change, neither `gofumpt` nor `gopls` were being used for formatting. (not in 10.0 and also not before that)

So I agree that it's probably best to merge this PR and get back to how it was before.

AnhQuanTrl:
Just want to give my thought on this but imo the current way of handling formatters is less flexible than the old one. The old way allowed us to disable null-ls in favour of lsp in some languages. `Go` is a prime example of such languages. With the current way, we have no mechanism to do so since the `priority` field is hard coded in the LazyVim config.

folke:
That's not true. It's pretty much the  same as with null-ls. You can change the conform formatters_by_ft for go and set it to an empty list so that the lsp will be used instead. Thats' the same as it was before

fredrikaverpil:
In case it helps anyone, here's my conform setup for go where I can remove/extend/replace per file type: https://github.com/fredrikaverpil/dotfiles/blob/3d58a40bf2a9cc13546cf9e1aaaef93750c47c5e/nvim-lazyvim/lua/plugins/lsp.lua#L163-L189

AnhQuanTrl:
Thank you sir @folke. This is awesome.

PR #1698, opened 2023-10-13
fix(keymaps): no diagnostic keymaps w/o lsp attached

lkhphuc:
Sometime, a buffer has a linter but does not have lsp attached (e.g bash/zsh file with shellcheck only). Some diagnostic keymaps are not available in those case since they are registered with LSP only.

I propose to move diagnostic keymaps to the more general keymaps file similar to formatting and inlay_hint functionality.


folke:
Makes sense. Thanks!

PR #1703, opened 2023-10-13
fix: add title to Spectre panel in edgy

kevintraver:
Title could also be `Search/Replace`, because the Spectre panel shows *Nvim Spectre* in the panel as well.


folke:
Thanks!

PR #1718, opened 2023-10-14
feat(lang): add markdown support

jakub-kozlowicz:


lua/lazyvim/plugins/extras/lang/markdown.lua
```diff
@@ -0,0 +1,52 @@
...
+    },
+  },
+  {
+    "iamcco/markdown-preview.nvim",
+    cmd = { "MarkdownPreviewToggle", "MarkdownPreview", "MarkdownPreviewStop" },
+    ft = { "markdown" },
+    build = function()
+      vim.fn["mkdp#util#install"]()
+    end,
+  },
```
folke, line 51: I would suggest `peek.nvim` instead, which is a lua plugin that provides similar functionality, but better in my opinion.

```lua
  -- markdown preview
  {
    "toppair/peek.nvim",
    build = "deno task --quiet build:fast",
    keys = {
      {
        "<leader>cp",
        ft = "markdown",
        function()
          local peek = require("peek")
          if peek.is_open() then
            peek.close()
          else
            peek.open()
          end
        end,
        desc = "Peek (Markdown Preview)",
      },
    },
    opts = { theme = "light" },
  },
```

jakub-kozlowicz: I've been looking into this plugin but still decided to choose the markdown preview. It is because you do not have to install any additional toolchains to build this plugin. Markdown preview can be installed from pre-build binary.

folke: I know but it's a lua plugin and actually renders the markdown exactly as github.

I just added a new commit that adds `peek.nvim`.
When `deno` is not on the user's system, the plugin will not be enabled and a warning will be shown.

Should be good enough

dpetka2001: @folke You should be aware of [issue 47](https://github.com/toppair/peek.nvim/issues/47), which might give new users difficulties. The version of deno I used to build the plugin was 1.25.x . If `peek.nvim` does make it into this PR, appropriate instructions should be given in the LazyVim docs. There also exists [PR 50](https://github.com/toppair/peek.nvim/pull/50), which fixes the issue, but there hasn't been any movement from the maintainer since April 1 in the repo, so I don't know how one can consider this plugin still maintained.

folke: Yeah, that's not good indeed.
Ok, I'll look into markdown-preview.nvim again to see how it has improved since I last checked.

dpetka2001: You could still use it, just be sure to point it out to the user in the Lazyvim docs. I'm just mentioning this because I've seen quite a few topics lately about stuff that are not Lazyvim related and this one should be definitely mentioned in the docs as a workaround, should you decide to go this way. However, in my personal opinion, such a plugin should not be included in an extra if it needs a workaround to be build, except if Lazyvim itself provides the workaround. You could always just provide support for anything else Markdown related, except from the previewer, which you would leave it up to the user to decide.

folke: Just changed it back to markdown-preview.nvim and added a keymap.
The plugin also no longer lazy-loads on `ft=markdown`.
It will only load when a `cmd` is used or on the keymap.

It does require this fix in `lazy.nvim` https://github.com/folke/lazy.nvim/pull/1117
Making a new release now.

dpetka2001: @folke sorry for pinging you for no good reason. I just wanted to express my gratitude towards the work you put in for all these amazing Lazyvim features (and of course other plugins that I'm less acquainted with). I wasn't around when you requested for beta testers, so I just saw the changes yesterday and upgraded. Now it's even way snappier than before. Thank you!!

folke: Thanks! Hopefully the upgrade didn't break too much :)

folke:
Merged. Thanks!

lua/lazyvim/plugins/extras/lang/markdown.lua
abeldekat: Issue [47](https://github.com/toppair/peek.nvim/issues/47) in `peak.nvim` has been fixed.

xclidongbo: > I would suggest `peek.nvim` instead, which is a lua plugin that provides similar functionality, but better in my opinion.
> 
> ```lua
>   -- markdown preview
>   {
>     "toppair/peek.nvim",
>     build = "deno task --quiet build:fast",
>     keys = {
>       {
>         "<leader>cp",
>         ft = "markdown",
>         function()
>           local peek = require("peek")
>           if peek.is_open() then
>             peek.close()
>           else
>             peek.open()
>           end
>         end,
>         desc = "Peek (Markdown Preview)",
>       },
>     },
>     opts = { theme = "light" },
>   },
> ```

not work. always loading .

PR #1724, opened 2023-10-14
Add toggle for welcome message

curtbushko:
Adds a toggle for the welcome message when no news is available.


folke:
That's not how it works. You're only supposed to see that message on a brand new LazyVim install.
After the first restart it would be gone.

Issue #1726, opened 2023-10-14
feature: ability to configure linter(s) for all filetypes.


m0lson84:
### Did you check the docs?

- [X] I have read all the LazyVim docs

### Is your feature request related to a problem? Please describe.

I've run into a case when configuring `nvim-lint` where I would like to have the particular linter enabled (`cspell`) for all filetypes. In `conform.nvim` this is supported by using the `*` filetype, but the same functionality is not available in `nvim-lint`.

### Describe the solution you'd like

Taking a look at [plugins/linting.lua](https://github.com/LazyVim/LazyVim/blob/f29543f2c9323ae84d0979f36598c0fb78ed767a/lua/lazyvim/plugins/linting.lua#L50) I think that this could be enabled by adjusting a single line. My proposed solution would be to update

```lua
local names = lint.linters_by_ft[vim.bo.filetype] or {}
```

to 

```lua
local names = vim.tbl_flatten({ lint.linters_by_ft[vim.bo.filetype], lint.linters_by_ft["*"] }) or {}
```

### Describe alternatives you've considered

I looked at creating an issue in `nvim-lint` but based on this [close issue](https://github.com/mfussenegger/nvim-lint/issues/355) it doesn't look like this functionality is a priority there.

### Additional context

Happy to submit a pull request for this change.

folke:
Feel free to submit a PR!

And would be great to also add the same conform functionality for `_`

PR #1727, opened 2023-10-14
feat(linting): ability to configure global and fallback linters

m0lson84:
# Description

Add ability to configure a linter to run either for all filetypes or as a fallback if no others are configured.

## Related issues

- https://github.com/LazyVim/LazyVim/issues/1726


folke:
I've changed it a bit to use `nvim-lint`s logic related to multi-filetype buffers.

folke:
Thanks!

PR #1740, opened 2023-10-15
feat(sql): add SQL extra

igorgue:


nomnivore:
I had something pretty similar in my config, but with lazy loading per their documentation:

```lua
"kristijanhusak/vim-dadbod-ui",
lazy = true,
dependencies = {
  { "tpope/vim-dadbod", lazy = true },
  { "kristijanhusak/vim-dadbod-completion", ft = { "sql", "mysql", "plsql" }, lazy = true },
  { "jsborjesson/vim-uppercase-sql", lazy = true },
},

cmd = {
  "DBUI",
  "DBUIToggle",
  "DBUIAddConnection",
  "DBUIFindBuffer",
},

keys = {
  {
    "<leader>cD",
    "<cmd>DBUI<CR>",
    desc = "Dadbod UI",
  },
},

```

And with a recipe for integration with the `edgy` extra:

```lua
{
  "folke/edgy.nvim",
  optional = true,

  opts = function(_, opts)
    table.insert(opts.right, {
      title = "Database",
      ft = "dbui",
      pinned = true,
      open = function()
        vim.cmd("DBUI")
      end,
    })

    table.insert(opts.bottom, {
      title = "DB Query Result",
      ft = "dbout",
    })
  end,
},

```

I haven't looked into custom `neo-tree` mappings yet, but it might also be nice to have a mapping that opens a selected sqlite db (`.sqlite, .sqlite3, .db, .db3, .s3db, .sl3`) in DBUI, streamlining the process of adding fully-qualified filepaths to the AddConnection prompt by hand.

Anyway, just wanted to share my own ideas in regards to a sql extra using `vim-dadbod`!

igorgue:
@nomnivore you don't need the `lazy = true,` on the dependencies nor the main spec, since you do:

```lua
cmd = {
  "DBUI",
  "DBUIToggle",
  "DBUIAddConnection",
  "DBUIFindBuffer",
},

keys = {
  {
    "<leader>cD",
    "<cmd>DBUI<CR>",
    desc = "Dadbod UI",
  },
},```

That would mean the plugin lazyloads on `cmd` and `keys`.

I've included all your suggestions, lets see if it gets approved, it's pretty solid in my setup.

**Comment by nomnivore** [comment]
Oh, good to know. I must've missed that in the LazyNvim documentation. Thanks for the tip!

**Comment by igorgue** [comment]
I just updated it to use nvim notify instead of their own notifications, I wonder if I missed some step to get this approved...

File: lua/lazyvim/plugins/extras/lang/sql.lua [hunk 1]
```diff
@@ -0,0 +1,55 @@
...
+      vim.g.db_ui_save_location = vim.fn.stdpath("data") .. "/db_ui"
+      vim.g.db_ui_use_nerd_fonts = true
+      vim.g.db_ui_execute_on_save = false
+      vim.g.db_ui_use_nvim_notify = true
+
+      vim.api.nvim_create_autocmd("FileType", {
+        pattern = sql_ft,
+        callback = function()
+          ---@diagnostic disable-next-line: missing-fields
+          require("cmp").setup.buffer({ sources = { { name = "vim-dadbod-completion" } } })
```
fredrikaverpil, line 50: This will make other completion engines unavailable which you might want to keep, effectively breaking completion for e.g. snippets, buffer text or keywords.

See this discussion: https://github.com/kristijanhusak/vim-dadbod-completion/issues/53

It would be desirable to _extend_ the list of completion engines with `"vim-dadbod-completion"` rather than hardcode it into becoming the only one. I'm unfortunately not sure how/if it's possible... and I asked about it in the above linked discussion.

This is what my setup currently looks like: https://github.com/fredrikaverpil/dotfiles/blob/5b2a1bcb793e60a60f6aeb3f94bb3c758c9d4b02/nvim-lazyvim/lua/plugins/sql.lua

igorgue:
@fredrikaverpil thanks just tried it, and I saw now I have snippets and others :+1:

fredrikaverpil (approved):
LGTM! 😊

rubiin (approved):

rubiin (approved):

rubiin:
I think this can be merged as well :)

rubiin (approved):

rubiin (approved):

lua/lazyvim/plugins/extras/lang/sql.lua
```diff
@@ -0,0 +1,71 @@
...
+    cmd = { "DBUI", "DBUIToggle", "DBUIAddConnection", "DBUIFindBuffer" },
+    dependencies = {
+      "kristijanhusak/vim-dadbod-ui",
+      { "kristijanhusak/vim-dadbod-completion", ft = sql_ft },
+      { "jsborjesson/vim-uppercase-sql", ft = sql_ft },
+    },
+    init = function()
+      vim.g.db_ui_save_location = vim.fn.stdpath("data") .. "/db_ui"
+      vim.g.db_ui_use_nerd_fonts = true
+      vim.g.db_ui_execute_on_save = false
```
kevinrobayna, line 43: why `db_ui_execute_on_save` = false by default? 

pardon my lack of knowledge on how dadbod works but how would we run the query otherwise? 😅

igorgue: I found that running on save can lead to Neovim crashing mistakes, for example, if you forget to put a `LIMIT 100` at the end of the query and your table is massive.

So I use the key bind, which are only applied to when this plugin is active.

Leader Shift+s to run the current buffer's query.

Is worth learning also Leader Shift+w as well to save the queries to disk.

rubiin: Yeah manually triggering query is better . This will make sure we are sure about the query . This will shine when you are writing mutations

kevinrobayna: should we add a comment here on the keymap/command to execute the buffer's query?

igorgue: @kevinrobayna hmm yeah, let me see, not only the comment but which key shows something like this:

`S ➜ <Plug>(DBUI_ExecuteQuery)`

I can try to add those descriptions without breaking anything since this only applies to the plugin, probably by using keys, let me give it shot.

kevinrobayna: ah! i see just saw it here! 
image

kevinrobayna:
Tried this locally works as expected!

@folke it would be awesome to get this merged I think. Currently have it locally by copying what @igorgue has on the pr 😅

dpetka2001:
Why are you using `vim-dadbod-ui` as dependency to `vim-dadbod`? The docs of `vim-dadbod-ui` suggest the other way around, so that `vim-dadbod` will be loaded first. 

Even if this is something minor, you define the `vim.g.db_ui_*` options in the `init` function of `vim-dadbod`. Even if this works, it isn't really good plugin structure in my opinion. These options should be defined in the `init` function of `vim-dadbod-ui` as they are pertained to the `vim-dadbod-ui` configuration. Consider someone will disable `vim-dadbod-ui` in his personal configuration and instead just run the `:DB` commands that `vim-dadbod` provides to run queries. As it is currently, I believe there will be issues as the `vim.g.db_ui_*` options are defined in the `init` function of `vim-dadbod`.

igorgue:
@dpetka2001 this is a great point, thanks! I'll fix it.

igorgue:
@dpetka2001 done, if you want to review it again.

@kevinrobayna thanks! Check the PR again if you want to try it, I changed the data directory though, it's now `dadbod_ui` instead of `db_ui` if you had setup anything at all.

I wonder if all this attention is because of @tjdevries' video.

dpetka2001:
@igorgue I was wrong about the dependency. Keep it as it is. If you make `vim-dadbod` a dependency of `vim-dadbod-ui`, if you try to disable `vim-dadbod-ui` in your personal config, it will also disable `vim-dadbod`. So, just keep the dependencies as you had them. Can you try the following and see if everything still works on your side
```lua
  {
    "tpope/vim-dadbod",
    dependencies = {
      {
        "kristijanhusak/vim-dadbod-completion",
        ft = sql_ft,
        init = function()
            pattern = sql_ft,
            callback = function()
              ---@diagnostic disable-next-line: missing-fields
              local cmp = require("cmp")
              local global_sources = cmp.get_config().sources
              local buffer_sources = {}

              -- add globally defined sources (see separate nvim-cmp config)
              -- this makes e.g. luasnip snippets available since luasnip is configured globally
              if global_sources then
                for _, source in ipairs(global_sources) do
                  table.insert(buffer_sources, { name = source.name })
                end
              end

              -- add vim-dadbod-completion source
              table.insert(buffer_sources, { name = "vim-dadbod-completion" })

              -- update sources for the current buffer
              cmp.setup.buffer({ sources = buffer_sources })
            end,
          })
        end,
      },
      { "jsborjesson/vim-uppercase-sql", ft = sql_ft },
    },
  },
  {
    "kristijanhusak/vim-dadbod-ui",
    cmd = { "DBUI", "DBUIToggle", "DBUIAddConnection", "DBUIFindBuffer" },
    dependencies = {
      "tpope/vim-dadbod",
    },
    init = function()
      local data_path = vim.fn.stdpath("data")

      vim.g.db_ui_save_location = data_path .. "/dadbod_ui"
      vim.g.db_ui_tmp_query_location = data_path .. "/dadbod_ui/tmp"

      vim.g.db_ui_use_nerd_fonts = true
      vim.g.db_ui_show_database_icon = true
      vim.g.db_ui_use_nvim_notify = true

      -- NOTE: The default behavior of auto-execution of queries on save is disabled
      -- this is useful when you have a big query that you don't want to run every time
      -- you save the file running those queries can crash neovim to run use the
      -- default keymap: <leader>S
      vim.g.db_ui_execute_on_save = false
    end,
  },
```
Sorry about that. I hope I didn't cause you too much trouble.

igorgue:
@dpetka2001 no worries, I was also wondering why I had it like I did because I've been using this for over a year there should've been a reason why...

igorgue:
@dpetka2001 okay check again.

dpetka2001:
@igorgue I'm just a simple user who started learning programming about 4-5 months ago, I just have a bit more experience with configuring LazyVim (since been using it for a bit over a year now). So, I really don't think you should ask me to review this. People who use it and do SQL should test it and share their opinions instead. I believe that will be a better metric. I can only offer some insight from a configuration perspective, that's all. I don't do SQL myself yet. So, if it works on your side I will trust your word :smile:

igorgue:
@dpetka2001 lmao, good job though, you raised great points, as somebody with over 20 years of experience (using Vim too), you have good instincts.

And yes it works, thank you!

rubiin:
@dpetka2001 is an op user for sure . He is being humble :D

igorgue:
@folke Fixed the CI by adding the `recommended` key on `sql` fts like I saw on other langs.

igorgue (commented):
@folke copied the config locally, configured `sqlfluff` and ran some queries. It's all good to go!

Thank you! I also have other extras in my config I gonna push later on after some cleanups, I like the recommended key!

igorgue:
For the subscribed to this issue, this `~/.sqlfluff` works for me, upper on the right things :)

```ini
[sqlfluff]
templater = jinja
sql_file_exts = .sql,.sql.j2,.dml,.ddl

[sqlfluff:indentation]
indented_joins = False
indented_using_on = True
template_blocks_indent = False

[sqlfluff:templater]
unwrap_wrapped_queries = True

[sqlfluff:templater:jinja]
apply_dbt_builtins = True

[sqlfluff:rules:capitalisation.keywords]
capitalisation_policy = upper

[sqlfluff:rules:capitalisation.identifiers]
extended_capitalisation_policy = lower

[sqlfluff:rules:capitalisation.functions]
extended_capitalisation_policy = upper

[sqlfluff:rules:capitalisation.literals]
capitalisation_policy = lower

[sqlfluff:rules:capitalisation.types]
extended_capitalisation_policy = upper

# vim:ft=config:
```

folke:
Merged. Thanks!

muneebusmani:
@igorgue are you getting completions with sqlfluff?

igorgue:
@muneebusmani no, only the cmp plugin, isn't that just a formatted though?

muneebusmani:
Oh my bad, I thought it was a lsp , anyways I tried setting up lsp but it doesn't provide completions in the query window, I used sqlls

igorgue:
@muneebusmani maybe one day I can make that happen but only the diagnostics are worth it, the completion works with the current extra, the problem is that you must configure the LSP, it just doesn't work with the current Dadbod connection, to do that you have to parse the Json file that stores the dbs on Dadbod to configure the LSP.

muneebusmani:
Could you elaborate with a snippet
@igorgue

igorgue:
@muneebusmani not really, if I had a snippet it'll be in the PR, I don't understand your problem, completions work with this extra.

PR #1743, opened 2023-10-15
feat(ui): add keymap to close other buffers

kevintraver:


kevintraver:
Open to other keymaps besides `<leader>bx` if something else makes more sense.

folke:
Changed it to `bo` and also added keymaps for br (right) and bl (left. Thanks!

PR #1744, opened 2023-10-15
feat(lualine): use gitsigns for diff source

edshamis:
We can use the git diff info from gitsigns in lualine. Anyway, it exists and can be configured to handle worktrees ($GIT_DIR other than .git)


folke:
Thanks!

PR #1772, opened 2023-10-17
feat(fold): expand fold to previous line when reopen buffer

lkhphuc:
Modified the last_loc autocommand to always reveal the current line if it's in a fold.
Not a problem with current setting of foldlevel=99 but it's a bit more convenient when use with default foldlevel to a low value.


lkhphuc:
Wait this actually doesn't work yet.

PR #1788, opened 2023-10-18
fix(tabnine): run `:CmpTabnineHub` automatically on build

andresatierf:
add the `:CmpTabnineHub` command to the build process, similar to copilot and codeium


folke:
thanks!

PR #1789, opened 2023-10-18
feat: added better fold ui extra

leisurelicht:
better fold ui

image

better preview the folding

image

### DIFF
--- a/lua/lazyvim/plugins/extras/ui/fold.lua
+++ b/lua/lazyvim/plugins/extras/ui/fold.lua
@@ -1,0 +1,116 @@
+return {
+  {
+    "lewis6991/gitsigns.nvim",
+    optional = true,
+    opts = {
+      signs = {
+        add = { text = "┃" },
+        change = { text = "┃" },
+        delete = { text = "󱈸" },
+        topdelete = { text = "󱈸" },
+        changedelete = { text = "┃" },
+        untracked = { text = "┇" },
+      },
+      signcolumn = true,
+      numhl = true,
+    },
+  },
+  {
+    "neovim/nvim-lspconfig",
+    optional = true,
+    init = function()
+      local keys = require("lazyvim.plugins.lsp.keymaps").get()
+      keys[#keys + 1] = { "K", false }
+    end,
+  },
+  { -- better statuscolumn
+    "luukvbaal/statuscol.nvim",
+    event = { "BufRead", "BufNewFile" },
+    opts = function()
+      local builtin = require("statuscol.builtin")
+      return {
+        ft_ignore = { "neo-tree", "neo-tree-popup", "alpha", "lazy", "mason" },
+        segments = {
+          { text = { builtin.lnumfunc }, click = "v:lua.ScLa" },
+          {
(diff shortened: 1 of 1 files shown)


lkhphuc:
This is quite intrusive and major change to be added to base LazyVim. You might consider contributing this as an extra since it's added quite a few plugin.
Since folke just added treesitter fold and a custom statuscolumn, I don't think this will be merged. In fact, I myself remove ufo and statuscol.nvim plugins to use the lightweight's native version.

leisurelicht:
> This is quite intrusive and major change to be added to base LazyVim. You might consider contributing this as an extra since it's added quite a few plugin. Since folke just added treesitter fold and a custom statuscolumn, I don't think this will be merged. In fact, I myself remove ufo and statuscol.nvim plugins to use the lightweight's native version. P/s: About preview fold, I previously had it configured but rarely use it. I found that a good mapping of toggling fold (`za`, `zv`) is a lot better than some popup window.

I did contribute it as an extra. I write all config in one file and put it in extra folder. Is that wrong?

not only fold ui, I also custom statuscolumn to make it appear narrower.

lkhphuc:
Ah my bad. I mistaken it as modifying the built-in. Cheers.

MystikReasons:
Love that idea. 
What do you think on adding a number suffix of folded lines?

grafik

dpetka2001:
In my personal opinion, this is way too intrusive for an Extra. 

`gitsigns` changes shouldn't be made here, instead it should be left up to the user to change the defaults in his personal configuration if he likes. 

`statuscol.nvim` shouldn't be made a hard requirement of this Extra. Users might prefer the default `statuscolumn` LazyVim provides. Maybe a different extra on its own would be better or even left to the user to install it in his personal configuration. 

The keymap should use `vim.keymap.set` instead of `LazyVim.safe_keymap_set` and could also be moved just before the `return` in the `opts` table, thus making the `config` function redundant, since it will be called either way.

leisurelicht:
> `gitsigns` changes shouldn't be made here, instead it should be left up to the user to change the defaults in his personal configuration if he likes.
> 
> `statuscol.nvim` shouldn't be made a hard requirement of this Extra. Users might prefer the default `statuscolumn` LazyVim provides. Maybe a different extra on its own would be better or even left to the user to install it in his personal configuration.

But the configuration of `statuscol` and `gitsigns` here is to better display the folds.
I use `statuscol` to display the fold symbols. If I remove statucol, the fold symbols will be gone.  
And I change `gitsigns` because I try to make the git symbols, line numbers, and fold symbols more harmonious and not overlap each other.  

> The keymap should use `vim.keymap.set` instead of `LazyVim.safe_keymap_set` and could also be moved just before the `return` in the `opts` table, thus making the `config` function redundant, since it will be called either way.

thanks for your reminder. I will change it.

folke:
Closing this in favor of #2923 that I just merged.

You can configure it in your `config/options.lua` with:

```lua
-- Options for the LazyVim statuscolumn
vim.g.lazyvim_statuscolumn = {
  folds_open = true, -- show fold sign when fold is open
  folds_githl = true, -- highlight fold sign with git sign color
}
```

PR #1792, opened 2023-10-18
refactor(typescript): split typescript dap from lang

andresatierf:


andresatierf:
You're right, I was only looking at the typescript extra. My reason was due to the lua `dap` being inside the dap folder, it made sense to me to do the same for the other languages.

dpetka2001:
I believe the lua `dap` was put in there, because there is not a `lua` extra, since it's inherent.

andresatierf:
I guess so... I still believe the dap configurations could be in its directory but I'll close this for now

dpetka2001:
I'm not the one who'll make the decision on this. The maintainer will be. Just leave it open until he decides. I just asked for the logic behind this, maybe it was something I didn't understand.

folke:
Makes more sense to stay in typescript lang.
And the lua thing is indeed confusing. Maybe I should move the nlua to a lua extra instead

PR #1794, opened 2023-10-18
feat(typescript): remove unused imports

agriffis:
Previously used typescript.nvim for this (`:TypescriptRemoveUnused` or somesuch)

Not sure if `<leader>cR` is the best binding but it seems reasonable. I like caps for stuff that's destructive.

### DIFF
--- a/lua/lazyvim/plugins/extras/lang/typescript.lua
+++ b/lua/lazyvim/plugins/extras/lang/typescript.lua
@@ -32,6 +32,19 @@ return {
               end,
               desc = "Organize Imports",
             },
+            {
+              "<leader>cR",
+              function()
+                vim.lsp.buf.code_action({
+                  apply = true,
+                  context = {
+                    only = { "source.removeUnused.ts" },
+                    diagnostics = {},
+                  },
+                })
+              end,
+              desc = "Remove Unused Imports",
+            },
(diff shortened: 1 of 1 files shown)


folke:
I'm not sure we should add this one as well. Did you know about `<leader>cA`?
It shows source code-actions instead of regular ones.

I'm actually thinking of making the default `<leader>ca` show both the regular ones and the source actions.

agriffis:
That sounds reasonable to me.

I can always add a local plugin for the additional keymap, although if I add more than one of these, I should make a util helper for it:

```lua
return {
  {
    'neovim/nvim-lspconfig',
    opts = function(_, opts)
      vim.list_extend(opts.servers.tsserver.keys, {
        {
          '<leader>cR',
          function()
            vim.lsp.buf.code_action {
              apply = true,
              context = {
                only = { 'source.removeUnused.ts' },
                diagnostics = {},
              },
            }
          end,
          desc = 'Remove unused imports',
        },
      })
    end,
  },
}
```

agriffis:
Feel free to close this in deference to your preferred approach

folke:
added for now :) Thanks!

agriffis:
Oh! What made you change your mind?

PR #1797, opened 2023-10-18
feat(eslint): make eslint primary formatter

mikeduminy:
## Summary
This change makes the eslint formatter the only formatter when using the eslint extra, however this code should probably be more configurable than this but I'm raising it here anyway to highlight the problem.

## Problem
Using the eslint extra and saving a typescript file in a project that uses eslint to format (via `prettier/prettier` rules) the formatting of `EslintFixAll` and `tsserver` (LSP) overlap and `tsserver` often wins, which causes eslint failures.
CleanShot 2023-10-18 at 23 03 45

I previously disabled this by manually disabling the formatting provider of tsserver but this no longer works.
```lua
{
  "neovim/nvim-lspconfig",
  opts = {
    servers = {
      tsserver = {
        -- redacted for brevity
      }
    },
    setup = {
      tsserver = function(_, opts)
        opts.capabilities.documentFormattingProvider = false
      end,
    },
  },
}
```

## Other ideas
- Enable configuration of the `primary` and `priority` fields in the eslint formatter.
- When multiple formatters exist, call each synchronously and ensure the last one is the main one.
- Support for disabling formatting of specific LSP servers.

### DIFF
--- a/lua/lazyvim/plugins/extras/linting/eslint.lua
+++ b/lua/lazyvim/plugins/extras/linting/eslint.lua
@@ -20,7 +20,7 @@ return {
 
           local formatter = require("lazyvim.util").lsp.formatter({
             name = "eslint: lsp",
-            primary = false,
+            primary = true,
             priority = 200,
             filter = "eslint",
           })


mikeduminy:
Ignore this, I just found https://www.lazyvim.org/configuration/recipes#add-eslint-and-use-it-for-formatting which seems to do the trick for me.

PR #1815, opened 2023-10-20
feat(neo-tree): show clickable selector on winbar

leisurelicht:
show clickable tabs to switch between sources
image


folke:
Not a fan of this. Thank you anyway. Closing.

PR #1826, opened 2023-10-21
fix(lang): add cmakelint to ensure installed list

konosubakonoakua:
1. mason/cmakelang install bin/cmake-lint
2. nvim-lint only recognize bin/cmakelint
3. mason/cmakelint install bin/cmakelint

so we need mason/cmakelint installed to using nvim-lint to perform formatting


folke:
thanks!

PR #1833, opened 2023-10-22
feat(nvim): extend j/k enhancements to up/down arrows

minusfive:
Some of us using alternate keyboard layouts (e.g. Colemak-DH) may rely on arrows instead of `j/k`. This extends the `j/k` enhancements for soft-wrapped lines navigation to the `<Up>/<Down>` arrows as well.

I'm a total vim noob, any reason I shouldn't do this?


folke:
Thanks!

Issue #1503, opened 2023-09-27
bug: LazyVim throws an error related to DAP on launch


AndersNS:
### Did you check docs and existing issues?

- [X] I have read all the LazyVim docs
- [X] I have searched the existing issues of LazyVim
- [X] I have searched the existing issues of plugins related to this issue

### Neovim version (nvim -v)

0.9.2

### Operating system/version

MacOS 13.6

### Describe the bug

<img width="1182" alt="image" src="https://github.com/LazyVim/LazyVim/assets/1052521/08b1224d-2a76-4d64-862b-babed3b94ad9">

Bug when DAP is enabled. 

Seems related to this change: https://github.com/LazyVim/LazyVim/commit/f1af34e88525cc291b7bacc76d915b642aef192b

I downgraded to t 6.3.0 and it works again. 

### Steps To Reproduce

1. Add `{ import = "lazyvim.plugins.extras.dap.core" }` to lazy.lua
2. Launch nvim
3. Observe error

### Expected Behavior

That I dont get an error

### Repro

```Lua
-- DO NOT change the paths and don't remove the colorscheme
local root = vim.fn.fnamemodify("./.repro", ":p")

-- set stdpaths to use .repro
for _, name in ipairs({ "config", "data", "state", "cache" }) do
  vim.env[("XDG_%s_HOME"):format(name:upper())] = root .. "/" .. name
end

-- bootstrap lazy
local lazypath = root .. "/plugins/lazy.nvim"
if not vim.loop.fs_stat(lazypath) then
  vim.fn.system({ "git", "clone", "--filter=blob:none", "https://github.com/folke/lazy.nvim.git", lazypath, })
end
vim.opt.runtimepath:prepend(lazypath)

-- install plugins
local plugins = {
  "folke/tokyonight.nvim",
  "folke/LazyVim",
  -- add any other plugins here
}
require("lazy").setup(plugins, {
  root = root .. "/plugins",
 spec = { -- not sure if this is correct
    -- add LazyVim and import its plugins
    { "LazyVim/LazyVim", import = "lazyvim.plugins" },
    -- import any extras modules here
    -- { import = "lazyvim.plugins.extras.lang.typescript" },
    { import = "lazyvim.plugins.extras.dap.core" },
    { import = "lazyvim.plugins.extras.lang.rust" },
    -- { import = "lazyvim.plugins.extras.lang.omnisharp" },
    { import = "lazyvim.plugins.extras.lang.tailwind" },
[...]

PR #1839, opened 2023-10-22
fix(dap): load vscode launch files with jsonc parser

MoetaYuko:
This seems to be the proper fix for #1503. jsonc ensures compatibility with native vscode.

Ref: https://github.com/mfussenegger/nvim-dap/issues/964


sid-6581:
An even simpler way is using neoconf's bundled jsonc parser, which saves an extra dependency:

```lua
require("dap.ext.vscode").json_decode = require("neoconf.json.jsonc").decode_jsonc
```

folke:
Cool, that's indeed a nice way to fix this. Thanks!

PR #1845, opened 2023-10-22
fix(extras.python-semshi): improve highlights

amaanq:
I don't use python too heavily (definitely not daily) but these are improvements I use locally because I find them to be worse than tree-sitter's semantically

Ping @lkhphuc since you added the extra

### DIFF
--- a/lua/lazyvim/plugins/extras/lang/python-semshi.lua
+++ b/lua/lazyvim/plugins/extras/lang/python-semshi.lua
@@ -17,13 +17,14 @@ return {
         -- Only add style, inherit or link to the LSP's colors
         vim.cmd([[
             highlight! semshiGlobal gui=italic
-            highlight! semshiImported gui=bold
+            highlight! link semshiImported @none
             highlight! link semshiParameter @lsp.type.parameter
             highlight! link semshiParameterUnused DiagnosticUnnecessary
             highlight! link semshiBuiltin @function.builtin
-            highlight! link semshiAttribute @attribute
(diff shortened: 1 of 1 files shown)


folke:
Thanks!

lkhphuc:
Hi @amaanq,
about `semshiFree` I'm not even sure what it is referring to so I don't have a comment.
But about `semshiImported`, the way you use link it to @none, it will clear all highlight of the imported symbols, i.e you would not be able to tell if the imported symbol is a function, module, constructor etc.

 Instead of this from Treesitter
Screenshot 2023-11-07 at 17 36 46

now you only have this
Screenshot 2023-11-07 at 17 34 40

I'm not sure if this is your intended purpose or not, but if you only want disregard effect of semshi on imported and free nodes, then the best way to that would be `hi! semshiImported gui=NONE`.

amaanq:
That's a bizarre one - no other LSP/semantic highlighter highlights imported identifiers differently from other identifiers, so I found that to be better disable

lkhphuc:
> but did it differentiate all 3 from each other

TS differentiates all kinds these types (functions, modules, ...), including in the import statement.
So as I said, it seems you just want to disable semshi (style like bold, italic), but not treesitter (the color), you should not link it to @none highlight group, but override the `gui` attribute of semshi like: `highlight! semshiImported gui=NONE`.

Anyway, this extra hopefully will be redundant when newer version of semshi is released https://github.com/wookayin/semshi/issues/16

PR #1858, opened 2023-10-23
fix(extras/lang/python): Add nvim-treesitter as dependency of neotest in extras/lang/python.lua

ohseunghyeon:
neotest couldn't find any python tests
so added nvim-treesitter as a dependency and it worked well

### DIFF
--- a/lua/lazyvim/plugins/extras/lang/python.lua
+++ b/lua/lazyvim/plugins/extras/lang/python.lua
@@ -47,6 +47,7 @@ return {
     optional = true,
     dependencies = {
       "nvim-neotest/neotest-python",
+      "nvim-treesitter/nvim-treesitter",
     },
     opts = {
       adapters = {


dpetka2001:
I'm not exactly sure why this is needed, but it seems to be indeed the solution for Neovim stable. In Neovim nightly that I use, it worked fine without the treesitter dependency. As discussed in #1856 I didn't observe the problem initially because I was using nightly and I only discovered it in the end by chance.

abeldekat:
Strange though. `Treesitter` is loaded on `VeryLazy`. How could there be a difference between nightly and 0.9 in that regard?

dpetka2001:
@abeldekat Would you mind trying it out to see if you can confirm? Although I also used a fresh starter template to make sure nothing of my personal config might give different results.

dpetka2001:
So, users will have to wait for Neovim v10 stable release or will it get taken care of in 0.9.5 if there's such a release? I'm guessing nightly has `vim.treesitter` latest API changes and doesn't have this problem?

abeldekat:
@dpetka2001, I have no idea. I don't understand why adding `nvim-treesitter` as a dep fixes the issue in 0.9.4, as `treesitter` is loaded before I invoke the test.

abeldekat:
on 0.9.4 this works as well:

```lua
    "nvim-neotest/neotest",
    tag = "v3.4.7",
```

dpetka2001:
Yes, they did mention that the PR linked in your comment was going to take place from neotest v4 onwards, so I'd guess pinning neotest to an older version prior to v4 would make sense. Maybe Folke could pin neotest to this version for Neovim stable like he had done in the past for telescope for users with Neovim 0.8?

ohseunghyeon:
Hi. i don't know what to do with this PR
should I just leave it or update this commit to be like below?
or maybe should i close it?

```
    "nvim-neotest/neotest",
    tag = "v3.4.7",
```

abeldekat:
Hello @ohseunghyeon,

I think we should wait for Folke to decide.

Please allow me to summarize my thoughts:

Relevant:
[neotest PR](https://github.com/nvim-neotest/neotest/pull/298) 
[neotest issue](https://github.com/nvim-neotest/neotest/issues/303)

The PR at hand adds `nvim-treesitter` as a dependency to `neotest`, otherwise tests don't run on neovim 0.9.4. Alternatively, it's also possible to pin neotest to v3.4.7. 
The change is not necessary for Neovim nightly.

I think it's very strange that on Neovim 0.9.4 the tests won't run, unless the dependency on `nvim-treesitter` is made explicit. 
The `neotest` plugin is loaded when pressing for example `leader t t`.
However, `nvim-treesitter` should have already been loaded, suggesting something is not quite right when loading `nvim-treesitter`.

I dug a little deeper into that loading:
```lua
  {
    version = false, -- last release is way too old and doesn't work on Windows
    build = ":TSUpdate",
    event = { "LazyFile", "VeryLazy" },
    -- truncated...
   }
```

1. change: `event = function() return { "VimEnter" } end` -> the tests run fine
2. add: `lazy = false` -> the tests run fine

@folke: Is the handling of `LazyFile` or `VeryLazy` different in 0.9.4, as opposed to nightly?

Best regards!

folke:
I'll look into it.
With the neotest change, I can only assume that it would fail id nvim-treesitter was not yet loaded when neotest loads, since the `queries` would not have been on the rtp yet.

But nvim-treesitter should have been loaded. Will see what's happening here

folke:
Was able to properly fix this for all plugins and related issues (like headlines.nvim).

See the commit for more details.

abeldekat:
That's great! Thanks!

abeldekat:
@folke,

It works like a charm, but I don't understand why...)

The `neotest` plugin is only activated on keypress. The UI shows that `nvim-treesitter` is loaded before I activate the tests.

When I load `nvim-treesitter` eagerly(`lazy = false`), the tests ran fine. Thus, it must have also loaded the `query-predicates`.

On `event = {"LazyFile", "VeryLazy"}` no tests where found. You fixed this by eagerly loading `nvim-treesitter.query_predicates`, using `init`.


I would like to understand: Why does it not load those `nvim-treesitter.query_predicates` on for example `VeryLazy`? A timing issue is not likely, as I wait a few seconds before executing the tests.

folke:
The python neotest adapter loads on filetype python, and that will load neotest as well.
The problem is mostly when opening a python file with `nvim foo.py`.

Due to `LazyFile`, the file will be rendered as quickly as possible and delay loading the plugins with `event=LazyFile`.

Since that neotest adpater has `ft=py`, it will load anyway since that's not on `LazyFile`. So it will load, then create it's queries and since treesitter has not yet been loaded, it won't have any queries.

This all used to work when they still did `require("nvim-treesitter")` in their code.

Since that was removed, this is no longer the case. Same thing with some other plugins like headlines.nvim.

Tbh, those plugins should also just be fixed in that they should not assume all queries are on the rtp when their setup function is called.

See headlines.nvim for example: https://github.com/lukas-reineke/headlines.nvim/blob/master/lua/headlines/init.lua#L20

Just requiring the init module will set the queries forever. If some queries become available after that on the rtp, they will never be used.

folke:
`nvim-treesitter` should also really be split in two separate plugins (that work is underway, but takes a while):

- treesitter-queries: simple plugin that puts queries on the rtp, allows building grammers and adds custom directives. Basically what I'm doing now in the `init()` function
- treesitter-modules: the additional functionality it currently provides like move etc

abeldekat:
Thanks for the explanation! 
I think I now understand the core of the problem, at least regarding `headlines.nvim`.

> The python neotest adapter loads on filetype python, and that will load neotest as well

I must be overlooking something. When I open a python file, `neotest` and the python adapter are not loaded... In the code of the adapter I do not see `ft=py`.

folke:
That's indeed odd. I didn't actually look into detail in the python issue, since I assumed it was caused by the same problem.

Hmm, maybe because some other code first gets the lang before nvim-treesitter was loaded? And NEovim itself doesn't get it again after adding nvim-treesitter to the rtp?

https://github.com/nvim-neotest/neotest/commit/b04332c2a07fbc13f4162b4b3f1820b433c428e7#diff-e59a70a3a602ce80307634f6bc6d3d6b26425eadba94eae59e3abfa946471062L117

abeldekat:
That's possible I think. I will try without using any other plugins than strictly necessary.

Also, perhaps worth noting:
When I modify `lazy.core.util`, the `verylazy` method, to use `VimEnter` instead of `UIEnter`, the tests also ran...

```lua
        vim.api.nvim_create_autocmd("VimEnter", {
          once = true,
          callback = function()
            _load()
          end,
        })
```

abeldekat:
Using the latest lazy.nvim and LazyVim, only loading the following plugins:

```lua
  Loaded (16)
    ● lazy.nvim 8.47ms  init.lua
    ● lazyflex.nvim 0.05ms  start
    ● LazyVim 3.43ms  start
    ● mason-lspconfig.nvim 0.07ms  nvim-lspconfig
    ● mason.nvim 10.51ms  nvim-lspconfig
    ● neoconf.nvim 0.17ms  nvim-lspconfig
    ● neodev.nvim 1.86ms  nvim-lspconfig
    ● neotest 11.49ms  <leader>tt
    ● neotest-python 0.65ms  neotest
    ● nvim-lspconfig 20.32ms  LazyFile
    ● nvim-treesitter 29.35ms  LazyFile
    ● nvim-treesitter-context 16.9ms  LazyFile
    ● nvim-treesitter-textobjects 1.72ms  nvim-treesitter
    ● nvim-ts-autotag 2.11ms  LazyFile
    ● plenary.nvim 0.28ms 󰢱 plenary.path  neotest
    ● tokyonight.nvim 1.07ms colorscheme tokyonight  LazyVim
```

... and the tests will not run without the new init property on the treesitter spec.

However, when I change lazy.nvim:
```lua
  vim.api.nvim_create_autocmd("User", {
    pattern = "LazyDone",
    once = true,
    callback = function()
      if vim.v.vim_did_enter == 1 then
        _load()
      else
        vim.api.nvim_create_autocmd("VimEnter", {
          once = true,
          callback = function()
            _load()
          end,
        })
      end
    end,
  })
end
```

... when using VimEnter instead of UIEnter, the tests run without the `init` part of the `treesitter` spec.

abeldekat:
When reverting the following [change](https://github.com/nvim-neotest/neotest/pull/298/commits/315d5d43fb159d073f8dec4fb960a42b0addb55f#diff-fb1101b1066a71d55d6c50df170de1a5800326cc8a5a58af69ecc887b4858534L69) in neotest's `subprocess.lua`, the tests run without needing the `init` part of the `treesitter` spec:

Add:
`nio.fn.rpcrequest(child_chan, "nvim_exec_lua", "require('nvim-treesitter')", {})`

folke:
Right! Now I understand. Neotest runs the tests in a new Neovim instance and nvim-treesitter would not have been loaded there automatically. That's why it failed.

folke:
Probably best to report this, since a lot of users (non LazyVim) may have the same issue when they lazy-load nvim-treesitter

abeldekat:
I also think that the core of the problem is the new `Neovim` instance.

However, `nvim-treesitter` *is* loaded automatically, when its spec is changed into:

- `lazy = false`
- or: `event = "VimEnter"`

It also loads when changing the implementation of `VeryLazy` itself, using `VimEnter` instead of `UIEnter`.

Perhaps this situation "uncovers" a hidden limitation in the current `VeryLazy` and `LazyFile` [implementation](https://github.com/folke/lazy.nvim/blob/16603c6917435d8446f7357cb61095138a417085/lua/lazy/core/util.lua#L145) . Is is possible that `UIEnter` never fires in the new `Neovim` instance and `nvim-treesitter` is not loaded because of that?

folke:
Again it will load nvim-treesitter, but it will load it too late. neotest expects it to be loaded when Neovim starts so on VimEnter. (Neotest loads itself on vimenter when in a running test) The tests are not running in your current Neovim instance, but in a new one. See that code that you linked to.

So no, this is not a limitation of lazy.

And yes, it works on nightly because it ships queries for python that will be on the rtp from the start.

abeldekat:
> Again it will load nvim-treesitter, but it will load it too late. neotest expects it to be loaded when Neovim starts so on VimEnter. (Neotest loads itself on vimenter when in a running test) 

I understand!

> Probably best to report this, since a lot of users (non LazyVim) may have the same issue when they lazy-load nvim-treesitter

Folke, would you be willing to do so?

Best regards and many thanks for answering my questions!

abeldekat:
@folke,

If you agree, I am planning to create an issue for `neotest` later today.

folke:
sure, go for it. Or maybe just add a comment to this issue here https://github.com/nvim-neotest/neotest/issues/303

abeldekat:
done.

PR #1864, opened 2023-10-24
fix: fix jdtls not spwaning in windows

dragove:
on windows system, it seems the executeble in cmd path can not be invoke directly from lua, using vim.fn.exepath can solve this problem.


folke:
thanks!

PR #1871, opened 2023-10-25
fix(mini.indentscope): remove duplicated filetype

andresatierf:


folke:
Look again. It's **not** a duplicate and it's there for a reason

andresatierf:
@folke I think you're referring to the capitalization, but this is the exact value of line 263

PR #1877, opened 2023-10-26
fix(util): pcall reference to deleted buffer

lparry:
See #1863

This does stop the error I was seeing, but I'm not super familiar with the language or the APIs to know if it's remotely correct.


folke:
I just checked using the plugin itself and added a proper fix. Now I also understand why you hit that other problem with deleting the augroup. All works as expected now

lparry:
Thank you so much @folke, love your work! ❤️

PR #1879, opened 2023-10-26
fix(telescope): anonymous keymap

andresatierf:
the map function receives a opts table but the desc is overridden by telescope. until that changes, this seems like a good alternative


andresatierf:
I've updated it. Thanks for that explanation, I've never really understood what `cwd = Util.root()` and `cwd = false` was for since I mostly navigate to the target directory before opening nvim

dpetka2001:
The `desc` "Open root directory" is actually the opposite of what it does. When setting `cwd = false` it actually disregards the root directory and instead searches in the current working directory.

folke:
Thanks!

PR #1882, opened 2023-10-26
братик ну по факту это лучше commit one

vova17s:


dpetka2001:
Quite a few trolls lately? Another candidate for the blocking list maybe?

amaanq:
Second repo that ik suffering from the bots, just like tree-sitter...

folke:
blocked

PR #1885, opened 2023-10-26
pre commit hook

jamescodegen:


folke:
?

amaanq:
These are just spam bots that GH doesn't seem to fix - tree-sitter core has the same issue (check closed PRs/issues that are not planned/locked -_-)

PR #1909, opened 2023-10-30
fix(keymaps): better up/down keymaps description

andresatierf:
j/k are shown correctly but I added a description anyways for consistency
image


folke:
Thanks!

PR #1930, opened 2023-11-02
prevent "nvim Undefined global `vim`."

v-sim:
when create new lua-file:
…
vim.fn.…● nvim Undefined global `vim`.

### DIFF
--- a/lua/lazyvim/plugins/lsp/init.lua
+++ b/lua/lazyvim/plugins/lsp/init.lua
@@ -59,6 +59,10 @@ return {
               completion = {
                 callSnippet = "Replace",
               },
+              diagnostics = {
+                -- Get the language server to recognize the `vim` global
+                globals = { "vim", "require" },
+              },
             },
           },
         },


dpetka2001:
This is already handled by `lua_ls` and `neodev` in Lazyvim. Do you happen to have changed the default configuration in your config and needed to add this? I don't believe there's a need for this in the default Lazyvim configuration.

v-sim:
> This is already handled by `lua_ls` and `neodev` in Lazyvim. Do you happen to have changed the default configuration in your config and needed to add this? I don't believe there's a need for this in the default Lazyvim configuration. Or if you mean a file outside of neovim config directory, then you just need to add a `.neoconf` file with the same configuration as the one that comes with Lazyvim.

And then why 'lua_ls =' in the default config at all? This is already handled by `lua_ls` and `neodev`.

I would add more:
```
              telemetry = {
                -- Do not send telemetry data containing a randomized but unique identifier
                enable = false,
              },
```
to be turned off by default

folke:
Neodev handles that.
telemetry is disabled by default, so does not need to be set.

PR #1935, opened 2023-11-02
Add keymaps for BufferLineMovePrev / BufferLineMoveNext

thedalelakes:
LazyVim's Bufferline config currently does not expose any keymaps by default to re-order tabs via BufferLineMovePrev / BufferLineMoveNext.

This PR simply adds keymaps to support moving the current buffer left or right.


agriffis:
Would this make more sense as `]B` and `[B`? Otherwise it seems like it could be easy to accidentally close buffers (existing mapping for `<leader>bl` and `<leader>bh`) when attempting to move them.

dale-lakes:
@agriffis I agree. Updated.

folke:
Thanks!

PR #1945, opened 2023-11-05
Change default project keymap

Gleland:
The current keymap conflicts with the `open neovim config` keymap, so users must either add their own override or type out the full command each time. This change proposes an unused keymap that's still intuitive. 

Thanks for your time and consideration!

### DIFF
--- a/lua/lazyvim/plugins/extras/util/project.lua
+++ b/lua/lazyvim/plugins/extras/util/project.lua
@@ -16,7 +16,7 @@ return {
           end)
         end,
         keys = {
-          { "<leader>fp", "<Cmd>Telescope projects<CR>", desc = "Projects" },
+          { "<leader>p", "<Cmd>Telescope projects<CR>", desc = "Projects" },
         },
       },
     },


dpetka2001:
Excuse my, perhaps, ignorance, but I fail to see the conflict in the initial keymap. `<leader>fp` doesn't seem to be used elsewhere from a simple search that i did. Or am i missing something?

Gleland:
Ah, this was a slight mistake on my part. I had originally found this suggestion for [Find Plugin File](https://github.com/LazyVim/lazyvim.github.io/blob/98f9d695bedfdf8d573c991b8579e9144c8298e8/docs/configuration/plugins.md?plain=1#L101C1-L105) and later edited it in my dotfiles. The snippet for reference is:

```
    {
      "<leader>fp",
      function() require("telescope.builtin").find_files({ cwd = require("lazy.core.config").options.root }) end,
      desc = "Find Plugin File",
    },
  ```

I think it might still be beneficial to adjust the keymap so it's not conflicting with existing guidance on the website for another action.

dpetka2001:
My guess is that the keymap you're referring to in the docs was meant to serve as an example. It doesn't have an immediate impact on the default Lazyvim configuration itself. So, I would suggest to either make a PR in the docs repo to correct that if you like or a new keymap for the action that you're referring to in the docs. I don't believe there's a reason to change an extra's keymap if there's no conflict. And if you do decide to pick a new keymap, I think it would be beneficial to follow the convention of keeping it in the `+file/find` section with something like `<leader>fx` (x can be whatever keymap that doesn't cause conflict).

Gleland:
> I don't believe there's a reason to change an extra's keymap if there's no conflict

Yes, I thought about this some more, and I agree. I'll close this as there's no conflict. Thanks for your time!

PR #1949, opened 2023-11-05
feat(keymaps): add toggle for mouse

aofei:
I personally find this toggle very useful, especially when I'm working on my MBP (the touchpad is just too large). Hope others find it useful too.


folke:
Not really something I think is needed. Thanks

PR #1958, opened 2023-11-07
docs(pt): add readme in portuguese (pt-pt)

andresatierf:


folke:
Sorry for the delay. Thanks!

PR #1966, opened 2023-11-07
Add ruff-format formatter in extras

martijnkruiten:
Ruff can be used as a drop-in replacement for Black. This code seems to work to integrate it into the LazyVim environment as an alternative to Black, but I'm not sure if everything here makes sense, especially the Mason part.

### DIFF
--- a/lua/lazyvim/plugins/extras/formatting/ruff_format.lua
+++ b/lua/lazyvim/plugins/extras/formatting/ruff_format.lua
@@ -1,0 +1,26 @@
+return {
+  {
+    "williamboman/mason.nvim",
+    opts = function(_, opts)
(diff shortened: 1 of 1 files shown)


dpetka2001:
Doesn't `ruff-lsp` already do this? Doesn't it use the `ruff` cli format tool to provide formatting (from what I can see in this [commit](https://github.com/astral-sh/ruff-lsp/pull/230)). I believe that if you have the `python` extra enabled and you don't enable any formatter via `conform/none-ls`, then `ruff-lsp` would be used by default for formatting, which would use the `ruff format` command (based on the commit that I mentioned). Of course, I'm not 100% sure myself either, so I'd appreciate the input of another more experienced user as well.

PR #1980, opened 2023-11-09
perf: Gitsigns travels when using wrap=true

grChad:
When using HTML and Markdown with the set.opt.wrap option set to 'true', **Gitsigns** in the _StatusColumn_ bar moves and doesn't stay in place.

**before:** _HTML_ and _Markdown_
example01

example02

**after:**
example03
example04


sibouras:
@folke this also fixes the statuscolumn when doing `Gitsigns diffthis`

before:
gitsigns-before

after:
gitsigns-after

PR #1989, opened 2023-11-10
fix(extras): Extend conform and nvim-lint formatters/linters dynamically

ItsDrike:
In many of the lang and formatter extras, the settings for `nvim-lint` and `conform.nvim` hard-code the `formatters_by_ft` or `linters_by_ft`  for the affected filetypes in opts. This could however potentially mean that once multiple extras are added that would both add a linter/formatter for the same filetype, they would start overwriting each other. (as far as I know, lazy.nvim won't merge lists).

This change fixes this and instead extends the list of formatters/linters for given filetype dynamically, using `vim.list_extend` or `table.insert`.

### DIFF
--- a/lua/lazyvim/plugins/extras/formatting/black.lua
+++ b/lua/lazyvim/plugins/extras/formatting/black.lua
@@ -17,10 +17,10 @@ return {
   {
     "stevearc/conform.nvim",
     optional = true,
-    opts = {
-      formatters_by_ft = {
-        ["python"] = { "black" },
-      },
-    },
+    opts = function(_, opts)
+      opts.formatters_by_ft.python = opts.formatters_by_ft.python or {}
+      table.insert(opts.formatters_by_ft.python, "black")
+      return opts
+    end,
   },
 }
--- a/lua/lazyvim/plugins/extras/formatting/prettier.lua
+++ b/lua/lazyvim/plugins/extras/formatting/prettier.lua
@@ -17,8 +17,20 @@ return {
   {
     "stevearc/conform.nvim",
     optional = true,
-    opts = {
-      formatters_by_ft = {
+    opts = function(_, opts)
+      --- Extend the conform plugin config and add given formatters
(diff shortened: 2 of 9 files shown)


Jomik:
I don't really think this is needed? LazyVim official extras should not have the problem, as they are more or less grouped per filetype?
And if users modify the list, then they should be doing this, as they normally do.
My worry with this change is that extras could, if loaded after your own changes, add things that you do not want them to.
E.g. if I do not want `goimports`, I need to make sure the extra is loaded first.
Extras usually are loaded first though.

ItsDrike:
Hm, I assumed it was determined that the extras will get loaded first. Especially considering this exact extending logic is used for none-ls already anyway:

https://github.com/LazyVim/LazyVim/blob/68ff818a5bb7549f90b05e412b76fe448f605ffb/lua/lazyvim/plugins/extras/formatting/black.lua#L9-L16

dpetka2001:
I don't think it matters what you use. Generally you use a function if you want to require other things as well (like `require("null-ls")`) according to what I have seen being done in Lazyvim's configuration (but I know very basic Lua, so someone should correct me in case I'm wrong). The tables in `opts` in the `conform` spec are dictionaries, not lists. `formatters_by_ft` and `["python"]` are both keys and they accept the values on the right. `lazy.nvim` already handles the `opts` [here](https://github.com/folke/lazy.nvim/blob/96584866b9c5e998cbae300594d0ccfd0c464627/lua/lazy/core/plugin.lua#L637) depending on if they're a list or a table. In this case, since it's not a list `Util.merge` (which is similar to `vim.tbl_deep_extend`) will be used to merge the user customizations with the default ones. But anyway I believe Folke is the one more appropriate to give a definitive answer to this. This is just my opinion, which might as well be wrong.

folke:
I don't see why this is needed. Loading of specs is always in this order:
- LazyVim core plugins
- LazyVim extras
- User plugins

Issue #2005, opened 2023-11-13
bug: inlay hints stopped working with latest neovim nightly


mati865:
### Did you check docs and existing issues?

- [X] I have read all the LazyVim docs
- [X] I have searched the existing issues of LazyVim
- [X] I have searched the existing issues of plugins related to this issue

### Neovim version (nvim -v)

NVIM v0.10.0-dev-748eae6

### Operating system/version

Ubuntu 22.04

### Describe the bug

I've been using inlay hints for a long time with Rust code but it broke in recent nightly and shows this stack trace when attaching to the file:
```
   Error  16:08:58 msg_show.lua_error Error detected while processing LspAttach Autocommands for "*":
16:08:58 msg_show Error executing lua callback: ...share/nvim/lazy/LazyVim/lua/lazyvim/plugins/lsp/init.lua:123: attempt to call upvalue 'inlay_hint' (a table value)
stack traceback:
	...share/nvim/lazy/LazyVim/lua/lazyvim/plugins/lsp/init.lua:123: in function 'on_attach'
	.../.local/share/nvim/lazy/LazyVim/lua/lazyvim/util/lsp.lua:32: in function <.../.local/share/nvim/lazy/LazyVim/lua/lazyvim/util/lsp.lua:29>
	[C]: in function 'nvim_exec_autocmds'
	.../nightly/nvim-linux64/share/nvim/runtime/lua/vim/lsp.lua:1722: in function '_on_attach'
	.../nightly/nvim-linux64/share/nvim/runtime/lua/vim/lsp.lua:1472: in function ''
	vim/_editor.lua: in function <vim/_editor.lua:0>
```
It's definitely related to https://github.com/neovim/neovim/commit/448907f65d6709fa234d8366053e33311a01bdb9 maybe there were other changes as well.

### Steps To Reproduce

1. Enable inlay hints
2. Using recent nightly neovim open a file will trigger LSP attach

### Expected Behavior

Inlay hints working properly.

### Repro

```Lua
-- DO NOT change the paths and don't remove the colorscheme
local root = vim.fn.fnamemodify("./.repro", ":p")

-- set stdpaths to use .repro
for _, name in ipairs({ "config", "data", "state", "cache" }) do
  vim.env[("XDG_%s_HOME"):format(name:upper())] = root .. "/" .. name
end

-- bootstrap lazy
local lazypath = root .. "/plugins/lazy.nvim"
if not vim.loop.fs_stat(lazypath) then
[...]

PR #2007, opened 2023-11-13
Detect if using nvim-0.10 and use new inlay_hint.enable method

garymjr:
Neovim introduced a breaking change to how inlay-hints were handled in [448907ff](https://github.com/neovim/neovim/commit/448907f65d6709fa234d8366053e33311a01bdb9). Check to see if the user is using nightly and if so use the newly added `.enable` method.

### DIFF
--- a/lua/lazyvim/config/keymaps.lua
+++ b/lua/lazyvim/config/keymaps.lua
@@ -120,7 +120,7 @@ map("n", "<leader>ud", function() Util.toggle.diagnostics() end, { desc = "Toggl
 local conceallevel = vim.o.conceallevel > 0 and vim.o.conceallevel or 3
 map("n", "<leader>uc", function() Util.toggle("conceallevel", false, {0, conceallevel}) end, { desc = "Toggle Conceal" })
 if vim.lsp.buf.inlay_hint or vim.lsp.inlay_hint then
-  map( "n", "<leader>uh", function() Util.lsp.toggle_inline_hint() end, { desc = "Toggle Inlay Hints" })
+  map( "n", "<leader>uh", function() Util.toggle.inlay_hints() end, { desc = "Toggle Inlay Hints" })
 end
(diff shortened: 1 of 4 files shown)


dpetka2001:
You might also change the keymap as I posted in the recently opened [issue](https://github.com/LazyVim/LazyVim/issues/2005#issuecomment-1808523152). But you might need to adjust it to take into consideration Neovim stable as well, as I only posted the solution for nightly as requested in the issue.

snelling-a:
change line 117 to 
```lua
	local inlay_hint = vim.lsp.buf.inlay_hint or vim.lsp.inlay_hint.enable
```
the function signatures are the same

garymjr:
> @dpetka2001  You might also change the keymap as I posted in the recently opened [issue](https://github.com/LazyVim/LazyVim/issues/2005#issuecomment-1808523152). But you might need to adjust it to take into consideration Neovim stable as well, as I only posted the solution for nightly as requested in the issue.

Thanks for catching that. Updated the keymap as well

lua/lazyvim/config/keymaps.lua
```diff
@@ -120,7 +120,7 @@ map("n", "<leader>ud", function() Util.toggle.diagnostics() end, { desc = "Toggl
 local conceallevel = vim.o.conceallevel > 0 and vim.o.conceallevel or 3
 map("n", "<leader>uc", function() Util.toggle("conceallevel", false, {0, conceallevel}) end, { desc = "Toggle Conceal" })
 if vim.lsp.inlay_hint then
```
dpetka2001, line 122: Should we maybe change this to `if vim.lsp.buf.inlay_hint or vim.lsp.inlay_hint`, so that older nightly versions benefit from the keymap as well? As far as I understand, the keymap will only be available if `vim.lsp.inlay_hint` is `true` right now. It doesn't take into account older nightly Neovim versions or am I wrong? Except if Folke decides to drop `vim.lsp.buf.inlay_hint` support all together.

garymjr: I _think_ that should be fine. I'm not sure if @folke had a reason for only checking against `vim.lsp` here and not also including `vim.lsp.buf`.

PR #2052, opened 2023-11-22
feat(lang): add Haskell config

mrcjkb:
As requested by @gacallea in https://github.com/mrcjkb/haskell-tools.nvim/issues/290.

@gacallea could you please test this? :smile:

### DIFF
--- a/lua/lazyvim/plugins/extras/lang/haskell.lua
+++ b/lua/lazyvim/plugins/extras/lang/haskell.lua
@@ -1,0 +1,98 @@
+return {
+
+  -- Add Haskell to treesitter
+  {
+    "nvim-treesitter/nvim-treesitter",
+    opts = function(_, opts)
+      if type(opts.ensure_installed) == "table" then
+        vim.list_extend(opts.ensure_installed, { "haskell" })
+      end
+    end,
+  },
+
+  {
+    "mrcjkb/haskell-tools.nvim",
+    version = "^3",
+    ft = { 'haskell', 'lhaskell', 'cabal', 'cabalproject' },
+    dependencies = {
+      { "nvim-telescope/telescope.nvim", optional = true },
+    },
+    config = function()
+      local ok, telescope = pcall(require, 'telescope')
+      if ok then
+        telescope.load_extension('ht')
+      end
+    end,
+  },
+
+  {
+    "williamboman/mason.nvim",
+    opts = function(_, opts)
+      opts.ensure_installed = opts.ensure_installed or {}
+      vim.list_extend(opts.ensure_installed, { "haskell-language-server" })
+    end,
+  },
+
+  { 
+    "mfussenegger/nvim-dap", 
+    optional = true,
+    dependencies = {
+      {
+        "williamboman/mason.nvim",
+        opts = function(_, opts)
+          opts.ensure_installed = opts.ensure_installed or {}
+          vim.list_extend(opts.ensure_installed, { "haskell-debug-adapter" })
+        end,
+      },
+    },
+  },
+
+  {
+    "nvim-neotest/neotest",
+    optional = true,
+    dependencies = {
+      { "mrcjkb/neotest-haskell", }
+    },
+    opts = {
+      adapters = {
+        ["neotest-haskell"] = {},
+      },
+    },
+  },
+
+  {
+    "mrcjkb/haskell-snippets.nvim",
+    dependencies = { "L3MON4D3/LuaSnip", },
+    config = function()
+      local haskell_snippets = require('haskell-snippets').all
+      require('luasnip').add_snippets('haskell', haskell_snippets, { key = 'haskell' })
+    end,
+  },
+
+  {
+    "luc-tielen/telescope_hoogle",
+    ft = { 'haskell', 'lhaskell', 'cabal', 'cabalproject' },
+    dependencies = {
+      { "nvim-telescope/telescope.nvim" },
+    },
+    config = function()
+      local ok, telescope = pcall(require, 'telescope')
+      if ok then
+        telescope.load_extension('hoogle')
+      end
+    end,
+  },
+
+  -- Make sure lspconfig doesn't start hls,
+  -- as it conflicts with haskell-tools
+  {
+    "neovim/nvim-lspconfig",
+    opts = {
+      setup = {
+        hls = function()
+          return true 
+        end,
+      },
+    },
+  },
+}


mrcjkb:
Marking as draft until tested.

gacallea:
Hi @mrcjkb , thank you :)

It works! With 2 issues:

- the top local variable doesn't work. I just removed it and placed the full lines where it is used.
- when I open an `.hs` file the nvim interface goes blank and I have to CTRL+C to see the file and get nvim back.

After that everything seems to be working just fine. The language server and completions is what I could quickly check.

If there is anything an Haskell noob could test here, let me know! :)

mrcjkb:
> It works! With 2 issues:
>
> - the top local variable doesn't work. I just removed it and placed the full lines where it is used.
> - when I open an .hs file the nvim interface goes blank and I have to CTRL+C to see the file and get nvim back.

Thanks for testing.

- I've inlined the top local variable.
- I tried out a local install (can't test it properly, because mason.nvim doesn't seem to be able to install haskell-language-server on NixOS), but I haven't been able to reproduce the blank interface.
  Can you try disabling plugins one-by-one to find out which one might be causing the issue?

gacallea:
Going to test now. I removed `haskell.lua` entirely and open an `.hs` file just to make sure this caused it. Without it I don't get the "blank nvim" issue. Going one by one now.

gacallea:
So, it is definitely something to do with `haskell-tools`. I tried one by one and left it for last. When I enable it I get the "blank" screen until I hit CTRL+C. This only happens if I open `.hs` files. I guess it's because it's lazy loaded. If there is a log to look into, I'm happy to assist.

Btw Mason installs hls fine here.

gacallea:
new development: I refrained from hitting CTRL+C and after a long while it opens the file. I guess some haskell-tools bootstrap process or something is causing this.

gacallea:
I also noticed (but may be unrelated?): Mason tries to install hls and kinda fails whenever I open a hs file? Tomorrow I'll be more torough. it is late now.

mrcjkb:
> I also noticed (but may be unrelated?): Mason tries to install hls and kinda fails whenever I open a hs file? Tomorrow I'll be more torough. it is late now.

Hmm, what happens if you comment out the lspconfig spec in the haskell.lua file? Does the "blank nvim" issue still occur?

gacallea:
> Hmm, what happens if you comment out the lspconfig spec in the haskell.lua file? Does the "blank nvim" issue still occur?

yes. I did try that already and I forgot to mention.

mrcjkb:
I'll have to see if I have time to continue this after work today. In the meantime, do you have a link to your config, so I can try to find potential conflicts?

gacallea:
> I'll have to see if I have time to continue this after work today. In the meantime, do you have a link to your config, so I can try to find potential conflicts?

thank you so much and sorry for the trouble.

my lazyvim repo: https://github.com/gacallea/lazyvim

I will try to remove everything but haskell.lua and see if it works. If something comes up I will let you know, otherwise we know already.

mrcjkb:
It might also be a good idea to troubleshoot without LazyVim using the [minimal config provided by haskell-tools](https://github.com/mrcjkb/haskell-tools.nvim#minimal-config).

gacallea:
So, I reverted to basically default lazyvim config + haskell.lua only: same issue. It seems to work but there is that "bootstrap" glitch that blanks the screen.

mrcjkb:
I think the mason.nvim issue is because [GHCup](https://www.haskell.org/ghcup/), which it uses to install hls, isn't installed.

I tried with a clean LazyVim install + the haskell.lua file and hls + ghc installed (GHCup is broken on NixOS, but mason can detect the existing hls installation), and it works fine for me.

gacallea:
@mrcjkb 

> I think the mason.nvim issue is because [GHCup](https://www.haskell.org/ghcup/), which it uses to install hls, isn't installed. The error message says "command `ghcup` not found". Unfortunately, mason doesn't seem to have a health check.

If I explicit mason in haskell.lua it works fine. Not sure this is the best, but it's a start:

```
  {
    "williamboman/mason.nvim",
    opts = function(_, opts)
      vim.list_extend(opts.ensure_installed, {
        "haskell-language-server",
      })
    end,
  },
```

> I tried with a clean LazyVim install + the haskell.lua file and hls + ghc installed (GHCup is broken on NixOS, but mason can detect the existing hls installation), and it works fine for me.

No matter what I try I still get the buggy behavior. I even tried the native Terminal app, same thing.

Is there a way to debug/monitor a nvim session in real time?

mrcjkb:
@gacallea I suggest we continue this discussion in https://github.com/mrcjkb/haskell-tools.nvim/issues/290 for now, as this seems specific to haskell-tools.nvim and your OS or setup.

Could you please do the following:

- Try to reproduce the issue using the [troubleshooting guide with plain nvim and a minimal confing](https://github.com/mrcjkb/haskell-tools.nvim#minimal-config).
- Run `:checkhealth haskell-tools` and post the output to https://github.com/mrcjkb/haskell-tools.nvim/issues/290
- Follow the [guide to enable debug logging and get the log files](https://github.com/mrcjkb/haskell-tools.nvim#logs):
  - You can set `vim.g.haskell_tools` anywhere in your config, e.g. in the `config` or `init` function of the plugin spec.
  - Post the log files to https://github.com/mrcjkb/haskell-tools.nvim/issues/290
  
I'm away for the weekend, so I might not be able to pick this up again until Monday evening :disappointed:

mrcjkb:
Tested and confirmed working by @gacallea :pray:

trevarj:
Should `fourmolu` be added in the mason ensure_installed list since it is the haskell-tools default formatter?

mrcjkb:
> Should `fourmolu` be added in the mason ensure_installed list since it is the haskell-tools default formatter?

The [hls-fourmolu-plugin](https://hackage.haskell.org/package/hls-fourmolu-plugin) uses fourmolu as a library, so you don't need it to be installed for it to work.

trevarj:
@mrcjkb I'm actually having trouble getting it working with the config on the PR 😳

mrcjkb:
> More context:
Getting this warning in the log file:
Warning | No plugin enabled for "textDocument/formatting" 

This means that your haskell-language-server installation doesn't have the fourmolu plugin enabled, or you're using a ghc version that can't compile the fourmolu plugin for the version of haskell-language-server you're using.

See also: https://haskell-language-server.readthedocs.io/en/latest/support/plugin-support.html#current-plugin-support-tiers

haskell-language-server doesn't shell out to a fourmolu installation. So if the mason fourmolu install is working for you, it's probably another plugin doing the formatting (maybe nil-ls or conform.nvim), not haskell-language-server.
If that's the case, I would advise against adding fourmolu to the mason `ensure_installed` list, because not everybody might want to use that formatter.

trevarj:
@mrcjkb yep, that was it, thanks! Classic user being too stupid for Haskell 😖 !

mrcjkb:
> @mrcjkb yep, that was it, thanks! Classic user being too stupid for Haskell 😖 ! Downgraded to an earlier GHC and now formatting works but the eval plugin doesn't....

haskell-language-server can be finicky when it comes to plugin support for recent GHC versions :disappointed:

mrcjkb:
Does anyone who was involved in this PR still use the `extras.lang.haskell` module? 
The plugin pins are quite outdated, so I'm led to believe this module is unmaintained.

Every now and then I get LazyVim users opening issues in the haskell-tools repo, so if `lang.haskell` doesn't have a maintainer, I'd recommend to remove it (or at least remove my plugins and replace haskell-tools with nvim-lspconfig).

cc @folke 

Cheers :)

andreacfromtheapp:
Hi, not using Haskell nor this tool anymore. Thank you for your tools, help, and time :)

PR #2053, opened 2023-11-22
Change default conceal level to 2

awilkins:
3 hides all concealed text, even if there are replacement chars defined

e.g. : for Markdown, 3 hides list item markers, dots for asterisk lists and em-dashes for dash lists

If replacements are defined as defaults in a plugin, the "Laziest™" config would seem to be displaying them.

Level 1 only seems useful when alignment really matters.


Mayrixon (approved):
I agree with this change.

I wrote the extra config for lang/tex.lua in LazyVim, and found `conceallevel=3` would reduce readability of long equations in LaTeX. In my personal configurations, I set `conceallevel=2` as the global config and found no effect on markdown files.

folke:
Makes sense. Thanks!

PR #2071, opened 2023-11-27
fix: correct eslint working directories setting

nickhstr:
The eslint LSP has changed one of the setting names from `workingDirectory` to `workingDirectories`.
See its documentation [here](https://github.com/microsoft/vscode-eslint/blob/17e79c0f7bcd38f3033d0fbc346a5607fbb4c60f/README.md?plain=1#L218).


folke:
thanks!

PR #2088, opened 2023-11-30
feat(ui): add function to toggle between light/dark backgrounds

NickSager:
Added a function in the ui menu to toggle between light and dark background themes. Helpful for themes like 'gruvbox' where color schemes are in one package and can't be changed with the theme switcher menu, but also works with other themes that support it.

I find this useful so I wanted to share in case someone else might too.


folke:
Thanks!

ugur-a:
This seems to be reset anytime I close LazyVim. Is this expected behaviour or should I open an issue?

dpetka2001:
@ugur-a This function just toggles the background for the current running session of Neovim. If you want a permanent change you should change `vim.opt.background` in your `/lua/config/options.lua` according to the background you'd like. See `:h 'background'` for more info.

Issue #1139, opened 2023-07-15
feature: Which Font do you use exactly in LazyVim


kuki-xie:
### Did you check the docs?

- [X] I have read all the LazyVim docs

### Is your feature request related to a problem? Please describe.

# Please list the exactly suggested font in the lazyVim website.
Now when i install the lazyvim, there exits a problem on the icons because the nerd-font i download.

### Describe the solution you'd like

list the exactly suggested font in the lazyVim website

### Describe alternatives you've considered

Just tell us the nerd-font you suggest to install.

folke:
It's listed in the requirements on the website

kuki-xie:
yeah i have noticed that but when i download nerd font such as Hack Nerd Font，3270 nerd font which version> 3.0. it has problems
 To show the icon before Find file(when I type nvim and loaded in the window)

发自我的 iPhone

在 2023年7月15日，17:09，Folke Lemaitre ***@***.***> 写道：

﻿

It's listed in the requirements on the website

kuki-xie:
[截屏2023-07-15 17.17.01.png]
The font I use is above.
But there exits problems to show the icon before Find File
[截屏2023-07-15 17.18.55.png]

2023年7月15日 17:08，Folke Lemaitre ***@***.***> 写道：



Closed #1139<https://github.com/LazyVim/LazyVim/issues/1139> as not planned.

folke:
Then it means your terminal is not using the Nerd Font v3

Siddaarth:
I think it is a issue when using iTerm2, since when I launch lazy vim on my default mac terminal, I do not experience the search icon loads normally. I also saw someone on discussion with the exact issue. This is despite using Nerd Font V3.

tjwelde:
Same issue for me using iterm2. Works in macos Terminal.
Has anyone found a solution to this?

tjwelde:
Seems like there is a fix in a nightly build of iTerm2: https://github.com/ryanoasis/nerd-fonts/issues/1149#issuecomment-1475762974

cerebrl:
I personally just change the config for the "dashboard" to use a glyph that renders without issue in my iTerm2: https://github.com/cerebrl/nvim/blob/main/lua/plugins/dashboard.lua. For the "magnifying glass" icon specifically, I use `F0349` from NerdFont symbols. 

[...]

PR #2102, opened 2023-12-01
fix: Can't display search icon on dashboard in iTerm2

greycodee:
截屏2023-12-01 20 43 25


greycodee:
截屏2023-12-01 21 28 31

dpetka2001:
You only changed this for `alpha` dashboard. Do the rest of the dashboards appear fine in iterm2 (I would guess not, since this seems to be a code point issue)? If not, maybe you could do the same for the other dashboards, since users might use different dashboards that come with LazyVim?

greycodee:
fix: https://github.com/LazyVim/LazyVim/issues/1139#issuecomment-1806858433

PR #2118, opened 2023-12-03
Fixing telescope find files keymaps.

ckt114:
Fixing telescope find files keymaps that seems to be mixed up.

### DIFF
--- a/lua/lazyvim/plugins/editor.lua
+++ b/lua/lazyvim/plugins/editor.lua
@@ -135,12 +135,12 @@ return {
       },
       { "<leader>/", Util.telescope("live_grep"), desc = "Grep (root dir)" },
       { "<leader>:", "<cmd>Telescope command_history<cr>", desc = "Command History" },
-      { "<leader><space>", Util.telescope("files"), desc = "Find Files (root dir)" },
+      { "<leader><space>", Util.telescope("files", { cwd = false }), desc = "Find Files (root dir)" },
       -- find
       { "<leader>fb", "<cmd>Telescope buffers sort_mru=true sort_lastused=true<cr>", desc = "Buffers" },
       { "<leader>fc", Util.telescope.config_files(), desc = "Find Config File" },
-      { "<leader>ff", Util.telescope("files"), desc = "Find Files (root dir)" },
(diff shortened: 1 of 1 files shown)


dpetka2001:
They're not mixed up. By default, in `Util.telescope` `cwd` is set to `Util.root`, which searches in `root_dir`. If you set `cwd = false`, then it defaults to `vim.loop.cwd` and searches in current working directory.

ckt114:
Could you please try those keys again? When I open a file in subfolders of a git repo and do `<leader><space>` the first time it works correctly but when I did that again it only showed files in the folder when the current file is opened.

dpetka2001:
Do you happen to have some kind of patterns in your subfolders that are being considered by LSP? What does `:LazyRoot` tell you? It sounds to me like LSP misconfiguration. I will check tomorrow, cuz it's pretty late here. I'm sure the maintainer will also share his thoughts starting tomorrow.

ckt114:
Using your `:LazyRoot` I was able to determine that the `.terraform` folder in the same folder as the opening file is causing this. Is there a way for me to tell it to not use `lsp` folder as the root? Use `.git` folder as root instead?

image

dpetka2001:
1. You could try changing `vim.g.root_spec` in your `options.lua` file, but this will apply to all LSP servers
2. Specifically for the LSP server that you use, you can try the following
```lua
return {
  {
    "neovim/nvim-lspconfig",
    opts = {
      servers = {
        name_of_lsp_server = {
          root_dir = require("lspconfig.util").root_pattern(".git", ".terraform"), -- here order matters so make sure `.git` is before `.terraform`
        },
      },
    },
  },
}
```
3. Alternatively you could change slightly your workflow and open Neovim at the root of the project (I would assume where your `.git` folder is located in your case) and then you can use `<leader><space>` for `root_dir` search and `<leader>fF` for `cwd` search, which would be the directory in which you started Neovim (where your `.git` folder is located).

ckt114:
For #3 I did start nvim from the root git dir but it still having that issue. I'm going to go with your #2 suggestion. Thank you very much for your help.

PR #2134, opened 2023-12-05
fix wierd Tmux/terminal esc-j/k line move issue

rjshrjndrn:
More details: https://github.com/LazyVim/LazyVim/discussions/163

### DIFF
--- a/lua/lazyvim/config/options.lua
+++ b/lua/lazyvim/config/options.lua
@@ -49,7 +49,6 @@ opt.splitkeep = "screen"
 opt.splitright = true -- Put new windows right of current
 opt.tabstop = 2 -- Number of spaces tabs count for
 opt.termguicolors = true -- True color support
-opt.timeoutlen = 300
 opt.undofile = true
 opt.undolevels = 10000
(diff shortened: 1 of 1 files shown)


dpetka2001:
Are you sure this is the correct solution? From Neovim's [wiki](https://github.com/neovim/neovim/wiki/FAQ#esc-in-tmux-or-gnu-screen-is-delayed) they suggest setting `escape-time` in `.tmux.conf`. `timeoutlen` seems like a personal choice and I prefer it being low, otherwise `which-key` gets really slow.

rjshrjndrn:
I think so. Because all the other solutions I tried, have `esc j/k` issue. 
I think it's okay to keep it as default, and if someone is okay with the occasional esc j/k issue or doesn't use esc key much, can tweak the value to be smaller?

dpetka2001:
I see what you're saying. Anyway I'm not the one to decide. Just expressed my opinion. The maintainer will make the decision. Even if it does end up being default, I'm ok with lowering the value myself since I don't use tmux that much.

folke:
Won't merge this, since it would change the behavior of whichkey. closing

PR #2137, opened 2023-12-05
Adds copy file name command to Neo-Tree with 'Y' binding

pr3c0g:
I was missing an easy way to copy the file name from within the Neo-Tree window. I think others might find this useful and wanted to contribute :)

This PR simply adds the suggestion [here](https://github.com/nvim-neo-tree/neo-tree.nvim/issues/359#issuecomment-1164667350).

Thank you!


folke:
Awesome, thanks!

PR #2178, opened 2023-12-12
Update gitsigns.nvim preview command

skoch13:
Considering the update `preview_hunk_inline` is the default option(https://github.com/lewis6991/gitsigns.nvim/commit/d195f0c35ced5174d3ecce1c4c8ebb3b5bc23fa9) now I suggest to update the keybinding used for hunks preview


folke:
Thanks!

PR #2192, opened 2023-12-14
Add diagnostic signs to lsp options

garymjr:
As of [8122470](https://github.com/neovim/neovim/commit/8122470f8310ae34bcd5e436e8474f9255eb16f2) diagnostic signs are configured using `vim.diagnostic.config`. This adds the signs to the diagnostic options that are passing to the diagnostics config.

### DIFF
--- a/lua/lazyvim/plugins/lsp/init.lua
+++ b/lua/lazyvim/plugins/lsp/init.lua
@@ -26,6 +26,14 @@ return {
           -- prefix = "icons",
         },
         severity_sort = true,
+        signs = {
+          text = {
+            [vim.diagnostic.severity.ERROR] = require("lazyvim.config").icons.diagnostics.Error,
+            [vim.diagnostic.severity.WARN] = require("lazyvim.config").icons.diagnostics.Warn,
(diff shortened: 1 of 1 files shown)


dpetka2001:
@garymjr You might also want to remove these [lines](https://github.com/LazyVim/LazyVim/blob/879e29504d43e9f178d967ecc34d482f902e5a91/lua/lazyvim/plugins/lsp/init.lua#L112-L116) along with the rest of your PR, since they won't be used any more.

garymjr:
> @garymjr You might also want to remove these [lines](https://github.com/LazyVim/LazyVim/blob/879e29504d43e9f178d967ecc34d482f902e5a91/lua/lazyvim/plugins/lsp/init.lua#L112-L116) along with the rest of your PR, since they won't be used any more.

I had considered removing them, but since keeping them doesn't hurt anything leaving them there maintains backwards compatibility with older versions.

dpetka2001:
Oh, I see, you're right. Hadn't considered that. Thanks for clarifying.

folke:
Thanks!

PR #2198, opened 2023-12-16
feat(lang) replace rust-tools.nvim with rustacean.nvim - fixes #2113

### LINKED ISSUE #2113: feature: rust-tools.nvim is no longer maintained. Should be switched to Rustacean.nvim
### Did you check the docs?

rust-tools.nvim is no longer maintained, you can see that discussion [here](https://github.com/simrat39/rust-tools.nvim/issues/422). There is a new maintained fork called Rustacean [here](https://github.com/mrcjkb/rustaceanvim/tree/master).

Remove rust-tools from LazyVim and replace with rustacean.

Do nothing.

### Additional context

_No response_

appelgriebsch:


dpetka2001:
@appelgriebsch A question about potentially the user wanting to modify the settings. I have no idea how global vim variables work. In this specific case it seems to be a table. Do they have the ability to get merged? If a user were to modify some settings, does he have to override the whole `vim.g.rustaceanvim` configuration in this PR? If the variable is put in `opts` instead of `config` does this allow for more granular user control (merging the customized subtables with the table of the variable introduced in this PR)?

appelgriebsch:
updated PR and moved most of the configuration into the `opts`, so users can overwrite it. Still need to check on the key mappings though.

lua/lazyvim/plugins/extras/lang/rust.lua
```diff
@@ -45,82 +45,60 @@ return {
...
     "neovim/nvim-lspconfig",
     opts = {
       servers = {
-        -- Ensure mason installs the server
-        rust_analyzer = {
-          keys = {
-            { "K", "<cmd>RustHoverActions<cr>", desc = "Hover Actions (Rust)" },
-            { "<leader>cR", "<cmd>RustCodeAction<cr>", desc = "Code Action (Rust)" },
-            { "<leader>dr", "<cmd>RustDebuggables<cr>", desc = "Run Debuggables (Rust)" },
-          },
```
dpetka2001, line 99: I hope you don't take it as me bothering you, but maybe it would be better for the end user to expose the keys of this PR in `lspconfig` instead of `rustaceanvim`'s `on_attach` function, so it's easier for them to change them if they want to? Otherwise the user would have to mess with the `config` function to do that?

dpetka2001:
After your last commit the keymaps will also be available as `LazyKeys` handlers, which means that the plugin will be lazy-loaded if the user presses the keymap even not in a Rust file (I tried it locally even though it throws an error about the command, but the plugin loads in the Lazy UI). I also tried the `keys` section in `lspconfig.servers.rust-analyzer`, but it doesn't take effect as with the `rust-tools` (maybe due to the fact that the `setup` function in lspconfig for `rust-analyzer` just returns instead of it being setup by `rust-tools` like in the previous iteration). So, I ended up putting the keymaps in the `on_attach` function (like you did initially) but in the `opts` instead of the `config` and they are only available when `rust-analyzer` is attached and don't operate as `LazyKeys` handlers any more. 
Maybe you would like to take that into account or not.

lua/lazyvim/plugins/extras/lang/rust.lua
alejandrodnm: I'm not very familiar with how lazy works, but rustaceanvim is opinionated in the sense that it doesn't want `lspconfig` to handle anything related to `rust_analyzer`. 

Would disabling everything but the keys in the `lspconfig` work?

lua/lazyvim/plugins/extras/lang/rust.lua
```diff
@@ -45,82 +45,58 @@ return {
...
-                    autocmd CursorHold                      *.rs silent! lua vim.lsp.buf.document_highlight()
-                    autocmd CursorMoved,InsertEnter         *.rs silent! lua vim.lsp.buf.clear_references()
-                    autocmd BufEnter,CursorHold,InsertLeave *.rs silent! lua vim.lsp.codelens.refresh()
-                  augroup END
-                ]])
-          end,
+    "mrcjkb/rustaceanvim",
+    version = "^3", -- Recommended
+    ft = { "rust" },
+    opts = {
```
alejandrodnm, line 51: Could you add the `dap` and `tools` sections from 

https://github.com/LazyVim/LazyVim/discussions/1976#discussioncomment-7912177

lua/lazyvim/plugins/extras/lang/rust.lua
```diff
@@ -45,82 +45,58 @@ return {
...
+      server = {
+        on_attach = function(client, bufnr)
+          -- register which-key mappings
```
alejandrodnm, line 54: Could you add: `vim.lsp.inlay_hint.enable(bufnr, true)` to have inlay hints enable by default, like with `rust-analyzer`. 

This will only work with neovim's builtin inlay hints, which I think are only in nightly. That's why @trevarj was saying in the discussion "it is probably not worth migrating until nvim 0.10".

lua/lazyvim/plugins/extras/lang/rust.lua
dpetka2001: I tried it and didn't work, that's why the final commit put them in the `on_attach` function instead.

lua/lazyvim/plugins/extras/lang/rust.lua
appelgriebsch: `dap` is automatically configured in rustacean.nvim when installed via mason (similar to what we had before) -> https://github.com/mrcjkb/rustaceanvim/blob/master/lua/rustaceanvim/config/internal.lua#L207

lua/lazyvim/plugins/extras/lang/rust.lua
appelgriebsch: shouldn't that be part of core LazyVim LspAttach functionality when it's running with neovim >= 0.10.x? see here: https://github.com/LazyVim/LazyVim/blob/879e29504d43e9f178d967ecc34d482f902e5a91/lua/lazyvim/plugins/lsp/init.lua#L119-L121

alejandrodnm: Are they being enabled by default for you? I tried late last night, and they weren't for me. Maybe I had other issues, I'll try again.

lua/lazyvim/plugins/extras/lang/rust.lua
alejandrodnm: Awesome. When I tried it, it wasn't by default 👍

lua/lazyvim/plugins/extras/lang/rust.lua
appelgriebsch: hmmm might be related to this: https://github.com/appelgriebsch/Nv/blob/main/lua/plugins/lang.lua#L6

alejandrodnm: When I call `:lua print(vim.fn.has('nvim-0.10'))` it prints `1`

```
λ ➜ nvim --version
NVIM v0.10.0-dev-1890+g1cf51a07a
Build type: RelWithDebInfo
LuaJIT 2.1.1702233742
Run "nvim -V1 -v" for more info
```

`:lua print(vim.fn.has('v0.10.0-dev-1890+g1cf51a07a'))` returns `0`

lua/lazyvim/plugins/extras/lang/rust.lua
alejandrodnm: Tried it and works great 🔥

lua/lazyvim/plugins/extras/lang/rust.lua
trevarj: They only enable for me when I start typing in insert mode and then <esc>ing and saving

trevarj:
Am I understanding correctly that this change won't yet use the `rust-analyzer` installed by mason because [rustacean sets the cmd](https://github.com/mrcjkb/rustaceanvim/blob/8d735ebafa1924d0918c3aa5bc9a970e24d4b81a/lua/rustaceanvim/config/internal.lua#L180-L182) to use whatever is installed on the path? This was confusing to me at first cause I had an old version of rust-analyzer installed manually.

appelgriebsch:
> Am I understanding correctly that this change won't yet use the `rust-analyzer` installed by mason because [rustacean sets the cmd](https://github.com/mrcjkb/rustaceanvim/blob/8d735ebafa1924d0918c3aa5bc9a970e24d4b81a/lua/rustaceanvim/config/internal.lua#L180-L182) to use whatever is installed on the path? This was confusing to me at first cause I had an old version of rust-analyzer installed manually.

Not sure what you are referring to, but it works for me with the `rust-analyzer` installed via mason.nvim…

lua/lazyvim/plugins/extras/lang/rust.lua
mrcjkb: For neovim < 0.10, there's [a separate plugin](https://github.com/lvimuser/lsp-inlayhints.nvim) that enables inlay hints.
There's also [one by simrat39](https://github.com/simrat39/inlay-hints.nvim), the rust-tools author, which is based on the implementation in rust-tools.

The existence of those two plugins + the built-in support after nvim 0.10 was the reason I decided to drop inlay hints from rustaceanvim.

I don't remember how good the performance was in rust-tools, but compared to the nvim 0.10 implementation, it [appears to be quite slow with both plugins](https://github.com/mrcjkb/rustaceanvim/discussions/46#discussioncomment-7697717).

I suppose you could in theory add one of the plugins as a plugin spec with `optional = vim.fn.has('nvim-0.10')`.

alejandrodnm:
Just an FYI rust-tools has been archived with a message to migrate to rustacean 

https://github.com/simrat39/rust-tools.nvim

MichaelSasser:
I just tried it out with [3.15.0](https://github.com/mrcjkb/rustaceanvim/releases/latest) and it works like a charm (though, I've only used it for a couple of minutes).

`:checkhealth rustaceanvim` gives me:

```plain
==============================================================================
rustaceanvim: require("rustaceanvim.health").check()

Checking for Lua dependencies ~
- OK [mfussenegger/nvim-dap](https://github.com/mfussenegger/nvim-dap) installed.

Checking external dependencies ~
- OK rust-analyzer: found rust-analyzer 0.3.1799-standalone
- OK Cargo: found cargo 1.77.0-nightly (2ce45605d 2024-01-04)
- OK rustc: found rustc 1.77.0-nightly (ca663b06c 2024-01-08)
- OK debug adapter: found codelldb 

Checking config ~
- OK No errors found in config.

Checking for conflicting plugins ~
- OK No conflicting plugins detected.
```

<details><summary>Code</summary>
<p>Basically just copy & past.

```lua
-- lua/plugins/rustacean.lua
return {
  { "simrat39/rust-tools.nvim", enable = false },
  {
    "mrcjkb/rustaceanvim",
    version = "^3", -- Recommended
    ft = { "rust" },
    opts = {
      server = {
        on_attach = function(client, bufnr)
          -- register which-key mappings
          local wk = require("which-key")
          wk.register({
            ["<leader>cR"] = {
              function()
                vim.cmd.RustLsp("codeAction")
              end,
              "Code Action",
            },
            ["<leader>dr"] = {
              function()
                vim.cmd.RustLsp("debuggables")
              end,
              "Rust debuggables",
            },
          }, { mode = "n", buffer = bufnr })
        end,
        settings = {
          -- rust-analyzer language server configuration
          ["rust-analyzer"] = {
            cargo = {
              allFeatures = true,
              loadOutDirsFromCheck = true,
              runBuildScripts = true,
            },
            -- Add clippy lints for Rust.
            checkOnSave = {
              allFeatures = true,
              command = "clippy",
              extraArgs = { "--no-deps" },
            },
            procMacro = {
              enable = true,
              ignored = {
                ["async-trait"] = { "async_trait" },
                ["napi-derive"] = { "napi" },
                ["async-recursion"] = { "async_recursion" },
              },
            },
          },
        },
      },
    },
    config = function(_, opts)
      vim.g.rustaceanvim = vim.tbl_deep_extend("force", {}, opts or {})
    end,
  },
  {
    "neovim/nvim-lspconfig",
    opts = {
      servers = {
        rust_analyzer = {},
        taplo = {
          keys = {
            {
              "K",
              function()
                if vim.fn.expand("%:t") == "Cargo.toml" and require("crates").popup_available() then
                  require("crates").show_popup()
                else
                  vim.lsp.buf.hover()
                end
              end,
              desc = "Show Crate Documentation",
            },
          },
        },
      },
      setup = {
        rust_analyzer = function()
          return true
        end,
      },
    },
  },
}
```

</p>
</details>

lua/lazyvim/plugins/extras/lang/rust.lua
```diff
@@ -38,89 +37,64 @@ return {
...
+          local wk = require("which-key")
+          wk.register({
+            ["<leader>cR"] = { function() vim.cmd.RustLsp("codeAction") end, "Code Action" },
+            ["<leader>dr"] = { function() vim.cmd.RustLsp("debuggables") end, "Rust debuggables" },
+          }, { mode = "n", buffer = bufnr })
+        end,
+        settings = {
```
mrcjkb, line 59: suggestion: Change this key to `default_settings`, as this will allow users
to override it with a project-local rust-analyzer.json.

FYI I'm thinking about adding the clippy lints by default, if rustaceanvim detects a clippy installation.

amaanq: +1 for clippy, it's so useful

appelgriebsch: updated

jameshurst (commented):
These keymaps should probably be set using `vim.keymap.set` instead of which-key. This allows these keymaps to work even when which-key is disabled.

```lua
vim.keymap.set("n", "<leader>cR", "<Cmd>RustLsp codeAction<CR>", { desc = "Code Action", buffer = bufnr })
```

lua/lazyvim/plugins/extras/lang/rust.lua
```diff
@@ -150,13 +122,12 @@ return {
...
-      adapters = {
-        ["neotest-rust"] = {},
-      },
-    },
+    opts = function(_, opts)
+      opts.adapters = opts.adapters or {}
+      vim.list_extend(opts.adapters, {
+        require('rustaceanvim.neotest'),
+      })
+    end
```
jameshurst, line 130: This may want to be moved to the `dependencies` section of `mrcjkb/rustaceanvim`. 

Currently rustaceanvim will only use neotest as it's `test_executor` if the `rustaceanvim.neotest` module was loaded prior to rustaceanvim being configured. (https://github.com/mrcjkb/rustaceanvim/blob/4.7.0/lua/rustaceanvim/config/internal.lua#L42-L48)

JasterV:
Hi! Any updates regarding this? I'm highly interested on updating LazyVim to use rustacean instead of rust-tools

iamralch:
When should we expect this PR to get merged?

folke:
Sorry it took a while to merge this. I don't use rust often and was planning to fully test this first.

We'll see if any issues come up :)

Thanks!

dpetka2001:
@folke May I suggest to take into consideration this [comment](https://github.com/LazyVim/LazyVim/pull/2198#pullrequestreview-1873658780) and just use `vim.keymap.set` instead of registering the keymaps directly under `which-key`?

folke:
@dpetka2001 yes, makes sense! Feel free to make a PR to address this. Thanks

lua/lazyvim/plugins/extras/lang/rust.lua
mrcjkb: rustaceanvim loads its configuration lazily (when opening a rust file). So it should still pick up the neotest config as long as neotest isn't lazy loaded by lazy.nvim.

utkarshgupta137:
Was it really a good idea to go ahead with this, given that rustacean doesn't use rust-analyzer from mason.nvim (uses whatever is installed on the system) & also doesn't support neoconf. Me & my team had integrated with both the features pretty nicely & I was disappointed to see that both were broken after an update (there is a workaround for the first one). Also, it doesn't look like neoconf support is planned, given that https://github.com/mrcjkb/rustaceanvim/issues/53 was closed as won't fix. Hopefully, the new plugin fixed some issue for someone, but for me, rust-tools was also working very nicely. I've pinned to 10.10 for now, but I guess I'll have to stop using the LazyExtra to update...

shanesveller:
When my time permits I want to investigate whether it's straightforward to simply have Neoconf write any loaded customizations into the `vim.g.rustaceanvim` table with compatible contents, so that the authors' preferences on the topic don't force me to have one file in my projects to configure every other LSP, and one bespoke JSON file specifically for rust-analyzer.

dpetka2001:
You can just copy the previous configuration of `rust-tools` into `~/.config/nvim/lua/plugins/extras` and this way have it shown under `LazyExtras` as well. You just won't be using the Rust LazyExtra from LazyVim. It will be shown as a `User` Extra. I also have my own configuration for clang and it shows in the LazyExtras like `lang.clang   User  clangd_extensions.nvim  mason.nvim  nvim-lspconfig  nvim-treesitter` in addition to `lang.clangd  clangd_extensions.nvim  nvim-cmp  nvim-lspconfig  nvim-treesitter  mason.nvim  nvim-dap` (which is the default LazyVim `clangd` Extra). No need to pin down the LazyVim version to 10.10 just for this I believe.

utkarshgupta137:
> You can just copy the previous configuration of `rust-tools` into `~/.config/nvim/lua/plugins/extras` and this way have it shown under `LazyExtras` as well. You just won't be using the Rust LazyExtra from LazyVim. It will be shown as a `User` Extra. I also have my own configuration for clang and it shows in the LazyExtras like `lang.clang  User  clangd_extensions.nvim  mason.nvim  nvim-lspconfig  nvim-treesitter` in addition to `lang.clangd  clangd_extensions.nvim  nvim-cmp  nvim-lspconfig  nvim-treesitter  mason.nvim  nvim-dap` (which is the default LazyVim `clangd` Extra). No need to pin down the LazyVim version to 10.10 just for this I believe.

Yep that is what I had decided to do.

mrcjkb:
> When my time permits I want to investigate whether it's straightforward to simply have Neoconf write any loaded customizations into the `vim.g.rustaceanvim` table with compatible contents, so that the plugin author's preferences on the topic don't force me to have one file in my projects to configure every other LSP, and one single-tool JSON file specifically for rust-analyzer.

I don't use neoconf myself, and as I have elaborated in https://github.com/mrcjkb/rustaceanvim/issues/53, there's not a strong enough use case for it.
Neoconf supports rust-tools, because rust-tools delegated configuring the LSP client to nvim-lspconfig (which comes with a bunch of issues; hence my rewrite).

I'm in the process of adding **experimental** support for `.vscode/settings.json` (which has a stronger use case, because it's a lot more widely used, and I can reuse some existing code). But I don't have the capacity to support and maintain everybody's edge case.

If you want to adapt rustaceanvim to support neoconf, you could probably override `vim.g.rustaceanvim.server.settings`, which can be a function that since the latest release can look something like this:

```lua
vim.g.rustaceanvim = {
  server = {
    settings = function(project_root_dir, default_settings)
         local results = vim.fn.glob(project_root .. "/.neoconf.json", true, true)
         if vim.tbl_isempty(results) then
           return default_settings
         end
         -- load neoconf json, decode with vim.json.decode, and merge
         -- or override the settings
    end,
  },
}
```

utkarshgupta137:
> I don't use neoconf myself, and as I have elaborated in [mrcjkb/rustaceanvim#53](https://github.com/mrcjkb/rustaceanvim/issues/53), there's not a strong enough use case for it.

I disagree, but I respect your opinion. My question was for the LazyVim maintainers, who've obviously worked a lot to integrate LazyVim & neoconf/neodev nicely (both also made by folke).

> Neoconf supports rust-tools, because rust-tools delegated configuring the LSP client to nvim-lspconfig (which comes with a bunch of issues; hence my rewrite).

I'm sure you've answered this before, but could you share some examples?

> If you want to adapt rustaceanvim to support neoconf, you could probably use `vim.g.rustaceanvim.server.settings`, which can be a function that since the latest release can look something like this:
> 
> ```lua
> vim.g.rustaceanvim = {
>   server = {
>     settings = function(project_root_dir, default_settings)
>          local results = vim.fn.glob(project_root .. "/.neoconf.json", true, true)
>          if vim.tbl_isempty(results) then
>            return default_settings
>          end
>          -- load neoconf json, decode with vim.json.decode, and merge
>          -- or override the settings
>     end,
>   },
> }
> ```

Yeah, but I still don't get automatic config merging using this. I guess I could do anything in this function & add merging myself, but it is still a hassle.
Also, there won't be any live-reloading or key bindings to edit files or view merged configs. Adding all those niceties will require re-inventing the wheel.

> I'm in the process of adding **experimental** support for `.vscode/settings.json` (which has a stronger use case, because it's a lot more widely used, and I can reuse some existing code). But I don't have the capacity to support and maintain everybody's edge case.

Hopefully, you'll consider support for both global + local support, otherwise it won't be useful for me. Since neoconf supports it too, I wouldn't mind migrating all my projects (& other LSPs) to it.

mrcjkb:
> I disagree, but I respect your opinion. 

Respectfully, it's not a subjective opinion.
It's easy to disagree when you don't know the code base and you don't have to maintain it yourself.
If LazyVim integrates with neoconf, then there's a use case for adding it to LazyVim, but not rustaceanvim.
It is not feasible for me to explicitly support every feature of every neovim distribution out there.
But I can and have provided an API.

> I'm sure you've answered this before, but could you share some examples?

For the general case, I have [written about it in detail](https://mrcjkb.dev/posts/2023-08-22-setup.html). For rustaceanvim specifically: lspconfig has to be configured eagerly at startup, with a `setup` function. And it sources those settings only once.
rustaceanvim, as a filetype plugin, lazily and dynamically loads its configurations. This means it has no impact on startup time (also for people to don't lazy load it) and it can automatically detect different settings for different workspaces based on what is available (with and without settings files), and share resources, all within the same session. It also works a lot nicer with things like direnv and nix devShells.
A lot of rustaceanvim's features would not be possible (or would be a lot more messy) with the lspconfig approach.

> I guess I could do anything in this function & add merging myself, but it is still a hassle.

I left that snippet as a tip for anyone who wants to implement neoconf support for the LazyVim extension.

> Adding all those niceties will require re-inventing the wheel.

I have no interest in adding those niceties to rustaceanvim.
It should be possible to use the neoconf API within the function I provided. I should note that doing so will break any built-in support for `.vscode/settings.json`. But if neoconf supports it, that shouldn't be a problem.

> Hopefully, you'll consider support for both global + local support

As of today, I don't see a strong enough use case for global support in rustaceanvim. The use case for local `.vscode/settings.json` support is to locally override the check command in large monorepos.

> given that rustacean doesn't use rust-analyzer from mason.nvim

To pick up on this: There's a good reason rustaceanvim doesn't automatically pick up a mason.nvim installation. It will most likely be built with a different toolchain than the one your project uses, often leading to discrepancies and subtle bugs.
It's easy to configure rustaceanvim to use mason.nvim if you really want it, but I generally adhere to the YAGNI principle.

utkarshgupta137:
> > I disagree, but I respect your opinion.
> 
> Respectfully, it's not a subjective opinion. It's easy to disagree when you don't know the code base and you don't have to maintain it yourself. If LazyVim integrates with neoconf, then there's a use case for adding it to LazyVim, but not rustaceanvim. It is not feasible for me to explicitly support every feature of every neovim distribution out there. But I can and have provided an API that distros can use to inject custom behaviour.

I disagree on the usefulness of the features you've removed, not the technical reasons. You're right I don't maintain the code, I'm just the user. But the usefulness of something is a subjective opinion.

> > I'm sure you've answered this before, but could you share some examples?
> 
> For the general case, I have [written about it in detail](https://mrcjkb.dev/posts/2023-08-22-setup.html). For rustaceanvim specifically: lspconfig has to be configured eagerly at startup, with a `setup` function. And it sources those settings only once. rustaceanvim, as a filetype plugin, lazily and dynamically loads its configurations. This means it has no impact on startup time (also for people who don't lazy load it) and it can automatically detect different settings for different workspaces based on what is available (with and without settings files), and share resources, all within the same session. It also works a lot nicer with things like direnv and nix devShells. A lot of rustaceanvim's features would not be possible (or would be a lot more messy) with the lspconfig approach.
> 
> > I guess I could do anything in this function & add merging myself, but it is still a hassle.
> 
> I left that snippet as a tip for anyone who wants to implement neoconf support for the LazyVim extension.
> 
> > Adding all those niceties will require re-inventing the wheel.
> 
> I have no interest in adding those niceties to rustaceanvim. It should be possible to use the neoconf API within the function I provided.

I'm not saying rustaceanvim should have certain features or not. I'm sure there are good reasons & trade offs for your decisions. Hence I'm having the discussion here, rather than in the rustaceanvim repo. My question isn't whether rustaceanvim should have these features, my question is whether LazyVim should use rustaceanvim, given that it removes a lot of niceties users have come to expect.

> > Hopefully, you'll consider support for both global + local support
> 
> As of today, I don't see a strong enough use case for global support in rustaceanvim. The use case for local `.vscode/settings.json` support is to locally override the check command in large monorepos.

Well there are a lot of use cases for this. For example, there are a lot of options which are related to personal preferences like semantic highlighting etc - these should be global. But there are a lot of options that are project specific like using nightly formatter or check commands - these should be local.

> > given that rustacean doesn't use rust-analyzer from mason.nvim
> 
> To pick up on this: There's a good reason rustaceanvim doesn't automatically pick up a mason.nvim installation. It will most likely be built with a different toolchain than the one your project uses, often leading to discrepancies and subtle bugs. It's easy to configure rustaceanvim to use mason.nvim if you really want it, but I generally adhere to the YAGNI principle.

I did not realize the rust-analyzer was toolchain specific. I always assumed it was toolchain agnostic because I never see mentions of rustc version requirements in the change logs. It is true that rustup packages a versioned rust-analyzer distribution. My team personally always stays on the latest version, so maybe that is why we never noticed any bugs. Out of curiosity, are there any instances where the suggested fix for older tool chain versions is to use an older version of rust-analyzer?

That said, I think the discussion has gone off-topic now, but hopefully you understand that my objective is not to beg you for features & expect you to maintain it. My question is directed at LazyVim maintainers.

mrcjkb:
>  For example, there are a lot of options which are related to personal preferences like semantic highlighting etc - these should be global

And they can be declared in the rustaceanvim config - no need to load a global settings file.

> are there any instances where the suggested fix for older tool chain versions is to use an older version of rust-analyzer?

I'm not sure, because I personally use nix for pretty much everything.
I have heard people mention that you'll have to compile twice if you're using different toolchains, regardless of the version (but I haven't tried it out myself).

Anyways, I agree we're drifting off here. Let's end this discussion to avoid spamming people with notifications.

I'm happy to help if there are any issues getting neoconf working with rustaceanvim. I just wanted to provide clarity about my choice not to implement it in the plugin :smile:

mikavilpas:
For anyone else wondering why rust-analyzer isn't starting up, it seems it's now required to be installed separately.

I'm not sure what the best method of doing this is but I just tried this and it seems to have worked https://rust-analyzer.github.io/manual.html#rustup

folke:
Yes, we changed that because rustacenean doesnt always work properly with Mason's rust-analyzer installation.

lua/lazyvim/plugins/extras/lang/rust.lua
```diff
@@ -38,89 +37,64 @@ return {
...
+            -- Add clippy lints for Rust.
+            checkOnSave = {
+              allFeatures = true,
+              command = "clippy",
+              extraArgs = { "--no-deps" },
+            },
+            procMacro = {
+              enable = true,
+              ignored = {
+                ["async-trait"] = { "async_trait" },
```
edevil, line 76: I'm curious about why these 3 macros were singled out?

PR #2211, opened 2023-12-18
feat(lang) extension for alternative `nvim-java` distribution

appelgriebsch:
Might not be on-par with `nvim-jdtls` right now, but is more opinionated and setup is way easier.


abeldekat:
I think an alternative would be great. We might have both in the extras?

appelgriebsch:
> I think an alternative would be great. We might have both in the extras?

Yeah, that's why I gave it a different name here (nvim-java). Eventually we want to rename the original extension as nvim-jdtls..?!?

PR #2217, opened 2023-12-19
feat(toggle-diagnostics): change logic based on `vim.diagnostic.is_di…

dpetka2001:
…sabled`

After discussion in #2215, I thought maybe it would be beneficial if we could change the logic of the `toggle-diagnostics` function based on if the user has disabled diagnostics in his own configuration.

### DIFF
--- a/lua/lazyvim/util/toggle.lua
+++ b/lua/lazyvim/util/toggle.lua
@@ -41,10 +41,10 @@ function M.number()
   end
 end
 
-local enabled = true
+local status = vim.diagnostic.is_disabled()
 function M.diagnostics()
-  enabled = not enabled
-  if enabled then
+  status = not status
+  if not status then
     vim.diagnostic.enable()
     Util.info("Enabled diagnostics", { title = "Diagnostics" })
   else
--- a/lua/lazyvim/util/toggle.lua
+++ b/lua/lazyvim/util/toggle.lua
@@ -41,9 +41,15 @@ function M.number()
   end
 end
 
-local enabled = not vim.diagnostic.is_disabled()
+local enabled = true
 function M.diagnostics()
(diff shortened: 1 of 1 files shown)


lua/lazyvim/util/toggle.lua
```diff
@@ -41,10 +41,10 @@ function M.number()
   end
 end
 
-local enabled = true
+local status = vim.diagnostic.is_disabled()
 function M.diagnostics()
-  enabled = not enabled
-  if enabled then
+  status = not status
+  if not status then
```
jyuan0, line 47: It might be clearer if you just change to:
```lua
local enabled = not vim.diagnostic.is_disabled()
```
and then I think the rest of the code can stay the same?

Super minor either way, please do what you think is best.

dpetka2001: Thank you @jyuan0. 

Always keep thinking it in my mind as status, but no need to make unnecessary changes to existing terminology.

abeldekat:
Nice!

konosubakonoakua:
IMO we don't need to use variable `enabled` to cache the status. 
#2297

```lua
function M.diagnostics()
  if vim.diagnostic.is_disabled() then
    vim.diagnostic.enable()
    Util.info("Enabled diagnostics", { title = "Diagnostics" })
  else
    vim.diagnostic.disable()
    Util.warn("Disabled diagnostics", { title = "Diagnostics" })
  end
end
```

dpetka2001:
We don't, I just tried to make minimal changes to the existing code. Maintainer can decide how to incorporate it. Both are fine.

folke:
Changed it a bit for backward compatibility and merged. 
Thanks!

PR #2235, opened 2023-12-21
feat(lang): add telescope terraform integrations

### LINKED ISSUE #2234: feature: add `cappyzawa/telescope-terraform.nvim` and `ANGkeith/telescope-terraform-doc.nvim` to `lazyvim.plugins.extras.lang.terraform`
### Did you check the docs?

`cappyzawa/telescope-terraform.nvim` provides information about terraform modules, resources, & data through Telescope.

`ANGkeith/telescope-terraform-doc.nvim` provides terraform documentation information through Telescope

Enabling `lazyvim.plugins.extras.lang.terraform` should automatically setup these telescope extensions 

Overriding telescope plugin to manually setup additional extensions.

### Additional context

_No response_

pbnj:
Closes #2234

Add telescope integrations for:

- https://github.com/ANGkeith/telescope-terraform-doc.nvim
- https://github.com/cappyzawa/telescope-terraform.nvim

### DIFF
--- a/lua/lazyvim/plugins/extras/lang/terraform.lua
+++ b/lua/lazyvim/plugins/extras/lang/terraform.lua
@@ -1,3 +1,5 @@
+local Util = require("lazyvim.util")
+
 vim.api.nvim_create_autocmd("FileType", {
   pattern = { "hcl", "terraform" },
   desc = "terraform/hcl commentstring configuration",
@@ -53,6 +55,29 @@ return {
         terraform = { "terraform_fmt" },
         tf = { "terraform_fmt" },
         ["terraform-vars"] = { "terraform_fmt" },
+      },
+    },
+  },
+  {
+    "nvim-telescope/telescope.nvim",
+    dependencies = {
+      {
+        "ANGkeith/telescope-terraform-doc.nvim",
+        event = "VeryLazy",
+        config = function()
+          Util.on_load("telescope.nvim", function()
+            require("telescope").load_extension("terraform_doc")
+          end)
+        end,
+      },
+      {
+        "cappyzawa/telescope-terraform.nvim",
+        event = "VeryLazy",
+        config = function()
+          Util.on_load("telescope.nvim", function()
+            require("telescope").load_extension("terraform")
+          end)
+        end,
       },
     },
   },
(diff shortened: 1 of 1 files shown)


dpetka2001:
Is there a reason you load both these plugins in `VeryLazy`? Since they will load either way with Telescope, I don't believe it is necessary. And less plugins loading at startup.

pbnj-dragon:
@dpetka2001 - thank you for the feedback.

> Is there a reason you load both these plugins in VeryLazy? 

I copied existing patterns in other Extras that setup telescope extensions, like `extras/util/project.lua`.

Nonetheless, I removed `event = "VeryLazy"`.

pbnj:
@dpetka2001 - thank you for the feedback.

> Is there a reason you load both these plugins in VeryLazy? 

I copied existing patterns in other Extras that setup telescope extensions, like `extras/util/project.lua`.

Nonetheless, I removed `event = "VeryLazy"`.

dpetka2001:
Oh I don't use that and hadn't noticed. I think the removal shouldn't affect anything. I checked out your PR locally and removing those events didn't seem to affect anything. Of course, I'm just a user like you and also don't know anything about Terraform, that's why I asked if you noticed any problems. 
With the removal of the events, Telescope pickers still seemed to work fine for these 2 plugins. 
If you do notice that something is amiss, regarding functionality of the plugins, feel free to add them back. 
But from my observations, it didn't seem to cause any wrong to have them lazy-loaded along Telescope.

pbnj:
I agree. I don't think lazy loading Telescope extensions are necessary.

In my previous (lazy.nvim) setup, I loaded the extensions along with Telescope.

pbnj:
I question if they should be `optional = true`.

If so, how do users opt-in or opt-out of these extensions?

dpetka2001:
`optional` is not for users to opt in or out of plugins. To quote from `lazy.nvim`'s docs
> When a spec is tagged optional, it will only be included in the final spec, when the same plugin has been specified at least once somewhere else without optional. This is mainly useful for Neovim distros, to allow setting options on plugins that may/may not be part of the user's plugins

So, for example there is the `extras.dap.core` which provides the core functionality for DAP config in LazyVim. And then in the `extras.lang.*`, you will see that `nvim-dap` is specified with `optional = true`. That essentially means that if you enable the language extra and you don't have the core dap enabled, that part of the dap spec in the language extra won't get added in the user's configuration.
If a user that enables the `extras.lang.terraform` would like to not have these 2 plugins, he would have to explicitly disable them in his own personal configuration.

pbnj:
Makes sense. Thank you for clarifying.

folke:
Thanks!

PR #2238, opened 2023-12-22
feat(util) create gitui extension

appelgriebsch:


sqwxl:
I've found lazygit to be slow and buggy, but still useful in some cases. gitui seems like a worthwhile alternative.

appelgriebsch:
Yeah, also was surprised to see that they have the gitui binary as part of mason registry, which makes installing/updating it a snap...

neuromaancer:
Unfortunately, I just copy and paste the code, it doesn't work for me when I click <leader>gg. It doesn't start properly, there seems to be a window, but it flashes back.

neuromaancer:
Yeah, I am sure.

dpetka2001:
Strange, it works for me.

neuromaancer:
There are two things to verified:

1.did you use edgy.nvim? I think it might affect the window layout.

2. I only installed the gitui but not by external tool for example: homebrew. I don't know if this will cause it?

dpetka2001:
I don't use edgy and I just used the code in this PR to install `gitui` via Mason.

neuromaancer:
Okay, I have figured out why. I need to first uninstall edgy and install `gitui`, lastly reinstall edgy. I think edgy does something for the registered window app, but if it is new installed one. It might flash back.

folke:
Thanks!

PR #2240, opened 2023-12-23
feat(ui): dashboard files searches git files if in git repo

pcattori:
just like `<leader>ff` keymap for telescope


andresatierf:
Since you're changing the find_files, you could also change the other ones, like oldfiles and live_grep

folke:
makes sense. Thanks!

PR #2254, opened 2023-12-26
fix(typescript): don't hardcode values in lua table

dpetka2001:
This started as an attempt from me to change the `shiftwidth` value for `.js` files in order for the `tsserver` to format the files using the new value.

However, that didn't seem to work. I tried to make the change both in `after/ftplugin/javascript.lua` and with an auto command on `FileType/BufReadPost` events. None of them worked, `tsserver` continued to format with the value of `shiftwidth` it would read from `options.lua`. I didn't want to change the top-level value of `shiftwidth` in `options.lua` (as I generally prefer it that way for the rest of my LazyVim usage). The other way would be to manually override the values of `tsserver` format settings in my personal `lspconfig` configuration.

My first attempt (first commit) was to change the `opts` into a function, so that the values wouldn't be evaluated when the module loaded, but instead defer the evaluation until `lspconfig` loaded. 

That was successful at first glance, when I noticed that although the auto commands would work as expected, the `after/ftplugin` files still wouldn't. Changes in settings would only take effect when invoking Neovim with args (thus `LazyFile` would be delayed executed on `User` event), but not when invoking Neovim without args (`LazyFile` would then be aliased to `BufReadPost` according to the [docs](https://github.com/LazyVim/LazyVim/blob/879e29504d43e9f178d967ecc34d482f902e5a91/lua/lazyvim/util/plugin.lua#L71-L73)). 

Further trying to bisect the awkward behavior for `after/ftplugin/javascript.lua` with and without args, I found out that commenting out the `indentSize` in `tsserver` format settings, allowed formatting correctly based on `shiftwidth` no matter where it was set. Trying to figure out this behavior led me to this [comment](https://github.com/stevearc/conform.nvim/issues/102#issuecomment-1837141104) in `conform`'s repo which stated that LSP servers are aware of the `tabsize` and further digging led to these [lines](https://github.com/neovim/neovim/blob/master/runtime/lua/vim/lsp/util.lua#L2083-L2110) in Neovim code, which seems to back up the user's claims in the `conform`'s issue that I had found. 

So, my final attempt (2nd commit) was to simply remove the format settings from Typescript extra and let `tsserver` figure out the correct `tabsize` from `shiftwidth` that the user is setting. [This](https://github.com/LazyVim/LazyVim/commit/eb92903342a7ac12185c62bb63b5caa4c80ce7dc) is the commit where Folke introduced this change, but I could not understand the reason why (from the message of the commit or that it wasn't really linked to any issue that might solve a problem), after the new things that I learned. Of course, I might be missing something.

Huge shoutouts to @abeldekat, who helped me through this endeavour (he was the one who pointed out to me the `opts` should be a function, as due to my poor Lua knowledge I thought that using closures was a mean to defer `require`s, so that parsing error from `lazy.nvim` could be avoided, never thought I could also apply it to defer the evaluation of values until plugin loads and he also was the one that found about uncommenting the setting in the extra would allow `tsserver` to format correctly and the piece of code in Neovim where that was happening).

@folke Sorry for the lengthy description, just wanted to properly explain the reasons behind my proposal for this PR.

### DIFF
--- a/lua/lazyvim/plugins/extras/lang/typescript.lua
+++ b/lua/lazyvim/plugins/extras/lang/typescript.lua
@@ -13,62 +13,59 @@ return {
   -- correctly setup lspconfig
   {
     "neovim/nvim-lspconfig",
-    opts = {
-      -- make sure mason installs the server
-      servers = {
-        ---@type lspconfig.options.tsserver
-        tsserver = {
-          keys = {
-            {
-              "<leader>co",
-              function()
-                vim.lsp.buf.code_action({
-                  apply = true,
-                  context = {
-                    only = { "source.organizeImports.ts" },
-                    diagnostics = {},
-                  },
-                })
-              end,
-              desc = "Organize Imports",
-            },
-            {
-              "<leader>cR",
-              function()
-                vim.lsp.buf.code_action({
-                  apply = true,
-                  context = {
-                    only = { "source.removeUnused.ts" },
-                    diagnostics = {},
-                  },
-                })
-              end,
-              desc = "Remove Unused Imports",
-            },
+    opts = function(_, opts)
+      opts.servers.tsserver = {
+        keys = {
+          {
+            "<leader>co",
+            function()
+              vim.lsp.buf.code_action({
+                apply = true,
+                context = {
+                  only = { "source.organizeImports.ts" },
+                  diagnostics = {},
+                },
+              })
+            end,
+            desc = "Organize Imports",
           },
-          settings = {
-            typescript = {
-              format = {
-                indentSize = vim.o.shiftwidth,
-                convertTabsToSpaces = vim.o.expandtab,
-                tabSize = vim.o.tabstop,
-              },
-            },
-            javascript = {
-              format = {
-                indentSize = vim.o.shiftwidth,
-                convertTabsToSpaces = vim.o.expandtab,
-                tabSize = vim.o.tabstop,
-              },
-            },
-            completions = {
-              completeFunctionCalls = true,
-            },
+          {
+            "<leader>cR",
+            function()
+              vim.lsp.buf.code_action({
+                apply = true,
+                context = {
+                  only = { "source.removeUnused.ts" },
+                  diagnostics = {},
+                },
+              })
+            end,
+            desc = "Remove Unused Imports",
           },
         },
-      },
-    },
+        settings = {
+          typescript = {
+            format = {
+              indentSize = vim.o.shiftwidth,
+              convertTabsToSpaces = vim.o.expandtab,
+              tabSize = vim.o.tabstop,
+            },
+          },
+          javascript = {
+            format = {
+              indentSize = vim.o.shiftwidth,
+              convertTabsToSpaces = vim.o.expandtab,
+              tabSize = vim.o.tabstop,
+            },
+          },
+          completions = {
+            completeFunctionCalls = true,
+          },
+        },
+      }
+    end,
   },
+
   {
     "mfussenegger/nvim-dap",
     optional = true,
--- a/lua/lazyvim/plugins/extras/lang/typescript.lua
+++ b/lua/lazyvim/plugins/extras/lang/typescript.lua
@@ -14,7 +14,10 @@ return {
   {
     "neovim/nvim-lspconfig",
     opts = {
+      -- make sure mason installs the server
       servers = {
+        ---@type lspconfig.options.tsserver
+        ---@diagnostic disable-next-line: missing-fields
         tsserver = {
           keys = {
             {
--- a/lua/lazyvim/plugins/extras/lang/typescript.lua
+++ b/lua/lazyvim/plugins/extras/lang/typescript.lua
@@ -13,57 +13,40 @@ return {
   -- correctly setup lspconfig
   {
     "neovim/nvim-lspconfig",
-    opts = function(_, opts)
-      opts.servers.tsserver = {
-        keys = {
-          {
-            "<leader>co",
-            function()
-              vim.lsp.buf.code_action({
-                apply = true,
-                context = {
-                  only = { "source.organizeImports.ts" },
-                  diagnostics = {},
-                },
-              })
-            end,
-            desc = "Organize Imports",
-          },
-          {
-            "<leader>cR",
-            function()
-              vim.lsp.buf.code_action({
-                apply = true,
-                context = {
-                  only = { "source.removeUnused.ts" },
-                  diagnostics = {},
-                },
-              })
-            end,
-            desc = "Remove Unused Imports",
-          },
-        },
-        settings = {
-          typescript = {
-            format = {
-              indentSize = vim.o.shiftwidth,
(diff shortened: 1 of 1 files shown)


Frederick888:
I also just bumped into this.

> let tsserver figure out the correct tabsize from shiftwidth that the user is setting

I believe it's actually [Neovim that's telling tsserver these settings](https://github.com/neovim/neovim/blob/a7550a20e0c3084eacd2b4ede1e6a94f282c2fb8/runtime/lua/vim/lsp/util.lua#L2105-L2108) according to [LSP specs](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#documentFormattingOptions).

But perhaps due to some [historical reasons](https://github.com/typescript-language-server/typescript-language-server/pull/43/files#diff-2a5096ffdb93efd5ef847c28e268f1e1e824fa2894ecd611c1a1f1164c4f4c63L424-R430), [convertTabsToSpaces and indentSize take precedence over insertSpaces and tabSize respectively](https://github.com/typescript-language-server/typescript-language-server/blob/80ef6149cd64b9859e402adb1ca3e7c410ab52b5/src/features/fileConfigurationManager.ts#L280-L285).

We can still have a default `tabSize` FWIW, since it exists in both [`FormattingOptions`](https://github.com/typescript-language-server/typescript-language-server/blob/80ef6149cd64b9859e402adb1ca3e7c410ab52b5/node_modules/vscode-languageserver-types/lib/umd/main.d.ts#L2314) and [`EditorSettings`](https://github.com/typescript-language-server/typescript-language-server/blob/80ef6149cd64b9859e402adb1ca3e7c410ab52b5/node_modules/typescript/lib/typescript.d.ts#L2783).

dpetka2001:
Yes, it is Neovim that is telling the LSP servers of these settings, as noted both by your first link and the [link](https://github.com/neovim/neovim/blob/master/runtime/lua/vim/lsp/util.lua#L2083-L2110) I provided in my first post. 

The problem from my investigation seemed to derive from the application of `LazyFile` being invoked too late or correctly in time for the settings to take effect, when the user set the `shiftwidth` option in an `after/ftplugin` file. When setting the option with an autocommand, it correctly respected the user's setting when invoking Neovim both with args and without args.

Removing the settings from `tsserver` completely was the only solution I could find that the user's setting of `shiftwidth` in both an autocommand and `after/ftplugin/` files would be respected. If someone has a better solution for this, I would definitely appreciate some input that I might have missed.

Frederick888:
Sorry I overlooked your link (well, it was a long post as you mentioned :P).

> The problem from my investigation seemed to derive from the application of `LazyFile` being invoked too late or correctly in time for the settings to take effect, when the user set the `shiftwidth` option in an `after/ftplugin` file. When setting the option with an autocommand, it correctly respected the user's setting when invoking Neovim both with args and without args.

I think that's because the tsserver `opts` is currently written as a plain table, so it's evaluated as soon as the file is read, before your other stuff like `ftplugin`s.

You can make it (somewhat) work while still keeping all the default indent configs: simply change it to a callback: (see at the bottom of the comment). Now `opts` is only evaluated when lazy.nvim loads lspconfig, and that's usually after `ftplugin`s.

But I really don't think this is how things should work, at least not in Neovim. And if we consider projects with mixed indent styles, even so. These default settings should be interpreted as '_**enforcing**_ these indent rules in the workspace' instead of fallbacks.

```diff
diff --git a/lua/lazyvim/plugins/extras/lang/typescript.lua b/lua/lazyvim/plugins/extras/lang/typescript.lua
index 4f53ae8..6c6483e 100644
--- a/lua/lazyvim/plugins/extras/lang/typescript.lua
+++ b/lua/lazyvim/plugins/extras/lang/typescript.lua
@@ -12,63 +12,65 @@ return {
 
   -- correctly setup lspconfig
   {
     "neovim/nvim-lspconfig",
-    opts = {
-      -- make sure mason installs the server
-      servers = {
-        ---@type lspconfig.options.tsserver
-        tsserver = {
-          keys = {
-            {
-              "<leader>co",
-              function()
-                vim.lsp.buf.code_action({
-                  apply = true,
-                  context = {
-                    only = { "source.organizeImports.ts" },
-                    diagnostics = {},
-                  },
-                })
-              end,
-              desc = "Organize Imports",
-            },
-            {
-              "<leader>cR",
-              function()
-                vim.lsp.buf.code_action({
-                  apply = true,
-                  context = {
-                    only = { "source.removeUnused.ts" },
-                    diagnostics = {},
-                  },
-                })
-              end,
-              desc = "Remove Unused Imports",
-            },
-          },
-          settings = {
-            typescript = {
-              format = {
-                indentSize = vim.o.shiftwidth,
-                convertTabsToSpaces = vim.o.expandtab,
-                tabSize = vim.o.tabstop,
+    opts = function(_, opts)
+      return vim.tbl_deep_extend("force", opts or {}, {
+        -- make sure mason installs the server
+        servers = {
+          ---@type lspconfig.options.tsserver
+          tsserver = {
+            keys = {
+              {
+                "<leader>co",
+                function()
+                  vim.lsp.buf.code_action({
+                    apply = true,
+                    context = {
+                      only = { "source.organizeImports.ts" },
+                      diagnostics = {},
+                    },
+                  })
+                end,
+                desc = "Organize Imports",
               },
-            },
-            javascript = {
-              format = {
-                indentSize = vim.o.shiftwidth,
-                convertTabsToSpaces = vim.o.expandtab,
-                tabSize = vim.o.tabstop,
+              {
+                "<leader>cR",
+                function()
+                  vim.lsp.buf.code_action({
+                    apply = true,
+                    context = {
+                      only = { "source.removeUnused.ts" },
+                      diagnostics = {},
+                    },
+                  })
+                end,
+                desc = "Remove Unused Imports",
               },
             },
-            completions = {
-              completeFunctionCalls = true,
+            settings = {
+              typescript = {
+                format = {
+                  indentSize = vim.o.shiftwidth,
+                  convertTabsToSpaces = vim.o.expandtab,
+                  tabSize = vim.o.tabstop,
+                },
+              },
+              javascript = {
+                format = {
+                  indentSize = vim.o.shiftwidth,
+                  convertTabsToSpaces = vim.o.expandtab,
+                  tabSize = vim.o.tabstop,
+                },
+              },
+              completions = {
+                completeFunctionCalls = true,
+              },
             },
           },
         },
-      },
-    },
+      })
+    end,
   },
   {
     "mfussenegger/nvim-dap",
     optional = true,
```

dpetka2001:
@Frederick888 

That's what I did in my first commit. Changed the `opts` into a closure. The problem however still remained when setting the option in the `after/ftplugin` files. I explained everything in detail about the steps I took in my initial post. According to my understanding the `after/ftplugin` are being sourced after `lazy.nvim` loads the plugins.

folke:
Thank you for the elaborate explanation. I wasn't aware that Neovim always passes formatting options based on the actual settings.

I vaguely remember in the past, there were formatting issues with snippets coming from completions.
That's probably why I added those settings. But to be fair, that might have just been a limitation of tsserver snippets (them not being formatted correctly).

So I'll merge this for now.

Thanks!

PR #2270, opened 2023-12-29
fix(autocmds): avoid checktime when buftype=nofile

Frederick888:
It triggered an 'E11: Invalid in command-line window; <CR> executes,
CTRL-C quits' error when Neovim was in Command Line window and
FocusGained.


folke:
thanks!

PR #2213, opened 2023-12-18
feat(lang) - improve java extension, fixes #1984

appelgriebsch:
I also added a new mason-registry entry to be able to fetch latest java-debug/java-test adapter plugins whose binaries are no longer provided by Microsoft on GH

Issue #1984, opened 2023-11-09
feature: add neotest-java to


rcasia:
### Did you check the docs?

- [X] I have read all the LazyVim docs

### Is your feature request related to a problem? Please describe.

I recently added the [neotest-java](https://github.com/rcasia/neotest-java) adapter. Today I started using LazyVim and I wanted to try running tests in java. When I tried it I got a "No tests found" message.

### Describe the solution you'd like

It would be nice to have neotest-java integrated to `lang.java`. (I found out LazyVim currently runs tests with mason as a workaround because there was no neotest adapter for java).

### Describe alternatives you've considered

What's working for me now is to use the `lang.java` extra and having neotest-java as an user plugin.

```lua
-- ~/.config/nvim/lua/plugins/test.lua
return {
  { "rcasia/neotest-java" },
  {
    "nvim-neotest/neotest",
    opts = { adapters = { "neotest-java" },
    },
  },
}
```

### Additional context

At present, neotest-java does not provide DAP support, but it will be implemented in the future.

PR #2285, opened 2024-01-01
feat(java): allow opting out default keymaps

Frederick888:
These keymaps cannot be disabled in the { keys = { '{lhs}', false } }
fashion. So instead, this patch offers a chance to disable them in
opts.keys.

It makes which-key.nvim an optional dependency too.

### DIFF
--- a/lua/lazyvim/plugins/extras/lang/java.lua
+++ b/lua/lazyvim/plugins/extras/lang/java.lua
@@ -62,7 +62,7 @@ return {
   -- Set up nvim-jdtls to attach to java files.
   {
     "mfussenegger/nvim-jdtls",
-    dependencies = { "folke/which-key.nvim" },
+    dependencies = { { "folke/which-key.nvim", optional = true } },
     ft = java_filetypes,
     opts = function()
       return {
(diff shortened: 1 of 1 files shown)


abeldekat:
Hello @Frederick888,

During the development of plugin [lazyvim-menu-addon](https://github.com/abeldekat/lazyvim-menu-addon) I struggled and failed to incorporate the java extra.

The bottleneck is documented [here](https://github.com/abeldekat/lazyvim-menu-addon#java-extra)

In your PR, an option is added to prevent **LazyVim** to create the hard-coded key-mappings.

```lua
local keys = vim.tbl_extend("force", {
  action = ok,
  test = ok,
}, opts.keys or {})
```

A flag to entirely opt out is very reasonable.

I would suggest a different approach though, using **LazyVim**'s lsp pattern:

[customizing lsp keymaps](https://www.lazyvim.org/plugins/lsp#%EF%B8%8F-customizing-lsp-keymaps)

Outline:

```lua
local M = {}
local M.keys = {}

---@type LazyKeysLspSpec[]|nil
M.Keys._keys = nil

-- Allow the keymaps to be overridden using the init method on the spec for nvim-jdtls...
-- Similar to lazyvim plugin.lsp.keymaps.lua, on_attach
---@return LazyKeysLspSpec[]
function M.keys.get()
  if M.Keys._keys then
    return M._keys
  end
  -- _keys could also hold subcategories: keys, keys_for_test, keys_for_dap
  M.Keys._keys =  {
    -- lots of keys
    {"<leader>cxv", function() require("jdtls").extract_variable_all() end, "Extract Variable" },
    -- more keys
  }
  return M.Keys._keys
end

return M
```

I think it would be better to use `vim.keymap.set` for defining actual keys, instead of `which-key.nvim`.
The user might have disabled `which-key.nvim`. Personally, I have a hotkey to load `which-key.nvim`, only when I need it.

It's not necessary to add which-key as a dependency.
For example, see [noice in ui.lua](https://github.com/LazyVim/LazyVim/blob/879e29504d43e9f178d967ecc34d482f902e5a91/lua/lazyvim/plugins/ui.lua#L277)

@appelgriebsch, as your [PR #1984](https://github.com/LazyVim/LazyVim/pull/2213) is also contributing to `extras.lang.java`:
What's your opinion?

Frederick888:
@abeldekat I can certainly do that. And I was also thinking maybe instead of categories, I can add a `cond = function(lazyPlugin, autocmdArgs)` or something so that users can have more flexibility. But yup, it'd be nice to have some confirmation before I start :)

abeldekat:
That's great @Frederick888.

However, as I am not using the Java extra, it might be good to also have the feedback from the people that contributed directly.

PR #2297, opened 2024-01-03
fix(toggle): use `vim.diagnostic.is_disabled()`

konosubakonoakua:
Use `vim.diagnostic.is_disabled()` instead of local variable `enable` to tracking diagnostic status.

If keeping using local var `enable`, toggle will **filp** at first time when the user disable diagnostic by default.


dpetka2001:
Duplicate of #2217

folke:
dup, so closing

PR #2321, opened 2024-01-06
docs(README-DE): remove .txt extension

tippfehlr:
Probably some windows editor that added this automatically + known extensions hidden.


folke:
thanks

PR #2332, opened 2024-01-10
fix(rust): avoid resetting cmp group_index to 1

Frederick888:
For example, 'buffer' source by default has group_index = 2 [1], however
if we take all the sources plus crates and run them through
cmp.config.sources() again, since it's a one-dimensional list, all
group_indexes will be reset to 1.

[1] https://github.com/LazyVim/LazyVim/blob/879e29504d43e9f178d967ecc34d482f902e5a91/lua/lazyvim/plugins/coding.lua#L74-L80

### DIFF
--- a/lua/lazyvim/plugins/extras/lang/rust.lua
+++ b/lua/lazyvim/plugins/extras/lang/rust.lua
@@ -16,10 +16,8 @@ return {
     },
     ---@param opts cmp.ConfigSchema
     opts = function(_, opts)
-      local cmp = require("cmp")
-      opts.sources = cmp.config.sources(vim.list_extend(opts.sources, {
-        { name = "crates" },
-      }))
+      opts.sources = opts.sources or {}
+      table.insert(opts.sources, { name = "crates" })
     end,
   },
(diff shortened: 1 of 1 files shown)


lua/lazyvim/plugins/extras/lang/rust.lua
```diff
@@ -17,9 +17,12 @@ return {
...
       local cmp = require("cmp")
-      opts.sources = cmp.config.sources(vim.list_extend(opts.sources, {
-        { name = "crates" },
-      }))
+      opts.sources = vim.list_extend(
+        opts.sources or {},
+        cmp.config.sources({
+          { name = "crates" },
+        })
+      )
```
folke, line 25: This should actually just be:

```lua
      table.insert(opts.sources, { name = "crates" })
```

That will automatically set grouop_index to 1

Frederick888: Updated. Thanks for reviewing :)

folke:
thanks!

PR #2336, opened 2024-01-10
feat(terraform): ensure tflint is installed

pbnj:
Ensures `tflint` (terraform linter) is installed when `lazyvim.plugins.extras.lang.terraform` is enabled


folke:
Thanks!

PR #2353, opened 2024-01-14
Add <leader>fg for finding files using git-files

yzhuang:
For working with git repositories, I found this to be more useful than the current <leader>ff and <leader>fF.
Contributing this shortcut so others can use <leader>fg as well, and I can remove it from my custom configs.

Thanks!

### DIFF
--- a/lua/lazyvim/plugins/editor.lua
+++ b/lua/lazyvim/plugins/editor.lua
@@ -141,6 +141,7 @@ return {
       { "<leader>fc", Util.telescope.config_files(), desc = "Find Config File" },
       { "<leader>ff", Util.telescope("files"), desc = "Find Files (root dir)" },
       { "<leader>fF", Util.telescope("files", { cwd = false }), desc = "Find Files (cwd)" },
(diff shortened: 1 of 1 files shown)


folke:
`<leader>ff` already uses git-files instead of `files` when in a git repo.

dpetka2001:
@folke I'm sorry if it seems that I'm nitpicking and am not entirely sure I'm right, but it seems that `<leader>ff` only uses `git_files` when the `root_dir` is `.git`. If you open a file in a subdirectory that has an LSP attached, the `root_dir` will naturally be that of the subfolder and `<leader>ff` will use `find_files` instead. If there's no LSP attached to the file you open, then `root_dir` will be the same as the root of the repo where `.git` resides and will use `git_files`. So, the behavior doesn't seem to be consistent to what a user would like. I believe having a dedicated key for `:Telescope git_files` wouldn't harm. Not that one couldn't do that in his own personal configuration. Just wanted to point out the different behavior of `lazyvim.util.telescope`.

folke:
@dpetka2001 yeah, you're right. Merged!

PR #2386, opened 2024-01-20
fix: java lombok support

imkratos:
when shell config JDTLS_JVM_ARGS environment ,

auto config agrs


binhtran432k:
Hi @imkratos,

Your title doesn't quite reflect the details of your implementation. If you're simply aiming to use Lombok, check out [this approach](https://github.com/LazyVim/LazyVim/discussions/275#discussioncomment-7457295). It leverages the Lombok shipped with Mason, so you don't need to do anything else.

Issue #2330, opened 2024-01-09
bug: tree-sitter-hypr fails to build


MoetaYuko:
### Did you check docs and existing issues?

- [X] I have read all the LazyVim docs
- [X] I have searched the existing issues of LazyVim
- [X] I have searched the existing issues of plugins related to this issue

### Neovim version (nvim -v)

v0.9.5

### Operating system/version

Arch

### Describe the bug

1. When installing before tree-sitter loads (e.g. auto-install on nvim starts), the build fails due to `:TSUpdate` not being available. This appears to be a lazy.nvim bug since the command trigger isn't effective at this stage.
2. When installing after tree-sitter loads, it throws the following error: 
```
[nvim-treesitter] [0/1] Downloading tree-sitter-hypr...
[nvim-treesitter] [0/1] Creating temporary directory
[nvim-treesitter] [0/1] Extracting tree-sitter-hypr...
nvim-treesitter[hypr]: Failed to execute the following command:
{
  cmd = "mv",
  opts = {
    args = { "-f", "tree-sitter-hypr-tmp/tree-sitter-hypr-master", "tree-sitter-hypr" },
    cwd = "/home/xxx/.local/share/nvim",
    stdio = {
      [2] = <userdata 1>,
      [3] = <userdata 2>
    }
  }
}
mv: cannot stat 'tree-sitter-hypr-tmp/tree-sitter-hypr-master': No such file or directory
```

### Steps To Reproduce

Just normally install the plugin

### Expected Behavior

plugin builds

### Repro

```Lua
-- DO NOT change the paths and don't remove the colorscheme
local root = vim.fn.fnamemodify("./.repro", ":p")

-- set stdpaths to use .repro
for _, name in ipairs({ "config", "data", "state", "cache" }) do
  vim.env[("XDG_%s_HOME"):format(name:upper())] = root .. "/" .. name
end

-- bootstrap lazy
local lazypath = root .. "/plugins/lazy.nvim"
if not vim.loop.fs_stat(lazypath) then
  vim.fn.system({ "git", "clone", "--filter=blob:none", "https://github.com/folke/lazy.nvim.git", lazypath, })
end
vim.opt.runtimepath:prepend(lazypath)

-- install plugins
local plugins = {
  "folke/tokyonight.nvim",
  "folke/LazyVim",
  -- add any other plugins here
  spec = {
    { "LazyVim/LazyVim", import = "lazyvim.plugins" },
[...]

PR #2404, opened 2024-01-22
fix(hyprlang): use lazyvim way to install

binhtran432k:
Hey @folke,

Apologies for the noise in #2330. I've double-checked the `tree-sitter-hyprlang` setup for Neovim users. It appears that the parser declaration is happening after Tree-sitter's initialization [init.lua](https://github.com/luckasRanarison/tree-sitter-hyprlang/blob/master/plugin/init.lua), preventing the Hyprlang parser from being installed during the install step. To address this, I've created this PR to leverage LazyVim's approach for Tree-sitter installation.

Please let me know if you have any questions or concerns.

Thanks!


folke:
Thanks!

PR #2409, opened 2024-01-23
fix(autocmds): apply conceal level change to local buffer options

m0lson84:
# Description

A recent commit added an auto command to set `conceallevel` to 0 (show normally) for JSON files. The issue is that this is applying the change at the window level. So other file types (like markdown) are also affected after a JSON file is opened. The proposed change is to apply the change to the local buffer options. This will preserve the default `conceallevel` set in both the LazyVim and user options.

## Motivation and Context

Restrict changes to `conceallevel` to the desired file types.

# How has this been tested?

Applied this change to my local configuration and opened up both json and markdown files to confirm that the desired conceal level is being applied.


folke:
Good catch. Thanks!

PR #2438, opened 2024-01-26
fix(toggle-diagnostics): correctly set initial state

dpetka2001:
With the previous implementation of `toggle-diagnostics` in 3cbfeab, if the user had disabled diagnostics, the first toggle call (using the keymap) would again disable diagnostics and the next one would enable them. This fixes it.


rubiin (approved):

dpetka2001:
My bad, I don't know how I tested this initially, but it seems to work as it should, so this is no longer needed.

PR #2447, opened 2024-01-28
feat(options): Disable foldtext for nightly

skoch13:
Considering https://github.com/neovim/neovim/pull/20750 was merged there is no need to utilize `foldtext()` helper for nightly


folke:
Thanks!

PR #2449, opened 2024-01-28
feat(http): add http extra lang

pbnj:


folke:
http and json are already added by default.
This extra just installs jq, which can be done manually though mason

PR #2450, opened 2024-01-28
feat(treesitter): add printf

amaanq:
```c
printf("%s\n", "Finally!");
```


folke:
I assume this is only working in c/cpp files? Maybe add it to clangd extra instead?

folke:
Merged!

It doesn't seem to work for me in lua though. With `:InspectTree`, I can see the correct parse

image

folke:
Nevermind. Fixed it in tokyonight :)

PR #2472, opened 2024-01-31
feat: turn off telemetry for lua ls

rubiin:
Telemetry is mostly seen as hostile and most users don't want their data to be collected


rubiin:
Might also use LuaJIT by default as it is 4-5X faster than lua current version. My Ram usage was cut down to below 1Gigs after I switched

madflow:
According to https://github.com/LuaLS/lua-language-server?tab=readme-ov-file#privacy, the telemetry is opt-in, and has been removed in v3.6.5.

PR #2475, opened 2024-01-31
fix(terraform): Add commentstring and proper syntax highlighting for tfvars files

reegnz:


reegnz:
#2526 obsoletes this pull request.

PR #2496, opened 2024-02-04
fix(which-key): add group description for folds

JefeDavis:
noticed, that the group description for folds was `+prefix`. I added a fix to set the description to `+fold`


rubiin (approved):

folke:
Thanks!

briandipalma:
It's also spelling? Not sure if many people use that feature though?

PR #2519, opened 2024-02-07
Use the more sane snippet session logic

keks:
When hitting tab, nvim very often jumps to random places in the code. This has been observed not only with LazyVim, but also AstroNvim and others: https://www.reddit.com/r/neovim/comments/12kurkl/why_does_tab_sometimes_send_me_to_a_random/

Apparently the problem is this: https://github.com/L3MON4D3/LuaSnip/issues/258

Conveniently, the issue provides configuration that prevents this from happening. I have tested it locally and it seems to work well for me. I am not sure if I should have guarded the autocmd with a check whether luasnip is available, but I didn't know how that would work.

### DIFF
--- a/lua/lazyvim/config/autocmds.lua
+++ b/lua/lazyvim/config/autocmds.lua
@@ -105,3 +105,17 @@ vim.api.nvim_create_autocmd({ "BufWritePre" }, {
     vim.fn.mkdir(vim.fn.fnamemodify(file, ":p:h"), "p")
   end,
 })
+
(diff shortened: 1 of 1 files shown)


HermesSantos:
unfortunately ive disabled the lemonade plugin due to this behavior.

rubiin (approved):

folke:
Closing, since luasnip is no longer the default snippet engine in LazyVim

PR #2520, opened 2024-02-08
fix(go): adding opts recursive_run

wellingtonlope:
Adding opts recursive_run to fix the error `no Go files in /path/project`. This issue is caused by a recent change in the `nvim-neotest/neotest-go` project, which now [defaults to non-recursively](https://github.com/nvim-neotest/neotest-go/pull/72)


folke:
thanks

PR #2535, opened 2024-02-10
feat: use outline.nvim instead of symbols-outline.nvim

neuromaancer:
Recently `symbols-outline.nvim` has been achieved, so use `outline.nvim` instead with more features.


rubiin (approved):

folke:
Thanks!

PR #2542, opened 2024-02-12
fix(util.ui.fg): Add `link=false` to show effective definition

### LINKED ISSUE #2540: bug(?): Function lazyvim.util.ui.fg returns nil for linked highlights

### Neovim version (nvim -v)

0.9.5

### Operating system/version

macOS 14.2.1

### Describe the bug

`require('lazyvim.util').ui.fg(hilightgroup)` returns `nil` for highlightgroups that are linked.

I added a question-mark in the title since [the function](https://github.com/LazyVim/LazyVim/blob/a50f92f7550fb6e9f21c0852e6cb190e6fcd50f5/lua/lazyvim/util/ui.lua#L142-L149) doesn't actually try to resolve linked groups. I'm just assuming it should.

### Steps To Reproduce

`:lua print(require('lazyvim.util').ui.fg("Special"))` returns a table, and 
`:lua print(require('lazyvim.util').ui.fg("Special")['fg'])` returns `#2ac3de`.

`:lua print(require('lazyvim.util').ui.fg("YankyPut"))` returns `nil`, but it is a highlight group, and running `:lua print(vim.api.nvim_get_hl(0, { name = "YankyPut" }))` will return a table with `{ ["link"] = IncSearch,}`

### Expected Behavior

I would expect the function to resolve the link, and return the correct value for fg (`#1d202f` in this case).

### Repro

```Lua

-- set stdpaths to use .repro
end

-- bootstrap lazy
end

-- install plugins
local plugins = {
  "folke/tokyonight.nvim",
  "folke/LazyVim",
  -- add any other plugins here
}
  root = root .. "/plugins",
})

```

dpetka2001:
Fixes #2540 by showing the effective definition instead of the linked group name.

### DIFF
--- a/lua/lazyvim/util/ui.lua
+++ b/lua/lazyvim/util/ui.lua
@@ -142,7 +142,8 @@ end
 function M.fg(name)
   ---@type {foreground?:number}?
   ---@diagnostic disable-next-line: deprecated
-  local hl = vim.api.nvim_get_hl and vim.api.nvim_get_hl(0, { name = name }) or vim.api.nvim_get_hl_by_name(name, true)
+  local hl = vim.api.nvim_get_hl and vim.api.nvim_get_hl(0, { name = name, link = false })
+    or vim.api.nvim_get_hl_by_name(name, true)
   ---@diagnostic disable-next-line: undefined-field
   local fg = hl and (hl.fg or hl.foreground)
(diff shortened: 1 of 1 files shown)


rubiin (approved):

folke:
Thanks!

PR #2543, opened 2024-02-13
fix(yanky): properly disable sqlite.lua on Windows

MoetaYuko:
When directly modifying its `enabled` property, the plugin will remain disabled even if required by another plugin.


folke:
Thanks!

PR #2562, opened 2024-02-16
feat(hipatterns): add shorthand hex color support

matadaniel:
Shorthand hex colors will have a matching background color. For example, `#f00` will have a red background.

I'm not very experienced with hipatterns' API. If anything at all can be done in a better way, please let me know.


rubiin (approved):

folke:
Thanks!

genmzy:
This is a bad idea, it will change default behavior, this will affect C/C++ `.h` files by default, such as `#def`(which included in `#define`)

genmzy:
May be this is not a bad idea, but a bad implementation

folke:
Fixed

PR #2581, opened 2024-02-21
feat(java): allow opts for setup_dap_main_class_configs()

Frederick888:
[1] https://github.com/mfussenegger/nvim-jdtls/blob/382b9f625861f47d95876bcfb4c261f3b96077cb/doc/jdtls.txt#L206-L210

---

Personally I needed this to configure different dap_main environment variables.


folke:
thanks

Issue #2593, opened 2024-02-22
bug: autocmd close-with-q does not work properly with man-files


dennisse:
### Did you check docs and existing issues?

- [X] I have read all the LazyVim docs
- [X] I have searched the existing issues of LazyVim
- [X] I have searched the existing issues of plugins related to this issue

### Neovim version (nvim -v)

0.9.1 (OpenBSD), 0.9.5 (macOS), 0.10.0-dev (Ubuntu)

### Operating system/version

OpenBSD 7.4, macOS 14.2.1, Ubuntu 22.04

### Describe the bug

Alright, this one's kinda two-parted.

This is about the [autocmd close_with_q](https://github.com/LazyVim/LazyVim/blob/91126b9896bebcea9a21bce43be4e613e7607164/lua/lazyvim/config/autocmds.lua#L54-L76).

1) The autocmd does not actually trigger on macOS or Ubuntu, when accessing man-pages with `MANPAGER='nvim +Man!'` set in env. Steps to reproduce below. This happens because man(1) on macOS/Ubuntu launches the MANPAGER in embedded mode, so the man-file is never passed as an argument to nvim, and [LazyVim lazyloads default autocmds if no arguments are provided](https://github.com/LazyVim/LazyVim/blob/91126b9896bebcea9a21bce43be4e613e7607164/lua/lazyvim/config/init.lua#L166). The autocmd is therefore set after the filetype has been determined, and it does not trigger. Pressing `q` now will work as beautifully,  since [this is default behaviour in neovim](https://github.com/neovim/neovim/blob/master/runtime/ftplugin/man.vim#L28).

2) After triggering the autocmd (it can be triggered by setting the filetype again - `:set ft=man`), pressing `q` produces the error `E444: Cannot close last window`.

The autocmd does trigger on OpenBSD, where man(1) passes `[man://man(1)]` as an argument when accessing man(1) (`man man`) with `MANPAGER='nvim +Man!'`. However, pressing `q` produces the same `E444`-error.

### Steps To Reproduce

(using repro.lua below - beware, I had to change root, even though the instructions said not to)
Prove autocmd does not trigger on macOS:
```
$ export MANPAGER='nvim +Man!'
$ man man
```
Now you should be inside nvim. `:verbose nmap q` should produce the following output:
```
n  q           *@:lclose<CR><C-W>q
        Last set from /opt/homebrew/Cellar/neovim/0.9.5/share/nvim/runtime/ftplugin/man.vim line 29
```
[...]

PR #2594, opened 2024-02-22
nvim already binds q to close for man-files

dennisse:
Proposed fix for #2593

Also, close-with-q is default for man-files in Neovim [1]. No need to fix this with a custom binding.

1: <https://github.com/neovim/neovim/blob/master/runtime/ftplugin/man.vim#L28>

### DIFF
--- a/lua/lazyvim/config/autocmds.lua
+++ b/lua/lazyvim/config/autocmds.lua
@@ -57,7 +57,6 @@ vim.api.nvim_create_autocmd("FileType", {
     "PlenaryTestPopup",
     "help",
     "lspinfo",
-    "man",
     "notify",
     "qf",
     "query",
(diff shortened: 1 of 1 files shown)


dpetka2001:
This has the side effect of not making the `man` buffer unlisted. So you have to additionally close the buffer as well after you press `q`, which is kind of annoying. Could I suggest to add the following to your PR
```lua
if vim.fn.has("mac") == 0 then
    group = "lazyvim_close_with_q",
    pattern = { "man" },
    callback = function(event)
      vim.bo[event.buf].buflisted = false
      vim.keymap.set("n", "q", "<cmd>close<cr>", { buffer = event.buf, silent = true })
    end,
  })
end
```
This way it will retain its current behavior on non-Mac OS.

dpetka2001:
Then how about just adding an additional autocmd that makes `man` buffers only unlisted and let the closing to the default Neovim behavior? So, something like
```lua
  group = augroup("man_unlisted"),
  pattern = { "man" },
  callback = function(event)
    vim.bo[event.buf].buflisted = false
  end,
})
```
Additionally to your removal of `man` pattern from `lazyvim_close_with_q` group?

folke:
Thanks!

PR #2597, opened 2024-02-23
Update dot.lua

StitiFatah:
automatic file detection for the hyprland config file


folke:
I just merged #2623 which also covered this.

PR #2616, opened 2024-02-27
fix(conform): make `lsp_fallback` option user configurable

dpetka2001:
Inspired by #2615, I thought this might be a good change hopefully.

Not 100% sure if this breaks anything else (though I tested it locally and seemed to work correctly), since I'm not that familiar with LazyVim formatting in depth, thus starting as a draft.


rubiin (approved):

folke:
Thanks!

PR #2623, opened 2024-02-28
fix(dot): remove hyprlang plugin

luckasRanarison:
hyprlang grammar is now included in nvim-treesitter and the plugin itself has been removed from the tree-sitter-hyprlang repo, so I've updated the file accordingly.


folke:
Thanks!

PR #2640, opened 2024-03-02
fix(dap): debugger config via launch.json for c/c++

bob3000:
# debugger config via launch.json for c/c++

nvim-dap allows to configure a debugger on the project level (see `:help dap-launch.json`). Currently this feature is not working because of a missing mapping between debug adapter and debug configuration name.

## the issue

Within `.vscode/launch.json` by default `configurations[].type` describes both, the name of the debug adapter and the debug configuration. Currently LazyVim configures a debug adapter `codellvm` for two configurations called `c` and `cpp`. The adapter name and the configuration are not the same.

## the fix

The adapter's name has to be mapped to the configuration names by calling the function `require("dap.ext.vscode").load_launchjs` with a table as the second argument that assigns the configuration names properly to the adapter name.

## how to test this

- create file named `main.c` and put in this contents
```c
#include <stdio.h>
int main() {
  printf("hello world!\n");
  return 0;
}
```
- compile with: `cc -g -O0 main.c`
- create a file `.vscode/launch.json` next to `main.c` and put in this contents:
```json
{
  "configurations": [
    {
      "type": "codelldb",
      "request": "launch",
      "name": "LaunchLLDB",
      "program": "${workspaceFolder}/${input:myInput}"
    }
  ],
  "inputs": [
    {
      "id": "myInput",
      "type": "pickString",
      "description": "Your input",
      "options": [
        "./a.out"
      ],
      "default": "./a.out"
    }
  ]
}
```
- open `main.c` in LazyVim, set a break point at line three (:DapToggleBreakpoint) and launch the debugger (:DapContinue)

With the added line from the PR there should be a debug configuration available called `LaunchLLDB`. Try without the added line and it won't show up.

## temporary workaround

Add this to your custom configuration as a temporary workaround:

```lua
  {
    "mfussenegger/nvim-dap",
    opts = function ()
      require("dap.ext.vscode").load_launchjs(nil, { codelldb = {'c', 'cpp'} })
    end
  },

```

### DIFF
--- a/lua/lazyvim/plugins/extras/lang/clangd.lua
+++ b/lua/lazyvim/plugins/extras/lang/clangd.lua
@@ -149,6 +149,7 @@ return {
           },
         }
       end
+      require("dap.ext.vscode").load_launchjs(nil, { codelldb = {'c', 'cpp'} })
     end,
   },
 }


folke:
I enabled this at some point in LazyVim, but unfortunately launch.json is json5, which supports comments.

However, the dap implementation does not, so a lot of users got errors about this.

I had to revert the change because of this.

Feel free to add this to your personal config.

bob3000:
Hi Folke,
fair point. However if that's the only issue there is actually a cure for that. From the nvim-dap help:
```
Unlike VS Code, nvim-dap supports standard JSON. Trailing commas on the
last item of a list are an error.

If you install a 3rd-party json5 parser you can override the json decode
function to support json5 features like trailing comma.
>
    require('dap.ext.vscode').json_decode = require'json5'.parse
<
(One json5 parser implementation is https://github.com/Joakker/lua-json5)
```
I just wanted to mention it because IMHO a per project debugger config is quite vital if you're used to work with a debugger. But certainly it's your call 🙏🏻

PR #2642, opened 2024-03-03
Add description to "Y" in Neo-Tree

Aman9das:
It copies the path of the file/directory

I was confused by the `"Y" : <function>` tooltip in neo-tree


folke:
Thanks!

PR #2660, opened 2024-03-07
fix(lang): Change keymaps to `vim.keymap.set` instead of `which-key`

dpetka2001:
Based on this [comment](https://github.com/LazyVim/LazyVim/pull/2198#pullrequestreview-1873658780).


folke:
Thanks!

PR #2676, opened 2024-03-08
Make Haskell-snippets lazy load for performance

uthmanmoh:
The `haskell-snippets.nvim` plugin wasn't being lazy loaded and was increasing my neovim startup time by 100ms.

- Added `ft = { <haskell_files> }` so it lazy loads only when entering a Haskell file.
- Also formatted the file

### DIFF
--- a/lua/lazyvim/plugins/extras/lang/haskell.lua
+++ b/lua/lazyvim/plugins/extras/lang/haskell.lua
@@ -13,14 +13,14 @@ return {
   {
     "mrcjkb/haskell-tools.nvim",
     version = "^3",
-    ft = { 'haskell', 'lhaskell', 'cabal', 'cabalproject' },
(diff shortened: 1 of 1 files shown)


lua/lazyvim/plugins/extras/lang/haskell.lua
```diff
@@ -62,23 +62,24 @@ return {
 
   {
     "mrcjkb/haskell-snippets.nvim",
-    dependencies = { "L3MON4D3/LuaSnip", },
+    dependencies = { "L3MON4D3/LuaSnip" },
+    ft = { "haskell", "lhaskell", "cabal", "cabalproject" },
```
uthmanmoh, line 66: This is the main change, adding the `ft` to the plugin

folke:
Thanks!

PR #2526, opened 2024-02-08
fix(extras): Drop unnecessary commentstring config for terraform

reegnz:
The commentstring for terraform and hcl files is now handled by
nvim-ts-context-commentstring:
https://github.com/JoosepAlviste/nvim-ts-context-commentstring/pull/94

PR #2680, opened 2024-03-08
fix(terraform-extra): Drop unnecessary commentstring config for terraform

NeckBeardPrince:
The commentstring for terraform and HCL files is now handled by nvim-ts-context-commentstring:
https://github.com/JoosepAlviste/nvim-ts-context-commentstring/pull/94

This PR replaces https://github.com/LazyVim/LazyVim/pull/2526.
Credit to https://github.com/reegnz

### DIFF
--- a/lua/lazyvim/plugins/extras/lang/terraform.lua
+++ b/lua/lazyvim/plugins/extras/lang/terraform.lua
@@ -52,4 +52,25 @@ return {
       },
     },
   },
+  {
+    "nvim-telescope/telescope.nvim",
+    dependencies = {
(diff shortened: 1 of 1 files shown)


folke:
you're also removing the telescoep integration which is not what you want to do i guess

NeckBeardPrince:
> you're also removing the telescoep integration which is not what you want to do i guess

Fixed. This is why I would make a horrible developer. I should just stick to being a YAML engineer and Terraform.

folke:
all good now, thanks!

PR #2687, opened 2024-03-09
feat(lualine): add more hl options to pretty_path

pkazmier:
Adds two additional options to pretty_path: `filename_hl` and `dirpath_hl`.

This allows users to customize the highlight group of both the directory component of the path name and the filename independently. modified_hl is still used when the buffer has been modified. 

This commit does not change LazyVim's default formatting of the lualine path name segment. 

Lazy's default:
image

With the new options:
image

When the buffer has been modified:
image

But, now users can achieve the above by passing the options to `pretty_path`:
```lua
local util = require("lazyvim.util")
return {
  {
    "nvim-lualine/lualine.nvim",
    opts = function(_, opts)
      opts.sections.lualine_c[4] =
        { util.lualine.pretty_path({ filename_hl = "Bold", modified_hl = "MatchParen", dirpath_hl = "Conceal" }) }
      return opts
    end,
  },
}
```

Thanks to @dpetka2001 for this `format` function when I inquired about how to achieve this in discussion  #2605.

### DIFF
--- a/lua/lazyvim/util/lualine.lua
+++ b/lua/lazyvim/util/lualine.lua
@@ -56,17 +56,36 @@ function M.format(component, text, hl_group)
   local lualine_hl_group = component.hl_cache[hl_group]
   if not lualine_hl_group then
     local utils = require("lualine.utils.utils")
-    lualine_hl_group = component:create_hl({ fg = utils.extract_highlight_colors(hl_group, "fg") }, "LV_" .. hl_group)
+    local mygui = function()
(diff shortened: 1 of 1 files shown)


folke:
Like it. Thanks!

Just made some small changes:
 - refactored the code a bit
 - renamed `dirpath_hl` to `directory_hl`
 - changed the defaults to:

```lua
{
    relative = "cwd",
    modified_hl = "MatchParen",
    directory_hl = "",
    filename_hl = "Bold",
  }
```

PR #2693, opened 2024-03-10
Description for `g.` keybinding in mini.files

dusty-phillips:
If you press `g` in a mini.files menu, the `.` key shows up but doesn't have a description. This adds a desc attribute to the binding so the menu shows a correct description.


dusty-phillips:
Note: This is also captured in #2695, but I'm keeping this one open as the other may be considered too invasive.

PR #2713, opened 2024-03-11
Add default formatters for yaml

AndesKrrrrrrrrrrr:
Yaml formatting has been a pain for some time now. I finally figured out that the yaml LSP formatter doesn't respect my prettier config. Since the sub-list searches for installed formatters, I think these are sane (non-invasive) defaults.


AndesKrrrrrrrrrrr:
Actually, I ended up setting the default formatter to Prettier

```
["*"] = { { "prettier" } }
```

PR #2720, opened 2024-03-12
fix(lang/rust): don't overwrite `vim.g.rustaceanvim` if it is defined

mrcjkb:
Hey :wave: 

A way to override rustaceanvim's settings per project is by using `:h exrc`.
This doesn't appear to work with LazyVim (see https://github.com/mrcjkb/rustaceanvim/issues/283), likely because it overwrites `vim.g.rustaceanvim` the settings in the `config` function.

This should fix that.


folke:
Makes sense. Thanks!

PR #2728, opened 2024-03-13
feat(git): add <leader>gf for lazygit commit history on current file

uthmanmoh:
This adds the `<leader>gf` (for Git File) keymap that lets you view commit history for the current file.

Lazygit supports this through their CLI (see [doc](https://github.com/jesseduffield/lazygit/blob/fd18db6ba285eb280aee1cb8c381295df6556cad/docs/Searching.md?plain=1#L20)) with a `-f <filepath>` flag which must be a path to the file from the git dir.

Looks like this:
image

Open to suggestions on different keymap or implementation

### DIFF
--- a/lua/lazyvim/config/keymaps.lua
+++ b/lua/lazyvim/config/keymaps.lua
@@ -129,10 +129,9 @@ map("n", "<leader>ub", function() Util.toggle("background", false, {"light", "da
 map("n", "<leader>gg", function() Util.terminal({ "lazygit" }, { cwd = Util.root(), esc_esc = false, ctrl_hjkl = false }) end, { desc = "Lazygit (root dir)" })
 map("n", "<leader>gG", function() Util.terminal({ "lazygit" }, {esc_esc = false, ctrl_hjkl = false}) end, { desc = "Lazygit (cwd)" })
 
--- Function to get the current file path from the root directory
-local current_file_path_from_root_dir = function()
+local current_file_path_from_git_root = function()
   local path = vim.fn.expand("%:p")
-  local root = Util.root()
+  local root = vim.fn.system("git -C " .. path .. " rev-parse --show-toplevel"):gsub("\n$", "")
   if not path:find(root, 1, true) then
     return path
   end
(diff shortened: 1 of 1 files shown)


lua/lazyvim/config/keymaps.lua
```diff
@@ -129,6 +129,20 @@ map("n", "<leader>ub", function() Util.toggle("background", false, {"light", "da
...
 
+-- Function to get the current file path from the root directory
+local current_file_path_from_root_dir = function()
+  local path = vim.fn.expand("%:p")
+  local root = Util.root()
+  if not path:find(root, 1, true) then
+    return path
+  end
+  return path:sub(#root + 2)
+end
```
uthmanmoh, line 140: This may not get the proper path if the lsp gives a different root than where the git repo is.

Is there a better way to do this (i.e. get the path to the file from the git root)?

lua/lazyvim/config/keymaps.lua
```diff
@@ -129,6 +129,19 @@ map("n", "<leader>ub", function() Util.toggle("background", false, {"light", "da
 map("n", "<leader>gg", function() Util.terminal({ "lazygit" }, { cwd = Util.root(), esc_esc = false, ctrl_hjkl = false }) end, { desc = "Lazygit (root dir)" })
 map("n", "<leader>gG", function() Util.terminal({ "lazygit" }, {esc_esc = false, ctrl_hjkl = false}) end, { desc = "Lazygit (cwd)" })
 
+local current_file_path_from_git_root = function()
```
uthmanmoh, line 132: Not sure if there's a better way to do this

folke:
Refactored it a bit and merged. Thanks!

briandipalma:
Hmm, would there be any interest in a diffview extra or adding it to the core set of plugins? IMO it's nicer to have diffview as it integrates with Neovim and can show LSP results and open the files in your current Neovim instance. For example this view I think is more useful than the view above and I can then open the files I'm interested in when I open a full repo history search:

image

skoch13:
@briandipalma totally! I use diffview for file history operation myself as it is closely tied to nvim with lsp's and etc as you mentioned

PR #2737, opened 2024-03-15
fix: format injected langs does not wait 300ms

Aman9das:
Some formatters like styler take a while to format (2s). They fail to format when run on injected code. This should help fix that :).


folke:
thanks!

Issue #2746, opened 2024-03-16
bug: LSP for Rust `extras.lang.rust` does not work even with LazyVim starter on Windows 11 Powershell


gabrielkryss:
### Did you check docs and existing issues?

- [X] I have read all the LazyVim docs
- [X] I have searched the existing issues of LazyVim
- [X] I have searched the existing issues of plugins related to this issue

### Neovim version (nvim -v)

v0.10.0-dev-2597+gd326e0486

### Operating system/version

Windows 11

### Describe the bug

My Lsp for rust (rust-analyzer) wasn't working on the LazyVim starter. I made sure that most of my `:LazyHealth` was relatively green with the right compilers. My Neovin nightly version (installed using scoop) but it is >=0.10. It seems like from the error logs below there is something to do with `man.lua` for the Shift+K behavior. Another observation is that the rust-analyzer may not support the tool chain  'stable-x86_64-pc-windows-msvc' from the LSP Logs shown below. 

Any help on this would be appreciated, the LSP for `extras.lang.rust` used to work fine but not sure what happened that it is not working now. I even tried with a new sLazyVim starter to make sure nothing else was the case, and I tried looking around and similar issues with the rust-analyzer community and cannot find a similar issue, but if anyone does let me know. I also tried with Golang and the LSP and inlay hints work fine.

### Steps To Reproduce

1. clone LazyVim starter
2. go to extras and enable lang.rust
3. open a rust project
4. Warning about rust analyzer exiting
![image](https://github.com/LazyVim/LazyVim/assets/20222391/1e71e96d-f602-4433-97cd-2e9e45e7fe10)
5. Tried doing Shift+K on println!() and see this error
![image](https://github.com/LazyVim/LazyVim/assets/20222391/486a0f81-0439-4790-8499-15896cba4115)
6. Tried <leader>uh to toggle inlay hints and see these errors
![image](https://github.com/LazyVim/LazyVim/assets/20222391/1e9ba893-c81b-4fb2-b10e-f9cc86f04d96)
7. `:LspInfo` is not recognized even restarting LSP with`:LspRestart`
![image](https://github.com/LazyVim/LazyVim/assets/20222391/753dc945-ec58-4042-8989-c876be83d6f9)
8. `:LspLog`
```
4-03-15 16:58:58] LSP logging initiated
[...]

PR #2755, opened 2024-03-17
fix(lang/rust): Remove `rust-analyzer` from `nvim-lspconfig`

dpetka2001:
According to the maintainer of `rustaceanvim` (see his comment [here](https://github.com/LazyVim/LazyVim/pull/2198#issuecomment-1999475044)) he says
> To pick up on this: There's a good reason rustaceanvim doesn't automatically pick up a mason.nvim installation. It will most likely be built with a different toolchain than the one your project uses, often leading to discrepancies and subtle bugs.
It's easy to configure rustaceanvim to use mason.nvim if you really want it, but I generally adhere to the YAGNI principle.

I tried locally and the removal of `rust-analyzer` from `nvim-lspconfig` doesn't seem to have any effect on how `rustaceanvim` behaves.

I propose to remove all instances of `rust-analyzer` from `nvim-lspconfig` to avoid any possible issues from users that don't have `rust-analyzer` installed in their toolchain (in this case it would pick up Mason's $PATH I believe), since they will think that since `rust-analyzer` is installed by Mason, there shouldn't be a problem and report issues as bugs. See for example #2746.

PS: Would appreciate further input from users, who actually use Rust if the proposed change breaks anything I'm not aware of.

### DIFF
--- a/lua/lazyvim/plugins/extras/lang/rust.lua
+++ b/lua/lazyvim/plugins/extras/lang/rust.lua
@@ -90,7 +90,6 @@ return {
     "neovim/nvim-lspconfig",
     opts = {
       servers = {
-        rust_analyzer = {},
         taplo = {
           keys = {
             {
@@ -106,11 +105,6 @@ return {
             },
           },
         },
-      },
-      setup = {
-        rust_analyzer = function()
-          return true
-        end,
       },
     },
   },


rubiin (approved):

amaanq:
I agree with this. If you do not consistently update rustc and r-a in mason in tandem, you can and probably will experience issues due to the version mismatch - using r-a provided by your rustup installation is ideal and should be the way to go imo. FWIW I use Rust daily and it's my main language

utkarshgupta137:
I think it depends on the user preference. I like to use the latest version of r-a, but with stable rust. The r-a packaged with stable rust can be quite old & I prefer to use the latest version so that we can contribute in testing & reporting bugs & also get new features earlier. AFAIK, r-a version isn't tied to rust version itself. Since rust is (almost) always backwards-compatible, I would be surprised if a new r-a version doesn't work with old rust. So I disagree that r-a provided by rustup is the best option. So I personally am not switching to rustaceanvim & would prefer we don't merge.
I'm open to be proven wrong though. Do we know how other major editors handle this?

dpetka2001:
@utkarshgupta137 

The point of this PR is based on the transition LazyVim made to use `rustaceanvim`. I've seen several example-issues where adding the rust-analyzer as a component to the Rust toolchain solved the issues users were having, since `rustaceanvim` doesn't actually use the package provided by Mason by default. What you're asking is different, I believe.

Users are having issues believing that the `rust-analyzer` installed by Mason, is actually being used, when that is not the case. That's what this PR actually tries to fix. More experienced users, who know what they're doing, can still install latest `rust-analyzer` and make necessary adjustments.

zjp-CN:
> So I disagree that r-a provided by rustup is the best option. So I personally am not switching to rustaceanvim & would prefer we don't merge.

rustaceanvim is the current active plugin targeting Rust, so it'd be wise to follow its design.

Personally speaking, I have RA component in rustup, and this is much better than that installed by mason:
* for nightly Rust users, regularly updating a nightly toolchain is a good habit, especially when you want more features from nightly RA. But it doesn't mean stable Rust users must install RA component: you can still config mason and nvim-lspconfig. So I think we need a key in opt table to let user choose a toolchain RA or a mason RA is wanted...
* mason has own registry, and seems to download RA from github. I tried to found a way to use proxy for mason, bacuase the network is usually poor for github in my country if without VPN.  But there are multiple rustup mirrors available, so it's faster to download. Github repos are usually small, and it's incremental, so it won't take too long to update plugins, but a RA binary is 


> I'm open to be proven wrong though.

People will forget to undate RA from mason. For myself, the frequency of updating toolchain is more often than updating RA in mason. And it's [not only me](https://github.com/rust-lang/rust-analyzer/issues/16306#issuecomment-1998738127).

folke:
Thanks!

xarthurx:
For those of you who want to use `mason` to manage `rust-analyzer`, you can just do:

```
require("lspconfig").rust_analyzer.setup = function()
	return true
end
```

Then the `lsp` settings will just skip adding another `RA` instances into your buffer.

PR #2758, opened 2024-03-18
feat(neo-tree): add mapping "O" to open with system default

uthmanmoh:
This adds the "O" mapping to Neo-tree to open the file in the system default app. I found this really useful for opening things like directories in the default file explorer, or images/pdfs since the terminal can't render those. There is an open [issue](https://github.com/nvim-neo-tree/neo-tree.nvim/issues/1224) in neo-tree but it's been labelled as wontfix so I think adding this to something like LazyVim is a good idea.

Let me know if you need me to add anything.

### DIFF
--- a/lua/lazyvim/plugins/editor.lua
+++ b/lua/lazyvim/plugins/editor.lua
@@ -69,17 +69,16 @@ return {
             end,
             desc = "copy path to clipboard",
           },
-          ["O"] = {
-            command = function(state)
-              if vim.ui.open then
-                local filepath = state.tree:get_node().path
-                vim.ui.open(filepath)
-              else
-                require("lazyvim.util").error("vim.ui.open is not available")
+          ["O"] = vim.ui.open and {
+            function(state)
+              local path = state.tree:get_node().path
+              local ret = vim.ui.open(path)
+              if ret and ret.code ~= 0 then
+                require("lazyvim.util").error("Failed to open file: " .. ret.stderr .. "\n .. " .. ret.stdout)
               end
             end,
-            desc = "open_with_system_default",
-          },
+            desc = "open with system default application",
+          } or nil,
         },
       },
       default_component_configs = {
--- a/lua/lazyvim/plugins/editor.lua
+++ b/lua/lazyvim/plugins/editor.lua
@@ -72,6 +72,12 @@ return {
           ["O"] = {
             command = function(state)
               local filepath = state.tree:get_node().path
+
+              if vim.ui.open then
+                vim.ui.open(filepath)
+                return
+              end
+
               local uname = vim.loop.os_uname()
               local OS = uname.sysname
               local is_mac = OS == "Darwin"
--- a/lua/lazyvim/plugins/editor.lua
+++ b/lua/lazyvim/plugins/editor.lua
@@ -72,8 +72,23 @@ return {
           ["O"] = {
             command = function(state)
               local filepath = state.tree:get_node().path
-
-              os.execute("open " .. string.format("'%s'", filepath))
+              local uname = vim.loop.os_uname()
+              local OS = uname.sysname
+              local is_mac = OS == "Darwin"
+              local is_linux = OS == "Linux"
+              local is_windows = OS:find("Windows") and true or false
+              local is_wsl = is_linux and uname.release:lower():find("microsoft") and true or false
+              if is_mac then
+                os.execute("open " .. string.format("'%s'", filepath))
+              elseif is_wsl then
+                os.execute("wslview " .. string.format("'%s'", filepath))
+              elseif is_linux then
+                os.execute("xdg-open " .. string.format("'%s'", filepath))
+              elseif is_windows then
+                os.execute("start " .. string.format("'%s'", filepath))
+              else
+                vim.notify("neo-tree: OS not detected", "error")
+              end
             end,
             desc = "open_with_system_defaults",
           },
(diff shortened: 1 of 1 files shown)


folke:
Can you change it to use `vim.ui.open`?

uthmanmoh:
> Can you change it to use `vim.ui.open`?

Wasn't `vim.ui.open` introduced in 0.10?

folke:
I'd rather not have that open code in there and just rely on `vim.ui.open`.

When `vim.ui.open` is not available, either just show a message with `LazyVim.error(...)`, or don't include the keymap.

uthmanmoh:
> When `vim.ui.open` is not available, either just show a message with `LazyVim.error(...)`, or don't include the keymap.

I'm not really sure how to not include the keymap based on a condition. Do you know if that's possible with neo-tree?

For now I just made it throw an error if `vim.ui.open` isn't available, but not having the keymap at all would be better

dpetka2001:
I tried several things to find a solution about a conditional keymap with no avail at the start and then thought of this
```lua
      if vim.ui.open then
        opts.window.mappings["O"] = {
          command = function(state)
            local filepath = state.tree:get_node().path
            vim.ui.open(filepath)
          end,
          desc = "open_with_system_default",
        }
      else
        opts.window.mappings["O"] = "none"
      end
```
and put it in the `config` function of neotree just before `require("neo-tree").setup(opts)`. Otherwise the whole `opts` could be refactored to a function and put this piece of code there, but haven't tried that, so not sure if it will work correctly.

dpetka2001:
The following change with refactored `opts` as a function also seems to work 
```lua
    opts = function(_, opts)
      opts.sources = { "filesystem", "buffers", "git_status", "document_symbols" }
      opts.open_files_do_not_replace_types = { "terminal", "Trouble", "trouble", "qf", "Outline" }
      opts.filesystem = {
        bind_to_cwd = false,
        follow_current_file = { enabled = true },
        use_libuv_file_watcher = true,
      }
      opts.window = {
        mappings = {
          ["<space>"] = "none",
          ["Y"] = {
            function(state)
              local node = state.tree:get_node()
              local path = node:get_id()
              vim.fn.setreg("+", path, "c")
            end,
            desc = "copy path to clipboard",
          },
        },
      }
      opts.default_component_configs = {
        indent = {
          with_expanders = true, -- if nil and file nesting is enabled, will enable expanders
          expander_collapsed = "",
          expander_expanded = "",
          expander_highlight = "NeoTreeExpander",
        },
      }
      if vim.ui.open then
        opts.window.mappings["O"] = {
          command = function(state)
            local filepath = state.tree:get_node().path
            vim.ui.open(filepath)
          end,
          desc = "open_with_system_default",
        }
      else
        opts.window.mappings["O"] = "none"
      end
    end,
```
Personally, I would prefer this over polluting the `config` function.

uthmanmoh:
Is changing all this worth it just to leave the mapping out? We could also just show the error message like it's doing now and that wouldn't be too bad.

I'll wait on @folke for his opinion

dpetka2001:
That depends on if further conditional keymaps might be added in the future and whether the maintainer wants to go forward this way. Personally, I think making `opts` a function is a small cost and provides more flexibility for stuff like that. It's being done in other LazyVim specs as well (lualine, telescope etc). 

Of course we could just throw an error and be done with it and I definitely agree about waiting on Folke's opinion on this. I just tried what I tried because I saw that there was a proposition about setting the key conditionally and that's what I tried to solve. By no means am I saying this is the best solution (it's just one of probably others as well that just works for this case). And maybe neotree does provide some way of setting conditional keys, but from the little that I searched on their repo I wasn't able to find anything relevant.

folke:
You can just do something like the below:

```lua
          ["O"] = vim.ui.open and {
            function(state)
              local path = state.tree:get_node().path
              local ret = vim.ui.open(path)
              if ret and ret.code ~= 0 then
                LazyVim.error("Failed to open file: " .. ret.stderr .. "\n .. " .. ret.stdout)
              end
            end,
            desc = "open with system default application",
          } or nil,
```

**BUT:** I tested with `vim.ui.open` and tbh, the way it's implemented in core makes zero sense to me.

They start the process detached (which is good), but then wait till it finished.

This makes no sense, since Neovim will literally be unresponsive until you close the app you just opened.

uthmanmoh:
> Neovim will literally be unresponsive until you close the app you just opened.

I just tested it on my Mac and that isn't the case for me. I can open the app and continue using neovim at the same time

uthmanmoh:
Based on my testing this works well for me

folke:
Might be just on Linux then, however the wait is in the code, so that really has to go:

https://github.com/neovim/neovim/blob/849d82b80b8584fe7918bffc6480f0ec0fda9b9b/runtime/lua/vim/ui.lua#L155

dpetka2001:
I'm on Linux Mint 21.3 (base Ubuntu 22.04), awesome WM, kitty terminal and the lock up is there when the system app is open. Main Neovim instance is unresponsive until the system app closes indeed.

folke:
lazy also has an `open` function, that I just changed a bit to allow forcing the ssytem app:

https://github.com/folke/lazy.nvim/commit/66466a2594ab0c446193772a68c236c7e4e02ade

Updated the PR to use lazy's implementation.

uthmanmoh:
Oh that looks even better, I like that. Eventually if the code gets fixed in core we can always just change lazy's implementation to start using that

folke:
Merged. Thanks!

PR #2769, opened 2024-03-19
feat(mini.ai): add more objects d,e,g,u,U

lkhphuc:
Proposing to add some more text objects I've been using quite frequently for a while.
These are all examples from the mini.ai document itself.

- `d` for digits, matches consecutive digits
- `e` because it's near `w`, for subword in camelCase and snake_case
- `g` for entire buffer, similar motion to `gg` and `G`. For example, now `dag` instead of `ggdG` or `ggvGd` will delete entire buffer.
- `u` for "Usage", to selection function and method call since `f` and `c` is already taken for function definition.
- `U` similar to `u`, but match without dot in function name


rubiin (approved):

dpetka2001:
Thanks for this. I believe it will be very useful if it gets merged :smile:

folke:
Great, thanks!

PR #2778, opened 2024-03-19
chore: remove outdated setting for flash.nvim

uyha:


dpetka2001:
What do you mean outdated?

uyha-kwz:
`vscode` as a setting doesn’t exist in `flash.nvim` anymore. You can run `LazyHealth` and see a warning regarding that.

dpetka2001:
That never existed in `flash.nvim`. It's purely a LazyVim implementation for the LazyExtra `vscode`. And this key just defines that if LazyVim is run inside of VsCode, then `flash.nvim` will be enabled. See the corresponding code in the [vscode extra](https://github.com/LazyVim/LazyVim/blob/1432f318b6b061d3da510ebd795a3292b10e636b/lua/lazyvim/plugins/extras/vscode.lua#L23-L25).

PR #2783, opened 2024-03-20
feat(rust): add `rustfmt` for Rust code formatting

2giosangmitom:


amaanq:
This doesn't make sense - rust-analyzer provides formatting already. This would *only* make sense if you disable the LSP, but I would say that this then belongs in your own config if you're doing that.

2giosangmitom:
@amaanq You're absolutely right, `rust-analyzer` directly utilizes `rustfmt` for code formatting functionalities. In essence, `rust-analyzer` LSP provides a convenient way to interact with the underlying `rustfmt` for automatic code formatting. There's no need to install or run `rustfmt` separately. My autoformatting isn't working because I haven't downloaded `rustfmt` and `rust-analyzer` using `rustup`, it's my mistake.

PR #2787, opened 2024-03-20
feat(python): support basedpyright as another option

leaxoy:
[basedpyright](https://github.com/DetachHead/basedpyright) has more capabilities than [pyright](https://github.com/microsoft/pyright), like: inlay hints and semantic highlights, so support it.

### DIFF
--- a/lua/lazyvim/plugins/extras/lang/based-python.lua
+++ b/lua/lazyvim/plugins/extras/lang/based-python.lua
@@ -1,0 +1,95 @@
+return {
+  {
+    "nvim-treesitter/nvim-treesitter",
+    opts = function(_, opts)
+      if type(opts.ensure_installed) == "table" then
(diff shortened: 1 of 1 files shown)


folke:
I added a better way to customize extras.

For the python extra, you can now add the below to your `options.lua`:

```lua
vim.g.lazyvim_python_lsp = "basedpyright"
```

The new docs also show the options an extra supports. They will be generated in a bit.

konosubakonoakua:
> I added a better way to customize extras.
> 
> For the python extra, you can now add the below to your `options.lua`:
> 
> ```lua
> vim.g.lazyvim_python_lsp = "basedpyright"
> ```
> 
> The new docs also show the options an extra supports. They will be generated in a bit.

maybe not using fallback lsp is better
```lua
        pyright = {
          enabled = vim.g.lazyvim_python_lsp == "pyright",
        },
        basedpyright = {
          enabled = vim.g.lazyvim_python_lsp == "basedpyright",
        },
```

PR #2789, opened 2024-03-20
fix(util.telescope): make `show_untracked` configurable

### LINKED ISSUE #2788: bug: lazyvim.util.telescope always sets show_untracked to true in a git repo, preveting recurse_submodules from working with it

### Neovim version (nvim -v)

NVIM v0.10.0-dev-2507+g3df1211eb

### Operating system/version

openSUSE Leap 15.5

### Describe the bug

LazyVim's Util.telescope always sets 'show_untracked' to true when executing git_files, passing 'recurse_submodules' is therefore not possible, as both cannot be set at the same time.

If `recurse_submodules` is passed explicitly, `show_untracked` should not be forced. Otherwise passing it explicitly should allow one to disable show_untracked without it being forcibly enabled.

Offending line [in `lazy/LazyVim/lua/lazyvim/util/telescope.lua` ](https://github.com/LazyVim/LazyVim/blob/main/lua/lazyvim/util/telescope.lua#L32).

### Steps To Reproduce

1. Add a keybinding for telescope to your `lua/plugins/telescope.lua` file like this:
```
local Util = require("lazyvim.util")

return {
        keys = {
                { "<leader>fa", Util.telescope("files", { recurse_submodules=true, show_untracked=false }), desc = "Find All Files (root dir)" },
        },
}
```
2. In normal mode press <leader>fa, while in a git repository and observe the error: `[telescope.builtin.git_files]: Git does not support both --others and --recurse-submodules` ; even though show_tracked=false is explicitly passed, it's overwritten by the plugin.

### Expected Behavior

I would expect this to work and execute `Telescope git_files recurse_submodules=True` .

### Repro

_No response_

dpetka2001:
Fixes #2788

### DIFF
--- a/lua/lazyvim/util/telescope.lua
+++ b/lua/lazyvim/util/telescope.lua
@@ -29,7 +29,9 @@ function M.telescope(builtin, opts)
         and not vim.loop.fs_stat((opts.cwd or vim.loop.cwd()) .. "/.ignore")
         and not vim.loop.fs_stat((opts.cwd or vim.loop.cwd()) .. "/.rgignore")
       then
-        opts.show_untracked = true
+        if opts.show_untracked == nil then
+          opts.show_untracked = true
+        end
         builtin = "git_files"
       else
         builtin = "find_files"


folke:
Thanks again! :)

PR #2795, opened 2024-03-21
feat(git): unnested hunk keymaps, add toggle &find

lkhphuc:
This PR is a long shot since it introduces breaking changes to gitsigns keymaps.
- Unpack the hunk-related keymap from `<leader>ghx` to just `<leader>gx`. These are most commonly used keymaps for gitsigns and I find it to be unnecessarily long, especially when there is (almost) no conflicting shorter keymap.
- New keymaps for "undo all staged hunks".
- Move the Telescope's git functionality inside `<leader>gf` for "git finds ..."
- Add new gitsigns' UI-related toggle inside `<leader>gt` for "git toggle  ..."
- Current file git history with lazygit from `<leader>gf` to `<leader>gh`

### DIFF
--- a/lua/lazyvim/config/keymaps.lua
+++ b/lua/lazyvim/config/keymaps.lua
@@ -129,7 +129,7 @@ map("n", "<leader>ub", function() Util.toggle("background", false, {"light", "da
 map("n", "<leader>gg", function() Util.terminal({ "lazygit" }, { cwd = Util.root.git(), esc_esc = false, ctrl_hjkl = false }) end, { desc = "Lazygit (root dir)" })
 map("n", "<leader>gG", function() Util.terminal({ "lazygit" }, {esc_esc = false, ctrl_hjkl = false}) end, { desc = "Lazygit (cwd)" })
 
-map("n", "<leader>gf", function()
+map("n", "<leader>gh", function()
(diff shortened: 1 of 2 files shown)


dpetka2001:
I'm not in favor of the gitsigns hunk change. I use for example `diffview.nvim` and have set the corresponding keymaps in the `<leader>g` git namespace and use `gd/gD` in the git namespace already. Also the `<leader>gf` change into a different namespace for `git find` doesn't take into consideration the latest addition of `Lazygit current file history` keymap. Also, this PR may be shortening the gitsigns hunk keybindings, but it unnecessarily makes the Telescope ones longer (status, commits) longer, which again in my own personal preference are more commonly used.

Personally, I think most of these changes should be made by the user in his own personal configuration.

lkhphuc:
> Also the `<leader>gf` change into a different namespace for `git find` doesn't take into consideration the latest addition of `Lazygit current file history` keymap

Ah thanks for pointing out. I missed this one since I also use Diffview.nvim and also bind `<leader>gh` for current file history and `<leader>gH` for repo history.
For the sake of this PR, I change `<leader>gf` to `<leader>gh` current file git's history with lazygit.

folke:
Not a fan of this. Closing...
Thank you anyway!

PR #2797, opened 2024-03-21
feat(treesitter): `v` for incremental node select

lkhphuc:
In case a breaking change get through #2795 , I propose another one for treesitter's textobject selection borrowed from from reddit.
Using `v` as increment keymap, now you can just spam `vvvvv` from normal mode to select bigger treesitter node, and `<bs>` to decrease node.
Everything behaves absolutely the same as before, `V` uses to change between line selection, etc.

### DIFF
--- a/lua/lazyvim/plugins/treesitter.lua
+++ b/lua/lazyvim/plugins/treesitter.lua
@@ -81,8 +81,7 @@ return {
       incremental_selection = {
         enable = true,
         keymaps = {
-          init_selection = "<C-space>",
-          node_incremental = "<C-space>",
+          node_incremental = "v",
           scope_incremental = false,
           node_decremental = "<bs>",
         },


dpetka2001:
What happens to the `v<text_object_motion>` usage (also in combination with `mini.ai` objects)? Does it still behave the same? I use it a lot in my workflow.

lkhphuc:
> What happens to the `v<text_object_motion>` usage (also in combination with `mini.ai` objects)? Does it still behave the same? I use it a lot in my workflow.

Yes it's still the same.
This change effective define a keymap in "v" in visual mode.
If you run `:verbose xmap v` and it's return no keymap then this change does not affect anything.

dpetka2001:
Cool thanks for the clarification.

folke:
Not a fan of this. Closing...
Thank you anyway!

PR #2798, opened 2024-03-21
feat(extras): add dial extra

Demianeen:
I just spent few hours trying to figure out how to set up dial.nvim, so I made this PR in the hope of saving other people's time)

I expect that people will modify the options and will create their own groups for dial. So I added few comments on how it works


folke:
Looks good!

I just went ahead and simplified some of the logic.

I also inverted the ft mapping similar to how it is with `conform.nvim` and `nvim-lint`

folke:
Merged! I was already using dial.nvim, but this config is much better. Thank you!

Demianeen:
> Merged! I was already using dial.nvim, but this config is much better. Thank you!

Pleasure!

Demianeen:
> Merged! I was already using dial.nvim, but this config is much better. Thank you!

Wow, just looked at your code. You are a genius seriously😂. I learnt a lot from it)

PR #2802, opened 2024-03-22
feat(extras): added gleam lang support

jluzny:
Added [Gleam lang](https://gleam.run/) support.

### DIFF
--- a/lua/lazyvim/plugins/extras/lang/gleam.lua
+++ b/lua/lazyvim/plugins/extras/lang/gleam.lua
@@ -14,13 +14,4 @@ return {
       require("lspconfig").gleam.setup({})
     end,
   },
-
-  {
-    "JoosepAlviste/nvim-ts-context-commentstring",
-    opts = {
-      languages = {
-        gleam = "// %s",
-      },
-    },
-  },
 }
--- a/lua/lazyvim/plugins/extras/lang/gleam.lua
+++ b/lua/lazyvim/plugins/extras/lang/gleam.lua
@@ -10,9 +10,6 @@ return {
 
   {
     "neovim/nvim-lspconfig",
-    init = function()
-      require("lspconfig").gleam.setup({})
-    end,
     opts = {
       servers = {
         gleam = {},
--- a/lua/lazyvim/plugins/extras/lang/gleam.lua
+++ b/lua/lazyvim/plugins/extras/lang/gleam.lua
@@ -1,0 +1,26 @@
+return {
+  {
+    "nvim-treesitter/nvim-treesitter",
+    opts = function(_, opts)
+      if type(opts.ensure_installed) == "table" then
+        vim.list_extend(opts.ensure_installed, { "gleam" })
+      end
+    end,
+  },
+
+  {
+    "neovim/nvim-lspconfig",
+    init = function()
+      require("lspconfig").gleam.setup({})
+    end,
+  },
+
+  {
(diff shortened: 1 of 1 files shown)


rubiin:
you should also add lsp support as well.

tkivela:
> you should also add lsp support as well. https://gleam.run/news/v0.21-introducing-the-gleam-language-server/ check here for more

Doesn't `nvim-lspconfig` configuration in the PR do effectively just that?

lua/lazyvim/plugins/extras/lang/gleam.lua
```diff
@@ -0,0 +1,17 @@
...
+        vim.list_extend(opts.ensure_installed, { "gleam" })
+      end
+    end,
+  },
+
+  {
+    "neovim/nvim-lspconfig",
+    init = function()
+      require("lspconfig").gleam.setup({})
+    end,
```
dpetka2001, line 15:
```suggestion
	opts = {
    	servers = {
    		gleam = {},
    	},
    },
```
This way the server will also get auto-installed by Mason.

dpetka2001: Oh it seems Mason has deprecated the installation for Gleam and it should be installed another way. Still, I believe the above changes make more sense in the spirit of how LazyVim configures LSP servers.

dpetka2001 (approved):

rubiin:
> > you should also add lsp support as well. [gleam.run/news/v0.21-introducing-the-gleam-language-server](https://gleam.run/news/v0.21-introducing-the-gleam-language-server/) check here for more
> 
> Doesn't `nvim-lspconfig` configuration in the PR do effectively just that?

Oh yeah , my bad . probably missed that

lua/lazyvim/plugins/extras/lang/gleam.lua
rubiin: True this looks sensible

rubiin (changes requested):
Add mason install for the gleam language server
`
  {
    "williamboman/mason.nvim",
    opts = function(_, opts)
      vim.list_extend(opts.ensure_installed, {
        "gleam",
      })
    end,
  },
`

lua/lazyvim/plugins/extras/lang/gleam.lua
```diff
@@ -0,0 +1,17 @@
+return {
+  {
+    "nvim-treesitter/nvim-treesitter",
+    opts = function(_, opts)
+      if type(opts.ensure_installed) == "table" then
+        vim.list_extend(opts.ensure_installed, { "gleam" })
+      end
+    end,
+  },
+
```
rubiin, line 10: {
    "williamboman/mason.nvim",
    opts = function(_, opts)
      vim.list_extend(opts.ensure_installed, {
        "elixir-ls",
      })
    end,
  },

lua/lazyvim/plugins/extras/lang/gleam.lua
jluzny: Indeed, the installation of LSP via mason has been removed per the request of gleam's team; see https://github.com/mason-org/mason-registry/pull/3872

dpetka2001:
@jluzny The lines with the `init` function is not needed at all (lines 13-15). See my [comment](https://github.com/LazyVim/LazyVim/pull/2802#discussion_r1544908976).

folke:
Thanks!

PR #2804, opened 2024-03-22
feat(lang/thrift): add thrift lang support

leaxoy:
This PR adds support for thriftls, the first line is because thrift is not among the neovim built-in file types.


amaanq:
The filetype is built in and has been for a while now

leaxoy:
My mistake, I will update this later.

leaxoy:
@amaanq A little late, but finally updated.

folke:
Updated and merged. Thanks!

PR #2806, opened 2024-03-22
feat(extras): Add extra for angular lspconfig

radvil:
Since treesitter has better support for angular parser recently that works for the legacy and the latest version of it, I am happy to suggest this extra configuration that works for well for me without any issue. And since lack of references about how to config the angular lsp inside of neovim out there, this will work fine as a starting point.

### DIFF
--- a/lua/lazyvim/plugins/extras/lang/angular.lua
+++ b/lua/lazyvim/plugins/extras/lang/angular.lua
@@ -1,0 +1,79 @@
+local goto_template_file = function()
+  local params = vim.lsp.util.make_position_params(0)
+  vim.lsp.buf_request(0, "angular/getTemplateLocationForComponent", params, function(_, result)
+    if result then
+      vim.lsp.util.jump_to_location(result, "utf-8")
+    end
+  end)
+end
+
+local goto_component_file = function()
+  local params = vim.lsp.util.make_position_params(0)
+  vim.lsp.buf_request(0, "angular/getComponentsWithTemplateFile", params, function(_, result)
+    if result then
+      if #result == 1 then
+        vim.lsp.util.jump_to_location(result[1], "utf-8")
+      else
+        vim.fn.setqflist({}, " ", {
+          title = "Angular Language Server",
+          items = vim.lsp.util.locations_to_items(result, "utf-8"),
+        })
+        vim.cmd.copen()
+      end
+    end
+  end)
+end
+
+---@param id string
+---@param cmd string | function
+---@param desc string
+local create_cmd = function(id, cmd, desc)
+  vim.api.nvim_create_user_command(string.format("Ng%s", id), cmd, {
+    desc = string.format("Angular » %s", desc),
+  })
+end
+
+return {
+  {
+    "nvim-treesitter",
+    optional = true,
+    opts = function(_, opts)
+      if type(opts.ensure_installed) == "table" then
+        vim.list_extend(opts.ensure_installed, { "angular" })
+      end
+    end,
+  },
+
+  {
+    "nvim-lspconfig",
+    opts = {
+      servers = {
+        angularls = {
+          root_dir = function(fname)
+            local util = require("lspconfig").util
+            local original = util.root_pattern("angular.json", "project.json")(fname)
+            -- set fallback for nx workspace support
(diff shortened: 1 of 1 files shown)


radvil:
Hi @folke I think the new `vtsls` lsp config pairs really well with this configuration, I am experiencing it's running smooth now. Can you review this ?

lua/lazyvim/plugins/extras/lang/angular.lua
```diff
@@ -0,0 +1,61 @@
+return {
+  recommended = function()
+    return LazyVim.extras.wants({
+      ft = { "html", "typescript" },
```
folke, line 4: Can you remove the filetypes here?
Otherwise angular will show up as recommended for any typescript/html file.
Having the root dir detection is enough

lua/lazyvim/plugins/extras/lang/angular.lua
```diff
@@ -0,0 +1,61 @@
...
+  {
+    "neovim/nvim-lspconfig",
+    opts = function(_, opts)
+      LazyVim.extend(opts.servers.vtsls, "settings.vtsls.tsserver.globalPlugins", {
+        {
+          name = "@angular/language-server",
+          location = LazyVim.get_pkg_path(
+            "angular-language-server",
+            "/node_modules/@angular/language-server",
+            { warn = false }
```
folke, line 54: You shoyld probably remove that warn=false?
I've only added that because vue/astro need an upstream mason PR merged to make it work

folke (changes requested):

lua/lazyvim/plugins/extras/lang/angular.lua
radvil: nice, thanks for the response. I ll update the PR, maybe I ll also remove the `nx.json` ? because nx can also be used to other frameworks.

lua/lazyvim/plugins/extras/lang/angular.lua
radvil: Thanks for this info, I ll update my loval config first and trynna play around with my angular codebases about one hour, then I ll update the PR again. Thanks..

radvil:
Hi @folke, just wanna let you know I've updated this PR based your recommendation. Also with a slight different on 
```typescript
enableForWorkspaceTypeScriptVersions = false
```
as per recommendation on the README here https://github.com/angular/vscode-ng-language-service to opt typescript version bundled with the language server itself as the default. However I want to try this config for on daily basis, maybe you should not merge this now, I ll report tomorrow if I found anything doesn't work as expected, but for now I am not having any issue with it.

folke:
Looks good. Thanks!

radvil:
> !

Thank, I love ur work man..

rofrol:
@radvil I have enabled the plugin. Restarted nvim.

But it seems it does not work for me. I am looking at these features https://angular.dev/tools/language-service

- [Autocompletion](https://angular.dev/tools/language-service#autocompletion) - works without this plugin
- [Error checking](https://angular.dev/tools/language-service#error-checking) - does not work, no error shown
- [Quick info and navigation](https://angular.dev/tools/language-service#quick-info-and-navigation) - no info popup, `<leader>gd` does nothing.

radvil:
> @radvil I have enabled the plugin. Restarted nvim.
> 
> But it seems it does not work for me. I am looking at these features https://angular.dev/tools/language-service
> 
> * [Autocompletion](https://angular.dev/tools/language-service#autocompletion) - works without this plugin
> * [Error checking](https://angular.dev/tools/language-service#error-checking) - does not work, no error shown
> * [Quick info and navigation](https://angular.dev/tools/language-service#quick-info-and-navigation) - no info popup, `<leader>gd` does nothing.
> 
> EDIT: Maybe I need to enable strictTemplates in tsconfig.json?

Thanks for your feedback on this, finally another angular developer ..😁

Strangely enough, because mine works just fine even without "strictTemplate".  Which version of angular ur project currently use?


https://github.com/LazyVim/LazyVim/assets/36059968/35d3277d-e051-4bbb-988b-60d4196c6508

radvil:
I am not so sure if I did this correctly, maybe we should add description for this as experimental? what do you think @folke ? I am not really a wizard in this, but I try to do it best. Thanks again for your feedback @rofrol.

PR #2837, opened 2024-03-26
feat(yanky): add yanky as completion source

kevintraver:
Add **yanky** as a completion source using [cmp_yanky](https://github.com/chrisgrieser/cmp_yanky) when **yanky** extra is enabled.

### DIFF
--- a/lua/lazyvim/plugins/extras/coding/yanky-completion.lua
+++ b/lua/lazyvim/plugins/extras/coding/yanky-completion.lua
@@ -1,0 +1,12 @@
+return {
+  -- add yanky as completion source
+  {
+    "hrsh7th/nvim-cmp",
+    dependencies = { "chrisgrieser/cmp_yanky" },
+    opts = function(_, opts)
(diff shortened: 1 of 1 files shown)


dpetka2001:
Personally I would prefer it if this could be made into an extra of its own. I use yanky but wouldn't want the `nvim-cmp` source of it. Of course, I could just remove it from the sources in my personal configuration, but I think things like these should be left to the user to opt-in, if possible. 

In the end, it's up to the maintainer to decide.

kevintraver:
> Personally I would prefer it if this could be made into an extra of its own. I use yanky but wouldn't want the `nvim-cmp` source of it. Of course, I could just remove it from the sources in my personal configuration, but I think things like these should be left to the user to opt-in, if possible.
> 
> In the end, it's up to the maintainer to decide.

That is a good point. I moved it into a separate extra.

folke:
Not a fan of this. Closing...
Thank you anyway!

PR #2838, opened 2024-03-27
fix(autocmds): remove query from q-to-quit autocmd

amaanq:
I fixed this before in #473 but maybe it was accidentally reintroduced. I can understand the benefits of being able to just use q to close the query editor, but query *files* (those in nvim-treesitter or any tree sitter repository) are actual files and become very annoying to edit with this shortcut if i accidentally hit q


rubiin (approved):

folke:
makes sense. Thanks!

PR #2839, opened 2024-03-27
fix(python-semshi): better highlights

amaanq:
I decided to just PR this in from my config for others that are using it, but I've switched to basedpyright for semantic highlighting now though :)


folke:
Thanks!

PR #2853, opened 2024-03-27
feat(extra): add repl and code cell

lkhphuc:
LazyVim extra for REPL functionality, emulate [VSCode's code cell](https://code.visualstudio.com/docs/python/jupyter-support-py#_jupyter-code-cells) features, using [slime](https://github.com/jpalardy/vim-slime?tab=readme-ov-file), treesitter and other plugins already in used by LazyVim.

## Edit script
1. Open a normal python file and write code as usual.
1. Defining code cell with the [percent](https://github.com/mwouts/jupytext/blob/main/docs/formats-scripts.md#the-percent-format) format 
1. Write markdown cell with proper highlight injection  (particularly relevant for conversion from  Screenshot 2024-03-27 at 22 49 33
1. Optionally conceal cell marker with `mini.hipatterns`, revealed on selection Screenshot 2024-03-27 at 22 47 36
1. Navigate to previous and next cell with `]r`, `[r`

## Run
1. Start a neovim terminal for repl
    - Run `:SlimeTarget wezterm` for example if you don't want to use neovim's terminal
1. Attach code buffer to target with `<leader>rC`, `:SlimeConfig` or just start send code
1. Select the correct target terminal if there are many based on terminal jobid Screenshot 2024-03-27 at 22 37 48
    - lualine component to show terminal's jobid for terminal buffer and connected code buffer via slime  Screenshot 2024-03-27 at 23 47 49
1. Send code to repl
    -  `<S-CR>` + motion in normal mode, `<S-CR>ip, <S-CR>af, <S-CR>ac` etc
    - Visually select + `<S-CR>` `<S-v><S-CR>` to send current line etc...
    - `<S-CR><S-CR>` to send and jump to next cell, similar to jupyter notebook's <S-CR>

## Remarks
- Use `leader<r>, ir, ar, ]r, [r` for "repl", as there is not many letter left for new text objects. 
- The functionality from treesitter is only for python for now, but can be extend to other languages
- The custom treesitter queries are bundled directly to LazyVim, outside of the extra. I don't know if there's a better way.
- This extra is just a collection of simple configuration for REPL-behaviour and code cell manipulation. Advanced functionality for jupyter like notebook like conversion, display output, kernel execution with completion, hover etc are out of scope. These can comes later as another LazyVim extra.

### DIFF
--- a/lua/lazyvim/plugins/extras/repl/slime.lua
+++ b/lua/lazyvim/plugins/extras/repl/slime.lua
@@ -1,0 +1,127 @@
+return {
+  -- General vim repl code send
+  {
+    "jpalardy/vim-slime",
+    init = function()
+      vim.g.slime_no_mappings = 1
+      vim.g.slime_bracketed_paste = 1
+      vim.g.slime_target = "neovim"
+      vim.g.slime_menu_config = true
+      vim.api.nvim_create_user_command("SlimeTarget", function(opts)
+        if opts.args ~= nil then
+          vim.b.slime_target = opts.args
+          vim.b.slime_config = nil
+        else
+          vim.b.slime_config = vim.g.slime_config
+          vim.b.slime_target = vim.g.slime_target
+        end
+        vim.notify("Slime send target is: " .. vim.b.slime_target)
+      end, { desc = "Change Slime target", nargs = 1 })
+    end,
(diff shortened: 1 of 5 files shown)


folke:
Not interested in this, but thank you anyway!

PR #2859, opened 2024-03-28
feat(extras): add chezmoi extra

Demianeen:
It is just a small extra for those that use Chezmoi)

### DIFF
--- a/lua/lazyvim/plugins/extras/util/chezmoi.lua
+++ b/lua/lazyvim/plugins/extras/util/chezmoi.lua
@@ -3,15 +3,10 @@ return {
   {
     -- highlighting for chezmoi files template files
     "alker0/chezmoi.vim",
-    opts = {
-      use_tmp_buffer = true,
-      source_dir_path = os.getenv("HOME") .. "/.local/share/chezmoi",
-    },
-    config = function(_, opts)
-      for key, value in pairs(opts) do
-        vim.g["chezmoi#" .. key] = value
-      end
-    end,
+	init = function()
+		vim.g.chezmoi#use_tmp_buffer = 1
+		vim.g.chezmoi#source_dir_path = os.getenv("HOME") .. "/.local/share/chezmoi",
+	end,
   },
   {
     "xvzc/chezmoi.nvim",
--- a/lua/lazyvim/plugins/extras/util/chezmoi.lua
+++ b/lua/lazyvim/plugins/extras/util/chezmoi.lua
@@ -1,3 +1,4 @@
+vim.notify("Hello")
 return {
   {
     -- highlighting for chezmoi files template files
@@ -38,7 +39,7 @@ return {
         select = { "<CR>" },
       },
     },
-    config = function(_, opts)
+    init = function()
       -- run chezmoi edit on file enter
       vim.api.nvim_create_autocmd({ "BufRead", "BufNewFile" }, {
         pattern = { os.getenv("HOME") .. "/.local/share/chezmoi/*" },
@@ -46,7 +47,8 @@ return {
           vim.schedule(require("chezmoi.commands.__edit").watch)
         end,
       })
-
+    end,
+    config = function(_, opts)
       require("chezmoi").setup(opts)
     end,
   },
--- a/lua/lazyvim/plugins/extras/util/chezmoi.lua
+++ b/lua/lazyvim/plugins/extras/util/chezmoi.lua
@@ -3,10 +3,10 @@ return {
   {
     -- highlighting for chezmoi files template files
     "alker0/chezmoi.vim",
-	init = function()
-		vim.g.chezmoi#use_tmp_buffer = 1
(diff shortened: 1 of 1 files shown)


Demianeen:
The bug was [fixed](https://github.com/alker0/chezmoi.vim/issues/72)

dpetka2001:
I believe you could just put atop the start of the file if you would like and it will be processed during the `lazy.nvim` parsing the specs phase I believe, if you don't want to tie it to the `config` function, or you could also put it in the `init` function instead (which will happen before the plugin loading). Of course the maintainer is more appropriate to answer as to what the more appropriate way should be.

Demianeen:
If I would add at the start of the file user wouldn't even be able to see it on lazyvim website or disable it. So, I am not sure what to do in this situation

dpetka2001:
That's why I recommended (in my personal opinion) to prefer the `init` function instead. But better wait for the maintainer's reply, as he knows best :smile:

Demianeen:
Okay @dpetka2001, I finally understood what you meant! The autocmd works now even without pressing the required key to load a plugin

lua/lazyvim/plugins/extras/util/chezmoi.lua
```diff
@@ -0,0 +1,79 @@
...
+    "alker0/chezmoi.vim",
+    opts = {
+      use_tmp_buffer = true,
+      source_dir_path = os.getenv("HOME") .. "/.local/share/chezmoi",
+    },
+    config = function(_, opts)
+      for key, value in pairs(opts) do
+        vim.g["chezmoi#" .. key] = value
+      end
+    end,
```
dpetka2001, line 14: `vim.g` variables usually go to the `init` function instead of config because they need to be available before the plugin loads.

lua/lazyvim/plugins/extras/util/chezmoi.lua
```diff
@@ -0,0 +1,79 @@
...
+      vim.api.nvim_create_autocmd({ "BufRead", "BufNewFile" }, {
+        pattern = { os.getenv("HOME") .. "/.local/share/chezmoi/*" },
+        callback = function()
+          vim.schedule(require("chezmoi.commands.__edit").watch)
+        end,
+      })
+    end,
+    config = function(_, opts)
+      require("chezmoi").setup(opts)
+    end,
```
dpetka2001, line 53:
```suggestion
```
This is not actually required since you already have `opts`. This `config` you do here is the default `config` that `lazy.nvim` will do on its own.

lua/lazyvim/plugins/extras/util/chezmoi.lua
dpetka2001, line 14:
```suggestion
	init = function()
		vim.g.use_tmp_buffer = 1
		vim.g.source_dir_path = os.getenv("HOME") .. "/.local/share/chezmoi",
	end,
```
This might be better?

dpetka2001:
@Demianeen Would appreciate it if you took anything I say with a grain of salt. I do have some experience with configuring LazyVim, but my knowledge is not end-all/be-all. I hope you tested out the changes I recommended and everything works properly on your end. The part about `vim.g` variables belonging in the `init` function I remember from reading it in the `lazy.nvim` docs. Also no need to list me as co-author. This is your PR and I have only made some minor suggestions for better LazyVim way of doing things. You're the one who put in the work to make this work and also fix the bug in one of the plugins.

Demianeen:
Hi @dpetka2001! That's fine, I did take a look at you changes. As for coauthoring, it was done automatically by GitHub because I accepted your changes. I also don't mind coauthoring changes. Hopefully you are alright with that as well😃

lua/lazyvim/plugins/extras/util/chezmoi.lua
```diff
@@ -0,0 +1,71 @@
+vim.notify("Hello")
```
abusch, line 1: That probably shouldn't be here...

lua/lazyvim/plugins/extras/util/chezmoi.lua
```diff
@@ -0,0 +1,70 @@
...
+    -- highlighting for chezmoi files template files
+    "alker0/chezmoi.vim",
+    init = function()
+      vim.g["chezmoi#use_tmp_buffer"] = 1
+      vim.g["chezmoi#source_dir_path"] = os.getenv("HOME") .. "/.local/share/chezmoi"
+    end,
+  },
+  {
+    "xvzc/chezmoi.nvim",
+    dependencies = { "nvim-lua/plenary.nvim" },
```
folke, line 12: this dep is not needed, so please remove

Demianeen: This dependency is provided in the [installation docs](https://github.com/xvzc/chezmoi.nvim#installation). Or do you mean that Lazyvim somehow provides this dep to all packages?

folke: already included. just remove it

folke:
Thanks!

PR #2865, opened 2024-03-28
feat(extras): propose to add mini.move as an extra instead of native move command

ammon134:
The main motivation for using mini.move instead of the native move command is to avoid polluting the undo history as mini.move allows undoing consecutive moves with one single `u`. And horizontal moves are nice too!

I set up the keybindings manually to overwrite the default keymaps, as well as to set up moving in InsertMode.

### DIFF
--- a/lua/lazyvim/plugins/extras/editor/mini-move.lua
+++ b/lua/lazyvim/plugins/extras/editor/mini-move.lua
@@ -1,0 +1,32 @@
+return {
+  {
+    "echasnovski/mini.move",
+    event = "VeryLazy",
+    opts = {},
+    keys = function()
+      local ret = {}
+      local directions = { "left", "down", "up", "right" }
+      local keys = { "h", "j", "k", "l" }
+      local move = require("mini.move")
+      for i, dir in ipairs(directions) do
+        ret[#ret + 1] = {
+          "<A-" .. keys[i] .. ">",
+          mode = { "i", "n" },
+          function()
+            move.move_line(dir)
+          end,
+        }
+      end
+      for i, dir in ipairs(directions) do
+        ret[#ret + 1] = {
+          "<A-" .. keys[i] .. ">",
+          mode = { "v" },
+          function()
+            move.move_selection(dir)
+          end,
+        }
+      end
+      return ret
+    end,
+  },
+}


folke:
Awesome, thanks!

theol0403:
In the quest for simplicity -- thoughts on removing the default keymaps, and allowing users to opt-in to the move bindings solely by using mini.move? Seems redundant to have "good" and "better" bindings with equivalent functionality.

folke:
In terms of simplicity, the included native keymaps cover 99% of the cases.
You can of course always add mini.move in your own config.

theol0403:
Well, the real reason for this suggestion is to fix https://github.com/LazyVim/LazyVim/pull/2945, because currently, enabling `mini.move` in the extras UI isn't enough to configure the plugin, the user also has to add custom configuration to prevent it from clashing with the default bindings. 

Additionally, I think `mini.move` is the better option for 99% of users, so adopting an "if you want move bindings, enable extras.mini.move" approach instead of "you already have move bindings, but if you want slightly better ones enable this extra" is more elegant.

folke:
Feel free to submit a PR for the extra.

theol0403:
@folke it already is an extra, thanks to this PR. Or do I misunderstand?

folke:
Sorry, I was confused. Now I remember. All the code in this PR turned out not to be needed.
And I've then changed the extra to:

```lua
return {
  {
    "echasnovski/mini.move",
    event = "VeryLazy",
    opts = {},
  },
}
```

So not event worth to be an extra tbh, but it's there now :)

dpetka2001:
@folke #2945 used to not work with the version of `lazyfile` function, but with the new I don't think there's any problem. So, maybe just loading `mini.move` on `LazyFile` instead will solve the issue of the default keymaps overriding the plugin's ones.

ammon134:
@dpetka2001 As you mentioned in #2945 loading `mini.move` on `LazyFile` won't work when opening nvim with an argument and the default keymap will still overwrite `mini.move`.

dpetka2001:
That was with the old `lazyfile` function. With the new one i don't believe there should be problems. Test it out locally to see if it works. Just do in your personal config
```lua
return {
  {
    "mini.move",
    event = function() 
      return "LazyFile" 
    end, 
  }, 
}
```

PR #2877, opened 2024-03-29
fix: optimizes harpoon keys

rubiin:
Basically uses loop for individual harpoon file key spec


folke:
thanks!

rubiin:
Damn that was fast :D 🚀

PR #2879, opened 2024-03-29
feat(dot): add bashls and shellcheck

rubiin:
I work with shell files a lot , so this basically adds support for shell files

### DIFF
--- a/lua/lazyvim/plugins/extras/util/dot.lua
+++ b/lua/lazyvim/plugins/extras/util/dot.lua
@@ -7,6 +7,23 @@ local function have(path)
 end
 
 return {
+  {
+    "neovim/nvim-lspconfig",
+    opts = {
+      servers = {
+        bashls = {},
+      },
+    },
+  },
+  {
+    "williamboman/mason.nvim",
+    opts = function(_, opts)
+      vim.list_extend(opts.ensure_installed or {}, {
+        "shfmt",
+        "shellcheck",
+      })
+    end,
+  },
   -- add some stuff to treesitter
   {
     "nvim-treesitter/nvim-treesitter",


lua/lazyvim/plugins/extras/lang/shell.lua
```diff
@@ -0,0 +1,28 @@
...
+      servers = {
+        bashls = {},
+      },
+    },
+  },
+  {
+    "nvim-treesitter/nvim-treesitter",
+    opts = function(_, opts)
+      if type(opts.ensure_installed) == "table" then
+        vim.list_extend(opts.ensure_installed, { "bash" })
```
folke, line 14: This is already included in LazyVim

lua/lazyvim/plugins/extras/lang/shell.lua
```diff
@@ -0,0 +1,28 @@
...
+      end
+    end,
+  },
+  {
+    "williamboman/mason.nvim",
+    opts = function(_, opts)
+      vim.list_extend(opts.ensure_installed or {}, {
+        -- formatters & linters
+        "shfmt",
```
folke, line 23: shfmt is not needed with bashls

folke (changes requested):
We already have an extra called `dot` for this kind of stuff.
Would be great if you can add bashls and shellcheck in there.

lua/lazyvim/plugins/extras/lang/shell.lua
folke: Scrap that. It **is** needed and I'm actually already using that for formatting :)

lua/lazyvim/plugins/extras/lang/shell.lua
rubiin: sure will do that instead

lua/lazyvim/plugins/extras/lang/shell.lua
rubiin: Moved the changes to dot file instead. I think I also need to change the PR title

folke:
Thanks!

PR #2881, opened 2024-03-29
feat(extras): add neogen

jakub-kozlowicz:


lua/lazyvim/plugins/extras/editor/neogen.lua
```diff
@@ -0,0 +1,45 @@
...
+          require("neogen").generate({ type = "type" })
+        end,
+        desc = "Generate documentation for type",
+      },
+    },
+  },
+
+  { -- which-key integration
+    "folke/which-key.nvim",
+    opts = {
```
dpetka2001, line 39: Probably want to guard it here with `optional = true`. Don't wanna bind this Extra to `which-key` explicitly I think, in case someone is using something else.

rabinadk1:
Can we add the following config for Neogen?

Differences with the proposed one.
1. One shortcut to generate the annotations. The plugin automatically selects the type of annotation.
2. Automatically select the snippet engine in the following priority order: "nvim", "luasnip", "snippy", and "vsnip"

```lua
return {
  "danymat/neogen",
  dependencies = "nvim-treesitter/nvim-treesitter",
  cmd = "Neogen",
  keys = {
    {
      "<leader>ga",
      function()
        require("neogen").generate({})
      end,
      desc = "Generate Annotations",
    },
  },
  ---@param _ LazyPlugin
  ---@param opts table
  opts = function(_, opts)
    local snippet_engine = nil

    if vim.snippet ~= nil then
      snippet_engine = "nvim"
    else
      local util = require("lazyvim.util")
      if util.has("luasnip") then
        snippet_engine = "luasnip"
      elseif util.has("snippy") then
        snippet_engine = "snippy"
      elseif vim.g.loaded_vsnip then
        snippet_engine = "vsnip"
      end
    end

    opts.snippet_engine = snippet_engine
  end,
}
```

dpetka2001:
@rabinadk1 This way when `vim.snippet` is available (which is on Neovim 0.10+) it will be hardcoded to use the native snippets engine by default, which some users might not like and prefer to use Luasnip instead. I believe it should be left up to the user to define what he wants. It's one line in the personal config after all.

rabinadk1:
> @rabinadk1 This way when `vim.snippet` is available (which is on Neovim 0.10+) it will be hardcoded to use the native snippets engine by default, which some users might not like and prefer to use Luasnip instead. I believe it should be left up to the user to define what he wants. It's one line in the personal config after all.

Since LazyVim already prefers native snippets on default for `nvim >=0.10`, I though it would be a sensible default than not using a snippet at all. As you said, you can change the config later according to their preferences.

dpetka2001:
Yes, but at least the checks could be reverted, so it first checks if any of the other snippet engines exist first (since the user would have explicitly enabled them) and as last fallback default to `vim.snippet`.

PR #2897, opened 2024-03-31
fix(lsp): check if `diagnostics.signs` is disabled by user

dpetka2001:
Based on #2896

### DIFF
--- a/lua/lazyvim/plugins/lsp/init.lua
+++ b/lua/lazyvim/plugins/lsp/init.lua
@@ -120,10 +120,12 @@ return {
 
       -- diagnostics signs
       if vim.fn.has("nvim-0.10.0") == 0 then
-        for severity, icon in pairs(opts.diagnostics.signs.text) do
-          local name = vim.diagnostic.severity[severity]:lower():gsub("^%l", string.upper)
-          name = "DiagnosticSign" .. name
-          vim.fn.sign_define(name, { text = icon, texthl = name, numhl = "" })
+        if type(opts.diagnostics.signs) ~= "boolean" then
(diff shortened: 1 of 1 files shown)


rubiin (approved):

lua/lazyvim/plugins/lsp/init.lua
```diff
@@ -120,10 +120,12 @@ return {
 
       -- diagnostics signs
       if vim.fn.has("nvim-0.10.0") == 0 then
-        for severity, icon in pairs(opts.diagnostics.signs.text) do
-          local name = vim.diagnostic.severity[severity]:lower():gsub("^%l", string.upper)
-          name = "DiagnosticSign" .. name
-          vim.fn.sign_define(name, { text = icon, texthl = name, numhl = "" })
+        if opts.diagnostics.signs then
```
SameAsMuli, line 123: Feels like it would be better to check the type of `opts.diagnostics.signs` here - if it's a boolean, but set to true, this will still cause issues. Alternatively, just revert the last part of [this changeset](https://github.com/LazyVim/LazyVim/commit/aaa10e717dbe76c5b6b0ab302698f1d151863dc6) that removed `vim.tbl_get(opts, "diagnostics", "signs", "text") or {}`.

dpetka2001: I didn't think that one would like to set `signs = true` to get the default icons instead. I've had in my mind that one would either set `signs = table_with_text_icons` or would set it to `false` to not get icons at all. 

Thank you :smile:

rubiin (approved):

folke:
Thanks!

Issue #2903, opened 2024-04-01
feature: yaml-language server autocompletion improvement


Dogacel:
### Did you check the docs?

- [X] I have read all the LazyVim docs

### Is your feature request related to a problem? Please describe.

My kubernetes yaml files did not show autocomplete out of the box. So I tried setting the schema manually however couldn't find many docs.

### Describe the solution you'd like

- Improve documentation to show how to set schemas.
  - AFAIK only way to set it is to put a comment on top such as `# yaml-language-server: $schema=https://kubernetesjsonschema.dev/v1.10.3-standalone/service-v1.json
` however there must be an easier way that also uses `b0o/SchemaStore.nvim"`. I feel like I am not using that schema store somehow.
- Make a LazyVim Extra plugin
  - The extra plugin can be setup as the following: https://github.com/someone-stole-my-name/yaml-companion.nvim/issues/44
  - Plugin contains both schema selector and lualine upgrade

### Describe alternatives you've considered

I am not aware of any alternatives.

### Additional context

Here is a PR to showcase:
- #2907

PR #2907, opened 2024-04-01
fix: add yaml companion to yaml setup

Dogacel:
Relates to #2903 

This adds the `yaml-companion` extension which auto-detects or allows users to manually select a yaml schema to make autocomplete work.

I think there are some documentation updates we gotta make as well. I haven't really looked into them, I am waiting this request to be approved 🙂

### DIFF
--- a/lua/lazyvim/plugins/extras/lang/yaml.lua
+++ b/lua/lazyvim/plugins/extras/lang/yaml.lua
@@ -74,4 +74,41 @@ return {
       },
     },
   },
+
+  -- support auto-detecting and selecting yaml schemas.
+  -- use ":Telescope yaml_schema" to change current schema.
+  {
+    "someone-stole-my-name/yaml-companion.nvim",
+    ft = { "yaml" },
+    dependencies = {
+      { "neovim/nvim-lspconfig" },
+      { "nvim-lua/plenary.nvim" },
+      { "nvim-telescope/telescope.nvim" },
+    },
+    config = function(_, opts)
+      local cfg = require("yaml-companion").setup(opts)
+      require("lspconfig")["yamlls"].setup(cfg)
+      require("telescope").load_extension("yaml_schema")
+    end,
+  },
+
+  -- Show currently used yaml-schema on lualine status bar
+  {
+    "nvim-lualine/lualine.nvim",
+    event = "VeryLazy",
+    opts = function(_, opts)
+      opts.sections.lualine_z = {
+        {
+          function()
+            local schema = require("yaml-companion").get_buf_schema(0)
+            if schema.result[1].name == "none" then
+              return ""
+            end
+            return schema.result[1].name
+          end,
+        },
+        opts.sections.lualine_z,
+      }
+    end,
+  },
 }


szechp:
its a bummer that this is plugin is not getting developed anymore, i recommend this fork:

https://github.com/msvechla/yaml-companion.nvim/tree/kubernetes_crd_detection

Dogacel:
> its a bummer that this is plugin is not getting developed anymore, i recommend this fork:
> 
> https://github.com/msvechla/yaml-companion.nvim/tree/kubernetes_crd_detection

Uh, are you sure? That fork is 2 commits behind ironically 😆

rubiin:
https://github.com/someone-stole-my-name/yaml-companion.nvim the last commit is 2 days old so I think this is pretty active

lua/lazyvim/plugins/extras/lang/yaml.lua
```diff
@@ -74,4 +74,41 @@ return {
...
+    },
+    config = function(_, opts)
+      local cfg = require("yaml-companion").setup(opts)
+      require("lspconfig")["yamlls"].setup(cfg)
+      require("telescope").load_extension("yaml_schema")
+    end,
+  },
+
+  -- Show currently used yaml-schema on lualine status bar
+  {
```
rubiin, line 96: I have some thoughts on this though. Instead of including it on default, maybe we should add it as optional.  Some users have their own lua line config like me and this will break things on them or make it look weird

dpetka2001: That's not what `optional` does. `optional` means that it will only be set up if there's another instance of the same plugin somewhere in another fragment of the plugin spec. 

I agree that the lualine spec in this PR should be `optional` however, as there might be other users that use a different statusline and it would be better to decouple the use of this PR from Lualine as a hard dependency with it.

dpetka2001: Yes, that's the same. You would do something like `{ "lualine.nvim", optional = true, ...}` here as well. Or am I misunderstanding what you're saying?

rubiin: Yeah. I was just talking about removing the lualine from the PR and adding a section on the docs similar to [lazyvim.org/extras/coding/copilot#lualinenvim-optional](https://www.lazyvim.org/extras/coding/copilot#lualinenvim-optional)

dpetka2001: There's no need to remove it. Just guard it with `optional` like all the rest of the extras do to take into consideration if the plugin in name is actually enabled or not.

And if you're worried about this PR messing with your lualine that's unavoidable. The only thing you can do (as far as I know) is remove in your personal configuration the lualine components you don't want to see. That's what I'm doing as well.

rubiin: Yeah I am thinking thats unavoidable. Will have to manually remove the component from lualine after this PR gets merged, atleast thats also the easiet solution I can think of atm

folke:
We already use schemastore, so this should not be needed.
For all the yaml files I edit that come with a schema, it just works.

If it doesn't work for some thing like kubernetes files, then user can always install this plugin.

Closing, but thank you anyway!

mikybars:
I was trying to have the current schema displayed in my lualine when I stumbled upon this... but I see that is closed T_T. Is there any setting in LazyVim that allows to do it?

PR #2890, opened 2024-03-31
feat(extras.lang): Add kotlin

arthur-mountain:
### Description:
  Adding kotlin support

---

### The plugins was added

1. `mason`
- [ktlint](https://github.com/pinterest/ktlint)
- [kotlin-language-server](https://github.com/neovim/nvim-lspconfig/blob/master/doc/server_configurations.md#kotlin_language_server)
- [kotlin-debug-adapter](https://github.com/fwcd/kotlin-debug-adapter)

---

2. `tree-sitter`
- kotlin

---

3. `nvim-lspconfig`
- kotlin-language-server

---

4. `nvim-lint`(optional)
- ktlint

---

5. `conform`(optional)
- ktlint

---

6. `none-ls`(optional)

---

7. `nvim-dap`(optional)
- [kotlin-debug-adapter](https://github.com/mfussenegger/nvim-dap/wiki/Debug-Adapter-installation#kotlin)

---
8. mini.comment/nvim-ts-context-commentstring
- I also opened a [pr](https://github.com/JoosepAlviste/nvim-ts-context-commentstring/pull/104) for kotlin support in nvim-ts-context-commentstring cause I found the commentstring is empty

Issue #2546, opened 2024-02-13
feature: Kotlin support


mairs8:
### Did you check the docs?

- [X] I have read all the LazyVim docs

### Is your feature request related to a problem? Please describe.

I would like to use LazyVim for Kotlin development but i do not see it being treated as first class citizen i.e. there is no kotlin in the 'plugins/lang' folder. 

### Describe the solution you'd like

Is it possible to add this?

### Describe alternatives you've considered

na

### Additional context

Please elaborate If there is a particular reason why kotlin has been blacklisted. Thank you

dpetka2001:
All the language Extras come from contributions via PRs from users of LazyVim. Kotlin is not there because no LazyVim user that potentially uses Kotlin has made a PR for Kotlin support. If you'd like feel free to make a PR for Kotlin support. So nothing is blacklisted as you erroneously have pointed out.

folke:
Feel free to provide an extra for Kotlin

PR #2908, opened 2024-04-01
feat: add kotlin language support with basic features

Dogacel:
I am pretty new to nvim toolchain. I have initially invested some time to setup a very basic kotlin config.

I would like to provide it here as a starting point.

### DIFF
--- a/lua/lazyvim/plugins/extras/lang/kotlin.lua
+++ b/lua/lazyvim/plugins/extras/lang/kotlin.lua
@@ -1,0 +1,26 @@
+return {
+  {
+    "nvim-treesitter/nvim-treesitter",
+    opts = function(_, opts)
+      vim.list_extend(opts.ensure_installed, {
+        "kotlin",
+      })
+    end,
+  },
+  {
+    "williamboman/mason.nvim",
+    opts = function(_, opts)
+      vim.list_extend(opts.ensure_installed, {
+        "kotlin-language-server",
+      })
+    end,
+  },
+  {
+    "neovim/nvim-lspconfig",
+    opts = {
+      servers = {
+        kotlin_language_server = {},
+      },
+    },
+  },
+}


dpetka2001:
There is also #2890. But in my opinion, your handling of nvim-treesitter and mason `opts.ensured_installed` is preferred. If you could also provide linting, formatting and dap configuration (according to the official [docs](https://github.com/mfussenegger/nvim-dap/wiki/Debug-Adapter-installation#kotlin), because the other PR doesn't actually provide a dap configuration for Kotlin) with `optional = true` for the corresponding plugins, I think your PR would be ready to merge. 

I'm just a regular user like you, so I don't know if you should collaborate with the OP of the other PR, so that there won't be duplicate PRs for the same thing. Surely the maintainer would be more appropriate to make that decision.

Dogacel:
@dpetka2001 Oops I actually didn't see that PR :).

Our plugins looks pretty much alike. Can you explain this? I couldn't see any difference other than semantics. Technically those arrays should be merged resulting in the same result?
> your handling of nvim-treesitter and mason opts.ensured_installed is preferred

cc @arthur-mountain if you wanna collab

dpetka2001:
nvim-treesitter and mason's `opts.ensure_installed` are list-like tables, not key-values-like tables, so the `opts` couldn't merge them (as they use `vim.tbl_deep_extend` in the background), so they would overwrite the default `opts.ensure_installed` of those plugins and that could lead to potential issues. So, `vim.list_extend` should be used instead to append them to the original `opts.ensure_installed`.

Dogacel:
> nvim-treesitter and mason's `opts.ensure_installed` are list-like tables, not key-values-like tables, so the `opts` couldn't merge them (as they use `vim.tbl_deep_extend` in the background, not really but a custom implementation in the `lazy.nvim` package manager that acts similarly), so they would overwrite the default `opts.ensure_installed` of those plugins and that could lead to potential issues. So, `vim.list_extend` should be used instead to append them to the original `opts.ensure_installed` of those plugins.

Very impressive, thanks for the explanation 🙇

dpetka2001:
Not impressive, i'm not even a programmer. And if you ask me about technical details i doubt i could answer you. It's just basic knowledge of Lua i aqcuired to be able to understand at least on the surface how i can configure my Neovim. I just tried examples of code and printed them out so i could see what they produce in order to be able to visualize what's happening (for example i tried a variant of mason's `ensure_installed` and tried to extend both with `vim.tbl_deep_extend` and `vim.list_extend` and saw the results for myself) . That's all :p

arthur-mountain:
@dpetka2001

Oh, I didn't realize `opts.ensured_installed` would overwrite; 

I thought it would merge. Otherwise, I would have updated the reference via a function.

arthur-mountain:
@Dogacel 
> Our plugins looks pretty much alike. Can you explain this? I couldn't see any difference other than semantics. Technically those arrays should be merged resulting in the same result?

I'm about to start learning Kotlin, but I found out that lazyvim doesn't support Kotlin.

I searched through the issues and found [this one](https://github.com/LazyVim/LazyVim/issues/2546) that was closed, but it seems there hasn't been any follow-up pull request.

So, I'm trying to provide the kotlin support.

Our plugins seem to be the same, except for the above, I might accidentally overwrite opts.ensured_installed >__<

arthur-mountain:
@Dogacel 
I'm sorry. What should I do in such a situation?

This is my first time contributing a PR.

I will also update my PR to avoid overwriting ensured_installed.

Dogacel:
> @Dogacel I'm sorry.>__<
> 
> What should I do in such a situation?
> 
> This is my first time contributing a PR.
> 
> I will also update my PR to avoid overwriting ensured_installed.

I want to say you opened the PR first and I missed it, so it would be better if you update your PR with the approach I did and we can continue from there. 🙂

arthur-mountain:
@Dogacel 
> I want to say you opened the PR first and I missed it, so it would be better if you update your PR with the approach I did and we can continue from there. 🙂

Thank you very much for your kindness and warmth.

I have made the following changes and updates on GitHub:
1. Updated settings for `mason` and `nvim-treesitter` to avoid overriding ensure_installed.
2. Added DAP configuration.

PR #2910, opened 2024-04-01
feat(extras.lang): add charp_roslyn support

cbosvik:
Add support for the new csharp lsp used in vscode

### DIFF
--- a/lua/lazyvim/plugins/extras/lang/charp_roslyn.lua
+++ b/lua/lazyvim/plugins/extras/lang/charp_roslyn.lua
@@ -1,0 +1,18 @@
+local capabilities = vim.lsp.protocol.make_client_capabilities()
+capabilities = require("cmp_nvim_lsp").default_capabilities(capabilities)
+local on_attach = function() end
+return {
+  {
+    "jmederosalvarado/roslyn.nvim",
+    ft = "cs",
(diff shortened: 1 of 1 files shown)


lua/lazyvim/plugins/extras/lang/charp_roslyn.lua
```diff
@@ -0,0 +1,18 @@
...
+return {
+  {
+    "jmederosalvarado/roslyn.nvim",
+    ft = "cs",
+    config = function()
+      require("roslyn").setup({
+
+        dotnet_cmd = "dotnet", -- this is the default
+        roslyn_version = "4.8.0-3.23475.7", -- this is the default
+        on_attach = on_attach,
```
dpetka2001, line 13: I would guess that here you would want to pass the LSP keymaps at least? Could you test with `on_attach = require("lazyvim.plugins.lsp.keymaps").on_attach(client, buffer)`? Does it actually work for you?

cbosvik: It does not seem necessary as the keymaps are working without explicitly passing them to the on_attach function. This is probably a very minimal implementation of the plugin but it works good and is much faster than Omnisharp.

PR #2923, opened 2024-04-04
feat(statuscol): fold open indicator

lkhphuc:
- Add to the status column the sign to indicate open fold, i.e where a fold will be triggered.
- Open and close fold sign will use the git sign color if it git info is present on the line (rel. line 5,4,2) and gitsigns' `signcolumn` is enabled (by default)
- otherwise close fold icon use "Folded" group, as before (line 7)
- open fold icon has no special highlight, the same as the number column (line 8, 3) 
Screenshot 2024-04-05 at 22 43 59

This is similar to VSCode on mouse hover
Screenshot 2024-04-04 at 17 28 00

### DIFF
--- a/lua/lazyvim/config/options.lua
+++ b/lua/lazyvim/config/options.lua
@@ -19,6 +19,12 @@ vim.g.root_spec = { "lsp", { ".git", "lua" }, "cwd" }
 -- Set to false to disable.
 vim.g.lazygit_config = true
 
+-- Options for the LazyVim statuscolumn
+vim.g.lazyvim_statuscolumn = {
+  folds_open = false, -- show fold sign when fold is open
+  folds_githl = false, -- highlight fold sign with git sign color
+}
+
 -- Optionally setup the terminal to use
 -- This sets `vim.o.shell` and does some additional configuration for:
 -- * pwsh
--- a/lua/lazyvim/util/ui.lua
+++ b/lua/lazyvim/util/ui.lua
@@ -101,10 +101,11 @@ function M.statuscolumn()
 
   if show_signs then
     ---@type Sign?,Sign?,Sign?
-    local left, right, fold
+    local left, right, fold, githl
     for _, s in ipairs(M.get_signs(buf, vim.v.lnum)) do
       if s.name and (s.name:find("GitSign") or s.name:find("MiniDiffSign")) then
         right = s
+        githl = s["texthl"]
       else
         left = s
       end
@@ -114,7 +115,9 @@ function M.statuscolumn()
     end
     vim.api.nvim_win_call(win, function()
       if vim.fn.foldclosed(vim.v.lnum) >= 0 then
-        fold = { text = vim.opt.fillchars:get().foldclose or "", texthl = "Folded" }
+        fold = { text = vim.opt.fillchars:get().foldclose or "", texthl = githl or "Folded" }
+      elseif not LazyVim.ui.skip_foldexpr[buf] and vim.treesitter.foldexpr(vim.v.lnum):sub(1, 1) == ">" then -- fold start
+        fold = { text = vim.opt.fillchars:get().foldopen or "", texthl = githl }
       end
     end)
     -- Left: mark or non-git sign
--- a/lua/lazyvim/util/ui.lua
+++ b/lua/lazyvim/util/ui.lua
@@ -99,13 +99,18 @@ function M.statuscolumn()
 
   local components = { "", "", "" } -- left, middle, right
 
+  local show_open_folds = vim.g.lazyvim_statuscolumn and vim.g.lazyvim_statuscolumn.folds_open
+  local use_githl = vim.g.lazyvim_statuscolumn and vim.g.lazyvim_statuscolumn.folds_githl
+
   if show_signs then
     ---@type Sign?,Sign?,Sign?
     local left, right, fold, githl
     for _, s in ipairs(M.get_signs(buf, vim.v.lnum)) do
       if s.name and (s.name:find("GitSign") or s.name:find("MiniDiffSign")) then
         right = s
-        githl = s["texthl"]
+        if use_githl then
+          githl = s["texthl"]
+        end
       else
         left = s
       end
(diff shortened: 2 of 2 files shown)


dpetka2001:
I like this addition, but it doesn't update the highlight for `foldopen` for me when I try locally your PR. Is there something else I'm missing? Maybe a setting in the user configuration?

lkhphuc:
Thanks. Glad you like it. 
But what do you means by updating the highlight for `foldopen`?
The modification is applied by default for LazyVim, as long as you don't overwrite the default
`set foldexpr?   foldexpr=v:lua.require'lazyvim.util'.ui.foldexpr()`

dpetka2001:
I mean that it doesn't inherit the color of gitsign. Following is a screencast

[Screencast 2024-04-04 19:07:24.webm](https://github.com/LazyVim/LazyVim/assets/12776461/73b7dd29-224a-4652-bd42-726d52cd53ca)

As you can see in the video, the `foldopen` icon doesn't inherit the gitsign color.

dpetka2001:
I see that you're using your custom `statuscolumn`. Can you try with default LazyVim installation and confirm?

dpetka2001:
I tried your dotfiles and it does indeed inherit the color from gitsigns with your custom `statuscolumn`. But default LazyVim `statuscolumn` does not do that. So, this PR, without any other changes, only adds the `foldopen` icon in the `statuscolumn`.

dpetka2001:
@lkhphuc 

I believe I managed to solve that with the following change
```lua
    vim.api.nvim_win_call(win, function()
      if vim.fn.foldclosed(vim.v.lnum) >= 0 then
        fold = { text = vim.opt.fillchars:get().foldclose or "", texthl = "Folded" }
      elseif not LazyVim.ui.skip_foldexpr[buf] and vim.treesitter.foldexpr(vim.v.lnum):sub(1, 1) == ">" then -- fold start
        for _, s in ipairs(M.get_signs(buf, vim.v.lnum)) do
          if s.name and (s.name:find("GitSign") or s.name:find("MiniDiffSign")) then
            fold = { text = vim.opt.fillchars:get().foldopen or "", texthl = s.texthl }
          end
        end
      end
    end)
```
The linenr is a setting from `gitsigns` itself, so please ignore that. But without this change in LazyVim's `statuscolumn`, your PR only provides the icon for `foldopen` without inhereting the `texthl` from `gitsigns`. Except if I'm missing something else.

dpetka2001:
Ok I deleted my previous post, because the solution was wrong. But this time I believe I got it right
```lua
    vim.api.nvim_win_call(win, function()
      if vim.fn.foldclosed(vim.v.lnum) >= 0 then
        fold = { text = vim.opt.fillchars:get().foldclose or "", texthl = "Folded" }
      elseif not M.skip_foldexpr[buf] and vim.treesitter.foldexpr(vim.v.lnum):sub(1, 1) == ">" then -- fold start
        if vim.tbl_isempty(M.get_signs(buf, vim.v.lnum)) then
          fold = { text = vim.opt.fillchars:get().foldopen or "" }
        else
          for _, s in ipairs(M.get_signs(buf, vim.v.lnum)) do
            if s.name and (s.name:find("GitSign") or s.name:find("MiniDiffSign")) then
              fold = { text = vim.opt.fillchars:get().foldopen or "", texthl = s.texthl }
            end
          end
        end
      end
    end)
```
Still not sure if it's the best way to do it, but it works as you said in your first post this way at least.

lkhphuc:
@dpetka2001 sorry my bad. I figure out why. I set up [numhl](https://github.com/lkhphuc/dotfiles/blob/ba27eab525a1ee9807912098cf22306cc3ff6402/nvim/lua/plugins/git.lua?plain=1#L15) in my gitsign.
In the current PR, the fold sign always inherit the color of num_hl, so it's only change color when gitsign change the color of number column.

dpetka2001:
Ohhh I see. I was trying to get it done without that setting. Thanks for clarifying that and good luck with your PR :smile:

dpetka2001:
Might I suggest we also do the following change
```lua
        if LazyVim.opts("gitsigns.nvim").numhl == true then
          fold = { text = vim.opt.fillchars:get().foldclose or "" }
        else
          fold = { text = vim.opt.fillchars:get().foldclose or "", texthl = "Folded" }
        end
```
so that the `foldclose` icon also inherits the gitsign `numhl` highlight if it's enabled in the user configuration?

rubiin (approved):

lkhphuc:
@dpetka2001 I updated the PR to use the git sign color when available, for both open and closed fold. Now it does not depends on gitsign's numhl feature.

dpetka2001:
@lkhphuc I appreciate your effort but I believe it would be wiser to leave it to inherit the `numhl` color if the user has enabled it in his configuration. This makes it more flexible to configure even for users who would not like this for example and would prefer to have default highlighting.

lkhphuc:
> I also try to think about how users can have a more flexible configuration based on their preferences.

Agree. And I think this behaviour of inherit gitsign's sign column is the correct and expected from user. The initial implementation inherit from number column was actually the wrong and counter intuitive one.

Technically, this PR add fold icon to the sign column, so ideally it should inherit the color of the thing it's overiding, here is gitsign's `signcolumn` feature. If you disabled `signcolumn` color from gitsigns, you also disable the color for fold icon and use "Folded" group or nothing as before.
Whatever the user configures for their number column, it does not affect the sign column, hence the fold icons here.

dpetka2001:
I understand what you mean and in hindsight, after reading your explanation, I totally agree. 

Thank you for this PR and also explaining to me your logic. Really appreciate it.

rubiin (approved):

folke:
I've made some changes to make this configurable.
The default config is as before. Check options.lua on how you can enable it in your own config.
Thanks!

AtifChy:
@dpetka2001 why doesn't foldopen icon show anymore?

dpetka2001:
https://github.com/folke/snacks.nvim/issues/1533 There's an issue with snacks statuscolumn.

dpetka2001:
@AtifChy The statuscolumn is provided by Snacks now. 

Did you enable the feature to show open folds? By default it's disabled. So, you should essentially do
```lua
return {
	{
		"folke/snacks.nvim",
		opts = {
			statuscolumn = {
				folds = {
					open = true,
				},
			},
		},
	},
}
```
Read also the Snacks docs for statuscolumn to see what other settings it has.

AtifChy:
> @AtifChy The statuscolumn is provided by Snacks now.
> 
> Did you enable the feature to show open folds? By default it's disabled. So, you should essentially do
> 
> ```lua
> return {
> 	{
> 		"folke/snacks.nvim",
> 		opts = {
> 			statuscolumn = {
> 				folds = {
> 					open = true,
> 				},
> 			},
> 		},
> 	},
> }
> ```
> 
> Read also the Snacks docs for statuscolumn to see what other settings it has.

yeah thanks. I was looking for the setting in LazyVim options. Also apply your fix it's working fine now.

PR #2924, opened 2024-04-04
Harpoon2: Use `add` to add the current buffer to the list instead of  `append`

rabinadk1:
Use `add` to add the current buffer to the list instead of `append.` The plugin warns about the deprecation of the latter in favor of the previous one.


dpetka2001:
Duplicate of #2920

rabinadk1:
> Duplicate of #2920

Closing then.

PR #2935, opened 2024-04-05
fix(gitsigns): update deprecated hunk nav mappings

lkhphuc:
https://github.com/lewis6991/gitsigns.nvim/commit/59bdc1851c7aba8a86ded87fe075ef6de499045c
Update deprecated functions, and add two keymaps to move to first and last hunk.


rubiin (approved):

folke:
Thanks!

PR #2944, opened 2024-04-07
feat(snippet): add friendly-snippets to native extra

lkhphuc:
I don't know if this is intentional but the extra for native snippet by default only expand the one return by LSP. 
For some language like Python, relies exclusively on friendly-snippets so I propose to add them here, especially if LazyVim intend to use native snippets as default.


folke:
Great, thanks!

PR #2945, opened 2024-04-07
fix(mini.move): <A-j> & <A-k> overridden by default global keybindings

ammon134:
The global keybindings <A-j> and <A-k> is still taking precedence over mini.move. This means that when one uses <A-j> or <A-k> with this setting they cannot use a single `u` to undo consecutive moves.

This fixes that.

### DIFF
--- a/lua/lazyvim/plugins/extras/editor/mini-move.lua
+++ b/lua/lazyvim/plugins/extras/editor/mini-move.lua
@@ -1,7 +1,7 @@
 return {
   {
     "echasnovski/mini.move",
-    event = "VeryLazy",
+    event = "LazyFile",
     opts = {},
   },
 }


dpetka2001:
This only works when you don't open Neovim with file arguments. If you open Neovim with file arguments like `nvim init.lua` it won't work. You could do something like
```lua
return {
  {
    "echasnovski/mini.move",
    event = "VeryLazy",
    config = function(_, opts)
      vim.schedule(function()
        require("mini.move").setup(opts)
      end)
    end,
  },
}
```
so that the internal mappings of `mini.move` always get set after `VeryLazy` (where `keymaps.lua` also gets loaded and are the last to be loaded I believe).

ammon134:
> This only works when you don't open Neovim with file arguments. If you open Neovim with file arguments like `nvim init.lua` it won't work. You could do something like
> 
> ```lua
> return {
>   {
>     "echasnovski/mini.move",
>     event = "VeryLazy",
>     config = function(_, opts)
>       vim.schedule(function()
>         require("mini.move").setup(opts)
>       end)
>     end,
>   },
> }
> ```
> 
> so that the internal mappings of `mini.move` always get set after `VeryLazy` (where `keymaps.lua` also gets loaded and are the last to be loaded I believe).

Thank you, indeed file arguments didn't work and your proposal seems to work for me, but it seems a bit... hacky? I'm not sure 😆  I'll let other chime in and ultimately the maintainer decide.

dpetka2001:
Sure, by no means I meant that my solution should be enforced. Leaving the decision to the maintainer is always wise :smile:

PR #2949, opened 2024-04-07
python: lsp_auto_brackets to set if brackets will be added

ShIRannx:
After v10.20.0, using python lsp `jedi_language_server` will got neat parentheses and relative hard to modify.
 image

### DIFF
--- a/lua/lazyvim/plugins/extras/lang/python.lua
+++ b/lua/lazyvim/plugins/extras/lang/python.lua
@@ -2,9 +2,11 @@ if lazyvim_docs then
   -- LSP Server to use for Python.
   -- Set to "basedpyright" to use basedpyright instead of pyright.
   vim.g.lazyvim_python_lsp = "pyright"
+  vim.g.lazyvim_python_lsp_auto_brackets = true
 end
 
 local lsp = vim.g.lazyvim_python_lsp or "pyright"
+local lsp_brackets = vim.g.lazyvim_python_lsp_auto_brackets == nil and true or false
 
 return {
   {
@@ -111,8 +113,10 @@ return {
   {
     "hrsh7th/nvim-cmp",
     opts = function(_, opts)
-      opts.auto_brackets = opts.auto_brackets or {}
-      table.insert(opts.auto_brackets, "python")
+      if lsp_brackets then
+        opts.auto_brackets = opts.auto_brackets or {}
+        table.insert(opts.auto_brackets, "python")
+      end
     end,
   },
 }


folke:
I just fixed this in another way. No need to configure anything. Thanks!

ShIRannx:
the problem persist after 10.25.0

folke:
You're right. Should be fixed now.
I fixed the `prev_char` code and also never add auto brackets for snippets.

ShIRannx:
the problem persist again now in a weird way. the print function is normal but other function, like sum, is not. 
image

folke:
You have two python lsps connected

ShIRannx:
image
i remove the jedi then auto completion gone. so there is only one lsp i believe.

folke:
Check with `:LspInfo`

ShIRannx:
image
with :LazyDev lsp

ShIRannx:
image
with :LspInfo

ShIRannx:
@folke  Hi, I hope you're doing well. I wanted to check in and see if you're still actively following the issue or if there's anything I can do to help move it forward. Thank you for your time.

crhf:
I recently started to use lazyvim. I have only ruff and jedi and am getting nested parentheses as well.

dpetka2001:
Can't you just remove `python` from `auto_brackets` if it messes with your setup? Something like 
```lua
  {
    "hrsh7th/nvim-cmp",
    opts = function(_, opts)
      for i, lang in ipairs(opts.auto_brackets) do
        if lang == "python" then
          table.remove(opts.auto_brackets, i)
        end
      end
    end,
  },
```
Haven't tested it, but I believe it should work.

crhf:
@dpetka2001 Works perfectly, thank you! I had no idea that cmp would do this. But I just now checked the default config here https://www.lazyvim.org/plugins/coding. It seems that `auto_brackets` is already empty. What could have added `python` to `auto_brackets`?

dpetka2001:
The [Python Extra](https://github.com/LazyVim/LazyVim/blob/78cf6ee024cbf6a17dc8406555eb131994cd8b63/lua/lazyvim/plugins/extras/lang/python.lua#L123-L129) as you can see here. It was just tested with `pyright` and `ruff-lsp` server, which are the defaults, as far as I know. But maybe some other LSP servers add parentheses by default and that causes the nested parentheses that you observe.

dpetka2001:
No idea if this is possible. Haven't looked into this myself and to be honest I don't understand the inner workings of `nvim-cmp` that well myself.

crhf:
Okay. Again, thanks for your help!

PR #2954, opened 2024-04-07
Fixed: Neotree isn't properly invoked when a directory is given by argument

mightywilly:
When a directory is given at launch through argument e.g. "nvim .", LazyVim opens Neo-tree side window instead of the dashboard. However, the instance isn't fully initialized: if one opens a file directly from Neotree, the tree buffer will be override by that file. And toggling Neotree will then result in error "failed to rename the buffer". Manually toggling Neotree can solve this issue.

I am still very new to Neovim and LazyVim, maybe there's a better solution :)

### DIFF
--- a/lua/lazyvim/plugins/editor.lua
+++ b/lua/lazyvim/plugins/editor.lua
@@ -45,6 +45,7 @@ return {
         local stat = vim.uv.fs_stat(vim.fn.argv(0))
         if stat and stat.type == "directory" then
           require("neo-tree")
+          vim.cmd("Neotree toggle")
         end
       end
     end,


dpetka2001:
I don't get what you're saying. For me it's opening correctly a file from neo-tree. Maybe you have something in your personal configuration that causes conflict? Screencast following with default LazyVim settings

[Screencast 2024-04-07 21:00:34.webm](https://github.com/LazyVim/LazyVim/assets/12776461/2f9472e0-aea9-434b-a467-c9ce58a67aae)

mightywilly:
> I don't get what you're saying. For me it's opening correctly a file from neo-tree. Maybe you have something in your personal configuration that causes conflict? Screencast following with default LazyVim settings
> Screencast.2024-04-07.21.00.34.webm

Thank you for the reply! I can confirm that the default configuration from LazyVim doesn't trigger this issue. I will research further more and keep this thread updated.

mightywilly:
Update: After disabling Oil plugin, Neotree works as expected without this modification.

PR #2956, opened 2024-04-08
feat(lang): add nix language support

izelnakri:
In the future we can add:
- dap / debugger support, if this is ever possible or needed.
- extend neotest, if this is ever needed, for most people I suspect not needed.
- linting, if this is ever needed.
- ~~maybe enhance nvim-cmp behavior with nixpkgs, rust lang extension crates.nvim does smt with nvim-cmp, other language extensions don't.~~

I think its already good to merge this PR, we can add these enhancements later, if they are ever needed.

### DIFF
--- a/lua/lazyvim/plugins/extras/lang/nix.lua
+++ b/lua/lazyvim/plugins/extras/lang/nix.lua
@@ -1,0 +1,35 @@
+return {
+  {
+    "nvim-treesitter/nvim-treesitter",
+    opts = function(_, opts)
+      opts.ensure_installed = opts.ensure_installed or {}
+      vim.list_extend(opts.ensure_installed, { "nix" })
+    end,
+  },
+  {
+    "williamboman/mason.nvim",
+    opts = function(_, opts)
+      vim.list_extend(opts.ensure_installed, {
+        "nixpkgs-fmt",
+        "nil",
+      })
+    end,
+  },
+  {
+    "neovim/nvim-lspconfig",
+    opts = {
+      servers = {
+        ["nil"] = {},
+      },
+    },
+  },
+  {
+    "stevearc/conform.nvim",
+    optional = true,
+    opts = {
+      formatters_by_ft = {
+        nix = { "nixfmt" },
+      },
+    },
+  },
+}
--- a/lua/lazyvim/plugins/extras/lang/nix.lua
+++ b/lua/lazyvim/plugins/extras/lang/nix.lua
@@ -7,18 +7,10 @@ return {
     end,
   },
   {
-    "williamboman/mason.nvim",
-    opts = function(_, opts)
-      vim.list_extend(opts.ensure_installed, {
-        "nil",
-      })
-    end,
-  },
-  {
     "neovim/nvim-lspconfig",
     opts = {
       servers = {
-        ["nil"] = {},
+        nil_ls = {},
       },
     },
   },
--- a/lua/lazyvim/plugins/extras/lang/nix.lua
+++ b/lua/lazyvim/plugins/extras/lang/nix.lua
@@ -10,7 +10,6 @@ return {
     "williamboman/mason.nvim",
     opts = function(_, opts)
       vim.list_extend(opts.ensure_installed, {
-        "nixpkgs-fmt",
         "nil",
       })
     end,
--- a/lua/lazyvim/plugins/extras/lang/nix.lua
+++ b/lua/lazyvim/plugins/extras/lang/nix.lua
@@ -1,4 +1,8 @@
 return {
+  recommended = {
+    ft = "nix",
+    root = "flake.nix",
+  },
   {
     "nvim-treesitter/nvim-treesitter",
     opts = function(_, opts)


lua/lazyvim/plugins/extras/lang/nix.lua
```diff
@@ -0,0 +1,35 @@
...
+        "nixpkgs-fmt",
+        "nil",
+      })
+    end,
+  },
+  {
+    "neovim/nvim-lspconfig",
+    opts = {
+      servers = {
+        ["nil"] = {},
```
rubiin, line 22: you can remove the double array syntax to just `nil`

rubiin (changes requested):

lua/lazyvim/plugins/extras/lang/nix.lua
```diff
@@ -0,0 +1,35 @@
...
+      opts.ensure_installed = opts.ensure_installed or {}
+      vim.list_extend(opts.ensure_installed, { "nix" })
+    end,
+  },
+  {
+    "williamboman/mason.nvim",
+    opts = function(_, opts)
+      vim.list_extend(opts.ensure_installed, {
+        "nixpkgs-fmt",
+        "nil",
```
shanesveller, line 14: Both of these concerns (formatter, LSP) have at least two successful, widespread community options and are highly subjective choices, particularly the formatter. I know a lot of folks really like `alejandra` for example and it produces _very_ different outcomes than `nixpkgs-fmt`. For flakes-based projects, IMO it's better to set `outputs.formatter.${system}` to the preferred package and not have an $EDITOR-wide selection at all.

There's also an open RFC attempting to implement a one-true-formatting-standard outcome:

https://github.com/NixOS/rfcs/pull/166

All that to say I don't think having this extra choose one particular option for these concerns is going to satisfy a great cross-section of Nix users.

lua/lazyvim/plugins/extras/lang/nix.lua
```diff
@@ -0,0 +1,35 @@
...
+        ["nil"] = {},
+      },
+    },
+  },
+  {
+    "stevearc/conform.nvim",
+    optional = true,
+    opts = {
+      formatters_by_ft = {
+        nix = { "nixfmt" },
```
shanesveller, line 31: As mentioned in my other comment, there are a lot of Nix formatters with no overwhelmingly dominant choice if you surveyed the community. More immediately, [nixfmt](https://github.com/NixOS/nixfmt) is a totally different project and binary than [nixpkgs-fmt](https://github.com/nix-community/nixpkgs-fmt) that you defined for Mason above.

shanesveller (commented):
All of my remarks are just bystander-to-bystander comments from someone who's been using Nix for around 8 years now but with no authority in this project or either ecosystem. **My TLDR is that I think the Nix ecosystem is so fragmented/factional/tribal that it would be difficult to have one single LazyVim extra with hardcoded choices successfully be pertinent to many established Nix users simultaneously.** If the LazyVim project wants to bless one single POV among that diversity and make it into an extra, in order to have _something_, that's of course their prerogative.

> extend neotest, if this is ever needed, for most people I suspect not needed.

I'm not aware of any broadly-adopted conventions or libraries for Nixlang unit testing within the community, which would be a prerequisite for Neotest support. I'm not omniscient, but I have looked for this in the past and have only found tiny projects attempting to solve for this with negligible adoption outside of the original authors.

> linting, if this is ever needed.

This concern does have [several community options](https://discourse.nixos.org/t/list-of-nix-linters/19279), I've personally worked with `nixpkgs-lint`, `statix` and `deadnix`, and continue to use the latter two regularly.

> maybe enhance nvim-cmp behavior with nixpkgs, rust lang extension crates.nvim does smt with nvim-cmp

Cargo/crates.io, being a relatively traditional package manager registry, lets you target your dependencies independently so long as as they self-describe as [cargo-flavored-SemVer](https://doc.rust-lang.org/cargo/reference/semver.html)-compatible with each other. In contrast, Nixpkgs is a large, holistic set of packages that are built and tested together by Hydra, much closer to a Stackage snapshot in concept.

If you need to mingle content from two revisions of Nixpkgs, such as mixing stable/unstable branches, you'll have to use multiple imports (or flake inputs) in the Nix project in question. Because of the syntax complexity and "many ways to skin this cat" of that concern, the specific functionality for Cargo.toml via [crates.nvim](https://github.com/Saecki/crates.nvim) (autocompleting package names, listing available package versions) doesn't have a very natural parallel in the Nix world.

Autocomplete of attribute paths, NixOS options, etc. while editing Nix syntax are already possible with one of the competing LSP options such as `nil`/`nixd`.

lua/lazyvim/plugins/extras/lang/nix.lua
izelnakri: My editor gives me this:
     └╴  <keyword> cannot be used as name. Lua Syntax Check. (keyword) [24, 8]

Resolving this issue now, please let me know if you'd like it to be reopened.

lua/lazyvim/plugins/extras/lang/nix.lua
izelnakri: I think having one formatter installed other than having no formatter installed is a better UX, we can remove it *when*/*if* there is a standard in the fragmented ecosystem. Users & non-NixOS home-manager users can always customize and remove these configurations. Please let me know if you'd like this issue to be reopened, I'm closing it now.

lua/lazyvim/plugins/extras/lang/nix.lua
dpetka2001, line 21: From what I'm seeing in [lspconfig docs](https://github.com/neovim/nvim-lspconfig/blob/master/doc/server_configurations.md#nil_ls), the name of the server is `nil_ls`. That also means that you don't need to specify it in Mason any more, since it will be installed automatically when declared in `nvim-lspconfig` configuration.

izelnakri:
@shanesveller thanks for the info & in-depth consideration of the changes! I agree that fragmentation isn't great. I think we should provide one blessed option at least installed and configured, users can change it if they want, because it is generally a better experience.

One exception to this is `nixpkgs-fmt` issue, as you described the reasonings I left that mason download out. Let me know if you'd like me to do any more changes.

lua/lazyvim/plugins/extras/lang/nix.lua
izelnakri: Hi @dpetka2001 , could you please provide me the specific code/documentation in `neovim/nvim-lspconfig` in where this happens? I don't want to remove it if I'm not sure, the link you've provided wasn't clear to me.

I'm not sure if changing this to `nil_ls` would work, I see that being the 2nd optional definition in `mason`, but not sure if `nvim-lspconfig` would pick this up. According to lspconfig docs:

```
Default values:

cmd :
{ "nil" }
```
Source:  https://github.com/neovim/nvim-lspconfig/blob/master/doc/server_configurations.md#nil_ls

Let me know if you've manually checked this, if this is verified we can change the syntax to nix_ls instead otherwise I'd like to keep it like this instead.

dpetka2001: The link to the docs that I provided, specifically say 
> Snippet to enable the language server:
`require'lspconfig'.nil_ls.setup{}`

`"nil"` is the name of the command that will invoke the language server. But in `nvim-lspconfig`, when you're setting up a server, you're passing the name of the language server in the `require`, not the `"cmd"`. You can try it out locally to see how it behaves.

izelnakri: :+1: I just manually tried it & it works, doing the change now. Thanks for pointing it out.

dpetka2001: I'm sorry, but I did not mean that you should remove the `nvim-lspconfig` spec. Maybe I wasn't clear enough. I meant you should do the following
```lua
  {
    "neovim/nvim-lspconfig",
    opts = {
      servers = {
        nil_ls = {},
      },
    },
  },
```
and remove the `mason.nvim` spec completely, since all it does at the moment is download the the LSP server, but that is not necessary since every server you specify in `nvim-lspconfig.opts.servers` will be automatically installed.

[Here](https://github.com/LazyVim/LazyVim/blob/bb36f71b77d8e15788a5b62c82a1c9ec7b209e49/lua/lazyvim/plugins/lsp/init.lua#L194-L216) you can find how LazyVim sets up the LSP servers. It checks `nvim-lspconfig.opts.servers` and adds them to `ensure_installed` and passes that table to `mason-lspconfig`, which takes care of the installation and the automatic default handlers for LSP servers.

lua/lazyvim/plugins/extras/lang/nix.lua
```diff
@@ -0,0 +1,26 @@
...
+    end,
+  },
+  {
+    "williamboman/mason.nvim",
+    opts = function(_, opts)
+      vim.list_extend(opts.ensure_installed, {
+        "nil",
+      })
+    end,
+  },
```
dpetka2001, line 16:
```suggestion
  {
    "neovim/nvim-lspconfig",
    opts = {
      servers = {
        nil_ls = {},
      },
    },
  },
```
This is my suggestion, so that there are no misunderstandings.

dpetka2001:
@izelnakri The `nil_ls` is not needed in Mason. I suggest you remove the entire Mason spec if all it's doing is install `nil_ls` LSP. If you specify a server in `nvim-lspconfig` it will be automatically installed without the need to also declare it in Mason's `ensure_installed`. The link I pointed in my previous [comment](https://github.com/LazyVim/LazyVim/pull/2956#discussion_r1566851809) describes how LazyVim sets LSP servers up.

folke:
Thanks!

PR #2959, opened 2024-04-08
feat(git): Change Commit Log to use Lazygit

georgeguimaraes:
I found that using Lazygit Commit Log is much more helpful. You can have all the Lazygit usefulness here.

I'm using it to browse the last commits, and then browse again the files that were changed in those commits, and using the `e` shortcut to open the file in the same Neovim instance.

### DIFF
--- a/lua/lazyvim/plugins/editor.lua
+++ b/lua/lazyvim/plugins/editor.lua
@@ -158,7 +158,6 @@ return {
       { "<leader>fr", "<cmd>Telescope oldfiles<cr>", desc = "Recent" },
       { "<leader>fR", LazyVim.telescope("oldfiles", { cwd = vim.uv.cwd() }), desc = "Recent (cwd)" },
       -- git
-      { "<leader>gc", "<cmd>Telescope git_commits<CR>", desc = "Commits" },
       { "<leader>gs", "<cmd>Telescope git_status<CR>", desc = "Status" },
       -- search
       { '<leader>s"', "<cmd>Telescope registers<cr>", desc = "Registers" },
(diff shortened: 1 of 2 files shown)


dpetka2001:
Really nice change. Hope it gets merged!! :smile:

rubiin (approved):

rubiin:
I am using lazygit  too so this makes sense to me as well to use it for most things :)

lua/lazyvim/plugins/editor.lua
```diff
@@ -158,7 +158,6 @@ return {
       { "<leader>fr", "<cmd>Telescope oldfiles<cr>", desc = "Recent" },
       { "<leader>fR", LazyVim.telescope("oldfiles", { cwd = vim.uv.cwd() }), desc = "Recent (cwd)" },
       -- git
-      { "<leader>gc", "<cmd>Telescope git_commits<CR>", desc = "Commits" },
```
jyuan0, line 161: If you'd like to not remove a mapping, `<leader>gl` is free and makes sense to me.

dpetka2001: But they do the same thing. They show the commits. It doesn't make sense to have 2 different mappings for 2 plugins that do the same thing essentially (to me personally at least).

rubiin: > But they do the same thing. They show the commits. It doesn't make sense to have 2 different mappings for 2 plugins that do the same thing essentially (to me personally at least).

I agree.

folke:
I kept the original `<leader>gc` and added a `<leader>gl` for LazyGit Log

PR #2974, opened 2024-04-12
feat(extras.lang): added jenkinsfile support

moisesmorillo:


idelice:
@folke  any comment on this?

Conarius:
I think it would be more practical to have a groovy lang extra and then a jenkinsfile in the dot files extra.

PR #2976, opened 2024-04-13
fix(harpoon2): replace deprecated append() with add()

mrflos:
lazyvim is showing a popup mentioning that append is deprecated while harpooning a file.  
This is the new function to use to fix this.


mrflos:
oup duplicate of #2920  , closing.

PR #2983, opened 2024-04-14
fix(extras.lang.typescript): support `node` debug type along with `pwa-node`

80avin:
`"pwa-"` prefixed debug types are deprecated in the debug adapter. ref: https://github.com/microsoft/vscode-js-debug/pull/1305

Also, vscode generates launch.json files with `"type": "node"` by default now, and warns if we use `"pwa-node"`

This PR adds support for `"node"` dap adapter/debugtype which maps to `"pwa-node"`. This behaviour is same as of the upstream debug adapter's supporting extension


folke:
Thanks!

PR #2984, opened 2024-04-14
fix(edgy): neo-tree doesn't open when leader isn't space

kylelaker:
The default mapping for this is to use `<leader>` which edgy should
respect. This ensures that neo-tree still opens even when leader has
been remapped. Otherwise, the top neo-tree window just stays in a
loading state.

Unfortunately, it seems the `<leader>` isn't a supported keycode for
`nvim_input` (or at least it isn't listed in `:help keycodes` and it
doesn't seem to work). Getting the leader first, then passing it to
`nvim_input` seems to clear this up.

### DIFF
--- a/lua/lazyvim/plugins/extras/ui/edgy.lua
+++ b/lua/lazyvim/plugins/extras/ui/edgy.lua
@@ -61,7 +61,8 @@ return {
             end,
             pinned = true,
             open = function()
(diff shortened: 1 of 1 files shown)


folke:
Not sure why that `open` was implemented like this. Just changed it to use the `neo-tree` API instead.
 Thanks

PR #2986, opened 2024-04-15
fix(keymaps): delete lazygit keymap when using gitui

rubiin:
Currently even when using gitui, lazy gits current file history keymap is still there. This deletes the keymap


folke:
Thanks!

PR #2987, opened 2024-04-15
feat(dot): add more filetypes

rubiin:
Adds dotenv files support for file names like
prod.env
.env
.env.xyz

Also adds configurationfiletypes rofi, wofi and vimfmrc


folke:
thanks!

PR #2989, opened 2024-04-15
fix(trouble_v3): show error only when `trouble_v3` is enabled

### LINKED ISSUE #2895: feature: allow disabling Trouble V3 Error messages
### Did you check the docs?

I want to use the Trouble v3 extra, but still prefer Outline for Trouble's document symbols (mostly for the out of the box support for JSX tags).

Functionally it all seems to work just fine, but I get an error message because of this:

https://github.com/LazyVim/LazyVim/blob/97480dc5d2dbb717b45a351e0b04835f138a9094/lua/lazyvim/plugins/extras/editor/trouble-v3.lua#L3-L13

Warning _every_ time seems a bit intrusive, and I think there should be an option to disable the warning.

Showing the warning just once or once in a while would also be okay, though I think an option to explicitly disable it is better.

### Additional context

_No response_

dpetka2001:
Fixes #2895

### DIFF
--- a/lua/lazyvim/plugins/extras/editor/trouble-v3.lua
+++ b/lua/lazyvim/plugins/extras/editor/trouble-v3.lua
@@ -1,14 +1,16 @@
 local Config = require("lazyvim.config")
 
-for _, other in ipairs({ "aerial", "outline" }) do
-  local extra = "lazyvim.plugins.extras.editor." .. other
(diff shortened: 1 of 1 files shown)


rubiin (approved):

folke:
thanks!

PR #2994, opened 2024-04-15
fix(harpoon): replace deprecated append call with add

XavierChanth:
`list():append()` function was deprecated in favour of `list():add()`, they have the same signature otherwise.

See https://github.com/ThePrimeagen/harpoon/blob/0378a6c428a0bed6a2781d459d7943843f374bce/lua/harpoon/list.lua#L102-L107


dpetka2001:
Duplicate of #2920

folke:
Merged that other PR. Thanks

PR #2995, opened 2024-04-15
feat(typescript): configure inlayHint

radvil:
since LazyVim have toggle inlayhint keymap, it make sense to configure them for each extra languages provided. In this commit I just add the configuration for typescript, since IDK how to configure the rest of lang extras that was provided.


mirsella:
+1 i also think inlay hints should be configured for supported languages, thanks for the contribution !

folke:
I've turned off some inlay hint configs and merged. Thanks!

PR #2999, opened 2024-04-17
changed append with add in harpoon2, because append was deprecates.

sayasuhendra:
image

changed append with add in harpoon2, because append was deprecates.


dpetka2001:
Duplicate of #2920

PR #3016, opened 2024-04-19
feat(python): use new `ruff` instead of `ruff_lsp`

lkhphuc:
Use the new builtin rust-based `ruff server` instead of the python-based `ruff_lsp` wrapper.
Ruff_lsp is not deprecated (yet) but the new rust-based server is faster and is (almost?) feature complete.
Though it's still in beta, it has been highlighted in the latest [ruff release](https://github.com/astral-sh/ruff/releases/tag/v0.4.0) and it has worked well for me for a while.

### DIFF
--- a/lua/lazyvim/plugins/extras/lang/python.lua
+++ b/lua/lazyvim/plugins/extras/lang/python.lua
@@ -28,7 +28,7 @@ return {
         [lsp] = {
           enabled = true,
         },
-        ruff_lsp = {
+        ruff = {
           keys = {
             {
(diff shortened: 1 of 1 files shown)


folke:
Thanks!

surmish:
Is this a replacement for all three: black/flake8/pyright (or basedpyright)?

tekulvw:
After updating I've lost the ability for ruff to correctly format imports and for ruff to complain about non formatted import block `I001`

PR #2883, opened 2024-03-30
feat(lang): add toml

gacallea:
adding toml support to PR on LazyVim

- tree-sitter: toml
- mason: taplo
- conform: taplo
- nvim-lint: taplo

PR #3022, opened 2024-04-20
feat(lang): add toml

gacallea:
adding toml support to PR on LazyVim

tree-sitter: toml
mason: taplo
conform: taplo
nvim-lint: taplo

originally here: https://github.com/LazyVim/LazyVim/pull/2883

### DIFF
--- a/lua/lazyvim/plugins/extras/lang/toml.lua
+++ b/lua/lazyvim/plugins/extras/lang/toml.lua
@@ -1,0 +1,30 @@
+return {
+  recommended = function()
+    return LazyVim.extras.wants({
+      ft = "toml",
+    })
+  end,
+  {
+    "nvim-treesitter/nvim-treesitter",
+    opts = function(_, opts)
+      if type(opts.ensure_installed) == "table" then
+        vim.list_extend(opts.ensure_installed, { "toml" })
+      end
+    end,
+  },
+  {
+    "williamboman/mason.nvim",
+    opts = function(_, opts)
+      opts.ensure_installed = opts.ensure_installed or {}
+      vim.list_extend(opts.ensure_installed, { "taplo" })
+    end,
+  },
+  {
+    "neovim/nvim-lspconfig",
+    opts = {
+      servers = {
+        taplo = {},
+      },
+    },
+  },
+}


lua/lazyvim/plugins/extras/lang/toml.lua
```diff
@@ -0,0 +1,34 @@
...
+    "stevearc/conform.nvim",
+    opts = {
+      formatters_by_ft = {
+        toml = { "taplo" },
+      },
+    },
+  },
+  {
+    "neovim/nvim-lspconfig",
+    ft = { "toml" },
```
dpetka2001, line 27: Why add `ft` to `nvim-lspconfig`? The LSP servers already have filetypes and patterns to search for to be loaded upon. Also, since the Mason spec only installs `taplo` LSP server, you can remove it as a whole, since every LSP server that gets defined in `nvim-lspconfig` will get automatically installed. Of course, this is just a suggestion from my part for the Mason spec and you're free to keep it around for reasons of clarity (but it doesn't really do anything).

gacallea: hi @dpetka2001 , apologies. I forgot to remove that. I use ft everywhere in my own configs.

gacallea: I removed ft and force pushed the changes. Regarding Mason, I like to have those for clarity in my configs.

dpetka2001: Please no need to apologize for anything. I'm just a regular user like you. I really don't feel comfortable. I just pointed out something that isn't needed and just a suggestion for Mason.

dpetka2001: You might want to do the same for your `elm` PR, since you already made the change here.

rubiin (approved):

rubiin:
Just noticed nvim-lint is not configured with taplo, but I guess we can use the LSP linting that taplo provides

gacallea:
> Just noticed nvim-lint is not configured with taplo, but I guess we can use the LSP linting that taplo provides

I could add it, no biggie. I may have missed that because I use toml with Zola CMS and so far so good. Other people my need  that. LMK.

rubiin:
yeah . I just checked nvim-lint but they dont support taplo by default. So my intuition says you have to use the custom linter settings .

gacallea:
> yeah . I just checked nvim-lint but they dont support taplo by default. So my intuition says you have to use the custom linter settings .

from a. quick search, it seems that LSP is the way to go: https://github.com/mfussenegger/nvim-lint/issues/247

rubiin:
yesh i did stumble upon that. lets ignore adding it

folke:
Doesnt taplo as lsp support formatting? If so the conform.nvim is not needed?

gacallea:
> Doesnt taplo as lsp support formatting? If so the conform.nvim is not needed?

from the docs it seems to take care of formatting as well: https://taplo.tamasfe.dev/cli/usage/formatting.html -- I added conform to make sure in case. not sure, but it could be removed? does it need extra config in your opinion?

dpetka2001:
There is not a formatter called `toml` in https://github.com/stevearc/conform.nvim/tree/master/lua/conform/formatters. And conform does fall back to lsp formatting if no formatter is available. `:LazyFormatInfo` and `:ConformInfo` should only show the LSP as the formatter, right? In that case the `conform` spec shouldn't be needed here.

gacallea:
> There is not a formatter called `toml` in https://github.com/stevearc/conform.nvim/tree/master/lua/conform/formatters. And conform does fall back to lsp formatting if no formatter is available. `:LazyFormatInfo` and `:ConformInfo` should only show the LSP as the formatter, right? In that case the `conform` spec shouldn't be needed here.
> 
> Not able to test it, so could you confirm what the above 2 commands show @gacallea ?

the formatter in my file is `taplo`, for the `toml` filetype :)

here's two screenies for the commands output with my configuration:

Screenshot 2024-05-19 at 6 40 02 PM

Screenshot 2024-05-19 at 6 40 09 PM

I hope this helps

dpetka2001:
If you remove the conform spec locally doesn't it work correctly?

gacallea:
> If you remove the conform spec locally doesn't it work correctly? I believe you shouldn't observe any different behavior.

I confirm this :) 

my bad. Will force push changes :)

dpetka2001:
Just make sure that formatting is enabled by default in taplo LSP and it doesn't need any additional configuration from the LSP server side.

gacallea:
> Just make sure that formatting is enabled by default in taplo LSP and it doesn't need any additional configuration from the LSP server side.

I did edit the same file by changing a few spoace here and there. it works.

folke:
Can you add the code below to the top? This way the extra will be shown as recommended for toml files.

```lua
  recommended = function()
    return LazyVim.extras.wants({
      ft = "toml",
    })
  end,
```

You can verify that it works by opening `:LazyExtras` and it should have a star next to the extra when opened from a toml file

lua/lazyvim/plugins/extras/lang/toml.lua
```diff
@@ -0,0 +1,31 @@
+recommended = function()
```
folke, line 1: This should be inside the returned table below :)

gacallea: oopsie.. I hope I got it right this time

folke:
Thanks!

gacallea:
> Thanks!

thank you all for lazyvim! honored to contribute.

rubiin:
image
image


Dont know why but getting this issue. LazyVim plugin is upto date

PR #3047, opened 2024-04-23
update the attribute name for process ID

xy-lin:
The attribute name for picked process id when attaching the debugger is `pid`, not `processId`.
https://github.com/vadimcn/codelldb/blob/master/MANUAL.md


rubiin (approved):

folke:
Thanks!

PR #3048, opened 2024-04-24
feat(inlay_hints): add notification for toggling

konosubakonoakua:


lua/lazyvim/util/toggle.lua
```diff
@@ -69,6 +69,12 @@ function M.inlay_hints(buf, value)
     end
     ih.enable(value, { bufnr = buf })
   end
+  
+  if value then
+    LazyVim.info("Enabled inlay_hints", { title = "inlay_hints" })
```
rubiin, line 74: Maybe make notification title to `Inlay hints` instead

dpetka2001: In my personal opinion, I don't believe this is a good change, because in every buffer you get the notification about hints being enabled and that's not good user experience. Ideally the notification would pop up only on toggle. This could be achieved if the notification is attached directly to the keymap. Something like 
```lua
if vim.lsp.buf.inlay_hint or vim.lsp.inlay_hint then
  map( "n", "<leader>uh", function()
    LazyVim.toggle.inlay_hints()
    if vim.lsp.inlay_hint.is_enabled() then
      LazyVim.info("Enabled inlay hints", {title = "Inlay hints"})
    else
      LazyVim.warn("Disabled inlay hints", {title = "Inlay hints"})
  end
end, { desc = "Toggle Inlay Hints" })
```
should work better I think.

rubiin: yeah i agree

dpetka2001: Oops, I just realized I did not reply to OP but you @rubiin. Sorry about that :stuck_out_tongue:

konosubakonoakua:
```lua
  {
    "neovim/nvim-lspconfig",
    opts = function(_, opts)
      opts.inlay_hints = {
        enabled = true,
      }
      return opts
    end
  },
```
with `inlay_hints.enabled = true` for `nvim-lspconfig`, I also encountered what you say.
Because:
```lua
      -- inlay hints
      if opts.inlay_hints.enabled then
        LazyVim.lsp.on_attach(function(client, buffer)
          if client.supports_method("textDocument/inlayHint") then
            LazyVim.toggle.inlay_hints(buffer, true)
          end
        end)
      end
```
It's really bad for notifying inlay_hints for every buffer.
Since the lspconfig automatically manage the status of inlay_hints, we surely should only notify the manual toggling trigged from our keymap, I'll close the pr, thanks for your reviewing and testing😆 @rubiin @dpetka2001

dpetka2001: @konosubakonoakua Just make a new one with the changes I suggested directly in the keymap itself and you should be good to go :smiley:

PR #3052, opened 2024-04-24
fix(dot): install when list is empty

matadaniel:
Assign `opts.ensure_installed` first because if it were to be nil, I believe that it would not be assigned nor extended with shellcheck.

Remove shfmt since LazyVim sets it up by default.

### DIFF
--- a/lua/lazyvim/plugins/extras/util/dot.lua
+++ b/lua/lazyvim/plugins/extras/util/dot.lua
@@ -18,10 +18,8 @@ return {
   {
     "williamboman/mason.nvim",
     opts = function(_, opts)
-      vim.list_extend(opts.ensure_installed or {}, {
-        "shfmt",
-        "shellcheck",
-      })
+      opts.ensure_installed = opts.ensure_installed or {}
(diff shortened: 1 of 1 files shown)


rubiin (changes requested):
shfmt is removed , might wanna add it back

dpetka2001:
`shfmt` is already installed [here by default](https://github.com/LazyVim/LazyVim/blob/f086bcde253c29be9a2b9c90b413a516f5d5a3b2/lua/lazyvim/plugins/lsp/init.lua#L238).

rubiin:
Oh cool will approve the pr then

rubiin (approved):

rubiin:
@amaanq Thanks for the headsup. By approval, it means it  passes checks . Since folke is a single collaborator with write access, without other people reviewing prs it would become cumbersome for him to manage all these prs.

folke:
I do agree with @amaanq RE the approvals.
Your intentions and enthusiasm are good, but it's indeed pretty annoying to get all those additional notifications :)

folke:
Thanks!

PR #3057, opened 2024-04-25
Revert "feat(python): use new `ruff` instead of `ruff_lsp`"

arafune:
Reverts LazyVim/LazyVim#3016
As described in ruff web page, ruff server is alpha.

I believe this change is too early. And its not only too early but there are real problems.  

For example, after updating this, command line ruff complains about non formatted import block "I001".

### DIFF
--- a/lua/lazyvim/plugins/extras/lang/python.lua
+++ b/lua/lazyvim/plugins/extras/lang/python.lua
@@ -28,7 +28,7 @@ return {
         [lsp] = {
           enabled = true,
         },
-        ruff = {
+        ruff_lsp = {
           keys = {
             {
               "<leader>co",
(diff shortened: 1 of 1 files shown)


fraenkel:
The above has been reported but not fixed yet.
https://github.com/astral-sh/ruff/issues/11207

akthe-at:
> The above has been reported but not fixed yet. [astral-sh/ruff#11207](https://github.com/astral-sh/ruff/issues/11207)

actually it has been fixed, and the PR mentioned in my issue looks to be approved and is scheduled for merge. This fix + the honoring of a config toml for the ruff server should push it pretty far in terms of usability.

folke:
Was fixed by #3060

PR #3060, opened 2024-04-25
fix(python): make both `ruff` and `ruff_lsp` available to user

dpetka2001:
Provide global variable to choose between `ruff` and `ruff_lsp` in accordance to `pyright`/`basedpyright`.

Pinging @arafune for #3057. Are you ok with something like this, so that users can decide which one to use? 
By default, `ruff_lsp` is used unless the user explicitly chooses `ruff` instead. If you would like to incorporate these changes
to your PR yourself, I don't have any problem closing this.

### DIFF
--- a/lua/lazyvim/plugins/extras/lang/python.lua
+++ b/lua/lazyvim/plugins/extras/lang/python.lua
@@ -2,9 +2,11 @@ if lazyvim_docs then
   -- LSP Server to use for Python.
   -- Set to "basedpyright" to use basedpyright instead of pyright.
   vim.g.lazyvim_python_lsp = "pyright"
+  vim.g.lazyvim_ruff = "ruff_lsp"
 end
 
 local lsp = vim.g.lazyvim_python_lsp or "pyright"
+local ruff = vim.g.lazyvim_ruff or "ruff_lsp"
 
 return {
   {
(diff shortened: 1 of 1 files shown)


tekulvw:
Bump, anything blocking this from merging?

dpetka2001:
When the maintainer has some time, he will surely look into this. No need to rush. Please be patient.

folke:
refactored it a bit and merged. Thanks!

btegs:
I know I am late to this @folke but I am looking at the documentation at https://www.lazyvim.org/extras/lang/python

If someone wants to use ruff for everything and not use pyright, do they switch out the pyright option in addition to LSP option? For example:

```
vim.g.lazyvim_python_lsp = "ruff"
vim.g.lazyvim_python_ruff = "ruff_lsp"
```

Documentation is unclear

dpetka2001:
Just use `vim.g.lazyvim_python_ruff = "ruff"` and make an lspconfig spec in your configuration which disables pyright explicitly, like so
```lua
  {
    "neovim/nvim-lspconfig",
    opts = {
      servers = {
        pyright = {
          enabled = false, -- disable server
          mason = false, -- don't install with mason
        },
      },
    },
  },
```

PR #3061, opened 2024-04-25
fix(extras): Rust-Analyzer cargo option

imzhongqi:
I saw the official manual said this. I thought our configuration was wrong. I fixed it.    
rust-analyzer manual:  
https://rust-analyzer.github.io/manual.html#nvim-lsp

### DIFF
--- a/lua/lazyvim/plugins/extras/lang/rust.lua
+++ b/lua/lazyvim/plugins/extras/lang/rust.lua
@@ -60,7 +60,7 @@ return {
             cargo = {
               allFeatures = true,
               loadOutDirsFromCheck = true,
-              runBuildScripts = true,
+              buildScripts = true,
             },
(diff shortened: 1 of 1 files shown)


dpetka2001:
I believe this should be `buildScripts.enable = true`. Reference [here](https://github.com/rust-lang/rust-analyzer/blob/56bee2ddafa6177b19c631eedc88d43366553223/crates/rust-analyzer/src/config/patch_old_style.rs#L43). We could either leave it as is (I don't know if the old style will get deprecated any time soon) or use the new syntax.

dpetka2001:
Yes it should be the table syntax as you noted, I just wrote it like that for abbreviation.

imzhongqi:
I updated the commit, but I didn’t look carefully. It’s a table.

folke:
Thanks!

PR #3062, opened 2024-04-25
feat(lualine): make path trimming configurable by user

### LINKED ISSUE #3059: feature: configurable lualine path trimming
### Did you check the docs?

In the lualine configuration, there is this snippet https://github.com/LazyVim/LazyVim/blob/main/lua/lazyvim/util/lualine.lua#L103-L105 which breaks down the displayed relative path from the current working directory if it is composed by more than three "parts", by keeping the first one and the last two. 
In my humble opinion, this is not an ideal behaviour especially with codebases that have a predictable structure. In Rust projects, this often just results in seeing the path `crates/.../src/lib.rs` or similar in the status line which doesn't give any real help in guessing where I am.

Ideally, I'd like trimming the path only if it reaches a certain configurable length or if there is no window space available. Completely removing those three lines for me is a good option too, but it's not probably what everyone would like.

Right now, I'm removing those three lines locally.

### Additional context

An example of the unwanted behaviour of lualine on Rust projects
image

dpetka2001:
Closes #3059

### DIFF
--- a/lua/lazyvim/util/lualine.lua
+++ b/lua/lazyvim/util/lualine.lua
@@ -79,6 +79,7 @@ function M.pretty_path(opts)
     directory_hl = "",
     filename_hl = "Bold",
     modified_sign = "",
+    length = 3,
   }, opts or {})
 
   return function(self)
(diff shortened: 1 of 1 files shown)


folke:
Thanks!

PR #3085, opened 2024-04-30
Say "local map =" in the comment under DO NOT USE THIS!

michaelbutleridt:
Really minor change.

Make the warning more clear on what "THIS" is -- LazyVim.safe_keymap_set.  Because when I saw this, it wasn't clear to me what I need to do.

I saw `local map = vim.keymap.set` in someone else's config.

### DIFF
--- a/lua/lazyvim/config/keymaps.lua
+++ b/lua/lazyvim/config/keymaps.lua
@@ -1,7 +1,7 @@
 -- This file is automatically loaded by lazyvim.config.init
 
--- DO NOT USE THIS IN YOU OWN CONFIG!!
--- use `local map = vim.keymap.set` instead
+-- DO NOT USE `LazyVim.safe_keymap_set` IN YOU OWN CONFIG!!
+-- use `vim.keymap.set` instead
 local map = LazyVim.safe_keymap_set
 
 -- better up/down
--- a/lua/lazyvim/config/keymaps.lua
+++ b/lua/lazyvim/config/keymaps.lua
@@ -1,7 +1,7 @@
 -- This file is automatically loaded by lazyvim.config.init
 
 -- DO NOT USE THIS IN YOU OWN CONFIG!!
--- use `vim.keymap.set` instead
+-- use `local map = vim.keymap.set` instead
 local map = LazyVim.safe_keymap_set
 
 -- better up/down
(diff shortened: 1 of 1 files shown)


dpetka2001:
Actually it's not optimal to include `local map`, because that is only assigning `vim.keymap.set` to a local variable map. But some people might use `vim.keymap.set` directly in their configuration (and that is also correct), so it would potentially become confusing that they have to assign it to a local variable.

michaelbutleridt:
but the rest of this file has stuff like

```
map({ "n", "x" }, "j", "v:count == 0 ? 'gj' : 'j'", { expr = true, silent = true })
map({ "n", "x" }, "<Down>", "v:count == 0 ? 'gj' : 'j'", { expr = true, silent = true })
map({ "n", "x" }, "k", "v:count == 0 ? 'gk' : 'k'", { expr = true, silent = true })
```
 
Only saying "use `vim.keymap.set` instead" isn't clear on what is the proper path forward. Maybe it's only me though.   

how about:

> Use `vim.keymap.set` instead of `LazyVim.safe_keymap_set`

Would that be better?  

Or where it says "DO NOT USE THIS IN YOUR OWN CONFIG!" it could say

> DO NOT USE `LazyVim.safe_keymap_set` IN YOUR OWN CONFIG! 
> Use `vim.keymap.set` instead

dpetka2001:
The correct way to do your keymaps with only reading the Neovim docs is to use `vim.keymap.set` directly (and `LazyVim.safe_keymap_set` also uses `vim.keymap.set` under the hood with some modifications to take into account the `LazyKeys` that `lazy.nvim` provides, but that is not necessary for the end user). So in your `/lua/config/keymaps.lua` you would have
```lua
vim.keymap.set({ "n", "x" }, "j", "v:count == 0 ? 'gj' : 'j'", { expr = true, silent = true })
vim.keymap.set({ "n", "x" }, "<Down>", "v:count == 0 ? 'gj' : 'j'", { expr = true, silent = true })
vim.keymap.set({ "n", "x" }, "k", "v:count == 0 ? 'gk' : 'k'", { expr = true, silent = true })
```
Simple as that. 

You can also assign `vim.keymap.set` to a local variable to avoid writing so many words and you do that with something like `local map = vim.keymap.set` and then use that to define your mappings.  But that is more confusing in my opinion for users who are not familiar with programming. So for such users, it would be preferable to directly use `vim.keymap.set` in their config until they become more familiar with Lua and know what they're doing.

michaelbutleridt:
thanks. i updated the comment. it's even simpler now

folke:
thanks!

Issue #3090, opened 2024-05-02
NeoTree & Telescope open in wrong CWD


juridiener:
### Did you check docs and existing issues?

- [X] I have read all the LazyVim docs
- [X] I have searched the existing issues of LazyVim
- [X] I have searched the existing issues of plugins related to this issue

### Neovim version (nvim -v)

NVIM v0.10.0-dev-2982+gea1c9f60e

### Operating system/version

MacoS

### Describe the bug

Hello, if i open a spesific directory it falls back always to the root kinda. So i opened in my terminal nvim joomla. The first time the file browser shows the right directory, then i press leader e or E to close and reopen, then it shows me everything above too  and in telescope i got not the find files picker but instead the git files. I don't know if I'm doing something wrong or is this the normal behavoir?? 
<img width="446" alt="Screenshot 2024-05-02 at 10 50 17" src="https://github.com/LazyVim/LazyVim/assets/92093874/72508914-2efd-4d13-bbde-7b896568bb60">
<img width="1591" alt="Screenshot 2024-05-02 at 10 50 50" src="https://github.com/LazyVim/LazyVim/assets/92093874/f3eec954-2fa6-4f0a-84c6-1fd312e4feac">

### Steps To Reproduce

Open a spesific directory with nvim folder,
press leader e twice
press leader ff

### Expected Behavior

Open neo-tree only in that folder as cwd
set that folder as cwd for telescope

### Repro

```Lua
-- DO NOT change the paths and don't remove the colorscheme
local root = vim.fn.fnamemodify("./.repro", ":p")

-- set stdpaths to use .repro
for _, name in ipairs({ "config", "data", "state", "cache" }) do
  vim.env[("XDG_%s_HOME"):format(name:upper())] = root .. "/" .. name
end

-- bootstrap lazy
local lazypath = root .. "/plugins/lazy.nvim"
if not vim.loop.fs_stat(lazypath) then
  vim.fn.system({ "git", "clone", "--filter=blob:none", "https://github.com/folke/lazy.nvim.git", lazypath, })
end
vim.opt.runtimepath:prepend(lazypath)

-- install plugins
local plugins = {
  "folke/tokyonight.nvim",
  "folke/LazyVim",
  -- add any other plugins here
}
require("lazy").setup(plugins, {
  root = root .. "/plugins",
})

vim.cmd.colorscheme("tokyonight")
-- add anything else here
```

dpetka2001:
[...]

Issue #3088, opened 2024-05-01
bug: neo-tree does not make directory current working directory when I click . on that directory


monoira:
### Did you check docs and existing issues?

- [X] I have read all the LazyVim docs
- [X] I have searched the existing issues of LazyVim
- [X] I have searched the existing issues of plugins related to this issue

### Neovim version (nvim -v)

0.9.5

### Operating system/version

EndeavourOS Linux x86_64

### Describe the bug

in other distributions like Astronvim and NvChad, when you press . on a directory when neo-tree is open, you automatically enter & cd into that directory and that directory becomes current working directory.

this is not case with LazyVim. this seems to be more of a LazyVim issue? because I deleted
 ~/.config/nvim && rm -rf ~/.local/share/nvim/ && rm -rf ~/.cache/nvim
 
and tried fresh install but it still doesn't cd into that directory upon . press.
 
 this is my lua/plugins/neo-tree.lua:
 
```lua
return {
  "nvim-neo-tree/neo-tree.nvim",

  opts = {
    sources = { "filesystem", "buffers", "git_status", "document_symbols" },
    open_files_do_not_replace_types = {
      "terminal",
      "Trouble",
      "trouble",
      "qf",
      "Outline",
    },
    filesystem = {
      bind_to_cwd = true,
      follow_current_file = { enabled = true },
      use_libuv_file_watcher = true,

      filtered_items = {
        visible = true,
        show_hidden_count = true,
        hide_dotfiles = false,
        hide_gitignored = false,
        hide_by_name = {
          ".git",
          ".DS_Store",
          "thumbs.db",
        },
        never_show = {},
      },
    },
    window = {
      mappings = {
        ["<space>"] = "none",
        ["Y"] = {
          function(state)
            local node = state.tree:get_node()
            local path = node:get_id()
            vim.fn.setreg("+", path, "c")
          end,
          desc = "Copy Path to Clipboard",
        },
        ["O"] = {
          function(state)
            require("lazy.util").open(
              state.tree:get_node().path,
              { system = true }
            )
          end,
          desc = "Open with System Application",
        },
      },
    },
    default_component_configs = {
      indent = {
        with_expanders = true, -- if nil and file nesting is enabled, will enable expanders
        expander_collapsed = "",
        expander_expanded = "",
        expander_highlight = "NeoTreeExpander",
      },
    },
  },
}
```
[...]

PR #3097, opened 2024-05-03
fix(neo-tree): correctly set up `cwd`

dpetka2001:
After taking another look at #3088 and #3090, I remembered another issue about Neo-tree, which used an autocommand to lazy-load it instead of directly requiring it at startup (see [this comment](https://github.com/LazyVim/LazyVim/discussions/2366#discussioncomment-8156257)). 

I then tested again with the initial LazyVim implementation in `init` function and saw that when you do `nvim some_directory/`, then it opens 2 buffers (neo-tree on the left and an empty unnamed buffer on the right). I did `verbose pwd` on both buffers and I saw different results. neo-tree buffer had `[tabpage] some_directory` (the same as the file argument passed when invoking nvim with a directory), but the empty unnamed buffer had `window current_working_directory` (the directory where Neovim was invoked from). Consequentially, if you close neo-tree and open it again it will open in the current working directory where Neovim was invoked from as opposed to where it was initially opened (the directory name of the file argument passed to Neovim). 

After that I tried to lazy-load neo-tree in LazyVim with an autocommand as suggested in the comment that I linked to and another issue at neo-tree repo for correctly hijacking netrw behavior without disabling netrw and saw that the results were different. With lazy-loading neo-tree both neo-tree buffer and the unnamed empty buffer show the same result when doing `verbose pwd`, namely `[tabpage] some_directory` (which is the name of the directory that was passed as file argument while invoking nvim). 

I also saw in Astronvim repo, which was mentioned as working correctly in one of the issues, and saw that instead of deleting the augroup of the autocommand it just uses `return true` as seen [here](https://github.com/AstroNvim/AstroNvim/commit/75458e693aecde45ab8d33cd55649c759e9873f3#diff-588c4f224d969c909997c8b380551855a56900cc105de17971aad31a7c1cc155), but I left the initial deletion of augroup the traditional way, because I honestly don't understand how `return true` accomplishes self-deletion of the augroup.

I also added `BufEnter` to clear the `root.cache`, because for some reason it was not being cleared when doing `set_root` from neo-tree (which i presumed that it should based on `DirChanged` event) and now when doing `set_root` you get the correct directory after closing and opening neo-tree again. Maybe `BufEnter` is triggered too often for this and this could be done differently? Unfortunately, I can't think of a better approach for this with my limited programming knowledge.

I'd appreciate further input from @folke, as I don't understand why directly requiring neo-tree in the `init` function shows 2 different `cwd` in the 2 buffers as opposed to lazy-loading it in an autocommand. 

Edit: the above tests were made with `vim.g.root_spec = { "cwd" }` and `filesystem.bind_to_cwd = true`, as that was what the OPs of the aforementioned issues wanted, I presume.

Edit 2: feel free to remove the comments. I just left them for clarity.

### DIFF
--- a/lua/lazyvim/plugins/editor.lua
+++ b/lua/lazyvim/plugins/editor.lua
@@ -46,15 +46,14 @@ return {
       vim.api.nvim_create_autocmd("BufEnter", {
         group = vim.api.nvim_create_augroup("Neotree_start_directory", { clear = true }),
         desc = "Start Neo-tree with directory",
+        once = true,
         callback = function()
           if package.loaded["neo-tree"] then
-            vim.api.nvim_del_augroup_by_name("Neotree_start_directory")
             return
           else
             local stats = vim.uv.fs_stat(vim.fn.argv(0))
             if stats and stats.type == "directory" then
               require("neo-tree")
-              vim.api.nvim_del_augroup_by_name("Neotree_start_directory")
             end
           end
         end,
--- a/lua/lazyvim/plugins/editor.lua
+++ b/lua/lazyvim/plugins/editor.lua
@@ -41,12 +41,24 @@ return {
       vim.cmd([[Neotree close]])
     end,
     init = function()
-      if vim.fn.argc(-1) == 1 then
-        local stat = vim.uv.fs_stat(vim.fn.argv(0))
-        if stat and stat.type == "directory" then
-          require("neo-tree")
-        end
-      end
+      -- FIX: use `autocmd` for lazy-loading neo-tree instead of directly requiring it,
+      -- because `cwd` is not set up properly.
+      vim.api.nvim_create_autocmd("BufEnter", {
+        group = vim.api.nvim_create_augroup("Neotree_start_directory", { clear = true }),
+        desc = "Start Neo-tree with directory",
+        callback = function()
+          if package.loaded["neo-tree"] then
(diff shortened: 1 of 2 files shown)


georgeguimaraes:
That’s great. I sometimes get a quirky behavior from neotree and cwd but never found out the problem. I’ll try this!

dpetka2001:
@georgeguimaraes Please do so and provide any necessary feedback as I might have missed something.

folke:
LGTM!

fyi, returning `true` from an autocmd callback deletes the autocmd.

But in this case, it's better to just use `once=true`

SooOverpowered:
I tested with a clean lazyvim repo and tried opening two buffers side by side, each buffer has a file in a separate directory. When I tried switching buffer then `cwd` doesn't change accordingly.

dpetka2001:
That's because default LazyVim installation uses https://github.com/LazyVim/LazyVim/blob/91fe7ef56815263b6a0f4a5c23c4bd1134c77849/lua/lazyvim/config/options.lua#L13
so the LSP `root_dir` is used. What are you trying to achieve?

SooOverpowered:
I'm trying to get the nvim-spectre plugin to work per-buffer, but somehow when I trigger nvim-spectre on the 2nd buffer it only search from within the 1st buffer, which is undesirable. I suppose it could be bc of some cwd problem.

dpetka2001:
In spectre you have to define the path yourself.

PR #3105, opened 2024-05-05
feat(extras/go): gopls check for shadow variables

victorkamoto:
This PR enables gopls' analyzer check for shadow variables.

https://pkg.go.dev/golang.org/x/tools/go/analysis/passes/shadow


amaanq:
I think this one could be pretty controversial to add, even in the linked docs it says this:

> (This definition can be refined; the module generates too many false positives and is not yet enabled by default.)

It should probably be up to the user to configure this themselves in their own config, and not be a LazyVim default imo, this could just easily generate a lot of false positives

victorkamoto:
Alright, reasonable.

PR #3107, opened 2024-05-05
fix(ui): don't lazy-load dashboard-nvim

mangkoran:
As https://github.com/nvimdev/dashboard-nvim/pull/450, dashboard-nvim shouldn't be lazy-loaded to properly handle stdin.

Not really sure if this is enough to make dashboard-nvim not lazy-loaded.

### DIFF
--- a/lua/lazyvim/plugins/ui.lua
+++ b/lua/lazyvim/plugins/ui.lua
@@ -346,9 +346,10 @@ return {
       return false
     end,
   },
+
   {
     "nvimdev/dashboard-nvim",
-    event = "VimEnter",
+    lazy = false, -- As https://github.com/nvimdev/dashboard-nvim/pull/450, dashboard-nvim shouldn't be lazy-loaded to properly handle stdin.
     opts = function()
       local logo = [[
            ██╗      █████╗ ███████╗██╗   ██╗██╗   ██╗██╗███╗   ███╗          Z


rubiin:
I think that should work . 
Other place which would by default enforce lazyloading would be 
```
  defaults = {
    lazy = false, -- should plugins be lazy-loaded?
```
but i think its turned off by default. 
Since some users may be using the lazy loading particularly for `dashboard.nvim`, it would be better to leave the link https://github.com/nvimdev/dashboard-nvim/pull/450 as a comment on the dashboard plugin spec so as to convey the info

mangkoran:
Updated. Please kindly check :D

rubiin:
That looks good to me :)

baod-rate:
does it make more sense to explicitly disable lazy-loading? 

```diff
     "nvimdev/dashboard-nvim",
-    event = "VimEnter",
+    lazy = false,
     opts = function()
```

as we do here:

https://github.com/LazyVim/LazyVim/blob/530e94a9fa19577401e968a9673282c3d79f01e3/lua/lazyvim/plugins/extras/lang/tex.lua#L27-L29

just removing the `event` param causes the dashboard to not load at all if the user has `require("lazy").setup({ defaults = { lazy = true }}`. AFAIK no other LazyVim-plugins behave like that

baod-rate:
Yes, I realize that `lazy = true` is not the default setup for lazy.nvim. But every other plugin in this repo, AFAIK, is implicitly `lazy` (whether through an `event`/`keys`/`cmd`) or explicity `lazy = true/false`. With this change, `dashboard-nvim` would be the **only** plugin in LazyVim whose `lazy` setting depends on the configured `default` for `lazy.nvim`. That means it stops working completely if the user sets `{ defaults = { lazy = true }`.

Regardless of whether you think `lazy` is an acceptable default (I agree it's better to be explicit about it). There's no reason to _not_ set `lazy = false` for the dashboard-nvim plugin spec (just as with `lervag/vimtex` spec, above). The plugin simply does not behave correctly if it is lazy-loaded. And if the user really wants, they can still [override that setting in their own config](http://www.lazyvim.org/configuration/plugins#%EF%B8%8F-customizing-plugin-specs).

dpetka2001:
Sure it definitely doesn't hurt to explicitly set it to `lazy = false` in that case.

mangkoran:
@dpetka2001 updated with your suggestion. That would make more sense :D

folke:
Thanks

PR #3128, opened 2024-05-09
fix(bufferline): prevent tabline pop-(in/out)

baod-rate:
bufferline.nvim updates the showtabline option in its `setup()`, so if showtabline is set to a conflicting value prior to loading bufferline, the user will see it flash in/out as the UI loads. Change the event that loads the plugin from `VeryLazy` to `UIEnter` to avoid this

e.g. set `vim.opt.showtabline = 2` in `lua/config/options.lua` and then open a single file. The tabline will display but only for less than a second

### DIFF
--- a/lua/lazyvim/plugins/ui.lua
+++ b/lua/lazyvim/plugins/ui.lua
@@ -56,7 +56,7 @@ return {
   -- tabs, which include filetype icons and close buttons.
   {
     "akinsho/bufferline.nvim",
-    event = "VeryLazy",
+    event = "UIEnter",
     keys = {
       { "<leader>bp", "<Cmd>BufferLineTogglePin<CR>", desc = "Toggle Pin" },
       { "<leader>bP", "<Cmd>BufferLineGroupClose ungrouped<CR>", desc = "Delete Non-Pinned Buffers" },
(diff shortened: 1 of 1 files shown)


dpetka2001:
Can this not be avoided if the user correctly uses the `always_show_bufferline` option? I also had this problem in the past and found out that if a user is using `bufferline`, then he should use `always_show_bufferline` option instead of setting `showtabline`. Even if you have `showtabline = 2` it doesn't really do anything if `always_show_bufferline` isn't set to `true`.

baod-rate:
That is only one of the cases. for ex ,if you set `always_show_bufferline = true` (and keep the default `showtabline = 1`) the bufferline will pop-in when you open a file. even if you use a completely unmodified LazyVim with default settings, the pop-in still happens if you open two buffers on startup, e.g. `nvim file1 file2`

dpetka2001:
Yes, that happens because `VeryLazy` happens after `UIEnter` when the TUI is ready. My concern was there might be some kind of performance hit, but I've tried it and I didn't observe anything meaningful to the human eye. I actually like this behavior better, because on `VeryLazy` it looks like some kind of a glitch, whereas your proposal makes it look more natural. 

Here's hoping that the maintainer will also be positive of this change :smile:

folke:
Not a fan of this. Thank you anyway. Closing.

PR #3134, opened 2024-05-10
feat(mason-lspconfig): allow `opts.ensure_installed` to be taken into account

dpetka2001:
In `mason-lspconfig` you can specify versions of packages to be installed (i.e eslint@4.8.0). The current `ensure_installed` in `lsp/init.lua` doesn't take that into account, rather only populates `local ensure_installed` with the server names declared in `nvim-lspconfig`. 

This change makes it possible to also take into account the `opts.ensure_installed` of `mason-lspconfig.nvim` in case users want to declare specific versions of packages to be installed.

### DIFF
--- a/lua/lazyvim/plugins/lsp/init.lua
+++ b/lua/lazyvim/plugins/lsp/init.lua
@@ -212,7 +212,14 @@ return {
       end
 
       if have_mason then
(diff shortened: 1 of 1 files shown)


folke:
Thanks!

PR #3145, opened 2024-05-12
fix(extras): env highlighting with dot extra

Demianeen:
There was a small issue that highlighting wasn't working with .env files because they were assigned to a new nvim filetype. This PR fixes that.

### DIFF
--- a/lua/lazyvim/plugins/extras/util/dot.lua
+++ b/lua/lazyvim/plugins/extras/util/dot.lua
@@ -47,6 +47,14 @@ return {
         },
       })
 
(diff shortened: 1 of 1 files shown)


dpetka2001:
When you do `nvim` without file arguments and then open the file you want via `Find file` from dashboard, does it get correct highlighting?

rubiin:
`dotenv` is a valid filetype but why are we setting to `sh`

folke:
That `dotenv` filetype was actually wrong. Changed it to `sh` instead, so this PR is no longer needed.

PR #3148, opened 2024-05-12
feat(extra): add undotree

rubiin:
undo history visualizer for VIM

### DIFF
--- a/lua/lazyvim/plugins/extras/editor/undotree.lua
+++ b/lua/lazyvim/plugins/extras/editor/undotree.lua
@@ -1,0 +1,26 @@
+return {
+  {
+    "mbbill/undotree",
+    cmd = "UndotreeToggle",
+    keys = {
+      -- [U]ndo [T]ree
+      { "<leader>uT", "<cmd>UndotreeToggle<cr>", desc = "Undo Tree Toggle" },
+    },
+    init = function()
+      -- Persist undo, refer https://github.com/mbbill/undotree#usage
+      local undodir = vim.fn.expand("~/.undo-nvim")
+
+      if vim.fn.has("persistent_undo") == 1 then
+        if vim.fn.isdirectory(undodir) == 0 then
+          os.execute("mkdir -p " .. undodir)
+        end
+
+        vim.opt.undodir = undodir
+        vim.opt.undofile = true
+      end
+
+      -- set layout style to 2, let g:undotree_WindowLayout = 2
+      vim.g.undotree_WindowLayout = 2
+    end,
+  },
+}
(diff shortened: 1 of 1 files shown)


yusufaktepe:
I suggest [telescope-undo.nvim](https://github.com/debugloop/telescope-undo.nvim). It might suit better since it's written in Lua and uses telescope.

rubiin:
> I suggest [telescope-undo.nvim](https://github.com/debugloop/telescope-undo.nvim). It might suit better since it's written in Lua and uses telescope.

Yeah I was gonna go for that but this seems to be more popular

dotfrag:
Thanks for the PR, I also would like to see this. However, note that the key conflicts with 'Toggle Treesitter Highlight'.

rubiin:
> Thanks for the PR, I also would like to see this. However, note that the key conflicts with 'Toggle Treesitter Highlight'.

thanks will remap to something else

dotfrag:
Thanks, that looks free indeed. Sorry but I just noticed.. I would recommend the following change (or similar) to avoid cluttering the home directory:

```lua
local undodir = vim.fn.stdpath("data") .. "/undotree" -- or /undodir
```

rubiin:
> Thanks, that looks free indeed. Sorry but I just noticed.. I would recommend the following change (or similar) to avoid cluttering the home directory:
> 
> ```lua
> local undodir = vim.fn.stdpath("data") .. "/undotree" -- or /undodir
> ```
> 
> (personal preference)
> 
> EDIT: My reasoning for the specific dir (`~/.local/share/nvim/`) is because other plugins use it for similar purposes (neo-tree logs, mason, project history, telescope history).
> 
> EDIT 2: The default `undodir` is `~/.local/state` (`stdpath('state')`).
> 
> See `:h stdpath()`:
> 
> ```
> 		state        String  Session state directory: storage for file
> 				     drafts, swap, undo, |shada|.
> ```
> 
> And `:h undodir`:
> 
> ```
> 'undodir' 'udir'	string	(default "$XDG_STATE_HOME/nvim/undo//")
> ```
> 
> Maybe this is a better dir, _unless_ it conflicts with the default undo files.
> 
> EDIT 3: I think the whole part can simply go away:
> 
> ```lua
>       local undodir = vim.fn.expand("~/.undo-nvim")
> 
>       if vim.fn.has("persistent_undo") == 1 then
>         if vim.fn.isdirectory(undodir) == 0 then
>           os.execute("mkdir -p " .. undodir)
>         end
> 
>         vim.opt.undodir = undodir
>         vim.opt.undofile = true
>       end
> ```
> 
> `undodir` already has a default from nvim and `undofile` is enabled in LazyVim
> 
> https://github.com/LazyVim/LazyVim/blob/ae381c91a1ec420cce082443501bace1ac7aaa3f/lua/lazyvim/config/options.lua#L85

yeah that makes sense to use whats already set on options

PR #3149, opened 2024-05-13
fix(extras.lang.rust): creates.nvim src option was renamed to completion

asashakira:
Recent changes to create.nvim displayed warning.
https://github.com/Saecki/crates.nvim/pull/120

Renamed `src` to `completion`


folke:
Thanks!

PR #3160, opened 2024-05-14
fix(lazyfile): exclude `editorconfig` from skips

### LINKED ISSUE #3153: bug: Editorconfig is ignored in Markdown files

### Neovim version (nvim -v)

neovim-0.10.0~ubuntu1+git202401272103-2cd76a758-971e32c87

### Operating system/version

Ubuntu 20.04.3

### Describe the bug

Editorconfig settings are ignored in Markdown files. Using nightly vim and relying on the native Editorconfig support.

Settings for e.g. js files appear to work OK.

### Steps To Reproduce

1. Create an editorconfig
```editorconfig
root = true

# Unix-style newlines with a newline ending every file
[*]
end_of_line = lf
insert_final_newline = true
indent_style = space
indent_size = 2

[*.js]
indent_size = 3

[*.md]
indent_size = 4
trim_trailing_whitespace = false
```
2. vim -u /tmp/nothing test.md
3. Type `i<Tab>indented<Esc>wq<Enter>`
4. vim test.md
5. Type `o<Tab>Indented<Esc>wq<Enter>`

I tried doing this by passing ex commands - ie

```
vim -c 'let @a = "o<tab>indented^["' -c "norm @a" -c "wq" -u /tmp/nothing test.md
vim -c 'let @a = "o<tab>indented^["' -c "norm @a" -c "wq" test.md
```

... but the Ex commands end up running before enough of LazyVim has loaded and the indents are both OK.

### Expected Behavior

Expected file content

```
    indented
    indented
```

Actual file content

```
    indented
  indented
```

### Repro

```Lua
-- My Lua skillz are not this mad
```

dpetka2001:
Also change the order of triggering events, so that `FileType` happens before `BufReadPost`, `BufNewFile` etc.

Fixes #3153

### DIFF
--- a/lua/lazyvim/util/plugin.lua
+++ b/lua/lazyvim/util/plugin.lua
@@ -88,7 +88,7 @@ function M.lazy_file()
     for _, event in ipairs(events) do
       local augroups = Event.get_augroups(event.event)
       local groups = vim.tbl_filter(function(t)
-        return not vim.tbl_contains({ t }, "filetypedetect")
+        return not vim.tbl_contains({ t }, "filetypedetect") and not vim.tbl_contains({ t }, "editorconfig")
       end, augroups)
       skips[event.event] = skips[event.event] or groups
     end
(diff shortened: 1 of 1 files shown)


iseth:
Thank you @dpetka2001 for fixing this!

folke:
I just removed all the code related to `lazy_file` and changed it in way where a file opened from the cmdline will be shown  quicker than before and without all the hacks related to `lazy_file`.

Related commits:

- https://github.com/LazyVim/LazyVim/commit/b29d169afb7560ed4a9276e4379e28ffaed1a171
- https://github.com/LazyVim/LazyVim/commit/965a469ca8cb1d58b49c4e5d8b85430e8c6c0a25
- https://github.com/LazyVim/LazyVim/commit/8301096c31f9855c74d5b732367b679bc8721b5e

While **time to first render** is quicker than before, the startup time will be way slower. But that of course is to be expected, since file based plugins will now no longer load lazily, but they will load after first render.

dpetka2001:
@folke If that's the case maybe you would also like to exclude my previous PR #3004 from the latest release, since it doesn't apply to the current logic anymore.

folke:
I'll just leave it in there, since I have no idea how to change release-please release notes :)

And thanks again for looking into this!

dpetka2001:
No problem. I really enjoyed trying to learn a bit more about how it worked under the hood :)

PR #3165, opened 2024-05-15
fix(dap): add debugger to filetypes mapping for launch.json

80avin:
The .vscode/launch.json detection added in https://github.com/LazyVim/LazyVim/pull/1839 doesn't support many filetypes

This adds the support of debug adapter to file types mappings.


folke:
Thanks!

PR #3169, opened 2024-05-16
refactor(autocmds): use autocmd in place of vim.api.nvim_create_autocmd

rubiin:
This makes the code more concise and readable imo similar to options


dpetka2001:
@rubiin Just a question. Wouldn't it be preferable to use another branch for the PRs you make and not main, since that's the one which is being synced with `upstream` when you pull?

rubiin:
> @rubiin Just a question. Wouldn't it be preferable to use another branch for the PRs you make and not main, since that's the one which is being synced with `upstream` when you pull?

yeah . Thats a mistake on my part . Only realized that this was on main after I created PR lol

folke:
This isnt really needed... Thanks

PR #3178, opened 2024-05-16
fix(treesitter-rewrite): show error in Extras only when enabled

dpetka2001:


folke:
This should not be needed, since the code in the extra will only be executed when the extra was enabled?

dpetka2001:
Every time I open the extras window from dashboard with `x` the error pops up in my case. Could you verify?

folke:
Right! Forgot about that. Thanks!

PR #3197, opened 2024-05-17
fix: bufferline does not refresh after close buffer

RayGuo-ergou:
### Description

Faced an issue that when close the buffer from bufferline (either right click or close icon). Some times the bufferline does not refresh after buffer close (not really sure why tho). This will cause a buffer id invalid error if try to close the buffer again.

I am not using lazyvim distro but I can reproduce this issue with lazyvim starter repo. 

image

### DIFF
--- a/lua/lazyvim/plugins/ui.lua
+++ b/lua/lazyvim/plugins/ui.lua
@@ -95,7 +95,7 @@ return {
     config = function(_, opts)
       require("bufferline").setup(opts)
       -- Fix bufferline when restoring a session
-      vim.api.nvim_create_autocmd("BufAdd", {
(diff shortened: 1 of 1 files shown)


folke:
This was due to something else. Just fixed that

RayGuo-ergou:
Hi @folke the new update does not seems to fix this issue.

GIF 5-18-2024 12-12-24 AM
_Yellow dot is left click and blue dot is right click_
image

folke:
I just added the `BufDelete` as in your PR. Didnt want to merge yours, since it also changed the formatting of the dashboard text.

Thanks!

RayGuo-ergou:
No problem. As long as the problem solved. Thanks!

PR #3201, opened 2024-05-17
fix(refactoring): add label to refactoring key group

Mawdac:
This adds a missing label for the key group in the refactoring extra.


rubiin:
was meaning to do it :D

folke:
Thanks!

PR #3211, opened 2024-05-17
fix(leap): add label to renamed surround mappings key group

stefanboca:
This adds a missing label for the remapped surround key group in the leap extra.

I was considering adding `["gs"] = { }` to the mappings, but I wasn't certain if this would be necessary or correct for removing the original `gs` key group (which doesn't appear in which-key anyway).


folke:
thanks!

PR #3229, opened 2024-05-19
Clarify vim.g.deprecation_warnings behavior

alichtman:
The previous wording was a little clunky (disable disabling). This removes the double negative.


folke:
Makes more sense indeed :)

PR #3230, opened 2024-05-19
fix(lazygit): Fix off-by-one issue in blame_line

flpcury:
**Example:**

I want to blame line 44 of file `myfile`.

**Current behavior:**

`cmd` will be `git log -n 3 -u -L 43,+1:myfile`, so I get the blame for line 43 instead.


folke:
Thanks!

PR #3234, opened 2024-05-19
fix(yanky): load yanky earlier to avoid losing deleted lines

dotfrag:
Currently, yanky is only loaded on yank and put operations. This is problematic because if you open a file and start deleting lines (`d<motion>`), you lose them because yanky is not loaded to save them.


folke:
Makes sense. Thanks!

PR #3235, opened 2024-05-19
feat: add dropbar for vscode like breadcrumbs

rubiin:
A polished, IDE-like, highly-customizable winbar for Neovim
with drop-down menu support and multiple backends

### DIFF
--- a/lua/lazyvim/plugins/extras/editor/dropbar.lua
+++ b/lua/lazyvim/plugins/extras/editor/dropbar.lua
@@ -1,10 +1,56 @@
 return {
   {
     "Bekaboo/dropbar.nvim",
-    event = { "BufReadPre", "BufNewFile" },
+    event = "LazyFile",
     dependencies = {
       "nvim-telescope/telescope-fzf-native.nvim",
     },
-    config = {},
+    keys = {
+      {
+        "<leader>D",
+        function()
+          require("dropbar.api").pick()
+        end,
+        desc = "Winbar pick",
+      },
+    },
+    opts = function()
+      local menu_utils = require("dropbar.utils.menu")
+
+      -- Closes all the windows in the current dropbar.
+      local function close()
+        local menu = menu_utils.get_current()
+        while menu and menu.prev_menu do
+          menu = menu.prev_menu
+        end
+        if menu then
+          menu:close()
+        end
+      end
+
+      return {
+        menu = {
+          preview = false,
+          keymaps = {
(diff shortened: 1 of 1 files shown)


dpetka2001:
@rubiin 

Some quality of life suggestions (which of course you're not obligated to follow)
```lua
return {
  {
    "Bekaboo/dropbar.nvim",
    event = "LazyFile",
    dependencies = {
      "nvim-telescope/telescope-fzf-native.nvim",
    },
    keys = {
      {
        "<leader>D",
        function()
          require("dropbar.api").pick()
        end,
        desc = "Winbar pick",
      },
    },
    opts = function()
      local menu_utils = require("dropbar.utils.menu")

      -- Closes all the windows in the current dropbar.
      local function close()
        local menu = menu_utils.get_current()
        while menu and menu.prev_menu do
          menu = menu.prev_menu
        end
        if menu then
          menu:close()
        end
      end

      return {
        menu = {
          preview = false,
          keymaps = {
            -- Navigate back to the parent menu.
            ["h"] = "<C-w>q",
            -- Expands the entry if possible.
            ["l"] = function()
              local menu = menu_utils.get_current()
              if not menu then
                return
              end
              local row = vim.api.nvim_win_get_cursor(menu.win)[1]
              local component = menu.entries[row]:first_clickable()
              if component then
                menu:click_on(component, nil, 1, "l")
              end
            end,
            ["q"] = close,
            ["<esc>"] = close,
          },
        },
      }
    end,
  },
}
```
Credit to old dotfiles from [MariaSolOs](https://github.com/MariaSolOs/dotfiles/tree/70356010f1e7518a60dc6a04950bc385f568c66a/.config/nvim) adapted to LazyVim way.

`h` and `l` are necessary in my opinion for navigating with keyboard and `<leader>D` (feel free to change this to whatever you like as long as there's no other keymap conflict, I just chose it for Dropbar since `d` is used for `Debug`) for invoking the Dropbar menu from keyboard as well (I rarely use mouse inside Neovim :stuck_out_tongue:). Also, `q` and `<esc>` are more intuitive for completely closing the Dropbar menu in my humble opinion.

rubiin:
Will play around and steal stuffs from this :D . I have a minimal dropbar config

dpetka2001:
I don't use it personally, so just had a look around googling for dotfiles, because when I tried it bare bones it looked to me that for most of its functionality you have to use the mouse. So, just wanted a few upgrades if this is going to make it as a LazyExtra.

PR #3244, opened 2024-05-19
feat(java): enable inlay hints

idelice:
I apologize if my Lua code is "bad", but I haven't really coded in Lua before, but this will make sure that the consumer can toggle inlay hints for java development by providing the opts:

```lua
opts.enableInlayHints = true
```

**EDIT**
enabled by default.

### DIFF
--- a/lua/lazyvim/plugins/extras/lang/java.lua
+++ b/lua/lazyvim/plugins/extras/lang/java.lua
@@ -193,8 +193,9 @@ return {
           local client = vim.lsp.get_client_by_id(args.data.client_id)
           if client and client.name == "jdtls" then
             -- INFO: without this the inlay hints start toggled off by default.
-            -- Maybe because `nvim-jdtls` attaches on a `FileType` event??
-            -- Still with this the inlay hints are toggled on by default
+            -- Probably because `nvim-lspconfig` is not used to set up `nvim-jdtls`
+            -- and `LazyVim.toggle.inlay_hints` doesn't execute, so we have
+            -- to execute it again ourselves in this autocmd in `LspAttach`
             LazyVim.toggle.inlay_hints(args.buf, true)
             local wk = require("which-key")
             wk.register({
--- a/lua/lazyvim/plugins/extras/lang/java.lua
+++ b/lua/lazyvim/plugins/extras/lang/java.lua
@@ -162,6 +162,21 @@ return {
           capabilities = LazyVim.has("cmp-nvim-lsp") and require("cmp_nvim_lsp").default_capabilities() or nil,
         }, opts.jdtls)
 
+        if opts.enableInlayHints then
+          local inlayHintsConfig = function()
+            local extendedClientCapabilities = require("jdtls").extendedClientCapabilities
+
+            return {
+              extendedClientCapabilities = extendedClientCapabilities,
+              parameterNames = {
+                enabled = "all",
+              },
+            }
+          end
+
+          extend_or_override(config, inlayHintsConfig())
+        end
+
         -- Existing server will be reused if the root_dir matches.
         require("jdtls").start_or_attach(config)
         -- not need to require("jdtls.setup").add_commands(), start automatically adds commands
--- a/lua/lazyvim/plugins/extras/lang/java.lua
+++ b/lua/lazyvim/plugins/extras/lang/java.lua
@@ -118,6 +118,15 @@ return {
         dap = { hotcodereplace = "auto", config_overrides = {} },
         dap_main = {},
         test = true,
+        settings = {
+          java = {
+            inlayHints = {
+              parameterNames = {
+                enabled = "all",
+              },
+            },
+          },
+        },
       }
     end,
     config = function()
@@ -158,22 +167,10 @@ return {
           init_options = {
             bundles = bundles,
           },
+          settings = opts.settings,
           -- enable CMP capabilities
           capabilities = LazyVim.has("cmp-nvim-lsp") and require("cmp_nvim_lsp").default_capabilities() or nil,
         }, opts.jdtls)
-
-        if opts.enableInlayHints == true then
-          local extendedClientCapabilities = require("jdtls").extendedClientCapabilities
-
-          extend_or_override(config, {
-            extendedClientCapabilities = extendedClientCapabilities,
-            inlayHints = {
-              parameterNames = {
-                enabled = "all",
-              },
-            },
-          })
-        end
 
         -- Existing server will be reused if the root_dir matches.
         require("jdtls").start_or_attach(config)
@@ -195,6 +192,10 @@ return {
         callback = function(args)
           local client = vim.lsp.get_client_by_id(args.data.client_id)
           if client and client.name == "jdtls" then
+            -- INFO: without this the inlay hints start toggled off by default.
+            -- Maybe because `nvim-jdtls` attaches on a `FileType` event??
+            -- Still with this the inlay hints are toggled on by default
+            LazyVim.toggle.inlay_hints(args.buf, true)
             local wk = require("which-key")
             wk.register({
               ["<leader>cx"] = { name = "+extract" },
--- a/lua/lazyvim/plugins/extras/lang/java.lua
+++ b/lua/lazyvim/plugins/extras/lang/java.lua
@@ -162,7 +162,7 @@ return {
           capabilities = LazyVim.has("cmp-nvim-lsp") and require("cmp_nvim_lsp").default_capabilities() or nil,
         }, opts.jdtls)
 
-        if opts.enableInlayHints then
+        if opts.enableInlayHints == true then
           local inlayHintsConfig = function()
             local extendedClientCapabilities = require("jdtls").extendedClientCapabilities
 
--- a/lua/lazyvim/plugins/extras/lang/java.lua
+++ b/lua/lazyvim/plugins/extras/lang/java.lua
@@ -167,8 +167,10 @@ return {
 
           extend_or_override(config, {
             extendedClientCapabilities = extendedClientCapabilities,
-            parameterNames = {
-              enabled = "all",
+            inlayHints = {
+              parameterNames = {
+                enabled = "all",
+              },
             },
           })
         end
--- a/lua/lazyvim/plugins/extras/lang/java.lua
+++ b/lua/lazyvim/plugins/extras/lang/java.lua
@@ -163,18 +163,14 @@ return {
         }, opts.jdtls)
 
         if opts.enableInlayHints == true then
-          local inlayHintsConfig = function()
-            local extendedClientCapabilities = require("jdtls").extendedClientCapabilities
+          local extendedClientCapabilities = require("jdtls").extendedClientCapabilities
 
-            return {
-              extendedClientCapabilities = extendedClientCapabilities,
-              parameterNames = {
-                enabled = "all",
-              },
-            }
-          end
-
-          extend_or_override(config, inlayHintsConfig())
+          extend_or_override(config, {
+            extendedClientCapabilities = extendedClientCapabilities,
+            parameterNames = {
+              enabled = "all",
+            },
+          })
         end
 
         -- Existing server will be reused if the root_dir matches.


dpetka2001:
So with this PR the user has to explicitly put `opts.enableInlayHints = true` in his personal configuration to enable the inlay hints?

idelice:
> So with this PR the user has to explicitly put `opts.enableInlayHints = true` in his personal configuration to enable the inlay hints?

Yes i added it as an option just in case someone doesn't want it - they are free to choose.

dpetka2001:
I don't think that's right. `inlay_hints` are enabled by default in LazyVim. So they should be enabled by default for Java as well. And the user can change that in his personal configuration. I also had started implementing a PR about this. And I believe I have solved these problems. I also found a somewhat hackish way to show the inlay hints by default as opposed to my comment in the other Discussion thread.

Please take a look [here](https://github.com/LazyVim/LazyVim/compare/main...dpetka2001:LazyVim:feat/java_inlay_hints) and try to incorporate it to your own PR.

idelice:
Even tho i don't agree with forcing consumers to use inlay hints, I've added your changes. Please review again

dpetka2001:
But the purpose of this PR is to enable inlay hints in Java by default, since they're already enabled by default in LazyVim. Otherwise there's no purpose for this PR and the users can still enable it in their personal configuration if they want.

lua/lazyvim/plugins/extras/lang/java.lua
```diff
@@ -182,6 +192,10 @@ return {
         callback = function(args)
           local client = vim.lsp.get_client_by_id(args.data.client_id)
           if client and client.name == "jdtls" then
+            -- INFO: without this the inlay hints start toggled off by default.
+            -- Maybe because `nvim-jdtls` attaches on a `FileType` event??
+            -- Still with this the inlay hints are toggled on by default
```
dpetka2001, line 197:
```suggestion
            -- INFO: without this the inlay hints start toggled off by default.
            -- Probably because `nvim-lspconfig` is not used to set up `nvim-jdtls`
            -- and `LazyVim.toggle.inlay_hints` doesn't execute, so we have
            -- to execute it again ourselves in this autocmd in `LspAttach`
```

dpetka2001 (changes requested):

lua/lazyvim/plugins/extras/lang/java.lua
dpetka2001: I believe this must be a more correct comment as to why we have to call it here again.

idelice:
@dpetka2001  i added your code changes

dpetka2001 (approved):

lua/lazyvim/plugins/extras/lang/java.lua
idelice: I've already added your comment changes.

dpetka2001:
Sorry missed that. I'm just a regular user like you, so now all that remains is for the maintainer to look into it.

idelice:
@folke  i guess this PR is good to go.

dpetka2001:
@idelice It works correctly on your side, right? Because I only tested on a crappy Java code snippet from GPT and I don't do Java as I've already mentioned. It would be even better if other users who actually do Java (unlike me) would share their experience with this. I hope I didn't miss anything.

idelice:
Yes it worked perfectly fine.

folke:
It doens't work right away since jdtls probably doesnt correctly advertise the capability or dynamic capability.
A better way is to force the capability on the client side, since we know it's supported.

I'll try to look into it tomorrow evening.

Either way, thank you for this PR!

folke:
After further investigating this, it seems that in LazyVim, we only check for inlay hint support for server capabilities, not later on when the server advertises it through a dynamic capability.

I'll have to work on a fix for this (and this may also fix jdtls)

dpetka2001:
@folke adding `LazyVim.toggle.hints(buffer, true)` in [here](https://github.com/LazyVim/LazyVim/blob/97862f425997bf89f581e6aeceed9aac85d90432/lua/lazyvim/plugins/lsp/init.lua#L126-L133) after the `lsp.keymaps.on_attach` and removing the one from the `LspAttach` autocmd in this PR also enables the inlay hints on a java file. I'm not familiar with LSP protocol at all, so maybe that's what you meant? I'm just assuming based on something similar I saw on other dotfiles about registering dynamic capabilities for the keymaps.

folke:
@dpetka2001 yeah indeed. I'm working on a fix. Almost ready.
There's some other things that also dont work for lsp servers with dynamic capabilities.
Like with tsserver, the document highlights dont work.

I have everything working locally. Will push changed in a bit

folke:
image

I've added an `on_supports_method`.
Similar to `on_attach` but also gets called on dynamic advertisemtns of capabilities/

So it will execute exactly once, either due to LspAttach or afterwards once the dynamic capability is ready.

lua/lazyvim/plugins/extras/lang/java.lua
```diff
@@ -182,6 +192,11 @@ return {
         callback = function(args)
           local client = vim.lsp.get_client_by_id(args.data.client_id)
           if client and client.name == "jdtls" then
+            -- INFO: without this the inlay hints start toggled off by default.
+            -- Probably because `nvim-lspconfig` is not used to set up `nvim-jdtls`
+            -- and `LazyVim.toggle.inlay_hints` doesn't execute, so we have
+            -- to execute it again ourselves in this autocmd in `LspAttach`
+            LazyVim.toggle.inlay_hints(args.buf, true)
```
folke, line 199: You can probably remove these lines now and inlay hints would still work if you have it enabled in lsp-config.
Would be great if you can test this.

Make sure you're using the main version of LAzyVim.  See tip at https://www.lazyvim.org/configuration/lazy.nvim

dpetka2001: I just tested it locally with crappy GPT java code and inlay hints are enabled by default (not from the PR but with personal configuration in `nvim-jdtls` spec like I had mentioned in the Discussion thread, so it should be the same). Will take a look at the new functionalities you added tomorrow. I'm in awe of your work as always :smile:

folke: I just refactored the code to be more vim like as well. Great that it works now!!

dpetka2001: Yes, but don't take my word for granted and wait for @idelice's input as he's the one who deals with Java and he (and maybe other users) should test the new behavior.

idelice: @folke I opened up two different java projects, and I could confirm that the inlay hints are being triggered as soon as the jdtls have been attached! 

Screenshot 2024-05-20 at 10 15 45

Screenshot 2024-05-20 at 10 15 59

But for some reason, in some places, the inlay hints are not working. I don't know if it's related to lazyvim or the api itself but here's a screenshot of where the inlay hints should be triggered (inside the `new User(....)`)

Screenshot 2024-05-20 at 10 16 34

dpetka2001: Did they get triggered with this PR and LazyVim release?

idelice: I used the latest development release as @folke asked me to, and it worked fine with my local jdtls changes:
 
```lua
      settings = {
        java = {
          inlayHints = { parameterNames = { enabled = "all" } },
```

I also tried to use local forked LazyVim in my lazy config:

```lua
return { "LazyVim/LazyVim", dev = true, dir = "path/to/local/lazyvim" }
```

But it returned this when opening nvim:
Screenshot 2024-05-20 at 11 47 22

dpetka2001: To use local forked repo I just have this in my `/lua/config/lazy.lua`
```lua
  dev = {
    path = "~/projects/plugins",
    patterns = { "LazyVim" },
  },
```
You can change the `path` to whatever local path you prefer.
I also have created another LazyVim profile under directory `~/.config/LazyVimDev` and call that with `NVIM_APPNAME=LazyVimDev nvim` when I want to use the local forked repo, so that it doesn't mess with my main LazyVim profile.

idelice: Ok I finally managed to find my way around the docs to run lazyvim locally... anyway, it seems that I'm now getting more inlay hints than before but with the cost of losing the inlay hints are the very first file it opens - which means, if i close and reopen the same buffer - the inlay hints will appear.

idelice: Screenshot 2024-05-20 at 16 34 44

idelice: @folke - your latest development release I tried before - actually triggered the inlay hints when jdtls first attached.

idelice: **Development branch**

After testing with the development branch, it seems that the inlay hints are not appearing in every place anymore. Please see attachment and compare with the attached image from my previous comment:

Screenshot 2024-05-20 at 16 58 19

There's clearly a big difference.

dpetka2001: Did you make any changes to make them appear as in the first image?

idelice: No. The _only_ difference is the branches i'm using.

The first one from `main` and the other `release-please--branches--main`.

dpetka2001: The `main` branch is the one you should be testing against. It contains the changes that have been merged in the latest HEAD and will be incorporated in the next release. So, with the `main` branch everything works as it should?

idelice: Yes

dpetka2001: Do you still have to close and reopen the first buffer? Can you instead do `:e`? Or maybe `:LspRestart`? One of the 2 I imagine should do the trick instead of closing and reopening the buffer.

idelice: > Do you still have to close and reopen the first buffer? Can you instead do `:e`? Or maybe `:LspRestart`? One of the 2 I imagine should do the trick instead of closing and reopening the buffer.

Yes unfortunately. I haven't tried :e or anything else. I can do that later and report back

dpetka2001: Did you try that with the suggestion of Folke by removing the suggested lines? Do you observe any different behavior if you don't remove the suggested lines? Whenever you have free time.

idelice: Removing that line does no difference at all from what i could see. Even in the main branch.

I still had to reopen the buffer or `:e` before inlay hints would appear.

The most noticable difference between the main and the dev branch is that the main branch somehow shows _more_ inlay hints - which is GOOD! 

But i still have to :e or reopen the buffer in order for inlay hints to appear or just open another java file.

folke:
I just fully tested jdtls and fixed inlay hints in the commit above.
All the default inlay hint stuff to enable it works as expected.

idelice:
> I just fully tested jdtls and fixed inlay hints in the commit above. All the default inlay hint stuff to enable it works as expected.

Why is it that the inlay hints are not showing everywhere with the current LazyVim but in the local LazyVim forked? There's litteraly no difference with the forked and the commit you just mentioned.

folke:
That line is not needed. the `opts` are already properly passed to nvim-jdtls.

Can you please remove any of your customizations and try again.

It really does work. There's also no reason why jdtls would be differernt than for other lsp servers, even when it's not managed by lspconfig.

I ran some simple teste myself and it all works as expected for me.

idelice:
> That line is not needed. the `opts` are already properly passed to nvim-jdtls.
> 
> Can you please remove any of your customizations and try again.
> 
> It really does work. There's also no reason why jdtls would be differernt than for other lsp servers, even when it's not managed by lspconfig.
> 
> I ran some simple teste myself and it all works as expected for me.

Yes it is in order for all inlay hints to trigger.

I copied your commit changes in to my local LazyVim with no extra modifications and it partially shows inlay hints.

Whenever I put that line that I've mentioned in my previous comment, it triggers *all* inlay hints.

idelice:
After pulling latest from main, I can confirm that inlay hints are triggered right away, but without the `settings = opts.settings` - it will only partially show inlay hints.

idelice:
> Yes, that was in the initial commit as well. Otherwise the user would have to do in his personal configuration
> 
> ```lua
> java = {
>             inlayHints = {
>               parameterNames = {
>                 enabled = "all",
>               },
>             },
>           },
> ```
> 
> because the default value is `literals` and the `local config` inside function `attach_jdtls` doesn't take `settings` into consideration, so it uses the default value.

I added the following in my jdtls opts in order to fully show inlay hints without adding that line i mentioned inside lazyvim:

```lua
      jdtls = {
        settings = {
          java = {
            inlayHints = { parameterNames = { enabled = "all" } },
          },
        },
      },
```

LazyVim has the same code which means we have to duplicate the exact same code seen in LazyVim in order to show all inlay hints. I don't really see that as the way to do it.

Why not just pass opts.settings in lazyvim and close this topic? then we wouldn't have to mirror whats inside lazyvim?

dpetka2001:
@idelice I agree with you. I mentioned that this was part of the initial commit in this PR and that it should also make it to the final version of Java Extra. I just analyzed the reason why this doesn't happen as things are now. Hopefully the maintainer will also agree to this.

folke:
Check my commit https://github.com/LazyVim/LazyVim/commit/e5ea8abd38c4b93edccd69c9ef4462377103a9c3

line 123

Before that change you indeed needed to add the `settings = opts.settings`.
With that change that really should no longer be needed?

dpetka2001:
I checked your commit. Look at these [lines](https://github.com/LazyVim/LazyVim/blob/e5ea8abd38c4b93edccd69c9ef4462377103a9c3/lua/lazyvim/plugins/extras/lang/java.lua#L162-L173). Here is where the `local config` gets evaluated. It merges the table 
```lua
{
          cmd = opts.full_cmd(opts),
          root_dir = opts.root_dir(fname),
          init_options = {
            bundles = bundles,
          },
          -- enable CMP capabilities
          capabilities = LazyVim.has("cmp-nvim-lsp") and require("cmp_nvim_lsp").default_capabilities() or nil,
        }
```
with `opts.jdtls`, but those `opts.jdtls` is a means to be used in the personal configuration from users to override/add whatever they want. This `local config` as of now doesn't take `opts.settings` into account. And that `local config` is passed to `require("jdtls").start_or_attach(config)` which does the initial setup of the server. I believe `settings = opts.settings` should be added inside the `local config` to be taken into account for the initial server setup.

folke:
ok, gotcha. I thouhgt it merged with all the opts. Just pushed a fix. Thanks!

PR #3250, opened 2024-05-20
feat(vscode): enable more compatible plugins

theol0403:
Enable dial.nvim, yanky.nvim, and mini.move, all extras which are compatible with vscode.


folke:
Thanks!

PR #3252, opened 2024-05-20
fix(bufferline): unpin bufferline with 0.9.x compatibility

lawrenceho:
Pin is no longer needed as bufferline added 0.9.x compatibility.


folke:
Thanks!

PR #3259, opened 2024-05-21
ci: change the release please action name to new

rubiin:
https://github.com/google-github-actions/release-please-action is deprecated(more info on the readme) and they suggest using 
https://github.com/googleapis/release-please-action.

Also there is no argument called `package-name`  and `extra-files `which was showing echoing an error on every action run

supported flags are : ['token', 'release-type', 'path', 'target-branch', 'config-file', 'manifest-file', 'repo-url', 'github-api-url', 'github-graphql-url', 'fork', 'include-component-in-tag', 'proxy-server', 'skip-github-release', 'skip-github-pull-request', 'changelog-host']

https://github.com/googleapis/release-please/blob/main/docs/customizing.md#updating-arbitrary-files for extra files


dpetka2001:
I also see `'extra-files'` mentioned in the warnings of `Actions`, so you might also want to remove that. 

But honestly not familiar at all with Github Actions and the maintainer might have some kind of custom setup locally, so he knows best.

rubiin:
> I also see `'extra-files'` mentioned in the warnings of `Actions`, so you might also want to remove that.
> 
> But honestly not familiar at all with Github Actions and the maintainer might have some kind of custom setup locally, so he knows best.

yeah that can be removed as well. Since that option no longer exists on the flags taken by the github action

folke:
I don't think it's as simple as this change. If I remember correctly two config files are now also needed.
I've gone through the process for one of my projects, but can't remember which one :)

folke:
This one: https://github.com/folke/lovelace-styler

Needs both:
- [release-please-config.json](https://github.com/folke/lovelace-styler/blob/main/release-please-config.json)
- https://github.com/folke/lovelace-styler/blob/main/.release-please-manifest.json

rubiin:
Will add them as well. I think the manifest file is only needed when you want to specify next release. And its only used once, this is only used when there is no prior release. 
But I will add both of them just in case

folke:
Merged. Fingers crossed :)

PR #3276, opened 2024-05-22
fix(dots): set markdown for LICENSE file

rubiin:


folke:
Thanks!

folke:
sorry, I was too quick. LICENSE is not always markdown

PR #3280, opened 2024-05-22
fix: extras.lang.python to enable venv select keys on ft

caidanw:
Update the Python extra for the `linux-cultist/venv-selector.nvim` plugin so that the key map to select a virtual environment is only present (in which-key) when the filetype of the buffer is `python`.

This is mostly a nice UI thing for when I'm not working in a Python project, so that which-key doesn't show me an option that is not currently relevant.


folke:
Thanks!

PR #3291, opened 2024-05-23
Fix/extras lang kotlin

arthur-mountain:
I am the author of this [PR](https://github.com/LazyVim/LazyVim/pull/2890). 

While reviewing my own code and checking the LazyVim official website, I noticed some inconsistencies, so I created a small PR to address them:

1. Each optional plugin has a comment message indicating that the plugin is optional, but DAP does not (as shown in the image). Since the official website already has an "optional" in title, **I removed the comment messages regarding "optional" to avoid confusion**.
Screenshot 2024-05-23 at 10 20 44 PM

---

2. I changed the DAP to use local variables for DAP to maintain consistency, even though both require statements reference the same module.


folke:
Great! Can you alsoremove the mason spec for only the debug adapter and move that as a dep of the dap config?
That way, the debug adapter would only be sintalled when dap is enabled.

arthur-mountain:
Thanks for the suggestion!!! Of course, it's a great idea. I've made the adjustments.

dpetka2001:
I don't think you need to move it as a dep to dap config? `mason-nvim-dap.nvim` has `automatic_installation = true`, so I believe it should be automatically installed if there's an adapter configuration set up for it in `nvim-dap` spec. Can you test it locally without that to make sure? I remember that being the case.

arthur-mountain:
> mason-nvim-dap.nvim has automatic_installation = true, so I believe it should be automatically installed if there's an adapter configuration set up for it in nvim-dap spec.

Yes, I see the automatic_installation is true in the mason-nvim-dap.nvim.

> Can you test it locally without adding that configuration to confirm? I remember it being like that.

I just tested it, and it seems to be working without that.

Thanks for the reminder! I learned something new that I didn't know before.

Thank you both very much for the review and suggestions!!!

folke:
You still have `"kotlin-debug-adapter"` in the PR? I think what @dpetka2001 was suggesting, is that you can remove that from the extra all-together

dpetka2001:
Did you uninstall the debugger from Mason and restart Neovim to see if it gets installed? Because I'm trying this now and it doesn't work for some reason on my end.

dpetka2001:
@folke For some reason the `automatic_installation` doesn't trigger for me. Could you verify?

As a workaround to trigger it i added the following in `mason-nvim-dap.nvim` spec
```lua
      config = function(_, opts)
        require("mason-nvim-dap").setup(opts)
        if LazyVim.opts("mason-nvim-dap.nvim").automatic_installation == true then
          vim.schedule(function()
            require("mason-nvim-dap.automatic_installation")()
          end)
        end
      end,
```
Do you have any insight as to why this happens?

arthur-mountain:
> You still have "kotlin-debug-adapter" in the PR?

I've updated the PR and I made a mistake earlier, my apologies😓!

arthur-mountain:
> Did you uninstall the debugger from Mason and restart Neovim to see if it gets installed?

You are right, I did not uninstall kotlin-debug-adapter on my end.

I just tested it again, and it seems that automatic_installation is not working on my end either.

dpetka2001:
Yes, that's what I've noticed as well. That's why I mentioned the workaround in `mason-nvim-dap` spec. With that it gets triggered correctly. But I don't understand the reason why this happens. It used to work correctly from what I can remember.

dpetka2001:
@folke Would you like me to make a PR for `mason-nvim-dap` with the workaround I mentioned? Although, I don't quite understand what the primary cause of it is.

dpetka2001:
Moving `mason-nvim-dap` in `extras.dap.core` out of `nvim-dap` dependencies also works. It does lazy-load on `mason-nvim-dap.mappings.filetypes  nvim-dap`.

x-osc:
suck my nuts

folke:
Thanks!

dpetka2001:
@folke This still doesn't install the kotlin adapter when setting a breakpoint. Did you take a look at my previous comment? Maybe `mason-nvim-dap` should be moved out of dependencies of `nvim-dap`?

PR #3293, opened 2024-05-23
feat(neo-tree): add h/l navigation and open file support

dotfrag:
Taken from https://github.com/nvim-neo-tree/neo-tree.nvim/wiki/Tips#navigation-with-hjkl and adjusted slightly to allow opening files.

`h`:
- if directory is opened -> close it
- if directory is closed or focused item is file -> focus parent directory

`l`:
- if focused item is file -> open it
- if focused item is directory and is closed -> open it
- if focused item is directory and is opened -> move focus to the first child

Unrelated but I think the `{}` are unnecessary for the `Y` and `O` keymaps but I didn't touch them.

EDIT: Force pushed an update to enable this for all source selectors (i.e. buffer/git), not filesystem only. (see https://github.com/nvim-neo-tree/neo-tree.nvim/discussions/163#discussioncomment-4747082)

### DIFF
--- a/lua/lazyvim/plugins/editor.lua
+++ b/lua/lazyvim/plugins/editor.lua
@@ -87,6 +87,26 @@ return {
             end,
             desc = "Open with System Application",
           },
+          ["h"] = function(state)
+            local node = state.tree:get_node()
+            if node:has_children() and node:is_expanded() then
+              state.commands.toggle_node(state)
+            else
+              require("neo-tree.ui.renderer").focus_node(state, node:get_parent_id())
+            end
+          end,
+          ["l"] = function(state)
+            local node = state.tree:get_node()
+            if node:has_children() then
+              if not node:is_expanded() then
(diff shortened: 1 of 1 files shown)


dpetka2001:
I just have 
```lua
      window = {
        mappings = {
          ["l"] = "open",
          ["h"] = "close_node",
        },
      },
```
Because in my experience, when I'm on a file, usually I want to go back and close the node to have more real estate of what else is in Neotree. With your implementation, `h` will first focus the parent directory and then you need to press `h` again to close it. In my personal opinion that seems rather redundant. Same for `l`, when you press it, it only makes sense to expand the current node if it's closed and if not just use `j/k` to go up down instead of pressing `l` again to focus the first child.

But this is just my opinion as a regular user and based on my personal preferences.

dpetka2001:
Sure, I already have the mappings in my own config, so it won't really affect me either way.

PR #3303, opened 2024-05-24
feat(ui): add codewindow (minimap)

gacallea:
[codewindow](https://github.com/gorbit99/codewindow.nvim) is a very light and unobtrusive minimap.

Of the many I tried this does the best job at both being a great minimap and not getting in a coders way.

I have also added keymappings `<leader>um` to toggle it.

### DIFF
--- a/lua/lazyvim/plugins/extras/ui/codewindow-nvim.lua
+++ b/lua/lazyvim/plugins/extras/ui/codewindow-nvim.lua
@@ -1,0 +1,37 @@
+return {
+  {
+    "gorbit99/codewindow.nvim",
+    keys = {
+      {
+        "<leader>um",
+        function()
+          require("codewindow").toggle_minimap()
+        end,
+        desc = "Toggle minimap",
+      },
+    },
+    lazy = false,
+    opts = {
+      auto_enable = true,
+      minimap_width = 20,
+      width_multiplier = 2,
+      screen_bounds = "background",
+      window_border = "none",
+      relative = "win",
+      exclude_filetypes = {
+        "dashboard",
+        "neo-tree",
+        "lazy",
+        "lazyterm",
+        "mason",
+        "help",
+        "checkhealth",
+        "lspinfo",
+        "noice",
+        "Trouble",
+        "fish",
+        "zsh",
+      },
+    },
+  },
+}


gacallea:
adding two screenshots:

- normal on window operation:

Screenshot 2024-05-24 at 4 41 40 PM

- when some panel is open, codewindow doesn't get in the way:

Screenshot 2024-05-24 at 4 41 47 PM

lua/lazyvim/plugins/extras/ui/codewindow-nvim.lua
```diff
@@ -0,0 +1,39 @@
...
+        "<leader>um",
+        function()
+          require("codewindow").toggle_minimap()
+        end,
+        desc = "Toggle minimap",
+      },
+    },
+    init = function()
+      require("codewindow")
+    end,
```
dpetka2001, line 15: I don't believe these lines are needed.

dpetka2001: Should've made a code suggestion. Sorry about that.

gacallea: I tested your changes and it stops working.

dpetka2001: Are you sure? Because I've also tested locally your PR and it works just fine.

gacallea: ok I see: it stops auto-opening but it is there. It bothers me having to manually open it when the setup should really do that for me. (opts auto enable). not sure what to think of this.

dpetka2001: You just press `<leader>um` and it opens up.

gacallea: that is exactly what I mean. I want it to open automagically for me without using the shortcut. auto enable should have that behavior but it doesn't work unless I add the init code.

dpetka2001: That happens because the plugin isn't loaded, but gets lazy-loaded on the press of keymap. You could put `lazy = false`, if you want that behavior. But LazyVim lazy-loads most of the prebuilt plugins. From `~/.config/nvim/lua/config/lazy.lua`
>     -- By default, only LazyVim plugins will be lazy-loaded. Your custom plugins will load during startup.

So, I believe we should adhere to LazyVim's default principles and let the user change that setting in his personal configuration if he likes. 

But I'm not the one who makes decisions, so let's just wait on the maintainer's input regarding this.

dpetka2001: But if you want that behavior, I'd suggest to replace those lines with `lazy = false` instead. It aligns better with how you set up `lazy.nvim`.

gacallea: I am going ahead and replacing those lines with lazy=false and see what maintainers think about this. I generally agree, but hope this gets approved. (unlikely).

dpetka2001: I don't mind either way. I was just stating how LazyVim operates regarding its prebuilt plugins. Please don't take my suggestion personally.

gacallea: no offense or taking personally here :) I perfectly understand your intentions and appreciate the clarification. I agree with you on all you said :) (but still hope to have one less plugin to customize think about).

MystikReasons:
> [codewindow](https://github.com/gorbit99/codewindow.nvim) is a very light and unobtrusive minimap.
> 
> Of the many I tried this does the best job at both being a great minimap and not getting in a coders way.
> 
> I have also added keymappings `<leader>um` to toggle it.

Love this idea!
I noticed that the package [codewindow.nvim](https://github.com/gorbit99/codewindow.nvim) hasn't been updated in 8 months. Is there a reason why you chose codewindow.nvim for example over [mini.map](https://github.com/echasnovski/mini.map)?
Lazyvim already has many plugins from mini integrated.

gacallea:
Hi @MystikReasons 

I can't recall if I tried mini-map but I chose codewindow over the few I tried over the years because it can work directly in the code window instead of needing a separate "panel". This way the right panel I can use for Symbols and still have the minimap in my code. I hope it makes sense.

dpetka2001:
Just tried `mini.map` and indeed it draws its window right above the Symbols panel when you have both open. You would have to do `toggle_side()` to change the side where `mini.map` appears.

PR #3315, opened 2024-05-26
fix(lsp): properly get buffer for dynamic capabilities

lawrenceho:
When getting `buffer` from `client.attached_buffers` with `ipairs`, the loop and the `LspDynamicCapability` autocmds may not run at all because buffer numbers may not be sequential numbers starting at 1. In such case, Goto Definition (`gd`), Rename (`<leader> cr`) and inlay hints are missing when jdtls is being loaded for the first time.

Getting `buffer` with `pairs` would fix the issue.


folke:
Thanks!

PR #3319, opened 2024-05-26
fix(vimtex): suppress treesitter warning

op3:
Currently, vimtex warns:

    syntax highlighting is controlled by Treesitter

Explicitly disabling syntax higlighting disables the warning.


folke:
The latex treesitter parser is disabled in the tex extra, so you should not see that warning.
The reason you see it is because you override the nvim-treesitter spec somewhere in your config.

cyscarinci:
On a fresh install of LazyVim, with only { import = "lazyvim.plugins.extras.lang.tex" } added to lazy.lua file, and no other custom configurations, I am also getting the Vimtex warning above. How can one disable the warning while using Vimtex's native syntax highlighting?

op3:
Are you sure that the latex treesitter parser is not installed? You can check using `:TSInstallInfo`. If it is, you can uninstall it using `:TSUninstall latex`, and then the native Vimtex syntax highlighting is used.

cyscarinci:
Got it! The parser was indeed installed...

Thanks for the help!!

PR #3320, opened 2024-05-26
feat(lsp): additional reference keymaps

dotfrag:
After getting rid of `vim-illluminate`, we lost the keymaps `<a-n>` and `<a-p>` for reference jumping. This might be a personal preference but I miss these mappings as they are easier to spam/repeat compared to `[[` and `]]`.

EDIT: Future note: With illuminate, pressing `<a-n>` while on the last reference, it would jump to the first one. This was a nice feature that is not supported by the `lsp.words.jump` method. Would be nice to have it.


folke:
I've refactored the underlying code a bit and implemented those keymaps with cycling.

Thanks!

PR #3323, opened 2024-05-26
fix(lsp.keymap) keymap desc typo

arthur-mountain:


folke:
that's intentional. not a typo

arthur-mountain:
Oh, I see. Thanks!!!

PR #3331, opened 2024-05-27
fix(typescript): invalid value in inlayHints options

benelan:
Change the [`includeInlayParameterNameHints`](https://github.com/microsoft/TypeScript/blob/27bcd4cb5a98bce46c9cdd749752703ead021a4b/src/server/protocol.ts#L3582) option's value to "literals" from "literal", which is invalid.

ref: https://github.com/typescript-language-server/typescript-language-server/blob/master/docs/configuration.md#workspacedidchangeconfiguration


folke:
Thanks!

PR #3338, opened 2024-05-27
fix(util.toggle): In accordance with #3202

dpetka2001:
With this change we can disable inlay hints only in a single buffer without disabling them globally (as the current state is).

Some things to consider:
  - Do we want to consider both `bufstate` and `globalstate`? In this case, maybe we should not merge this and  just add 2 keymaps, one where we pass `buffer` (`bufstate`) and one where we don't pass anything (`globalstate`)? But in `nvim-lspconfig` spec we pass `buffer` to toggle them on (`bufstate`), but I guess that only has effect on the `is_enabled` when checking for the value.
  - Or just don't merge this and keep enabling/disabling globally only.

### DIFF
--- a/lua/lazyvim/util/toggle.lua
+++ b/lua/lazyvim/util/toggle.lua
@@ -69,7 +69,7 @@ function M.inlay_hints(buf, value)
     if value == nil then
       value = not ih.is_enabled({ bufnr = buf or 0 })
     end
-    ih.enable(value, { bufnr = buf })
+    ih.enable(value, { bufnr = buf or 0 })
   end
 end
 


folke:
I added a way to exclude certain filetypes from auto inlay hints:
https://github.com/LazyVim/LazyVim/commit/e68ff6897eb8e90ec08f7400d3271a8b4fb1da1e

dpetka2001:
Yes, I certainly don't mind. I only found out about this when trying to come up with an autocmd to disable inlay hints only for vue types. Otherwise I wouldn't have noticed. And now that I know what's going on, it's really easy for someone to just create a keymap in their personal configuration to toggle them on/off only for the buffer if they'd like.

I just wasn't sure how LazyVim would want to handle this, that's why I shared my concerns about this as well. Thanks!!

PR #3366, opened 2024-05-30
fix: vtls add select ts version

rubiin:
Adds a keymap to select typescript version


rubiin:
Also noticed we deleted treesitter config on this file, was that intentional.

rubiin:
> `typescript` and `tsx` already included in the default [treesitter spec](https://github.com/LazyVim/LazyVim/blob/2e5e46d9ff8ea11c805a40dec5cdb2beb9272c6c/lua/lazyvim/plugins/treesitter.lua#L48-L49).

oh nice.

folke:
Thanks!

PR #3379, opened 2024-05-30
fix(extras): `underline = false` for setting diagnostics

dpetka2001:
Similar to https://github.com/folke/lazy.nvim/commit/ea7b9c3c3fd9026e1a5ae27950585df9a42ccd5b# 

I only noticed this, because in my main config I have some Extras not managed by `LazyExtras`.


folke:
Thanks!

PR #3385, opened 2024-05-30
feat(config): custom module location for config

vdbe:
Add an extra options to LazyVim for custom config module location.

For example this layout would set modname to `full.config`.

```
.
└── lua
    ├── full
    │   ├── config
    │   │   ├── autocmds.lua
    │   │   ├── init.lua
    │   │   ├── keymaps.lua
    │   │   └── options.lua
    │   └── plugins
    └── noPlugins
```

### DIFF
--- a/lua/lazyvim/config/init.lua
+++ b/lua/lazyvim/config/init.lua
@@ -20,6 +20,9 @@ local defaults = {
     -- lazyvim.config.options can't be configured here since that's loaded before lazyvim setup
     -- if you want to disable loading options, add `package.loaded["lazyvim.config.options"] = true` to the top of your init.lua
   },
+  -- module containing autocmds, keymaps and options
+  ---@type string
+  modname = "config",
   news = {
     -- When enabled, NEWS.md will be shown when changed.
     -- This only contains big new features and breaking changes.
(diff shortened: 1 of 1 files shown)


dpetka2001:
Why?

vdbe:
Because I have multiple configs which I like to keep separate (1 directory for each in ./lua). To do that without this change you need to set `default.{autocmds,keyamps}` to false which is fine you can load this manually later, the way to do this for options feels however a bit hacky (`package.loaded["lazyvim.config.options"] = true`)

dpetka2001:
To have multiple configs you take advantage of `:h $NVIM_APPNAME`. My humble personal opinion as a regular user like you, is that what you're proposing is out of scope. LazyVim is supposed to be a single point of user configuration.

vdbe:
When using $NVIM_APPNAME you need to symlink if you have shared parts in the config (and is a bit harder to contain in a single repo)

dpetka2001:
The maintainer has the final say, but in my opinion this seems out of scope with the purpose of a Neovim distro. In this case, you might just roll your own config or fork the LazyVim repo yourself while keeping up with latest changes.

folke:
yeah, not a fan of this.
Just use `$NVIM_APPNAME` like @dpetka2001 mentions

PR #3389, opened 2024-05-30
fix(rust): let rustaceans.nvim setup rust_analyzer

mirsella:
fix this warning
```
   Warn  11:52:45 PM notify.warn nvim-lspconfig.rust_analyzer has been setup.
This will likely lead to conflicts with the rustaceanvim LSP client.
See ':h rustaceanvim.mason'
```
fix taken from https://github.com/mrcjkb/rustaceanvim/blob/master/doc/mason.txt

### DIFF
--- a/lua/lazyvim/plugins/extras/lang/rust.lua
+++ b/lua/lazyvim/plugins/extras/lang/rust.lua
@@ -103,6 +103,11 @@ return {
   {
     "neovim/nvim-lspconfig",
     opts = {
+      setup = {
+        rust_analyzer = function()
+          return true
+        end,
+      },
       servers = {
         taplo = {
           keys = {


mirsella:
oh this actually doesn't work as it then use rust_analyzer in $PATH, I'll submit another PR with it explicitly using mason's rust_analyzer

dpetka2001:
This was explicitly removed by #2755, so that users install `rust-analyzer` via Rust toolchain.

gibfahn:
I hit the same issue, and what I needed to do to work around it was in nvim:

```vim
:MasonUninstall rust-analyzer
```

and then in the shell:

```shell
rustup component add rust-analyzer
```

<details><summary>Full output of error messages in case useful:</summary>

On updating I see:

```
⚠️   Warn  10:07:54 notify.warn nvim-lspconfig.rust_analyzer has been setup.
This will likely lead to conflicts with the rustaceanvim LSP client.
See ':h rustaceanvim.mason'
```

The help notes that you probably want to remove the Mason rust-analyzer installation, but also includes this block (which should probably be removed now https://github.com/LazyVim/LazyVim/pull/2755 has landed.

```
Using LazyVim:

>lua
  {
    'neovim/nvim-lspconfig',
    opts = {
      setup = {
        rust_analyzer = function()
          return true 
        end,
      },
    },
  }
```

Instead you want to uninstall rust-analyzer via Mason, and instead use the one that comes with rustup (as per the rustaceanvim help text):

```
:MasonUninstall rust-analyzer
```

---

However now I see this when I reopen:

```
 ⚠️   Warn  10:12:01 notify.warn Client rust-analyzer quit with exit code 1 and signal 0. Check log for errors: /Users/gib/.local/state/nvim/lsp.log
```

And the log says:

```
[ERROR][2024-01-26 08:51:21] .../vim/lsp/rpc.lua:31	"rpc"	"/Users/gib/.local/share/nvim/mason/bin/rust-analyzer"	"stderr"	'[ERROR rust_analyzer::dispatch] unknown request: Request { id: RequestId(I32(972)), method: "workspace/executeCommand", params: Object {"title": String("triggerParameterHints"), "command": String("rust-analyzer.triggerParameterHints")} }\n'
```

Which means something is caching the path to rust-analyzer that I just removed:

```console
❯ ls -l /Users/gib/.local/share/nvim/mason/bin/rust-analyzer
cannot access '/Users/gib/.local/share/nvim/mason/bin/rust-analyzer': No such file or directory
```

The fix for this ended up being that the `rust-analyzer` in my path was a broken link. Tested this by running `rust-analyzer --version`, which errored. Solution was to reinstall it using `rustup` with:

```shell
rustup component add rust-analyzer
```

</details>

folke:
This is indeed a good idea. Did something similar for typescript

folke:
This at least ensures the rust extra wont give errors no?
I also never had issues with the mason installed rust-analyzer.

Maybe a better approach then would be to check if `rust-analyzer` was installed through mason and show an eror that the user should uninstall?

folke:
Yep, you're right. Just reverted this PR!

jpx40:
@folke 

hey, i don't really like the direction rustaceanvim is going. I would start suggest forking Rust-Tools and start from there. Revert to the as Rust-Tools was used. I would be ok maintaining it, i wouldn't add any new features to  Rust-Tools just keep it working. The Rust LSP is good enough, we don't need rustaceanvim. I don't like it that they break compatibly with mason

dpetka2001:
They don't break compatibility with Mason. They just don't use it by default, because the maintainer thinks it's better to use the rust-analyzer installed by Rust toolchain. Users can still change in their personal configuration the `cmd` to point to the rust-analyzer installed by Mason (but they do that on their own risk and assuming they know how to solve potentially problems by mismatched version of rust-analyzer and the Rust toolchain).

The reason this was implemented like this in LazyVim, was to avoid problems for new users that thought it would work out of the box with the Mason rust-analyzer and to avoid any possible problems by using a different version of rust-analyzer than your current toolchain. In #2755 other users have also given their opinion on what potential problems might arise by having mismatched version of rust-analyzer and the toolchain itself. One user even linked a comment from an issue at rust-analyzer repo that arose due to mismatching versions of rust-analyzer and toolchain itself.

In my personal opinion, this is the easiest way to avoid issues from beginner users and those who are more advanced and know what they're doing can change the settings in their personal configuration.

jpx40:
> They don't break compatibility with Mason. They just don't use it by default, because the maintainer thinks it's better to use the rust-analyzer installed by Rust toolchain. Users can still change in their personal configuration the `cmd` to point to the rust-analyzer installed by Mason (but they do that on their own risk and assuming they know how to solve potentially problems by mismatched version of rust-analyzer and the Rust toolchain).
> 
> The reason this was implemented like this in LazyVim, was to avoid problems for new users that thought it would work out of the box with the Mason rust-analyzer and to avoid any possible problems by using a different version of rust-analyzer than your current toolchain. In #2755 other users have also given their opinion on what potential problems might arise by having mismatched version of rust-analyzer and the toolchain itself. One user even linked a comment from an issue at rust-analyzer repo that arose due to mismatching versions of rust-analyzer and toolchain itself.
> 
> In my personal opinion, this is the easiest way to avoid issues from beginner users and those who are more advanced and know what they're doing can change the settings in their personal configuration.

you are right it just annoying that i breaks my personal Config, i haven't experienced mature problems because of the Versions yet

dpetka2001:
You can do whatever you want in your personal config though. I just don't think this is a good default for a Neovim distro, given the number of issues i've seen created about this on this repo alone. It's better for beginners to just install rust-analyzer directly from rust toolchain. And whenever they update their toolchain, rust-analyzer will also get updated. Users who know what they're doing can just change defaults in their personal config. Less issues for everyone this way.

jpx40:
> You can do whatever you want in your personal config though. I just don't think this is a good default for a Neovim distro, given the number of issues i've seen created about this on this repo alone. It's better for beginners to just install rust-analyzer directly from rust toolchain. And whenever they update their toolchain, rust-analyzer will also get updated. Users who know what they're doing can just change defaults in their personal config. Less issues for everyone this way.

yeah i think it the best for me, i relatively new to Neovim (6 Month i think), my Problem is that i update quite a lot breaking Changes are annoying. I starting to realise it's the best if i try to do my own because i have my own Personal way of doing things so have not to Rely on other Plugins. That works best for me.

jpx40:
> You can do whatever you want in your personal config though. I just don't think this is a good default for a Neovim distro, given the number of issues i've seen created about this on this repo alone. It's better for beginners to just install rust-analyzer directly from rust toolchain. And whenever they update their toolchain, rust-analyzer will also get updated. Users who know what they're doing can just change defaults in their personal config. Less issues for everyone this way.

i disabled the Rust Plugin and implemented the Features i needed for me in my own Config i think i am fine now.

dpetka2001:
@jpx40 Please do keep in mind that I'm just a regular user like you. I'm just stating my personal opinion based on my experience. The maintainer has the final say. If you'd like to do another Extra with rust-tools forked and maintained by you, then by all means do so and the maintainer will decide. But if you choose to do so, make it at least a different Extra, so that users will be able to make their choice.

jpx40:
> @jpx40 Please do keep in mind that I'm just a regular user like you. I'm just stating my personal opinion based on my experience. The maintainer has the final say. If you'd like to do another Extra with rust-tools forked and maintained by you, then by all means do so and the maintainer will decide. But if you choose to do so, make it at least a different Extra, so that users will be able to make their choice.

oh no it was just an recomandation, i personally rely on some features like inline hints that were removed from rustaceanvim. Rust-tools had them still. There lsp  inline hints  were added with 0.10 of neovim but i found 0.10 a little buggy for so i will use 0.95 still for a while. It was just for my personal, but if other people wanna go back rust-tools they could use my fork if they want.

folke:
inlay hints work perfectly fine with rustaceanvim and as @dpetka2001 says, the incompatibility with the mason installed rust-analyzer is to be expected, since your local toolchain might be different.

jpx40:
> inlay hints work perfectly fine with rustaceanvim and as @dpetka2001 says, the incompatibility with the mason installed rust-analyzer is to be expected, since your local toolchain might be different.

yeah in 0.10 but i use 0.95 of neovim. My current config isn't compatible with. 0.10 of neovim. I made sure to disable this plugin and get all the stuff i need to my own Config so i am independent from the rust plugin now. It bothered me before because this broke a little bit of my config, but it's ok now for me.

RazaGR:
> I hit the same issue, and what I needed to do to work around it was in nvim:
> 
> ```viml
> :MasonUninstall rust-analyzer
> ```
> 
> and then in the shell:
> 
> ```shell
> rustup component add rust-analyzer
> ```
> 
> Full output of error messages in case useful:

I was seeing this issue and this helped

PR #3391, opened 2024-05-31
fix(ui): do not add trouble to lualine when aerial is enabled

Sped0n:
image


folke:
I've changed it so you can set `vim.g.trouble_lualine = false` in your config instead

PR #3397, opened 2024-05-31
feat(extras): Add PHP support

NoctuaDevelopment:
Add the required LSP for PHP with a condition on the php version to know which lsp to use.

### DIFF
--- a/lua/lazyvim/plugins/extras/lang/php.lua
+++ b/lua/lazyvim/plugins/extras/lang/php.lua
@@ -1,0 +1,71 @@
+local handle = io.popen("php -v")
+if handle ~= nil then
+    local result = handle:read("*a")
+    io.close(handle)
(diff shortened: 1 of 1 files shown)


folke:
This will run `php` every time Neovim starts, which is not acceptable.

Please change it similar to what we do in the python extra. The top section is added to the docs automaically.

So `vim.g.lazyvim_php_lsp = the default`.

In your own config you'd still be able to set that opton based on php -v

muneebusmani:
bruh there is no need to check php version for installing either phpactor or intelephense, because legacy php code is also supported by intelephense, they support down to php 5.3 , and i suggest adding psalm in ensure installed because it is eslint of php and add pretty-php instead of php cs-fixer because it is prettier for php , but there is an issue for psalm, you have to run psalm --init to make it work

PR #3401, opened 2024-05-31
fix(extras.lang.vue) missing typescript autocomplete in .vue files

eilefsen:
The existing Vue extra was not properly configuring vtsls to use the globalPlugin: `@vue/typescript-plugin`

This meant that even though the language server could load and provide hover info, it wouldn't provide any sort of autocomplete.

This commit fixes that by moving the `@vue/typescript-plugin` config to the vtsls settings and adding the line `configNamespace = "typescript"`. The fix does not work without the mentioned line.

I dont understand the intention behind LazyVim.extend() in the original code, given that the function takes two arguments and not three as provided. It also doesn't seem necessary to use any extend function to begin with either.

### DIFF
--- a/lua/lazyvim/plugins/extras/lang/vue.lua
+++ b/lua/lazyvim/plugins/extras/lang/vue.lua
@@ -39,12 +39,10 @@ return {
         },
       })
       LazyVim.extend(opts.servers.vtsls.settings.vtsls.tsserver.globalPlugins, {
-        {
-          name = "@vue/typescript-plugin",
-          location = vue_typescript_plugin,
-          languages = { "vue" },
-          configNamespace = "typescript",
-        },
+        name = "@vue/typescript-plugin",
+        location = vue_typescript_plugin,
+        languages = { "vue" },
+        configNamespace = "typescript",
       })
     end,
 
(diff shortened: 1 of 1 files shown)


folke:
`globalPlugins` is a list, so by defining it the way you do, it will overwrite any other definition of  `globalPlugins`.

I still need to add the correct tyescript plugin for astro, so when having both extras enabled in a way like you did, it wouldn't work as intended

lua/lazyvim/plugins/extras/lang/vue.lua
```diff
@@ -38,12 +38,16 @@ return {
           },
         },
       })
-
-      LazyVim.extend(opts.servers.vtsls, "settings.vtsls.tsserver.globalPlugins", {
-        name = "@vue/typescript-plugin",
-        location = vue_typescript_plugin,
-        languages = { "vue" },
+      LazyVim.extend(opts.servers.vtsls.settings.vtsls.tsserver.globalPlugins, {
+        {
```
folke, line 42: This should be one value, not a list of values, so you shoudl remove one set of `{}`

folke (changes requested):

lua/lazyvim/plugins/extras/lang/vue.lua
folke: Seems like I actually did it wrong, so that was my mistake

folke:
Merged, thanks!

PR #3403, opened 2024-05-31
feat(bigfile): add bigfile module

rubiin:
This plugin automatically disables certain features if the opened file is big. File size and features to disable are configurable.

Useful for viewing/editing bigfiles like logs

### DIFF
--- a/lua/lazyvim/plugins/extras/editor/bigfile.lua
+++ b/lua/lazyvim/plugins/extras/editor/bigfile.lua
@@ -1,19 +1,7 @@
 return {
-  "LunarVim/bigfile.nvim",
-  event = { "BufReadPre", "BufNewFile" },
-  opts = {
-    pattern = function(bufnr, filesize_mib)
-      if filesize_mib >= 2 then
-        return true
-      end
-      -- you can't use `nvim_buf_line_count` because this runs on BufReadPre
-      local file_contents = vim.fn.readfile(vim.api.nvim_buf_get_name(bufnr))
-      local file_length = #file_contents
-      if file_length > 5000 then
-        return true
-      end
-
-      return false
-    end,
-  },
+	"LunarVim/bigfile.nvim",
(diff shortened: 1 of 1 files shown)


lua/lazyvim/plugins/extras/editor/bigfile.lua
```diff
@@ -0,0 +1,19 @@
+return {
+  "LunarVim/bigfile.nvim",
+  event = { "BufReadPre", "BufNewFile" },
+  opts = {
+    pattern = function(bufnr, filesize_mib)
+      if filesize_mib >= 2 then
+        return true
+      end
```
dpetka2001, line 8:
```suggestion

```
This is already default and is used as fallback when the function returns `false`. Did you experience otherwise and decided to put this here? From `docs`
> You can add your own logic for detecting big files by setting pattern in the config to a function. If the function returns true file will be considered big, otherwise filesize will be used as a fallback

dpetka2001: Then you should use the `filesize` option, since it wouldn't be hardcoded in the pattern and the users can easily change it in their personal config in my personal opinion.

rubiin: thats a nice suggestion . I will do that

folke:
I just added support for big files in LazyVim.
tbh, my implementation is alot better than what `bigfile.nvim` does.
That plugin disables stuff after the fact.
My impementation prevents starting anything at all by ensuring it gets the filetype `bigfile`.

For more info, see https://github.com/LazyVim/LazyVim/commit/938a6718c6f0d5c6716a34bd3383758907820c52

rubiin:
Thats awesome :clinking_glasses:

ctruett:
How do I disable this? It's incorrectly detecting a 15 kb file as being huge and severely limiting what I can do inside of it.

PR #3411, opened 2024-05-31
gco and gcO create conflicts inside Neovim as both have gc prefix

rahulgavhar:
WhichKey: checking conflicting keymaps ~
- WARNING conflicting keymap exists for mode **"n"**, lhs: **"gc"**
- rhs: `

The above message is displayed at the very bottom, when :checkhealth is run in nvim.
As every user will face this conflict due to the "gc" prefix in both of them.


folke:
that's just a warning. all good

rahulgavhar:
How can I turn off this warning can you help out me

folke:
Why? It's a warning. It does literally nothing.

PR #3424, opened 2024-06-02
fix(todo-comments): use toggle for Trouble todo commands

0xN1nja:
Using `toggle` with `Trouble todo` so that it can close Trouble with the same keybinding.


folke:
Thanks!

PR #3430, opened 2024-06-03
Update mini-hipatterns.lua config to add support for heex, rust and svelte files

Alt-iOS:
added support for heex, rust and svelte to hipatterns files in the config for min.hipatterns


folke:
Thanks!

PR #3432, opened 2024-06-03
feat(editor-extra): Add inc-rename

briandipalma:
Not sure how to safely move the LSP keymap into this file `<leader>cr`...

Using

`LazyVim.lsp.on_attach(function(client, buffer)`?

but how do I ensure that the function in the config for this extra gets called after the default lsp keymap one?

### DIFF
--- a/lua/lazyvim/plugins/extras/editor/inc-rename.lua
+++ b/lua/lazyvim/plugins/extras/editor/inc-rename.lua
@@ -1,14 +1,19 @@
 return {
+
+  -- Rename with cmdpreview
+  recommended = true,
+  desc = "Incremental LSP renaming based on Neovim's command-preview feature",
   {
     "smjonas/inc-rename.nvim",
     cmd = "IncRename",
-    config = true,
+    opts = {},
   },
+
+  -- LSP Keymaps
   {
     "neovim/nvim-lspconfig",
-    init = function()
+    opts = function()
       local keys = require("lazyvim.plugins.lsp.keymaps").get()
-
       keys[#keys + 1] = {
         "<leader>cr",
         function()
@@ -16,9 +21,18 @@ return {
           return ":" .. inc_rename.config.cmd_name .. " " .. vim.fn.expand("<cword>")
         end,
         expr = true,
-        desc = "Rename",
+        desc = "Rename (inc-rename.nvim)",
         has = "rename",
       }
     end,
+  },
+
+  --- Noice integration
+  {
+    "folke/noice.nvim",
+    optional = true,
+    opts = {
+      presets = { inc_rename = true },
+    },
   },
 }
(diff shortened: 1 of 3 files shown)


dpetka2001:
Just add a `nvim-lspconfig` spec that changes the `<leader>cr` keymap and remove it from `lazyvim/plugins/lsp/keymaps.lua`? I think that should work.

folke:
LazyVim actually already includes special handling for inc-rename! :)

Just add 

```lua
  {
    "smjonas/inc-rename.nvim",
    cmd = "IncRename",
    config = true,
  },
```

Nothing else is needed.

dpetka2001:
I agree nothing else is needed, just made the suggestion because OP probably had in mind separation of concerns regarding the keymap, so that it would be introduced via this Extra instead of directly in `lsp.keymaps`.

briandipalma:
> Just add a `nvim-lspconfig` spec that changes the `<leader>cr` keymap and remove it from `lazyvim/plugins/lsp/keymaps.lua`? I think that should work. So something like
> 
> ```lua
>   {
>     "neovim/nvim-lspconfig",
>     init = function()
>       local keys = require("lazyvim.plugins.lsp.keymaps").get()
>       keys[#keys + 1] = {
>         "<leader>cr",
>         function()
>           local inc_rename = require("inc_rename")
>           return ":" .. inc_rename.config.cmd_name .. " " .. vim.fn.expand("<cword>")
>         end,
>         expr = true,
>         desc = "Rename",
>         has = "rename",
>       }
>     end,
>   },
> ```

I'm blind, I didn't see this:

```lua
  if M._keys then
    return M._keys
  end
```

thanks for the suggestion @dpetka2001 I've updated the PR. Does this clear down the old rename keymap though? I'm not too familiar with Lua.

dpetka2001:
It should overwrite it, like with other keys as well. But, I agree with the maintainer this is not really needed, since it's already present in `lsp.keymaps`. The maintainer has the final say.

folke:
I've refactored the code a bit and moved the noice integration.

folke:
Thanks!

lua/lazyvim/plugins/extras/editor/inc-rename.lua
```diff
@@ -1,24 +1,38 @@
...
     cmd = "IncRename",
-    config = true,
+    opts = {},
   },
+
+  -- LSP Keymaps
   {
     "neovim/nvim-lspconfig",
-    init = function()
+    opts = function()
```
dpetka2001, line 15: @folke Is it alright for it to be changed to `opts` instead of `init`? I thought the LSP keymaps had to be defined in `init` according to the docs.

folke: opts is fine if it's part of the lspconfig spec, since that will be evaluated before lspconfig's config will execute.

folke: I've updated the docs. Thanks!

dpetka2001: Cool!! Thank you for clarifying.

PR #3446, opened 2024-06-04
fix(lsp): Enable using `deno.enablePaths` in denols settings

eriklindebratt:
denols LSP client has support for enabling using Deno in _parts_ of a codebase using the `deno.enablePaths` setting. For example, the following setting would enable denols in all files matching the `./deno-files` path:

```lua
settings = {
  deno = {
    enablePaths = { "./deno-files" }
  }
}
```

If we return `true` in the callback to `LazyVim.lsp.disable("denols")` denols will be completely disabled and can never pick up the `deno.enablePaths` setting.

In this PR I've changed it so that we – when a Deno root can't be found, instead disable Deno using its LSP client setting `deno.enable`. This allows denols to still pick up `deno.enablePaths`, and that setting will then take precedence over `deno.enable` for the files matched.

I'm not sure this is the best way way to please let me know if there's a better way.


folke:
LGTM, thanks!

PR #3455, opened 2024-06-05
fix(extra.lang.scala): Fix the always initializing regardless of file…

gpwns3717:
…type


folke:
Thanks!

And I only now saw that vsnip got in there, so thank you for removing that!

PR #3458, opened 2024-06-05
feat(extras): Add extra for vim language

imroc:


lua/lazyvim/plugins/extras/lang/vim.lua
```diff
@@ -0,0 +1,30 @@
...
+      end
+    end,
+  },
+  {
+    "williamboman/mason.nvim",
+    opts = function(_, opts)
+      opts.ensure_installed = opts.ensure_installed or {}
+      vim.list_extend(opts.ensure_installed, { "vim-language-server" })
+    end,
+  },
```
folke, line 21: This is not needed. Will be installed automatically when `vimls` is in your lspconfig opts

lua/lazyvim/plugins/extras/lang/vim.lua
```diff
@@ -0,0 +1,30 @@
...
+    })
+  end,
+  {
+    "nvim-treesitter/nvim-treesitter",
+    opts = function(_, opts)
+      if type(opts.ensure_installed) == "table" then
+        vim.list_extend(opts.ensure_installed, { "vim", "vimdoc" })
+      end
+    end,
+  },
```
folke, line 14: `vim` and `vimdoc` (`help`) are already part of the installed treesitter langs

folke:
After resolving my comments, this extra would just install `vimls`.
So there's no real need for this to be an extra?
You can just install `vimls` with `:Mason` and you'd have the same setup as in this extra

imroc:
> After resolving my comments, this extra would just install `vimls`. So there's no real need for this to be an extra? You can just install `vimls` with `:Mason` and you'd have the same setup as in this extra

Yep, close PR then

PR #3464, opened 2024-06-05
feat: add undo tree

rubiin:
This is the continuation of  https://github.com/LazyVim/LazyVim/pull/3148
The old pr got deleted when I deleted the main branch

### DIFF
--- a/lua/lazyvim/plugins/extras/editor/undotree.lua
+++ b/lua/lazyvim/plugins/extras/editor/undotree.lua
@@ -1,0 +1,13 @@
+return {
+  {
+    "mbbill/undotree",
+    cmd = "UndotreeToggle",
+    keys = {
+      { "<leader>uu", "<cmd>UndotreeToggle<cr>", desc = "Undo Tree Toggle" },
+    },
+    init = function()
+      -- set layout style to 2, let g:undotree_WindowLayout = 2
+      vim.g.undotree_WindowLayout = 2
+    end,
+  },
+}


surmish:
Really don't think `LazyVim` should be adding this, `telescope-undo.nvim` is an option that people tend to use too. This looks like it belongs in personal config area.

rubiin:
Yeah I was gonna go for that but this seems to be more popular among the developers

muneebusmani:
Bruh telescope isn't everything, undotree is the goat, it can easily integrate into edgy

surmish:
> Bruh telescope isn't everything, undotree is the goat, it can easily integrate into edgy

Telescope-undo isn't a part of LazyVim either. See the requirements for adding an extra. Straightforward configurations don't belong in an extra.

rubiin:
If we were to add telescope-undo, I think it wont work with fzf as fzf replaces telescope. This pr is a continuation of https://github.com/LazyVim/LazyVim/pull/3148, the configurations that undotree requires was stripped down

dpetka2001:
Honestly, in my personal opinion, this is a really straightforward configuration and neither of the plugins mentioned here should belong in a Neovim distro I believe. It would be best if users could choose whichever one they like and just install it in their personal configuration. 

If it were to make it into an Extra I would also prefer something that's not telescope related. I've fully switched into fzf-lua Extra since it got merged yesterday. Maintainer has the final say.

kevintraver:
There is a lua version that might also be worth checking out as well: [https://github.com/jiaoshijie/undotree](https://github.com/jiaoshijie/undotree)

PR #3466, opened 2024-06-05
fix(svelte): added organize imports and prettier

matadaniel:
I'm not sure if prettier should be set up for the svelte filetype here in the svelte extra or in the formatting.prettier extra.


folke:
Thanks!

folke:
I reverted that prettier change, since that doesn't work out of the box.
You need to have prettier-plugin-svelte installed

PR #3468, opened 2024-06-05
fix(extra): fix lost ensured_installed issue after release

gpwns3717:
I found an issue after the 12.5.0 release where lists were not being merged in the type(opts.ensure_installed) == “table” condition.


dpetka2001:
Are you sure it's not something on your end? Because I tried with default LazyVim starter template and tested Python and clangd Extras and both Mason and nvim-treesitter `ensure_installed` got installed correctly.

gpwns3717:
@dpetka2001
You're right, in my personal treesitter parser I explicitly added it to ensured_install and it didn't work.
I'll close this PR, thanks.

PR #3469, opened 2024-06-05
fix(extra.lang.angular): treesitter not enabled for angular templates

mishasbtr:
## Problem
Treesitter was not working in angular html templates after enabling `lang.angular` extra.

## Solution
Added a change recommeded by angular treesitter developer to enable treesitter for components and containers.

### Reference to recommendation in treesitter docs
[Reference to the recommendation for this hack](https://github.com/dlvandenberg/tree-sitter-angular?tab=readme-ov-file#filetype).

### DIFF
--- a/lua/lazyvim/plugins/extras/lang/angular.lua
+++ b/lua/lazyvim/plugins/extras/lang/angular.lua
@@ -20,12 +20,7 @@ return {
           [".*%.container%.html"] = "angular.html",
         },
       })
-      vim.api.nvim_create_autocmd("FileType", {
-        pattern = "angular.html",
-        callback = function()
-          vim.treesitter.language.register("angular", "angular.html")
-        end,
-      })
+      vim.treesitter.language.register("angular", "angular.html")
     end,
   },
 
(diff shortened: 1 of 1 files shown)


lua/lazyvim/plugins/extras/lang/angular.lua
```diff
@@ -14,6 +14,18 @@ return {
...
+        pattern = {
+          [".*%.component%.html"] = "angular.html",
+          [".*%.container%.html"] = "angular.html",
+        },
+      })
+      vim.api.nvim_create_autocmd("FileType", {
+        pattern = "angular.html",
+        callback = function()
+          vim.treesitter.language.register("angular", "angular.html")
+        end,
```
folke, line 27: IS this really needed to do in an autocmd?
Why not just register it?

mishasbtr: Indeed, you are right, this is not necessary.
Removed the wrapping of registration with autocmd.

folke:
Thanks!

BBboy01:
is your lsp still working after setting the filetype to `angular.html`

mishasbtr:
> is your lsp still working after setting the filetype to `angular.html`

Thank you for catching that! Indeed, lsp stops working after changing the filetype.
Working on a fix.

mishasbtr:
@BBboy01 
PR with a fix: https://github.com/LazyVim/LazyVim/pull/3485

mishasbtr:
Re-fixed in https://github.com/LazyVim/LazyVim/pull/3488

PR #3481, opened 2024-06-05
fix(ui): remove extra space

aimuzov:
Remove extra space between icon and text (for source label).

## Before:
image

## After:
image

### DIFF
--- a/lua/lazyvim/util/extras.lua
+++ b/lua/lazyvim/util/extras.lua
@@ -267,7 +267,7 @@ function X:extra(extra)
     self.text:append(" "):append(LazyConfig.options.ui.icons.favorite or " ", "LazyCommit")
   end
   if extra.source.name ~= "LazyVim" then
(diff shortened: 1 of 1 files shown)


folke:
That extra space in on purpose. See how it looks like in wezterm:

image

lua/lazyvim/util/extras.lua
```diff
@@ -267,7 +267,7 @@ function X:extra(extra)
     self.text:append(" "):append(LazyConfig.options.ui.icons.favorite or " ", "LazyCommit")
   end
   if extra.source.name ~= "LazyVim" then
-    self.text:append(" "):append(LazyConfig.options.ui.icons.event .. " " .. extra.source.name, "LazyReasonEvent")
+    self.text:append(" "):append(LazyConfig.options.ui.icons.event .. "" .. extra.source.name, "LazyReasonEvent")
```
folke, line 270: ok, I see it now.
Then just remove the `.. "" ` as well?

folke (changes requested):

folke:
Thanks!

PR #3485, opened 2024-06-06
fix(extra-lang-angular): fix angular lsp not working for angular.html filetype

mishasbtr:
## Problem

After adding a new `angular.html` filetype in #3469 , angular LSP stopped working for the files marked with this filetype.

## Solution

Add `angular.html` to filetypes the LSP is enabled for.

Thank you @BBboy01 for catching this issue :heart:

### DIFF
--- a/lua/lazyvim/plugins/extras/lang/angular.lua
+++ b/lua/lazyvim/plugins/extras/lang/angular.lua
@@ -32,7 +32,13 @@ return {
     "neovim/nvim-lspconfig",
     opts = {
       servers = {
-        angularls = {},
+        angularls = {
+          filetypes = {
+            -- default filetypes
+            "typescript", "html", "typescriptreact", "typescript.tsx",
+            -- file type added in order to enable treesitter
+            "angular.html" },
+        },
       },
       setup = {
         angularls = function()
--- a/lua/lazyvim/plugins/extras/lang/angular.lua
+++ b/lua/lazyvim/plugins/extras/lang/angular.lua
@@ -35,9 +35,13 @@ return {
         angularls = {
           filetypes = {
             -- default filetypes
-            "typescript", "html", "typescriptreact", "typescript.tsx",
+            "typescript",
+            "html",
+            "typescriptreact",
+            "typescript.tsx",
             -- file type added in order to enable treesitter
-            "angular.html" },
+            "angular.html",
+          },
         },
       },
       setup = {


dpetka2001:
Maybe that was what the autocommand was trying to avoid? Because LSP servers usually load on FileType as well. Did you have problems with the autocommand you removed? Not at all sure what I'm talking about. Just a guess.

BBboy01:
there are many lsp running on html filetype, such as tailwindcss-language-server, html-lsp, css-lsp, css-variables-language-server, emmet-language-server, sonarlint-language-server... it's goona be a hard work to make all lsp compatible

i dont know but maybe have some easier way to do this

mishasbtr:
> Maybe that was what the autocommand was trying to avoid? Because LSP servers usually load on FileType as well. Did you have problems with the autocommand you removed regarding the LSP server attaching? Not at all sure what I'm talking about. Just a guess.

This problem occurs with autocommand as well.

folke:
I agree with @BBboy01 this won't fix any other LSP servers that trigger on `html`.
You should probably revert your earlier PR and see if you can fix treesitter in another way

mishasbtr:
> there are many lsp running on html filetype, such as tailwindcss-language-server, html-lsp, css-lsp, css-variables-language-server, emmet-language-server, sonarlint-language-server... it's goona be a hard work to make all lsp compatible
> 
> i dont know but maybe have some easier way to do this

Yes, you're right. It'll be hard to make all other lsp compatible.
Will try to find a workaround.

folke:
I've reverted that other PR for now

folke:
Do something like:

```lua
vim.api.nvim_create_autocmd("BufRead", {
  pattern = "*.component.html",
  callback = function()
     vim.treesitter.start(0, "angular")
  end
})
```

Untested, but something like this will probably work

mishasbtr:
> Do something like:
> 
> ```lua
> vim.api.nvim_create_autocmd("BufRead", {
>   pattern = "*.component.html",
>   callback = function()
>      vim.treesitter.start(0, "angular")
>   end
> })
> ```
> 
> Untested, but something like this will probably work

@folke , thank you for reverting the previous PR breaking the LSP, sorry for this issue.
Will try the solution you proposed.

mishasbtr:
@folke , opened a PR implementing your suggestion: https://github.com/LazyVim/LazyVim/pull/3488/
Verified LSP and treesitter to be working with this implementation.

PR #3488, opened 2024-06-06
fix(extra-lang-angular): re-fix angular html treesitter

mishasbtr:
This was previously attemped to be fixed by https://github.com/LazyVim/LazyVim/pull/3469, which caused an issue with angular LSP.
The PR was reverted as mentioned in https://github.com/LazyVim/LazyVim/pull/3485.

This PR implements suggestion by @folke, which fixes treesitter without breaking LSP.

### DIFF
--- a/lua/lazyvim/plugins/extras/lang/angular.lua
+++ b/lua/lazyvim/plugins/extras/lang/angular.lua
@@ -14,13 +14,12 @@ return {
       if type(opts.ensure_installed) == "table" then
         vim.list_extend(opts.ensure_installed, { "angular", "scss" })
       end
-      vim.filetype.add({
-        pattern = {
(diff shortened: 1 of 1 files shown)


BBboy01:
`vim.treesitter.language.register("angular", "html")`

how about just simply highlight all `html` filetype with  `angular`

https://github.com/nvim-treesitter/nvim-treesitter/blob/fc9ee79f64b18bf431e862c2462d4f76d488492a/queries/angular/highlights.scm#L1

`angular` highlight is inherits from `html`, so it won't break normal `html` highlight

folke:
@BBboy01 I prefer the way it's fixed in this PR.

folke:
Thanks!

PR #3491, opened 2024-06-06
feat(extras): add supermaven

Nybkox:
Add fast copilot alternative


dotfrag:
I believe this config is better, not for any particular reason, but because it matches what LazyVim already has for the Codeium extra:

```lua
return {
  "nvim-cmp",
  dependencies = {
    "supermaven-inc/supermaven-nvim",
    build = ":SupermavenUseFree", -- this line is optional, remove if you are using pro
    opts = {},
  },
  ---@param opts cmp.ConfigSchema
  opts = function(_, opts)
    table.insert(opts.sources, 1, {
      name = "supermaven",
      group_index = 1,
      priority = 100,
    })
  end,
}
```

Also see https://github.com/LazyVim/LazyVim/discussions/3471.

dpetka2001:
I believe the current implementation of the PR is better. It's decoupling the 2 plugins from each other and if you also safeguard `nvim-cmp` with `optional = true`, then all the better. That way if there are more sources in the future it will be easier to manage.

dotfrag:
Fair enough!

lua/lazyvim/plugins/extras/coding/supermaven.lua
```diff
@@ -0,0 +1,20 @@
...
+    "supermaven-inc/supermaven-nvim",
+    config = function()
+      require("supermaven-nvim").setup({
+        keymaps = {
+          accept_suggestion = "<C-y>",
+          clear_suggestion = "<C-]>",
+          accept_word = "<C-j>",
+        },
+      })
+    end,
```
dpetka2001, line 12: I believe it would be better if this would be in `opts` instead of `config`, so that users can overwrite them in their personal configuration easier instead of having to overwrite the whole `config` function.

dpetka2001:
Also, in retrospect, I believe @dotfrag's suggestion is better because it also takes into account lazy-loading the plugin when needed.

mirsella:
All the others "ai completion" we configure them explicitly with only nvim cmp, and no inline suggestion.
Shouldn't supermaven do that too ? So that it's the same across all the completion of lazyvim.

dpetka2001:
@mirsella Yes, what you're suggesting is more in line with other LazyVim completions. Although, personally, I'll be changing that in my config if I end up using the Extra because I prefer it the other way. I'm just waiting for them to implement some kind of API to enable/disable it like with Codeium (because in Codeium I have it only enabled when `nvim-cmp` menu is not open, so there's no overlap).

dpetka2001:
No, I'm not talking about this. I'm talking about something like [this](https://github.com/dpetka2001/dotfiles/blob/ca6bff6219348b95866502584ee75b206a5b4274/dot_config/nvim/lua/plugins/coding.lua#L97-L110), so I can directly enable/disable it via API commands.

mirsella:
ooh that's nice. Is codeium the only who supports this ? Sadly, it's not documented on their repo.

dpetka2001:
`codeium.vim` supports as well. You can see the commented out sections that are `codeium.vim` equivalent. `codeium.nvim` does not support it (I think?? Except if I'm wrong), since it's a source directly for `nvim-cmp`.

Miguel-A-Jara:
I don't know if this helps anybody, but this was my solution:



``` lua
-- .config/nvim/lua/plugins/nvim-cmp.lua
return {
  "nvim-cmp",
  keys = {
    {
      "<Tab>",
      function()
        local suggestion = require("supermaven-nvim.completion_preview")
        local isSnippetActive = vim.snippet.active({ direction = 1 })

        if isSnippetActive then
          return "<cmd>lua vim.snippet.jump(1)<cr>"
        end

        if suggestion.has_suggestion() then
          return '<cmd>lua require("supermaven-nvim.completion_preview").on_accept_suggestion()<cr>'
        end

        return "<Tab>"
      end,
      expr = true,
      silent = true,
      mode = { "i" },
    },
  },
}


```

dotfrag:
I'm still not happy with how supermaven works with my setup, but if anyone wants to suppress the annoying notifications:

```lua
  {
    "folke/noice.nvim",
    opts = function(_, opts)
      vim.list_extend(opts.routes, {
        {
          filter = {
            event = "msg_show",
            any = {
              { find = "Starting Supermaven" },
              { find = "Supermaven Free Tier" },
            },
          },
          skip = true,
        },
      })
    end,
  },
```
You can still view them with `<leader>sna`.

lua/lazyvim/plugins/extras/coding/supermaven.lua
```diff
@@ -0,0 +1,36 @@
...
+  {
+    "hrsh7th/nvim-cmp",
+    optional = true,
+    opts = function(_, opts)
+      table.insert(opts.sources, { name = "supermaven", priority = 100 })
+    end,
+  },
+  {
+    "folke/noice.nvim",
+    opts = function(_, opts)
```
folke, line 21: optional = true

folke (changes requested):
Also see codium plugin as a base on how to configure supermaven.
So please add a similar cmp spec and lualine

dotfrag:
> Also see codium plugin as a base on how to configure supermaven.
> So please add a similar cmp spec and lualine

That's how I started out using supermaven, but it works differently to codeium. I think following the same setup isn't ideal. The completion is super long making the UX confusing and also causes many unintended text overrides. There was another issue that I can't recall right now and I'm not on my pc. How have you tried using supermaven?

folke:
Before making changes, please pull changes from this PR first.

- [ ] The supermaven spec should be under nvim-cmp like how it's done for the others, otherwise supermaven won't work when a user has set opts.defaults.lazy = true
- [ ] the cmp icon is missing. See lazyvim config
- [ ] since we use cmp, the suggestions by supermaven should be removed, because now I see two ghost suggestions
- [ ] ideally, just disable the non-cmp usage completely
- [ ] the build step should probably be `:SupermavenUseFree`?

folke:
@dotfrag the lualine config wasnt working, because no icon was added yet. Same for the icon of the cmp completions. (lsp kind icon in lazyvim config).

As for the long code, we should probably configure cmp to trim the ghost text if at all possible.

Feel free to make any suggestions.

dotfrag:
Ah I see.. good luck finding an icon. We could use codeium's but that might cause confusion to the user as to which one is enabled.

I have no idea how to approach this with cmp, and I believe it will be quite difficult to get right, if at all possible. 

The completion doesn't only show the next word as a completion item, but sometimes the whole line.

FWIW, in my setup, I have not added cmp integration, and just `<Tab>` or `<C-j>` to accept suggestions from ghost text. I understand this doesn't follow LazyVim's convention, but this provides the best UX (for me at least).

dpetka2001:
I haven't used supermaven for about a 1+ month, so no idea about latest progress. When I used it, it was still a bit rough around the edges. So, my question would be, is it mature enough to be introduced as an Extra? 

@dotfrag Since you seem to be using it more, what do you think about whether it's mature enough to be introduced as an Extra? My experience with it from a month ago would discourage me personally to contribute an Extra about it.

dotfrag:
If it were up to me, I would not add it. It's still a bit rough and hard to modify its behaviour.

mirsella:
OP has already done most of the work, it's free feature for the project even if the plugin is still in early phases, so imo it can be merged when thoses checklist from folke are done

> Before making changes, please pull changes from this PR first.
> 
>     * [ ]  The supermaven spec should be under nvim-cmp like how it's done for the others, otherwise supermaven won't work when a user has set opts.defaults.lazy = true
> 
>     * [ ]  the cmp icon is missing. See lazyvim config
> 
>     * [ ]  since we use cmp, the suggestions by supermaven should be removed, because now I see two ghost suggestions
> 
>     * [ ]  ideally, just disable the non-cmp usage completely
> 
>     * [ ]  the build step should probably be `:SupermavenUseFree`?

folke:
@mirsella it's not as simple as that unfortunately.

Adding a new extra, means we'll also need to maintain it.

If it's buggy and can cause issue, then I'd rather wait till the project is more mature.

iangrayhazzard:
Supermaven is very stable now, using it daily in LazyVim. I prefer it to all of the other ai completion tools (it's much faster). I'd love to see this considered as an extra in the future!

tnthi115:
@dotfrag @folke @Nybkox The README provides a star icon [here](https://github.com/supermaven-inc/supermaven-nvim?tab=readme-ov-file#using-with-nvim-cmp), but not sure if this is something we'd like to use for LazyVim?

lua/lazyvim/plugins/extras/coding/supermaven.lua
```diff
@@ -0,0 +1,49 @@
...
+    "hrsh7th/nvim-cmp",
+    optional = true,
+    opts = function(_, opts)
+      table.insert(opts.sources, 1, {
+        name = "supermaven",
+        group_index = 1,
+        priority = 100,
+      })
+    end,
+  },
```
exsesx, line 22: Would love to see integration with `Saghen/blink.cmp`

folke:
Merged. Thank you!

PR #3493, opened 2024-06-06
fix(mason-nvim-dap): correctly configure to auto-install debug adapters

dpetka2001:
Tested with `php` and `kotlin` Extras. 

`kotlin` Extra doesn't have the debug adapter in Mason `ensure_installed` and doesn't install automatically on its own. 

`php` Extra has the debug adapter defined in Mason `ensure_installed` and does correctly install automatically. However, this should not be needed due to `mason-nvim-dap`. If you remove [these lines](https://github.com/LazyVim/LazyVim/blob/ed93ce9c15cfe121b9cfb946b06b0449ae44486a/lua/lazyvim/plugins/extras/lang/php.lua#L26-L33), then the debug adapter doesn't automatically install, but it should.

With the change introduced in this PR, both Extras automatically install the debug adapters. `kotlin` Extra doesn't need anything else and in `php` Extra the lines I mentioned previously can be removed or not. The point is not to remove previously Mason dependencies with debug adapters, but to correctly configure `mason-nvim-dap`, so in future Extras you don't have to define the debug adapter inside a Mason dependency.

### DIFF
--- a/lua/lazyvim/plugins/extras/dap/core.lua
+++ b/lua/lazyvim/plugins/extras/dap/core.lua
@@ -11,112 +11,114 @@ local function get_args(config)
 end
 
 return {
-  "mfussenegger/nvim-dap",
-  recommended = true,
-  desc = "Debugging support. Requires language specific adapters to be configured. (see lang extras)",
+  {
+    "mfussenegger/nvim-dap",
+    recommended = true,
+    desc = "Debugging support. Requires language specific adapters to be configured. (see lang extras)",
 
-  dependencies = {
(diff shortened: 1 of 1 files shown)


folke:
LGTM. Thank you for looking into this!

dpetka2001:
Honestly, i'm not quite sure why the previous configuration didn't work. I pinged you at the kotlin PR about my observations, but you obviously were quite busy and didn't want to disturb you again. That's why i took the liberty of creating this PR and thought if you had a better idea then you'd simply reject this.

PR #3497, opened 2024-06-06
fix(prettier): correctly recognize patterns in condition

dpetka2001:
Without it, it shows in `ConformInfo` that the formatter is available but the condition failed. I experimented with the pattern in `vim.fs.find` to see if it could find the `.prettierrc` at the root of my projects directory.

### DIFF
--- a/lua/lazyvim/plugins/extras/formatting/prettier.lua
+++ b/lua/lazyvim/plugins/extras/formatting/prettier.lua
@@ -58,7 +58,7 @@ return {
             end
             if enabled[ctx.filename] == nil then
               enabled[ctx.filename] = vim.fs.find(function(name, path)
-                return name:match("^%.prettierrc%.") or name:match("^prettier%.config%.")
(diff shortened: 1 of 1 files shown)


lua/lazyvim/plugins/extras/formatting/prettier.lua
```diff
@@ -58,7 +58,7 @@ return {
             end
             if enabled[ctx.filename] == nil then
               enabled[ctx.filename] = vim.fs.find(function(name, path)
-                return name:match("^%.prettierrc%.") or name:match("^prettier%.config%.")
+                return name:match("^%.prettierrc%.*.*") or name:match("^prettier%.config%.*.*")
```
folke, line 61: right!
but probably only the first pattern has to change. Just add a `?` and drop your other changes

folke:
Just fixed it. Good catch!

lua/lazyvim/plugins/extras/formatting/prettier.lua
dpetka2001: Sorry was outside and didn't see this until now.

PR #3503, opened 2024-06-06
feat(dropbar): add dropbar

rubiin:
A polished, IDE-like, highly-customizable winbar for Neovim
with drop-down menu support and multiple backends


iamxiaojianzheng:
@aimuzov 

Hello, could you please learn your configuration? I'm curious as to how this effect is achieved.

image

JelteF:
> This PR is stale because it has been open 60 days with no activity.

I think the github-actions bot cannot count, it's been only 30 days.

tienbg-zigexn:
@folke

How to disable the similar path in lualine? I mean easily without copy the whole lazyvim lualine config and remove that part.

I prefer dropbar just to see that path. Lazyvim's lualine or the navic extra won't display these path in `routes.rb` or `settings.yml` in a rails project for example, but dropbar works flawlessly.

aimuzov:
> @folke
> 
> How to disable the similar path in lualine? I mean easily without copy the whole lazyvim lualine config and remove that part.
> 
> I prefer dropbar just to see that path. Lazyvim's lualine or the navic extra won't display these path in `routes.rb` or `settings.yml` in a rails project for example, but dropbar works flawlessly.

@tienbg-zigexn You don't need to copy LazyVim's lualine config. `opts` as a function receives the config LazyVim already built, so you can just patch it
from your own spec.

`sections.lualine_c` in LazyVim is:

1. `root_dir`
2. `diagnostics`
3. filetype icon
4. `pretty_path` ← the path you want gone
5. trouble symbols (only when `vim.g.trouble_lualine` and `trouble.nvim` are enabled)

**Hide it** (keeps the table shape, so nothing else shifts):

```lua
-- ~/.config/nvim/lua/plugins/lualine.lua
return {
  "nvim-lualine/lualine.nvim",
  optional = true,

  opts = function(_, opts)
    local hidden = function() return false end

    opts.sections.lualine_c[4].cond = hidden -- pretty_path
    opts.sections.lualine_c[3].cond = hidden -- filetype icon
  end,
}
```

That's what I do: [lualine.lua#L79-L80](https://github.com/aimuzov/LazyVimx/blob/main/lua/lazyvimx/overrides/other/lualine.lua#L79-L80).

Or remove the components outright (highest index first):

```lua
table.remove(opts.sections.lualine_c, 4) -- pretty_path
table.remove(opts.sections.lualine_c, 3) -- filetype icon
```

If you also want the breadcrumb symbols gone, set `vim.g.trouble_lualine = false` before lualine loads.

**And to get the dropbar-like path back without dropbar** — put the same components into a winbar. `pretty_path` is just the buffer path, it doesn't go
through an LSP/navic symbol provider, so it shows up in `routes.rb` or `settings.yml` too:

```lua
return {
  "nvim-lualine/lualine.nvim",

  opts = function(_, opts)
    local winbar = {
      lualine_c = {
        { "filetype", icon_only = true, separator = "", padding = { left = 3, right = 1 } },
        { LazyVim.lualine.pretty_path({ length = 6, relative = "root" }) },
      },
    }

    opts.options.disabled_filetypes.winbar = { "neo-tree", "trouble", "snacks_dashboard", "qf", "noice" }
    opts.winbar = winbar
    opts.inactive_winbar = winbar
  end,
}
```

Full version (background color handling + a longer list of excluded filetypes):
[winbar.lua](https://github.com/aimuzov/LazyVimx/blob/main/lua/lazyvimx/extras/ui/winbar.lua). If you want IDE-style breadcrumbs on top of the path, append
trouble's `symbols.get` component to the same winbar section — that's the snippet from my earlier comment in this thread.

Whole config, in case it's useful: https://github.com/aimuzov/LazyVimx

tienbg-zigexn:
@aimuzov 

Thank you. I actually set `vim.g.trouble_lualine` to `false` when I installed navic. The duplicated thing I was talking about is actually the navic symbols displayed in lualine. I just forgot about the navic thing completely.

I removed navic, and lualine no longer show that.

PR #2695, opened 2024-03-10
feat(mini.files): add cwd and vertical/horizontal keybindings to mini.files

dusty-phillips:
* Makes the toggle_hidden keybinding configurable via mappings
* Adds new mini.files keybindings for opening files in a vertical split, with both go_in and go_in_plus modes (configurable via mappings)
* Adds new keybinding to change working directory from mini.files (configurable via mappings)

Closes #2692

PR #3507, opened 2024-06-06
fix(mini.files): fix error about `opts.mappings`

dpetka2001:


krishna-bala:
This resolves the issue identified in https://github.com/LazyVim/LazyVim/pull/2695#issuecomment-2153135134.

folke:
Thanks!

PR #3512, opened 2024-06-07
feat(lang): add bash support

0x2E:
Currently only bash/sh is supported. 

We can add zsh if the following issues are resolved:

- https://github.com/koalaman/shellcheck/issues/809
- https://github.com/bash-lsp/bash-language-server/issues/252


0x2E:
I saw the [comment](https://github.com/LazyVim/LazyVim/pull/2879#pullrequestreview-1969578539). This PR is duplicated.

PR #3514, opened 2024-06-07
fix(lualine): only show one Lualine component for symbols

### LINKED ISSUE #3511: bug: duplicate lsp symbol navigation for lualine

### Neovim version (nvim -v)

v0.10.0

### Operating system/version

Apple M1 Pro 14.5 (23F79)

### Describe the bug

It seems that the core plugin `trouble` will add LSP symbol navigation by default, and if you add another extra navigation plugin like `navic`, there will be duplicate LSP symbol navigation in lualine.

image

### Steps To Reproduce

1. import `LazyVim` and  ` lazyvim.plugins.extras.editor.navic ` and install.
2. install lsp server
3. start neovim

### Expected Behavior

I think we should disable the default symbol navigation in extra symbol navigation plugins like navic or aerial.

### Repro

```Lua

-- set stdpaths to use .repro
end

-- bootstrap lazy
end

-- install plugins
local plugins = {
  "folke/tokyonight.nvim",
  "folke/LazyVim",
  -- add any other plugins here
  { import = "lazyvim.plugins.extras.editor.navic" },
}
  root = root .. "/plugins",
})

-- add anything else here
```

dpetka2001:
Fixes #3511

### DIFF
--- a/lua/lazyvim/plugins/extras/editor/aerial.lua
+++ b/lua/lazyvim/plugins/extras/editor/aerial.lua
@@ -104,27 +104,29 @@ return {
     "nvim-lualine/lualine.nvim",
     optional = true,
     opts = function(_, opts)
-      table.insert(opts.sections.lualine_c, {
-        "aerial",
-        sep = " ", -- separator between symbols
-        sep_icon = "", -- separator between icon and symbol
+      if not vim.g.trouble_lualine then
+        table.insert(opts.sections.lualine_c, {
+          "aerial",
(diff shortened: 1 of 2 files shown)


dpetka2001:
Should we also introduce variables to keep track of `aerial` and `navic` Lualine components in case someone disables Trouble Lualine component, but ends up enabling both navic and aerial? Or that's a bit too much of a stretch?

folke:
This is a good solution. Thanks!

And no need to make those both also configurable

PR #3516, opened 2024-06-07
feat(autocmd): add autocmds to remember folds like vscode

rubiin:


cjalvis (approved):

folke:
I don't think it's a good idea to include this in core.
At some point I enabled something like this in my config, but it had issues and also potentially slows down opening files.
Besides, when you use sessions, this is a non-issue, since sessions already store folds as well.

I suggest you to post this in https://github.com/LazyVim/LazyVim/discussions/categories/recipes

Thank you

PR #3518, opened 2024-06-07
fix(editor): telescope init error when trouble.nvim is disabled

rayrw:
I got this error message when opening telescope.

```
Failed to run `config` for telescope.nvim

...l/share/nvim/lazy/LazyVim/lua/lazyvim/plugins/editor.lua:241: module 'trouble.sources.telescope' not found:
^Ino field package.preload['trouble.sources.telescope']
...
```

After investigating it, the reason was I have `trouble.nvim` disabled but the default base `opts` function is loading it.
I tried wrapping it with a function to lazy load the `trouble.nvim` module and it worked in my local setup.

### DIFF
--- a/lua/lazyvim/plugins/editor.lua
+++ b/lua/lazyvim/plugins/editor.lua
@@ -240,7 +240,9 @@ return {
     opts = function()
       local actions = require("telescope.actions")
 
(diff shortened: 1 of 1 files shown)


folke:
Thanks!

PR #3522, opened 2024-06-07
fix(extra-lang-angular): add BufNew event to attach angular treesitter

mishasbtr:
## Problem

Sometimes angular treesitter didn't get started, e.g. when creating a new file.

## Solution

- Add `BufNew` event to attach angular treesitter.
- Change `BufRead` to `BufReadPost` for better readability.

### Related Neovim docs
- [BufRead / BufReadPost docs](https://neovim.io/doc/user/autocmd.html#BufRead)
- [BufNew](https://neovim.io/doc/user/autocmd.html#BufNew)

### DIFF
--- a/lua/lazyvim/plugins/extras/lang/angular.lua
+++ b/lua/lazyvim/plugins/extras/lang/angular.lua
@@ -14,7 +14,7 @@ return {
       if type(opts.ensure_installed) == "table" then
         vim.list_extend(opts.ensure_installed, { "angular", "scss" })
       end
-      vim.api.nvim_create_autocmd("BufEnter", {
+      vim.api.nvim_create_autocmd({ "BufReadPost", "BufNewFile" }, {
         pattern = { "*.component.html", "*.container.html" },
         callback = function()
           vim.treesitter.start(nil, "angular")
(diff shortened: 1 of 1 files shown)


folke:
Sorry, but that's not a good solution for this, since this will restart treesitter highlighting whenever you enter the buffer

dpetka2001:
@mishasbtr Have you tried using `{ "BufReadPost", "BufNewFile" }` instead? `BufReadPost` is the same as `BufRead`, I'm just used to it as I can distinguish it easier versus `BufReadPre`.

mishasbtr:
> @mishasbtr Have you tried using `{ "BufReadPost", "BufNewFile" }` instead? `BufReadPost` is the same as `BufRead`, I'm just used to it as I can distinguish it easier versus `BufReadPre`.

@dpetka2001 this seems like a good solution, will try and test it out

mishasbtr:
@dpetka2001 thank you for your suggestion, it worked out very well :+1:

mishasbtr:
> Sorry, but that's not a good solution for this, since this will restart treesitter highlighting whenever you enter the buffer

@folke updated the implementation suggested by @dpetka2001 to include an additional `BufEnter ` event for attaching treesitter

PR #3527, opened 2024-06-07
fix(cmp): don't enable `snippets` source without `nvim-snippets`

dpetka2001:
Make it easier, so if users decide to disable `nvim-snippets`, they don't have to remove the `snippets` source manually in their config.


folke:
Thanks!

PR #3538, opened 2024-06-08
refactor(mini-files): mini-files

dpetka2001:
Incorporate additional boolean logic when setting keymaps to avoid creating `opts.mappings` table and leave that only from the user configuration to change mappings.

### DIFF
--- a/lua/lazyvim/plugins/extras/editor/mini-files.lua
+++ b/lua/lazyvim/plugins/extras/editor/mini-files.lua
@@ -1,14 +1,6 @@
 return {
   "echasnovski/mini.files",
   opts = {
-    mappings = {
-      toggle_hidden = "g.",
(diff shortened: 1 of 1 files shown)


mishasbtr:
@dpetka2001 
May I ask a question on the purpose of the refactor - as far as I understand, this would enable users to configure mappings without overwriting the default ones?

dpetka2001:
No they could already do that (since the `opts.mappings` in the user's personal configuration would get merged with the `opts.mappings` in the initial PR I did to fix a problem). This one is just to remove the `opts.mappings` table from the default spec, since it's not really needed if we apply the additional boolean logic directly when we create the mappings.

mishasbtr:
@dpetka2001 thank you for the explaination :+1:

folke:
Thanks!

PR #3541, opened 2024-06-08
feat(extras): enhance incremane

zeekrs:
bind C+u to clean renaming words


zeekrs:
i think , this is not a better way

Issue #3543, opened 2024-06-08
bug: neotest rustacean does not work with `lazyvim.plugins.extras.test.core`


zspher:
### Did you check docs and existing issues?

- [X] I have read all the LazyVim docs
- [X] I have searched the existing issues of LazyVim
- [X] I have searched the existing issues of plugins related to this issue

### Neovim version (nvim -v)

0.10.0

### Operating system/version

Linux 6.6.32, NixOS, 24.11 (Vicuna), 24.11.20240605.e8057b6

### Describe the bug

neotest does not load test from rust when using `lazyvim.plugins.extras.test.core`

when using the ff. config without `lazyvim.plugins.extras.test.core`, the tests are loaded
```lua
    {
        "nvim-neotest/neotest",
        dependencies = {
            "nvim-neotest/nvim-nio",
        },
        config = function()
            require("neotest").setup {
                adapters = {
                    require "rustaceanvim.neotest",
                },
            }
        end,
        keys = {
            {
                "<leader>tt",
                function() require("neotest").run.run() end,
                desc = "Run Nearest Test",
            },
        },
    },

```

### Steps To Reproduce

1. open a rust file that contains the ff
```rs
fn main() {}

#[test]
fn test() {}
```
2. run `:Neotest summary`

### Expected Behavior

neotest should load tests from the file
![Screenshot_2024-06-08-20-36-30](https://github.com/LazyVim/LazyVim/assets/66728045/7994929d-5e65-4157-8049-d5416cc6f6d5)

### Repro

```Lua
-- DO NOT change the paths and don't remove the colorscheme
local root = vim.fn.fnamemodify("./.repro", ":p")

-- set stdpaths to use .repro
for _, name in ipairs({ "config", "data", "state", "cache" }) do
  vim.env[("XDG_%s_HOME"):format(name:upper())] = root .. "/" .. name
end

-- bootstrap lazy
local lazypath = root .. "/plugins/lazy.nvim"
if not (vim.uv or vim.loop).fs_stat(lazypath) then
  vim.fn.system({ "git", "clone", "--filter=blob:none", "https://github.com/folke/lazy.nvim.git", lazypath, })
end
vim.opt.runtimepath:prepend(lazypath)

-- install plugins
local plugins = {
  "folke/tokyonight.nvim",
  "folke/LazyVim",
  -- add any other plugins here
}
require("lazy").setup(plugins, {
  root = root .. "/plugins",
})

[...]

PR #3544, opened 2024-06-08
fix(lang): neotest not loading rust tests

zspher:
fixes https://github.com/LazyVim/LazyVim/issues/3543

### DIFF
--- a/lua/lazyvim/plugins/extras/lang/rust.lua
+++ b/lua/lazyvim/plugins/extras/lang/rust.lua
@@ -119,12 +119,10 @@ return {
 
   {
     "nvim-neotest/neotest",
-    optional = true,
-    opts = function(_, opts)
-      opts.adapters = opts.adapters or {}
-      vim.list_extend(opts.adapters, {
-        require("rustaceanvim.neotest"),
-      })
-    end,
+    opts = {
+      adapters = {
+        ["rustaceanvim.neotest"] = {},
+      },
+    },
   },
 }


dpetka2001:
@folke I'm guessing this happens because users specify in their personal config `opts.adapters` as key-values table. I experienced the same thing. And when I do `:LazyVim.opts("neotest").adapters` I get the error
```
   Error  06:50:19 PM notify.error lazy.nvim Failed to run `config` for neotest

...im/lazy/LazyVim/lua/lazyvim/plugins/extras/lang/rust.lua:126: loop or previous error loading module 'rustaceanvim.neotest'

# stacktrace:
  - /LazyVim/lua/lazyvim/plugins/extras/lang/rust.lua:126 _in_ **values**
  - /rustaceanvim/lua/rustaceanvim/neotest/init.lua:33
  - /LazyVim/lua/lazyvim/plugins/extras/lang/rust.lua:126 _in_ **values**
  - lua:1
```
and only see the python adapter from my personal configuration printed out.

Maybe we should conform to a specific style, so that users don't experience this? Because the Rust extra extends `opts.adapters` as a list to append the rustacean.neotest adapter.

dpetka2001:
@zspher What happens if you also change [these lines in your config](https://github.com/zspher/neovim-conf/blob/32105d85af024d7bee83df7fdec0d4a93ce75d97/lua/plugins/lang.lua) according to what I said in my previous comment? I believe it should work then?

zspher:
@dpetka2001 it doesn't work
i think your change in neotest python doesn't actually affect the adapters
its the neotest python in `lang.python` that does

zspher:
if i had to guess as to why lists dont work, it's because of this https://github.com/LazyVim/LazyVim/blob/fbb1276ffd45683d4d64a2d99907bdbd3421c16f/lua/lazyvim/plugins/extras/test/core.lua#L74C3-L99C1

dpetka2001:
> i think your change in neotest python doesn't actually affect the adapters
its the neotest python in lang.python that does

I'm actually testing it locally with a forked LazyVim repo and I commented out the corresponding adapter from the Python Extra and copied that code to the personal configuration that I have for that repo. But, you 're right that code still won't work. To work you would have to do `vim.list_extend(opts.adapters, { require("neotest-python").setup({ -- whatever settings here }) })`. 

The code you mentioned tries to resolve both lists and tables from `opts.adapters`. The thing that I believe is at fault here (although I might be entirely wrong) is what actually happens with the merging of the specs. If `opts.adapters` in LazyVim spec is a list and the user uses a key-value table in his personal configuration, then I believe those 2 won't merge and the personal configuration will overwrite the default LazyVim one and that's why we get `no tests found`. 

Maintainer would definitely provide more insight as to what is actually happening here.

folke:
LazyVim added some code to make it easier to work with adapaters.

@zspher is right, it should be tables everywhere instead of lists.

dpetka2001:
@folke Then maybe [these lines in `test/core.lua`](https://github.com/LazyVim/LazyVim/blob/0567b37f61d5ad997793b950c38a6c8f73de3b9b/lua/lazyvim/plugins/extras/test/core.lua#L8-L11) should be changed to better explain what the user should do? Because they mention that it also could be a list like what neotest expects.

folke:
That is still all true.
The problem arises when someone uses `vim.llist_extend` to alter that option.
That will effectively remove the table keys.

dpetka2001:
So, in the Rust Extra it could also be 
```lua
opts = {
  adapters = {
    require("rustaceanvim.neotest"),
  },
}
```
?

folke:
I wouldn't do that since that would already load that module, while not needed yet

folke:
And since that would then be a list, it woudl override all other lists

dpetka2001:
Yes, that's what I meant in my previous post (but probably didn't express it correctly). That `opts.adapters` has to be a uniform data structure. It can't contain both a list of adapter names and a table with adapter names mapped to configs. It has to be one of the 2 for consistency so that `opts.adapters` get merged correctly. The code in `test.core` only makes it feasible that it recognizes both a list of adapter names and a table with adapter names mapped to configs. But it's actually `lazy.nvim` that will do the merging of `opts.adapters`. That's why I suggested to change the wording and maybe add some kind of guideline that LazyVim only uses a table with adapter names mapped to configs. Because if a user uses in his personal configuration a list of adapter names, that would overwrite the default LazyVim configuration. 

Except if I'm still not understanding correctly what's happening, in which case sorry for wasting your time.

folke:
> @folke I'm guessing this happens because users specify in their personal config `opts.adapters` as key-values table. I experienced the same thing. And when I do `:LazyVim.opts("neotest").adapters` I get the error
> 
> ```
>    Error  06:50:19 PM notify.error lazy.nvim Failed to run `config` for neotest
> 
> ...im/lazy/LazyVim/lua/lazyvim/plugins/extras/lang/rust.lua:126: loop or previous error loading module 'rustaceanvim.neotest'
> 
> # stacktrace:
>   - /LazyVim/lua/lazyvim/plugins/extras/lang/rust.lua:126 _in_ **values**
>   - /rustaceanvim/lua/rustaceanvim/neotest/init.lua:33
>   - /LazyVim/lua/lazyvim/plugins/extras/lang/rust.lua:126 _in_ **values**
>   - lua:1
> ```
> 
> and only see the python adapter from my personal configuration printed out.
> 
> Maybe we should conform to a specific style, so that users don't experience this? Because the Rust extra extends `opts.adapters` as a list to append the rustacean.neotest adapter.
> 
> But when I enable the Python Extra, this doesn't happen although the neotest adapter is also configured as a key-value table in LazyVim's Python Extra. I don't understand why this happens honestly.
> 
> Edit: Ok I changed my configuration to an `opts` function with the following
> 
> ```lua
>       opts.adapters = vim.list_extend(opts.adapters, {
>         ["neotest-python"] = {
> 			runner = "pytest",
>         },
>       })
> ```
> 
> and it seems to work now. So, I guess we should make it clearer as to what style the neotest `opts.adapters` should follow? But I don't understand why the LazyVim Python Extra, which doesn't do that doesn't cause any problems and it only happens in the user's personal configuration. Probably because it gets parsed before the Rust Extra and the Rust Extra then changes `opts.adapters` into a list?

That error is because of the `require`. You have a `require` lazy loading loop.

folke:
In all LazyVim specs, it should be a table.
The user can still add new entries with `vim.list_extend`. Just checked that `extend` code, and it **will** keep the other non-numeric keys.

dpetka2001:
Cool!! Thank you very much for bearing with me in trying to understand better. Really appreciate it.

PR #3548, opened 2024-06-09
feat(extras.editor): add media files support for telescope

arthur-mountain:
## What is this PR for?
Support media file preview for telescope.

When developing, using Telescope to find files or search for PNG or other binary files sometimes encounters issues with previewing.

Therefore, one has to use VSCode or OS preview to view them, which requires initiating additional tool processes.

More information can be found [here](https://github.com/LazyVim/LazyVim/pull/3548#issuecomment-2158768879) and [here](https://github.com/LazyVim/LazyVim/pull/3548#discussion_r1633446752).

- [x] No
- [ ] Yes, #<issue_number>

## ChangeLog
1. Create `extra/editor/media.lua`

2. Add `nvim-telescope/telescope-media-files.nvim`
    2.1 Ensure all required prerequisites are installed
    2.2 Verify that at least one of the required find commands is installed
    2.3 Check optional prerequisites and include their filetypes if installed
    2.4 Setup media_files extension for telescope

## Checklist

## Refs
https://github.com/nvim-telescope/telescope-media-files.nvim


lua/lazyvim/plugins/extras/editor/medias.lua
```diff
@@ -0,0 +1,112 @@
...
+          break
+        end
+      end
+
+      if not findCmd then
+        return disableWithWarnMessage(
+          "None of the required 'find cmd' prerequisites is installed.\n"
+            .. "Please install at least one of the following: `rd`, `rg`, or `find`."
+        )
+      end
```
dpetka2001, line 87:
```suggestion
    local Config = require("lazyvim.config")
    if vim.tbl_contains(Config.json.data.extras, "lazyvim.plugins.extras.editor.medias") then

      for _, tool in ipairs(prerequisites.required) do
        if vim.fn.executable(tool.cmd) == 0 then
          return disableWithWarnMessage(string.format("The required prerequisite not installed: `%s`", tool.name))
        else
          supportedFiletypeSet:add(tool.filetypes)
        end
      end

      -- Verify that at least one of the required find commands is installed
      for _, tool in ipairs(prerequisites.findCmd) do
        if vim.fn.executable(tool.cmd) == 1 then
          findCmd = tool.cmd
          break
        end
      end

      if not findCmd then
        return disableWithWarnMessage(
          "None of the required 'find cmd' prerequisites is installed.\n"
            .. "Please install at least one of the following: `rd`, `rg`, or `find`."
        )
      end
    end
```
With this it won't throw the warning message if this Extra is not enabled.

lua/lazyvim/plugins/extras/editor/medias.lua
```diff
@@ -0,0 +1,112 @@
...
+end
+
+return {
+  -- The enable function will add supported file types first, so we add recommended as a function.
+  recommended = function()
+    return supportedFiletypeSet:size() > 0 and LazyVim.extras.wants({ ft = supportedFiletypeSet:toList() }) or true
+  end,
+  {
+    "nvim-telescope/telescope-media-files.nvim",
+    event = "VeryLazy",
```
dpetka2001, line 31:
```suggestion
```
This is not needed, since you have `keys` defined. And if you leave this, then this will also load Telescope, since you define it as a dependency, which will mess up Telescope's lazy-loading which is currently loaded only on `keys`.

lua/lazyvim/plugins/extras/editor/medias.lua
```diff
@@ -0,0 +1,112 @@
...
+      if ok then
+        opts.extensions = vim.tbl_extend("force", opts.extensions or {}, {
+          media_files = { filetypes = supportedFiletypeSet:toList(), find_cmd = findCmd },
+        })
+      else
+        LazyVim.error("Failed to load the telescope extension for `telescope-media-files.nvim`:\n" .. err)
+      end
+    end,
+    keys = {
+      { "<leader>fm", "<cmd>Telescope media_files<cr>", desc = "Find Media Files" },
```
dpetka2001, line 109:
```suggestion
      { "<leader>fm", "<cmd>Telescope media_files<cr>", desc = "Find Media Files" },
```
Maybe you would like to choose a different keymap for this, since this is also used by `mini.files` Extra? If a user has both Extras enabled there will be conflict.

dpetka2001:
Personally not a fan of this Extra and believe users should install it in their personal configuration if they want to. But the maintainer has the final say and will decide.

lua/lazyvim/plugins/extras/editor/medias.lua
arthur-mountain: The extras are actively added by the user into their user config.

If the prerequisites are not installed correctly, we should prevent this plugin from being added to the main(final) spec, even if it will eventually be disabled.

So, if the user added this extra to the config, I hope to have a warning message here to notify them instead of no messages being displayed.

WDYT~~~?

lua/lazyvim/plugins/extras/editor/medias.lua
arthur-mountain: Got it, I'll delete it!

However, I have an idea here that I'm not sure is correct. What do you think?

Because we need to use prerequisites to determine file types, I was thinking of using a cond function in [lazy plugin spec](https://github.com/folke/lazy.nvim?tab=readme-ov-file#-plugin-spec). This would not only set whether the plugin is loaded or not, but if enabled, it would also add the file types to the plugin.


```lua
cond = function(plugin)
    -- ...same processing for check prerequisites

    plugin.ft = supportedFiletypeSet:toList()
    return true
end
```

lua/lazyvim/plugins/extras/editor/medias.lua
arthur-mountain: Ops, thanks!!! 

You're right. I'll make some adjustments. Thank you for noticing!

lua/lazyvim/plugins/extras/editor/medias.lua
```diff
@@ -0,0 +1,112 @@
...
+      end
+
+      return true
+    end,
+    config = function(_, opts)
+      local ok, err = pcall(require("telescope").load_extension, "media_files")
+      if ok then
+        opts.extensions = vim.tbl_extend("force", opts.extensions or {}, {
+          media_files = { filetypes = supportedFiletypeSet:toList(), find_cmd = findCmd },
+        })
```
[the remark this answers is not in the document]
arthur-mountain: Ops, thanks!!! 

You're right. I'll make some adjustments. Thank you for noticing!

lua/lazyvim/plugins/extras/editor/medias.lua
arthur-mountain: BTW, 

Actually, the keymap here is not really necessary because...

If possible, I would prefer to preview directly in Telescope, handling images, fonts, and other binary files decode within the Telescope previewers.

For example, when I type <leader>ff and encounter files like images and fonts, they should be previewed directly. However, this is not possible at the moment, and we see a grep preview instead.
image


However, this is not my area of expertise at the moment, so I can only follow the official recommendation of using the Telescope media_files command from telescope-media-files.

lua/lazyvim/plugins/extras/editor/medias.lua
dpetka2001: I must have done something wrong and I was getting the warning each time I was opening the LazyExtras UI. Now, I don't get that error any more. So, I don't believe what I suggested here is needed. Please disregard this.

dpetka2001: Actually no. I pulled the latest changes you made in the PR and every time I open LazyExtras UI, I get the warning about prerequisites not being installed even though the Extra is not enabled.

arthur-mountain: No worries!!! Thank you for your attention to detail and for always providing new suggestions that allow me to deep dive into the LazyVim source code.

In fact, I also encounter a warning message that appears every time. 

For some performance reasons, so I have set `vim.opt.lazyredraw = true`, which also results in a warning message each time I open Neovim~~~ 😆

dpetka2001: If you guard the warning messages with 
```lua
    local Config = require("lazyvim.config")
    if vim.tbl_contains(Config.json.data.extras, "lazyvim.plugins.extras.editor.medias") then
      -- logic to check prerequisites and display messages
    end
```
You won't get this problem.

lua/lazyvim/plugins/extras/editor/medias.lua
dpetka2001: This will still load the plugin on `ft` and consequently Telescope also (since you define it as a dependency to this plugin) and will mess up Telescope's lazy-loading which is entirely on `keys` and `cmd`. In my opinion this is not a desired behavior.

cjalvis (approved):

lua/lazyvim/plugins/extras/editor/medias.lua
arthur-mountain: Yes, you are right. I misunderstood your point above. 

I will add an extra check to avoid displaying warning messages repeatedly.

```lua
      if not vim.tbl_contains(LazyVim.config.json.data.extras, "lazyvim.plugins.extras.editor.medias") then
        return false
      end

-- logic to check prerequisites and display messages
```

lua/lazyvim/plugins/extras/editor/medias.lua
arthur-mountain: Sure, understood. I will remove the ft of the plugin.

arthur-mountain: Here, I would like to ask,

To avoid messing up the lazy loading of telescope, what if we change telescope-media-files to a telescope dependency and then add optional = true?

```lua
{
    optional = true,
    dependencies = {
      {
        "nvim-telescope/telescope-media-files.nvim",
        cond = function()
          -- logic to check prerequisites and display messages
        end,
        keys = {
          { "<leader>fm", "<cmd>Telescope media_files<cr>", desc = "Find Media Files" },
        },
      },
    },
    opts = function(_, opts)
      opts.extensions = vim.tbl_extend("force", opts.extensions or {}, {
        media_files = { filetypes = supportedFiletypeSet:toList(), find_cmd = findCmd },
      })
      return opts
    end,
  }

```

dpetka2001: Yes, sure. This is also how `telescope-fzf-native` is configured. And your comment about `LazyVim.on_load` is also good. See here for [fzf-native](https://github.com/LazyVim/LazyVim/blob/d1d24fa077add680bc532e08fd68aaa0841a14dc/lua/lazyvim/plugins/editor.lua#L153-L170).

lua/lazyvim/plugins/extras/editor/medias.lua
dpetka2001: The display of the messages should go inside the guarding conditional, I believe. Not just return `false`. Test it out on your end to see how it behaves.

dpetka2001: I just tested your latest changes and the warning message doesn't indeed pop up every time I open LazyExtras. So all good :smile:

lua/lazyvim/plugins/extras/editor/medias.lua
dpetka2001: Actually will it load correctly with the keymap if not Telescope is loaded first? Not sure. Test it out on your end.

daUnknownCoder:
hey @arthur-mountain, ik there isn't but is there any way i can view images in a normal telescope window? like if im in a project and i do `find_files` do we have an option to preview images?

arthur-mountain:
> hey @arthur-mountain, ik there isn't but is there any way i can view images in a normal telescope window? like if im in a project and i do `find_files` do we have an option to preview images?


I think it's possible. telescope-media-files also has related [issues](https://github.com/nvim-telescope/telescope-media-files.nvim/issues/49), but no one has implemented it yet.

I tried to implement it here. However, it currently doesn't work.

Below is the code I implemented (Note: It doesn't work correctly 😢). You can refer to it or help me see how to complete it.

But this part should not be included in this PR, as the official recommendation is still to use the telescope media_files method.

```lua
local telescope_previewer_extend = function(telescope, plugin)
  local previewers = require("telescope.previewers")
  local ft = telescope.extensions.media_files.filetypes

  -- preview maker using vim.fn.job
  local new_preview_maker = function(filepath, bufnr, opts)
    print("fts: " .. vim.inspect(ft))
    print("filepath: " .. vim.inspect(filepath))
    print("bufnr: " .. vim.inspect(bufnr))
    print("options: " .. vim.inspect(opts))

    opts = opts or {}
    local isSupportedBinary = vim.list_contains(ft, filepath:match("^.+%.(.+)$"))

    if isSupportedBinary then
      vim.fn.jobstart({
        plugin.dir .. "/scripts/vimg",
        filepath,
        -- TODO: args for vimg
        --  refs: https://github.com/nvim-telescope/telescope-media-files.nvim/blob/master/lua/telescope/_extensions/media_files.lua#L35-L41
      }, {
        stdout_buffered = true,
        on_stdout = function(_, data)
          if data then
            vim.api.nvim_buf_set_lines(bufnr, 0, -1, false, data)
          end
        end,
        on_stderr = function(_, data)
          if data then
            vim.api.nvim_err_writeln(table.concat(data, "\n"))
          end
        end,
        on_exit = function()
          vim.api.nvim_buf_call(bufnr, function()
            vim.cmd("filetype detect")
          end)
        end,
      })
    else
      previewers.buffer_previewer_maker(filepath, bufnr, opts)
    end
  end

  -- preview maker using plenary.job
  local new_maker = function(filepath, bufnr, opts)
    print("fts: " .. vim.inspect(ft))
    print("filepath: " .. vim.inspect(filepath))
    print("bufnr: " .. vim.inspect(bufnr))
    print("options: " .. vim.inspect(opts))
    
    local isSupportedBinary = vim.list_contains(ft, filepath:match("^.+%.(.+)$"))

    if isSupportedBinary then
      local Job = require("plenary.job")
      local winc = vim.api.nvim_win_get_config(opts.winid)

      Job:new({
        command = plugin.dir .. "/scripts/vimg",
        args = {
          filepath,
          -- TODO: args for vimg;
          --   refs:  https://github.com/nvim-telescope/telescope-media-files.nvim/blob/master/lua/telescope/_extensions/media_files.lua#L35-L41
        },
        on_stdout = vim.schedule_wrap(function(_, data)
          print("output: " .. data)
          if vim.api.nvim_buf_is_valid(bufnr) then
            vim.api.nvim_buf_set_lines(bufnr, 0, -1, false, { data })
          end
        end),
        on_stderr = vim.schedule_wrap(function(_, data)
          if vim.api.nvim_buf_is_valid(bufnr) then
            vim.api.nvim_buf_set_lines(bufnr, 0, -1, false, { data })
          end
        end),
        on_exit = vim.schedule_wrap(function()
          if vim.api.nvim_buf_is_valid(bufnr) then
            vim.api.nvim_set_option_value("filetype", "text", { buf = bufnr })
          end
        end),
      }):start()
    else
      previewers.buffer_previewer_maker(filepath, bufnr, opts)
    end
  end

  telescope.defaults = vim.tbl_extend("force", telescope.defaults or {}, {
    buffer_previewer_maker = new_maker,
  })
end
```

In telescope plugin spec:
```lua
config = function(plugin)
  LazyVim.on_load("telescope.nvim", function()
    local ok, telescope = pcall(require, "telescope")
    if ok then
      telescope.extensions["media_files"] = vim.tbl_extend("force", telescope.extensions["media_files"] or {}, {
        filetypes = supportedFiletypeSet:toList(),
        find_cmd = findCmd,
      })
      -- telescope_previewer_extend(telescope, plugin)
    else
      LazyVim.error("Failed to load telescope.nvim")
    end
  end)
end,
```

lua/lazyvim/plugins/extras/editor/medias.lua
arthur-mountain: > Yes, sure. This is also how telescope-fzf-native is configured. And your comment about LazyVim.on_load is also good. See here for [fzf-native](https://github.com/LazyVim/LazyVim/blob/d1d24fa077add680bc532e08fd68aaa0841a14dc/lua/lazyvim/plugins/editor.lua#L153-L170).


Thank you so much!!! 


> Actually will it load correctly with the keymap if not Telescope is loaded first? Not sure. Test it out on your end.


Sure!!! I will test it out and see. If you're interested, you can also pull the latest changes and give it a try.

lua/lazyvim/plugins/extras/editor/medias.lua
```diff
@@ -0,0 +1,111 @@
...
+            local ok, telescope = pcall(require, "telescope")
+            if ok then
+              telescope.extensions["media_files"] = vim.tbl_extend("force", telescope.extensions["media_files"] or {}, {
+                filetypes = supportedFiletypeSet:toList(),
+                find_cmd = findCmd,
+              })
+            else
+              LazyVim.error("Failed to load telescope.nvim")
+            end
+          end)
```
dpetka2001, line 101: This should be in the `config` function. As it is now in the `enabled` function I'm not sure you're setting up the plugin to load. Does it work like this on your side?

arthur-mountain: Yes! I have successfully tested loading the extensions!

As far as I understand, if I only subscribe to the telescope onload event, it should be fine to subscribe even when it's enabled.

---

Initially, I placed it in the config function, but I noticed something strange.

If it's done through the config function, the loading time is longer(by about 10ms), and I also need to store variables out of spec. I'm not sure if these variables will be freed when prerequisites is checked.

image

Screenshot 2024-06-12 at 12 03 06 PM

---

However, I still prefer to put these one-time variables into the function scope to ensure these variables can free memory when the function frame pops up. 
Therefore, I put the logic into the enabled function, and this approach seems to reduce a lot of loading time (about 0.5ms).
Screenshot 2024-06-12 at 12 05 08 PM

daUnknownCoder:
> I think it's possible. telescope-media-files also has related issues, but no one has implemented it yet.

There's this thing: https://github.com/nvim-telescope/telescope.nvim/wiki/Configuration-Recipes#use-terminal-image-viewer-to-preview-images

But Idk how to use it with `ueberzug` + not use it if it is not found in the system

lua/lazyvim/plugins/extras/editor/medias.lua
arthur-mountain: Because at the beginning, when I check prerequisites, and also need to add the file types that the specific prerequisites can handle into supportedFiletypes in enabled  function.

Therefore, I must store outside variables in the spec, so that in the config function, I can add supportedFiletypes into the telescope-media-files extension.


I have made some new adjustments here:

simply checks if all prerequisites have been installed in enabled.

add findCmd and supportedFiletypes into the telescope-media-files extension in config.

WDYT~~~

dpetka2001: Yes, this looks better I believe. 

@arthur-mountain Please don't ask me to review, because I'm just a simple user like you. We're having a discussion here each one with his own suggestions. The maintainer will be the one to make the final review and see if everything is ok. I might as well be wrong about something, so asking me to review just doesn't feel right. 

Please test to see if everything is working on your side with the latest changes correctly. I think with the latest changes you're in a good place to make your case to the maintainer for this potential Extra.

arthur-mountain: > Yes, this looks better I believe.

Thank you very much!


> Please don't ask me to review, because I'm just a simple user like you. We're having a discussion here each one with his own suggestions. The maintainer will be the one to make the final review and see if everything is ok. I might as well be wrong about something, so asking me to review just doesn't feel right.


Of course, but I still greatly appreciate your assistance and suggestions!


> Please test to see if everything is working on your side with the latest changes correctly. I think with the latest changes you're in a good place to make your case to the maintainer for this potential Extra. Good luck 😄

thanks~~ 😄 I did some final testing and confirmed that everything is fine for now. I'll leave it to the maintainers to make further judgments from here on out!

arthur-mountain:
> > I think it's possible. telescope-media-files also has related issues, but no one has implemented it yet.
> > @arthur-mountain
> > There's this thing: https://github.com/nvim-telescope/telescope.nvim/wiki/Configuration-Recipes#use-terminal-image-viewer-to-preview-images
> 
> But Idk how to use it with `ueberzug` + not use it if it is not found in the system


Thank you very much for the information! I didn't see that article. 🙃 

However, I looked at the telescope source code and tried [memo_hook](https://github.com/nvim-telescope/telescope.nvim/blob/master/lua/telescope/previewers/buffer_previewer.lua#L173-L175) in the initial implementation as well, but it doesn't seem to work.

I think my usage might be incorrect, but this could be explored further and a new PR could be opened.

Additionally, ueberzug is no longer being maintained, so i don't recommended to use it. You can try other options such as:
- chafa, which is also used in telescope-media-files.
- fzf, which supports [image preview](https://github.com/junegunn/fzf?tab=readme-ov-file#previewing-an-image) and has compatibility with two or three different terminals
- ...etc., for decoding image binaries that should be previewed in terminal tools.

daUnknownCoder:
> Additionally, ueberzug is no longer being maintained, so i don't recommended to use it.

hey @arthur-mountain, i bet you didnt check: [jstkdng/ueberzugpp](https://github.com/jstkdng/ueberzugpp?tab=readme-ov-file)

arthur-mountain:
> > Additionally, ueberzug is no longer being maintained, so i don't recommended to use it.
> 
> hey @arthur-mountain, i bet you didnt check: [jstkdng/ueberzugpp](https://github.com/jstkdng/ueberzugpp?tab=readme-ov-file)

Indeed, I couldn't find this project through my search. 

The main hope is that it is safe and reliable, but the final decision will be left to the maintainers to judge.

daUnknownCoder:
> > > Additionally, ueberzug is no longer being maintained, so i don't recommended to use it.
> > 
> > 
> > hey @arthur-mountain, i bet you didnt check: [jstkdng/ueberzugpp](https://github.com/jstkdng/ueberzugpp?tab=readme-ov-file)
> 
> Indeed, I couldn't find this project through my search.
> 
> The main hope is that it is safe and reliable, but the final decision will be left to the maintainers to decide.

it is safe and reliable: i am using yazi.nvim and i can view images without installing image.nvim and i just have it configured in the main yazi config

lua/lazyvim/plugins/extras/editor/telescope.lua
```diff
@@ -201,6 +201,93 @@ return {
...
+        enabled = function()
+          -- Check that the required prerequisites are installed.
+          if vim.fn.executable("chafa") == 0 then
+            return false
+          end
+
+          -- Check that at least one prerequisite is installed.
+          local hasFindCmd = vim.tbl_contains({ "find", "fd", "rg" }, function(cmd)
+            return vim.fn.executable(cmd) == 1
+          end, { predicate = true })
```
dpetka2001, line 221: This won't work on Neovim 0.9.5. Only on 0.10+. You might want to go the traditional route of iterating through the set and checking like you do on lines 234-239.

Also the way you made it now, if the user has the executable `chafa` on his OS, this will pull in the `telescope-media-files.nvim` plugin as a dependency to Telescope. I believe it would be better if it was a different Extra and the users had the ability to opt-in if they want it.

arthur-mountain: Oh, I understand!

I was thinking that since telescope is now considered an extra, telescope-media-files could also be a part of extra.telescope, rather than creating a new extra.media.

How about making nvim-telescope/telescope-media-files.nvim optional and including it in extra.telescope?

dpetka2001: `optional` means that it will only be included if there's another instance of the plugin spec defined somewhere else. I still think making it a different Extra is better, but you might as well leave it like that and see what the maintainer says. He might be willing to incorporate it like that. He's the one who has the final say.

arthur-mountain: Of course!!! No problem, thank you so much~~ 

Let’s leave it to the maintainer to decide!

folke:
not a fan of this. thank you anyway

PR #3551, opened 2024-06-09
feat(trouble): add open with trouble mapping telescope

rubiin:
Adds this by default , really handy when using trouble


folke:
That's already included. see line 243 of the same file

PR #3552, opened 2024-06-09
feat(util): add (git) blame-nvim

gacallea:
adding https://github.com/FabijanZulj/blame.nvim with default settings and overriding `<leader>gb` keymap

### DIFF
--- a/lua/lazyvim/plugins/extras/util/blame-nvim.lua
+++ b/lua/lazyvim/plugins/extras/util/blame-nvim.lua
@@ -1,0 +1,9 @@
+return {
+  "FabijanZulj/blame.nvim",
+  config = function()
+    require("blame").setup()
+  end,
+  keys = {
+    { mode = "n", "<leader>gb", "<cmd>BlameToggle<cr>", desc = "Git Blame file" },
+  },
+}
--- a/lua/lazyvim/plugins/extras/util/blame-nvim.lua
+++ b/lua/lazyvim/plugins/extras/util/blame-nvim.lua
@@ -1,8 +1,6 @@
 return {
   "FabijanZulj/blame.nvim",
-  config = function()
-    require("blame").setup()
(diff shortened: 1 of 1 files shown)


lua/lazyvim/plugins/extras/util/blame-nvim.lua
```diff
@@ -0,0 +1,9 @@
+return {
+  "FabijanZulj/blame.nvim",
+  config = function()
+    require("blame").setup()
+  end,
```
dpetka2001, line 5:
```suggestion
  opts = {},
```
That should be easier for users to change options in their personal config. Otherwise you would have to pass `opts` to the `config` function and also pass them to `setup`.

dpetka2001:
Also have you tried `f-person/git-blame.nvim`? That's what I'm using, but I'm interested if you've tried both and can do a comparison between them. This is not a suggestion to change the plugin, just a question from a user perspective if possible.

folke:
`<leader>gb` is a default keymap that shows a git history for the cursor location.

folke:
I also don't really see the point of this extra, since it's simply adding the plugin.
Users can easily do that themselves

gacallea:
> Also have you tried `f-person/git-blame.nvim`? That's what I'm using, but I'm interested if you've tried both and can do a comparison between them. This is not a suggestion to change the plugin, just a question from a user perspective if possible.

I didn't know about this. I'll give it a try :)

gacallea:
> `<leader>gb` is a default keymap that shows a git history for the cursor location.
it is labeled "Git Blame Line" so I thought it would make sense to "Git Blame File" instead :)

gacallea:
> I also don't really see the point of this extra, since it's simply adding the plugin. Users can easily do that themselves

oops! I thought it would make a nice addition. closing :)

PR #3555, opened 2024-06-09
feat(extras): added fzf-lua

folke:
New extra with an initial implementation of fzf-lua.

## Todo

- [x] check all places that currently depend on telescope and provide an alternative or disable
- [x] disable telescope spec completely. (currently only removed its keymaps)
- [x] trouble integration
- [x] https://github.com/ibhagwan/fzf-lua/issues/1241

### DIFF
--- a/lua/lazyvim/plugins/extras/editor/fzf.lua
+++ b/lua/lazyvim/plugins/extras/editor/fzf.lua
@@ -15,6 +15,11 @@ LazyVim.pick._open = function(command, opts)
   return require("fzf-lua")[command](opts)
 end
 
+local function symbols_filter(entry, ctx)
+  ctx.symbols_filter = ctx.symbols_filter or require("lazyvim.config").get_kind_filter(ctx.bufnr)
+  return vim.tbl_contains(ctx.symbols_filter, entry.kind)
+end
+
 return {
   {
     "nvim-telescope/telescope.nvim",
@@ -113,7 +118,7 @@ return {
         "<leader>ss",
         function()
           require("fzf-lua").lsp_document_symbols({
-            symbols = require("lazyvim.config").get_kind_filter(),
+            regex_filter = symbols_filter,
           })
         end,
         desc = "Goto Symbol",
@@ -122,7 +127,7 @@ return {
         "<leader>sS",
         function()
           require("fzf-lua").lsp_dynamic_workspace_symbols({
-            symbols = require("lazyvim.config").get_kind_filter(),
+            regex_filter = symbols_filter,
           })
         end,
         desc = "Goto Symbol (Workspace)",
--- a/lua/lazyvim/plugins/extras/editor/fzf.lua
+++ b/lua/lazyvim/plugins/extras/editor/fzf.lua
@@ -39,11 +39,6 @@ return {
     event = "VeryLazy",
     opts = {
       [1] = "default-title",
-      keymap = {
-        fzf = {
-          ["ctrl-q"] = "select-all+accept",
-        },
-      },
       fzf_colors = {
         ["fg"] = { "fg", "FzfLuaNormal" },
         ["bg"] = { "bg", "FzfLuaNormal" },
@@ -62,11 +57,11 @@ return {
       },
     },
     config = function(_, opts)
-      if opts.profile then
-        opts[1] = opts.profile
-        -- require("fzf-lua").setup({ opts.profile })
-        -- opts.profile = nil
-      end
+      local config = require("fzf-lua.config")
+      local actions = require("trouble.sources.fzf").actions
+      config.defaults.actions.files["ctrl-t"] = actions.open_all
+      config.defaults.actions.files["alt-t"] = actions.open
+      config.defaults.keymap.fzf["ctrl-q"] = "select-all+accept"
       -- vim.env.FZF_DEFAULT_OPTS = ""
       require("fzf-lua").setup(opts)
     end,
--- a/lua/lazyvim/plugins/extras/editor/fzf.lua
+++ b/lua/lazyvim/plugins/extras/editor/fzf.lua
@@ -9,7 +9,6 @@ LazyVim.pick.commands = {
 ---@param opts? FzfLuaOpts
 LazyVim.pick._open = function(command, opts)
   opts = opts or {}
-
   if opts.cmd == nil and command == "git_files" and opts.show_untracked then
     opts.cmd = "git ls-files --exclude-standard --cached --others"
   end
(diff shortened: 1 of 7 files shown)


lua/lazyvim/plugins/extras/editor/fzf.lua
```diff
@@ -0,0 +1,157 @@
...
+  {
+    "ibhagwan/fzf-lua",
+    event = "VeryLazy",
+    opts = {
+      [1] = "default-title",
+      keymap = {
+        fzf = {
+          ["ctrl-q"] = "select-all+accept",
+        },
+      },
```
dpetka2001, line 46: @folke Just a suggestion. By doing that you overwrite all the default keymaps as also mentioned in the [docs](https://github.com/ibhagwan/fzf-lua/blob/9ecc8f6a6555cd3988c21480069aa5da001cd9df/README.md?plain=1#L606-L609). If you don't intend to do this, then take a look at my [dots](https://github.com/dpetka2001/dotfiles/blob/a95a1cb3650c7e8a9f138b2067d1f4a544d7de88/dot_config/nvim/lua/plugins/editor.lua#L170-L200) for alternatives how to not completely overwrite them.

folke: Yeah, I actually know. Forgot that was still in there. Should be config.globals or config.defaults.

But I'll definitely check your dots!

I also want that bug I referenced fixed. Something odd is going on there. Or its just my limited understanding

lua/lazyvim/plugins/extras/editor/fzf.lua
```diff
@@ -0,0 +1,142 @@
...
+    },
+    config = function(_, opts)
+      local config = require("fzf-lua.config")
+      local actions = require("trouble.sources.fzf").actions
+      config.defaults.actions.files["ctrl-t"] = actions.open_all
+      config.defaults.actions.files["alt-t"] = actions.open
+      config.defaults.keymap.fzf["ctrl-q"] = "select-all+accept"
+      -- vim.env.FZF_DEFAULT_OPTS = ""
+      require("fzf-lua").setup(opts)
+    end,
```
dpetka2001, line 57:
```suggestion
    opts = function()
      local config = require("fzf-lua.config")
      local actions = require("trouble.sources.fzf").actions
      config.defaults.actions.files["ctrl-t"] = actions.open_all
      config.defaults.actions.files["alt-t"] = actions.open
      config.defaults.keymap.fzf["ctrl-q"] = "select-all+accept"

      return {
        [1] = "default-title",
        fzf_colors = {
          ["fg"] = { "fg", "FzfLuaNormal" },
          ["bg"] = { "bg", "FzfLuaNormal" },
          ["hl"] = { "fg", "Special" },
          ["fg+"] = { "fg", "Visual" },
          ["bg+"] = { "bg", "Visual" },
          ["hl+"] = { "fg", "Special" },
          ["info"] = { "fg", "NonText" },
          ["border"] = { "fg", "FzfLuaBorder" },
          ["gutter"] = { "bg", "FzfLuaNormal" },
          ["query"] = { "fg", "FzfLuaNormal" },
          ["prompt"] = { "fg", "FzfLuaNormal" },
          ["pointer"] = { "fg", "FzfLuaMarker" },
          ["marker"] = { "fg", "FzfLuaMarker" },
          ["header"] = { "fg", "FzfLuaNormal" },
        },
      }
    end,
```
I could change a keymap in a `opts` function in personal configuration, but could not delete the keymap by setting its value to `nil` because it was being set up again in `config` function. I know you still probably have work to do and maybe you already intended to do that in the final form of this Extra, just thought I'd share my thoughts.

folke: Thanks!

dpetka2001:
I have a question regarding keymaps. So, LazyVim has `<esc><esc>` in terminal mapped to the sequence to escape insert mode. Because of that any other `<esc>` sequence is lagging a bit. I noticed that you used `keys` for that in fzf-lua. In my dots I used `winopts.on_create` to create buffer local mapping for fzf-lua as per the docs. I also had created an `<esc>` mapping for closing the window, but was noticing the lag (didn't give it much thought until I saw this PR). 

I tried to make my mapping like yours by adding `nowait = true`, but still the lag was there. The only cases that worked correctly, were if I defined the keymap in `keys` (but with `ft = fzf` as well) or by creating a mapping in `/after/ftplugin/fzf.lua` with `nowait = true` (mostly same like `keys` but without lazy-loading the plugin). What is the difference between creating the mapping for the filetype and creating it for the window that `fzf-lua` creates? I would expect that `nowait = true` would also work in `winopts.on_create` function. 

If you could shed some light to this baffling behavior (for me at least :P) I'd appreciate it whenever you have the time.

folke:
No idea tbh :)

cjalvis (approved):

lua/lazyvim/plugins/extras/editor/fzf.lua
```diff
@@ -42,18 +42,31 @@ return {
...
+          if type(v) == "table" then
+            fix(v)
+          end
+        end
+      end
+      fix(defaults)
 
-      return {
-        [1] = "default-title",
+      return vim.tbl_deep_extend("force", opts, defaults, {
```
dpetka2001, line 66: Do we want the defaults to overwrite `opts`? Or the other way around?

cjalvis:
I think default to overwrite opts.

lua/lazyvim/plugins/extras/editor/fzf.lua
folke: opts will be empty in this case so it doesn't matter.
User's options are loaded later

cjalvis: Yup. Sounds good.

folke:
fyi, I just updated the fzf extra in tokyonight so that it now looks the same as fzf-lua inside Neovim.

Also renamed the extension to `.sh` instead of `.zsh`, in case you are already using it.

dpetka2001:
Please excuse my ignorance, but do I just copy [these contents](https://github.com/folke/tokyonight.nvim/blob/main/extras/fzf/tokyonight_moon.sh) into `FZF_DEFAULT_OPTS`? Or is there a way to directly use the file if I save it?

dpetka2001:
Cool!! Thanks a lot. I copied them straight in the `FZF_DEFAULT_OPTS` and it indeed looks awesome!! I'll try with the file that you suggested now :smile:

PR #3556, opened 2024-06-09
fix(typescript): clean up configuration of the vscode dap extension

mehalter:
`mason-nvim-dap.nvim` adds it's own adapter filetype mappings to the vscode extension in `nvim-dap` (as of https://github.com/jay-babu/mason-nvim-dap.nvim/pull/129)

This also moves the type mappings for the `pwa-node` and `node` adapters to where they are defined in the typescript pack.

Now that the default internal type mappings are configured we can rely on the automatic loading of the `.vscode/launch.json` that `nvim-dap` does


folke:
Thanks!

PR #3560, opened 2024-06-10
fix(neo-tree): use last window for preview

aimuzov:
Similar to "trouble.nvim" I want the preview to open in the last active window. Currently the neo-tree preview opens in a floating window, which (to me) doesn’t look very good.

What do you think about doing preview window the same everywhere (trouble & neo-tree)?

## Before:

https://github.com/LazyVim/LazyVim/assets/3313023/7d7c21c9-158d-40ef-82f2-62e0dc795555

## After:

https://github.com/LazyVim/LazyVim/assets/3313023/bafc5bae-20cb-4dd1-8e40-1358ada573d9


cjalvis (approved):

cjalvis (approved):

folke:
Makes sense. Thanks!

PR #3564, opened 2024-06-10
feat(extras): added fsharp support

muneebusmani:
Added fsharp support  #3470

### DIFF
--- a/lua/lazyvim/plugins/extras/lang/fsharp.lua
+++ b/lua/lazyvim/plugins/extras/lang/fsharp.lua
@@ -23,7 +23,7 @@ return {
   },
   {
     "ionide/Ionide-vim",
-    event = "VeryLazy",
+    ft = { "fs", "fsi", "fsx" },
   },
   {
     "mason.nvim",
--- a/lua/lazyvim/plugins/extras/lang/fsharp.lua
+++ b/lua/lazyvim/plugins/extras/lang/fsharp.lua
@@ -1,0 +1,64 @@
+-- requires dotnet sdk  to be installed
+return {
+  recommended = function()
+    return LazyVim.extras.wants({
+      ft = { "fs", "fsx", "fsi" },
+      root = { "*.fsproj" },
+    })
+  end,
+  {
+    "nvim-treesitter/nvim-treesitter",
+    config = function()
+      local parser_config = require("nvim-treesitter.parsers").get_parser_configs()
+      parser_config.fsharp = {
+        install_info = {
+          url = "https://github.com/ionide/tree-sitter-fsharp",
+          branch = "main",
+          files = { "src/scanner.c", "src/parser.c" },
+        },
+        filetype = "fsharp",
+      }
+    end,
+    opts = { ensure_installed = { "fsharp" } },
+  },
+  {
+    "ionide/Ionide-vim",
+    event = "VeryLazy",
+  },
+  {
+    "mason.nvim",
+    opts = { ensure_installed = {
+      "fsautocomplete",
+      "fantomas",
+    } },
+  },
+  {
+    "nvimtools/none-ls.nvim",
+    optional = true,
+    opts = function(_, opts)
+      local nls = require("null-ls")
+      opts.sources = vim.list_extend(opts.sources or {}, {
+        nls.builtins.formatting.fantomas,
+      })
+    end,
+  },
+  {
+    "stevearc/conform.nvim",
+    optional = true,
+    opts = {
+      formatters_by_ft = {
+        fs = { "fantomas" },
+        fsi = { "fantomas" },
+        fsx = { "fantomas" },
+      },
+    },
+  },
+  {
+    "neovim/nvim-lspconfig",
+    opts = {
+      servers = {
+        fsautocomplete = {},
+      },
+    },
+  },
+}
--- a/lua/lazyvim/plugins/extras/lang/fsharp.lua
+++ b/lua/lazyvim/plugins/extras/lang/fsharp.lua
@@ -28,7 +28,6 @@ return {
   {
     "mason.nvim",
     opts = { ensure_installed = {
-      "fsautocomplete",
       "fantomas",
     } },
   },


cjalvis (approved):

lua/lazyvim/plugins/extras/lang/fsharp.lua
```diff
@@ -0,0 +1,63 @@
...
+      parser_config.fsharp = {
+        install_info = {
+          url = "https://github.com/ionide/tree-sitter-fsharp",
+          branch = "main",
+          files = { "src/scanner.c", "src/parser.c" },
+        },
+        filetype = "fsharp",
+      }
+    end,
+    opts = { ensure_installed = { "fsharp" } },
```
folke, line 22: Would be best to have this in nvim-treesitter first

folke:
Would also be great if you could read https://github.com/LazyVim/LazyVim/blob/main/CONTRIBUTING.md
(if you haven't done so).

The doc is new

lua/lazyvim/plugins/extras/lang/fsharp.lua
```diff
@@ -0,0 +1,63 @@
...
+  },
+  {
+    "ionide/Ionide-vim",
```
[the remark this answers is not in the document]
muneebusmani: Yeah it provides the functionality which omnisharp-extended-lsp provides, like full project scanning and completions, which wouldn't work properly without this plugin, because the fsharp_language_server stated in server_configurations in nvim lspconfig is no longer maintained, and is stopped at .net 6 i guess, and current lts is .net 8

muneebusmani:
https://github.com/ionide/Ionide-vim?tab=readme-ov-file#requirements
Check the repo out , this plugin supports both vim and neovim and it is written in lua and vimscript both, to support both editors, and there is no lua only solution for fsharp, it is easily overridable , according to my knowledge

lua/lazyvim/plugins/extras/lang/fsharp.lua
muneebusmani: It is a custom parser defined above

amaanq: adding the parser is useless without any queries, this should be PRed first to nvim-treesitter if the parser is of good enough quality and can parse most of the F# spec, then can be added here

muneebusmani: > adding the parser is useless without any queries, this should be PRed first to nvim-treesitter if the parser is of good enough quality and can parse most of the F# spec, then can be added here

I didn't add any query, this plugin has queries in it's repo

amaanq: queries in its repo =/= queries that neovim supports and/or are in nvim-treesitter

muneebusmani: Is there a way to add queries with extras?

amaanq: Not directly, and you shouldn't. Again, it should be added to nvim-treesitter first if the parser is of good enough quality.

lua/lazyvim/plugins/extras/lang/fsharp.lua
```diff
@@ -0,0 +1,63 @@
...
+    },
+  },
+  {
+    "neovim/nvim-lspconfig",
+    opts = {
+      servers = {
+        fsautocomplete = {},
+      },
+    },
+  },
```
dpetka2001, line 62: Should this be enabled with the `Ionide-vim`? From its [README](https://github.com/ionide/Ionide-vim?tab=readme-ov-file#for-neovim-05)
> If you are using [neovim/nvim-lspconfig](https://github.com/neovim/nvim-lspconfig), do not enable fsautocomplete. Ionide-vim automatically integrates itself into nvim-lspconfig and will register itself as a server.

Would it be better to hold up on this PR until more people who do f-sharp weigh in their opinions?

muneebusmani: My mistake, I didn't read that part in the readme, if you are right, this pr needs to be holded

dpetka2001: Do you actually program in f-sharp? Or did you only want to provide support in LazyVim for f-sharp because there was a request? This is actually a naive question as I don't know anything about you. But I believe that language Extras should actually be provided by people with experience in the specific language domain.

lua/lazyvim/plugins/extras/lang/fsharp.lua
muneebusmani: Dude this thing works and that's what ppl care about who are new to neovim and LazyVim, idk if that parser is good or not because I don't do fsharp but have done csharp so I have a general idea of dotnet ecosystem,  there isn't any other person working on fsharp parser so there isn't a choice

lua/lazyvim/plugins/extras/lang/fsharp.lua
muneebusmani: I did csharp before and know a lil bit of fsharp , and have great idea of dotnet ecosystem so that's why I searched about these things and implemented them

dpetka2001: My personal opinion is to hold up on this PR since there are quite a few lacking areas that we aren't sure about. I'd prefer to wait for more people who do f-sharp to weigh in their opinions. But in the end it's up to the maintainer to decide.

folke: Good point.
Just updated the CONTRIBUTING doc.

@muneebusmani new languages really should only be added by people that use these languages already and have a lot of experience in configuring the language for Neovim.

You have to understand that when I merge a PR for a new plugin/language, then that also needs to be maintained. To lessen the burden on maintenance, the new lang should at least be configured by someone that knows the config well.

Thank you for the effort, but I'm going to close this for now.

lua/lazyvim/plugins/extras/lang/fsharp.lua
amaanq: > Dude this thing works

It does *not* work, and if you don't do f-sharp then maybe don't add the parser without bothering to understand how tree-sitter/nvim-treesitter work? Adding it here does *absolutely* nothing without proper queries

PR #3565, opened 2024-06-10
feat(extras): flutter/dart support

muneebusmani:
added flutter/dart support

### DIFF
--- a/lua/lazyvim/plugins/extras/lang/flutter.lua
+++ b/lua/lazyvim/plugins/extras/lang/flutter.lua
@@ -7,10 +7,6 @@ return {
   {
     "akinsho/flutter-tools.nvim",
     ft = "dart",
-    dependencies = {
-      "nvim-lua/plenary.nvim",
-      "stevearc/dressing.nvim",
-    },
     config = true,
   },
 
--- a/lua/lazyvim/plugins/extras/lang/flutter.lua
+++ b/lua/lazyvim/plugins/extras/lang/flutter.lua
@@ -6,7 +6,7 @@ return {
   },
   {
     "akinsho/flutter-tools.nvim",
-    event = "VeryLazy",
+    ft = "dart",
     dependencies = {
       "nvim-lua/plenary.nvim",
       "stevearc/dressing.nvim",
--- a/lua/lazyvim/plugins/extras/lang/flutter.lua
+++ b/lua/lazyvim/plugins/extras/lang/flutter.lua
@@ -1,0 +1,30 @@
+-- requires dart/flutter to be installed
+return {
+  recommended = {
+    ft = "dart",
+    root = { "pubspec.yaml" },
+  },
+  {
+    "akinsho/flutter-tools.nvim",
+    event = "VeryLazy",
+    dependencies = {
+      "nvim-lua/plenary.nvim",
+      "stevearc/dressing.nvim",
+    },
+    config = true,
+  },
+
+  {
+    "nvim-treesitter/nvim-treesitter",
+    opts = { ensure_installed = { "dart" } },
+  },
+  {
+    "nvimtools/none-ls.nvim",
+    optional = true,
+    opts = function(_, opts)
+      local nls = require("null-ls")
+      opts.sources = opts.sources or {}
+      table.insert(opts.sources, nls.builtins.formatting.dart_format)
+    end,
+  },
+}


lua/lazyvim/plugins/extras/lang/flutter.lua
```diff
@@ -0,0 +1,30 @@
+-- requires dart/flutter to be installed
+return {
+  recommended = {
+    ft = "dart",
+    root = { "pubspec.yaml" },
+  },
+  {
+    "akinsho/flutter-tools.nvim",
+    event = "VeryLazy",
```
dpetka2001, line 9: Should this really be loaded every time Neovim opens? Maybe better to load it on `ft` instead? Don't know anything about Dart, so not sure if there are any problems if you load on `ft` instead. Just a suggestion.

muneebusmani: let me see if ft works

muneebusmani: it works, i updated this on fsharp pr too

folke:
Just added this to the contributing doc:
- You should be familiar with the language you are adding.
- You should have experience with the language's ecosystem, including formatters,
  linters, and LSP servers.

I don't think the current extra is up to standard.
It also doesn't install the needed `dartls` lsp server.
But since I don't use dart/flutter myself, I can't really fully review this PR.

So do you have extensive experience in setting up dart/flutter in Neovim?

muneebusmani:
i currently do flutter and laravel so thats the main pr's i had in mind because it was a mess maintaining fragments in so many files in nvim directory

muneebusmani:
> Just added this to the contributing doc:
> 
>     * You should be familiar with the language you are adding.
> 
>     * You should have experience with the language's ecosystem, including formatters,
>       linters, and LSP servers.
> 
> 
> I don't think the current extra is up to standard. It also doesn't install the needed `dartls` lsp server. But since I don't use dart/flutter myself, I can't really fully review this PR.
> 
> So do you have extensive experience in setting up dart/flutter in Neovim?

this doesnt install dartls because it is available in the dart-sdk , it isnt available in mason registry because it requires whole sdk to work, i mentioned in a comment in start that this requires dart/flutter sdk to work

dpetka2001:
@folke It seems that `flutter-tools` kind of works like `rustaceanvim` (or maybe more like old `typescript.nvim`). See this in their [README](https://github.com/akinsho/flutter-tools.nvim?tab=readme-ov-file#warning)
> flutter tools does not depend on nvim-lspconfig. The two can co-exist but please ensure you do NOT configure dartls using lspconfig. It will be automatically set up by this plugin instead.

Now I don't do flutter at all, but I'm guessing if this is the case as with `rustaceanvim` it would be better to let users install the LSP server via the framework itself (like with Rust toolchain) and not install it via `nvim-lspconfig`. 

I'm saying this with every bit of caution one can say, as I've 0 experience with flutter.

lua/lazyvim/plugins/extras/lang/flutter.lua
```diff
@@ -0,0 +1,30 @@
...
+    ft = "dart",
+    root = { "pubspec.yaml" },
+  },
+  {
+    "akinsho/flutter-tools.nvim",
+    ft = "dart",
+    dependencies = {
+      "nvim-lua/plenary.nvim",
+      "stevearc/dressing.nvim",
+    },
```
folke, line 13: this can be removed

lua/lazyvim/plugins/extras/lang/flutter.lua
```diff
@@ -0,0 +1,30 @@
...
+    config = true,
```
folke, line 14: replace by `opts= {}`

lua/lazyvim/plugins/extras/lang/flutter.lua
```diff
@@ -0,0 +1,30 @@
...
+  },
+
+  {
+    "nvim-treesitter/nvim-treesitter",
+    opts = { ensure_installed = { "dart" } },
+  },
+  {
```
folke, line 21: Can you add a spec for coform?

folke:
> i currently do flutter and laravel so thats the main pr's i had in mind because it was a mess maintaining fragments in so many files in nvim directory, fyi flutter is a framwork on dart and this lsp works both with dart and flutter, i tested in a cli app in dart and a flutter app too

ok, thank you for the clarification

lua/lazyvim/plugins/extras/lang/flutter.lua
muneebusmani: thanks for pointing out , fixed it

lua/lazyvim/plugins/extras/lang/flutter.lua
muneebusmani: it is too unstable to add a separate spec for conform and nvim-lint because when i try to use it, without none-ls and flutter-tools.nvim, it just lags alot when lsp is analyzing and when formatter is running

muneebusmani: still testing , can't conclude if my previous statement is valid or not

dpetka2001: Can you try with saving like this `require("conform").format({ async = true })` (set it up with a custom keymap of yours) and see if there's any improvement. I read an issue about `dart` in conform repo getting timed out and the maintainer suggested to run it async.

muneebusmani:
dart lsp sucks really bad, both in manual configiration and in flutter-tools.nvim, idk if it is buggy for only me or not , but i will close this pr as the lsp has really bad completion,linting,formatting times, it analyzies a empty project for like 10 seconds then starts giving completions,  not really worth creating and maintaining, i will do flutter in vscode with neovim extension

PR #3566, opened 2024-06-10
feat(extras): added zsh support

muneebusmani:
Added support for  diagnostics, formatting and lsp support in zsh scripts

### DIFF
--- a/lua/lazyvim/plugins/extras/lang/zsh.lua
+++ b/lua/lazyvim/plugins/extras/lang/zsh.lua
@@ -7,10 +7,6 @@ return {
     })
   end,
   {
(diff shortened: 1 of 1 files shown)


dotfrag:
I believe `beautysh` is outdated and not used by many. `shfmt` has taken over and is already in LazyVim.

folke:
bashls also doesn't support zsh, so this extra doesn't really provide anything.
Thank you anyway! Closing

muneebusmani:
@folke ik it doesn't support but I configured it to work on zsh files, and I get completions for most things which are common in bash and zsh

folke:
See https://github.com/bash-lsp/bash-language-server/issues/252

PR #3567, opened 2024-06-10
feat(lualine): show readonly icon

dotfrag:
Adds a configurable lock icon when a file is opened as read only. Mimics `vi -R <file>` status line.


dotfrag:
Forced-push to add missing types

cjalvis (approved):

folke:
Thanks!

PR #3573, opened 2024-06-10
fix(extras.copilot-chat): small ui fix for which-key in visual mode

dangdd2003:
## What is this PR for?
- Just a small fix for which-key in visual mode that the authors forgot to add.

- As copilot-chat has just been added to extras so i think this is new.

## Checklist

### DIFF
--- a/lua/lazyvim/plugins/extras/coding/copilot-chat.lua
+++ b/lua/lazyvim/plugins/extras/coding/copilot-chat.lua
@@ -22,7 +22,7 @@ return {
       }
     end,
     keys = {
-      { "<leader>a", "", desc = "+ai" },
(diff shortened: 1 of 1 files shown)


cjalvis (approved):

folke:
Thanks!

cjalvis:
Yeah no problem. All good.

dpetka2001:
@cjalvis I believe you mean good, but it's not so desirable to only get notifications in the Inbox about someone approving some changes, since it doesn't really provide anything to the original PR. A simple reaction (thumbs up, hooray etc) would be more than enough to show your support for a PR. I just got 6 notifications about you approving changes without the state of the PRs being changed. Just a suggestion from a simple user like you, that you're under no obligation to follow.

PR #3580, opened 2024-06-11
fix(fzf): correctly pass `opts.cwd` when `root=false`

dpetka2001:
## What is this PR for?

Currently, `util.pick` does not pass correctly `opts.cwd` to `M._open` when `root = false`. The value of `vim.uv.getcwd` gets assigned to `local cwd` and then used to evaluate if `git_files` should be used. 

I noticed this when using `<leader>fR` (Recent cwd) which returned the same results as `<leader>fr`. The other pickers just start
the search in the directory of the current buffer if no `cwd` is passed, so not passing `opts.cwd` when `root = false` didn't affect their behavior.

If this is intended behavior to rely on the fact that most pickers start the search in the directory of the current buffer, then probably we should discard this PR, but remove `<leader>fR` (Recent cwd) from the commands since it doesn't provide different results than `<leader>fr`.

No

## Checklist

### DIFF
--- a/lua/lazyvim/util/pick.lua
+++ b/lua/lazyvim/util/pick.lua
@@ -34,13 +34,13 @@ function M.open(command, opts)
     opts.cwd = LazyVim.root({ buf = opts.buf })
   end
 
-  local cwd = opts.cwd or vim.uv.cwd()
+  opts.cwd = opts.cwd or vim.uv.cwd()
   if command == "auto" then
     command = "files"
     if
-      vim.uv.fs_stat(cwd .. "/.git")
-      and not vim.uv.fs_stat(cwd .. "/.ignore")
-      and not vim.uv.fs_stat(cwd .. "/.rgignore")
+      vim.uv.fs_stat(opts.cwd .. "/.git")
+      and not vim.uv.fs_stat(opts.cwd .. "/.ignore")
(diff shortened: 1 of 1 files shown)


dpetka2001:
Something else I noticed is that the custom mappings `<a-c>/<c-r>/<c-t>` are all described as functions when opening the help menu with `<F1>`. Tried assigning the function to a local variable and then use the variable in the mappings without success (like we did for Telescope). Maybe you know of a better solution for this (if possible at all)?

folke:
I had accidentally removed that `cwd = vim.uv.cwd()` when refactoring.
Added it back for recent files (cwd) in both fzf-lua and telescope

folke:
> Something else I noticed is that the custom mappings `<a-c>/<c-r>/<c-t>` are all described as functions when opening the help menu with `<F1>`. Tried assigning the function to a local variable and then use the variable in the mappings without success (like we did for Telescope). Maybe you know of a better solution for this (if possible at all)?

Not sure how to do that. Need to look into the fzf-lua code. But feel free to create a PR if you've figured it out :)

folke:
Added description for toggle root.
The one for trouble will work once https://github.com/ibhagwan/fzf-lua/pull/1251 is merged

dpetka2001:
Cool!! Thnx a lot :)

PR #3590, opened 2024-06-11
feat(copilot-chat): use fzf-lua if enabled in extras

akthe-at:
Adds the abiltiy for fzf-lua to be used as a picker for these copilot chat actions if it was enabled in LazyExtras

## What is this PR for?

This adds keybinds that exist currently for copilot-chat and Telescope but with the addition of fzf-lua as an extra, I noticed we needed the ability for the user to take advantage of them with FzfLua.

No

## Checklist


folke:
Thanks!

PR #3591, opened 2024-06-11
feat(fzf): add keymap for finding all plugin files

konosubakonoakua:
## What is this PR for?

Add keymap for finding plugin files under `stdpath("data").."/lazy"`

### DIFF
--- a/lua/lazyvim/plugins/extras/editor/fzf.lua
+++ b/lua/lazyvim/plugins/extras/editor/fzf.lua
@@ -144,6 +144,7 @@ return {
       { "<leader>ff", LazyVim.pick("auto"), desc = "Find Files (Root Dir)" },
       { "<leader>fF", LazyVim.pick("auto", { root = false }), desc = "Find Files (cwd)" },
       { "<leader>fg", "<cmd>FzfLua git_files<cr>", desc = "Find Files (git-files)" },
+      { "<leader>fP", LazyVim.pick.plugin_files(), desc = "Find Plugin File" },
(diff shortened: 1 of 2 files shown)


folke:
I'm not really sure it's a good idea to provide this keymap to users, since it makes it more likely they will accidentally (or on purpose) make changes to plugin files.

You may also find my `<leader>fl` mapping useful:

```lua
{
    "ibhagwan/fzf-lua",
    optional = true,
    keys = {
      {
        "<leader>fp",
        LazyVim.pick("files", {
          cwd = require("lazy.core.config").options.root,
        }),
        desc = "Find Plugin File",
      },
      {
        "<leader>fl",
        function()
          local files = {} ---@type table<string, string>
          for _, plugin in pairs(require("lazy.core.config").plugins) do
            repeat
              if plugin._.module then
                local info = vim.loader.find(plugin._.module)[1]
                if info then
                  files[info.modpath] = info.modpath
                end
              end
              plugin = plugin._.super
            until not plugin
          end
          local filespec = table.concat(vim.tbl_values(files), " ")
          require("fzf-lua").live_grep({ filespec = "-- " .. filespec, search = "/" })
        end,
        desc = "Find Lazy Plugin Spec",
      },
    },
  }
```

konosubakonoakua:
Yes, I can use your keymap to find where specs are referenced quickly. 
But it's convenient for me when I need to quickly find plugin files in case there's an error I 
 must figure it out, or I just want to do some quick experimental modifications. 
Anyway, we can map it anytime; it is not that necessary to merge into LazyVim.

PR #3606, opened 2024-06-12
feat(pick): move pickers to extras (telescope still the default)

folke:
## What is this PR for?

Refactoring of pickers in LazyVim:

- [x] telescope moved to extras
- [x] dressing was moved to the telescope extra. Not needed with fzf-lua and noice
- [x] when none of the two is enabled, then telescope will be enabled
- [x] when using `:LazyExtras` to enable fzf-lua, the telescope spec will never be parsed
- [x] when not using `:LazyExras`, the spec will be parsed, but one of the two will be disabled.
- [x] only one picker extra can be used to prevent issues
- [ ] cleanup lsp keymaps


lua/lazyvim/plugins/extras/editor/telescope.lua
```diff
@@ -0,0 +1,242 @@
...
+            },
+          },
+        },
+      }
+    end,
+  },
+
+  -- Flash Telescope config
+  {
+    "folke/flash.nvim",
```
dpetka2001, line 198: You probably want `nvim-telescope/telescope.nvim` here?

folke: Indeed. Fixed!

folke: I also changed the default formatter to a custom formatter, since `path.filename_first` messes with searching.
To search a file in a dir, you'd have to type the file part first and then dir. Far from ideal

dpetka2001: Yes I noticed that too and also saw a discussion on Reddit that Ibhagwan mentioned is an upstream problem with how fzf operates. He also made an issue in fzf repo to revisit this (apparently it had been discussed again but declined) but without success it seems. Of course I'm just repeating what I read without any deeper understanding of what's actually going on under the hoods.

PR #3615, opened 2024-06-12
fix(tailwind): use default init_options if none are provided

garymjr:
## What is this PR for?

The tailwind extra currently doesn't work with elixir ex/heex files. Using the default `init_options` fixes this issue.

## Checklist

### DIFF
--- a/lua/lazyvim/plugins/extras/lang/tailwind.lua
+++ b/lua/lazyvim/plugins/extras/lang/tailwind.lua
@@ -42,6 +42,15 @@ return {
 
           -- Add additional filetypes
           vim.list_extend(opts.filetypes, opts.filetypes_include or {})
+
+          -- Add init_options
+          opts.init_options = opts.init_options or {}
+          opts.init_options.userLanguages = vim.tbl_deep_extend(
+            "force",
+            {},
+            opts.init_options.userLanguages or {},
+            tw.default_config.init_options.userLanguages
+          )
         end,
       },
     },


folke:
Why is this needed?
Default options are always merged by lspconfig.
Can you first figure out why this is not working?

garymjr:
I was having issues getting tailwind suggestions to work and couldn't figure out what was different. When I printed out the final opts object from the tailwind extra I noticed `init_options` wasn't included even though it's in the [default config](https://github.com/neovim/nvim-lspconfig/blob/master/lua/lspconfig/server_configurations/tailwindcss.lua#L64) for tailwind. Adding this back resulted in suggestions working as expected.

dpetka2001:
The final opts object from `LazyVim.opts("plugin_name")` only outputs the `opts` that are defined in the plugin spec as parsed by `lazy.nvim` as far as I know. So, it won't show the default settings for the `nvim-lspconfig` servers. There's another command you can do that, but I can't remember right now.

dpetka2001:
I see that you're extending the `init_options` in your configuration [here](https://github.com/garymjr/nvim/blob/45a930992cce1412c6c72f293df19004d6d9ac5c/lua/plugins/lsp.lua#L12C5-L20C7) which adds these entries to the default tailwind config. I enabled the tailwind Extra and opened a `.html` file and doing `=LazyVim.opts("nvim-lspconfig").servers.tailwindcss` indeed doesn't show the default options from `nvim-lspconfig`, but I believe that's because of what I said in my previous post. However, doing `=vim.lsp.get_client_by_id(1).config.init_options` correctly shows that the default nvim-lspconfig `init_options` have been extended by what you have in your configuration (change `1` with the `client_id` shown for Tailwind in `:LspInfo`). So, maybe there's something else going on?

garymjr:
> I see that you're extending the `init_options` in your configuration [here](https://github.com/garymjr/nvim/blob/45a930992cce1412c6c72f293df19004d6d9ac5c/lua/plugins/lsp.lua#L12C5-L20C7) which adds these entries to the default tailwind config.
> 
> I enabled the tailwind Extra and opened a `.html` file and doing `=LazyVim.opts("nvim-lspconfig").servers.tailwindcss` indeed doesn't show the default options from `nvim-lspconfig`, but I believe that's because of what I said in my previous post.
> 
> However, doing `=vim.lsp.get_client_by_id(1).config.init_options` correctly shows that the default nvim-lspconfig `init_options` have been extended by what you have in your configuration (change `1` with the `client_id` shown for Tailwind in `:LspInfo`).
> 
> So, maybe there's something else going on
> 
> PS: I don't do web development, so I couldn't test the Tailwind auto-completion, I'm just detailing what I've observed from a user configuration point of view.

I added it to my config and then decided it could be helpful if I upstreamed the change so no one else would encounter this issue. As for not getting any suggestions, if you don't have a `tailwind.config.js` file in your project root you won't get any from tailwind.

dpetka2001:
But that doesn't mean you should add the `init_options` like this. If you want to upstream the changes, then if you simply added 
```lua
init_options = {
  userLanguages = {
    elixir = "html-eex",
	eelixir = "html-eex",
	heex = "html-eex",
  },
},
```
I believe should do the trick. These should get merged with the default `init_options` from Tailwind.

garymjr:
I didn't want to hardcode the values as they're already provided by the tailwind [lspconfig](https://github.com/neovim/nvim-lspconfig/blob/master/lua/lspconfig/server_configurations/tailwindcss.lua#L65). This way if that config is updated the changes would be included here. Of course if the intention of the tailwind extra is to only provide supported filetypes and leave the rest of the config up to the user I'm more than happy to keep this in my config and close this PR.

dpetka2001:
But whatever the user adds in his Tailwind `nvim-lspconfig` spec, it will be merged with the default Tailwind config. So, I don't get why we're extending `opts.init_options` with `tw.default_config.init_options`. They already get merged as far as I can tell. Or am I missing something?

garymjr:
That's the issue I noticed. For some reason all of the default tailwind settings seem to be merged, except for `init_options`.

dpetka2001:
What does `=vim.lsp.get_client_by_id(1).config.init_options` show? Change `1` to the client id for Tailwind that is shown in `LspInfo`.

garymjr:
Oddly enough, both with and without, that command shows `init_options` under `config`. Unless I add the settings to my config as well though tailwind lsp doesn't show suggestions. I don't think I completely understand what's happening here, so this PR may have been premature. I'll wait for feedback from folke, but I'm thinking I should just keep these settings in my config.

dpetka2001:
Yes the command shows `init_options` in both cases as it should. Without your customization it shows the default options from `nvim-lspconfig`. With your customizations it should show the `init_options` with both the default options and your custom ones.

Maybe you don't get auto-completions because you use a different filetype than the ones defined in default Tailwind nvim-lspconfig config and you need to add these to your personal configuration to have auto-completions. What exactly does the `userLanguages` field do? It seems to be mapping some kind of Tailwind specific filetypes to filetype extensions?

folke:
closing, since there's probably something else going on here.

The reason we merge filetypes, is because that's a list and as such won't be merged automatically.

PR #3632, opened 2024-06-14
Update-diagnostic-args

joetoth:
update args


folke:
That's not correct. The latest trouble doesnt have those anymore

PR #3637, opened 2024-06-14
feat(lsp): add lspRestart keymap

kevintraver:
- Add `<leader>cL` keymap to restart lsp server (`LspRestart`)

Open to other keymap suggestions!


folke:
Just use `:LspRestart`. I'd advise against restarting, so doesn't make sense to make this a keymap

PR #3643, opened 2024-06-14
fix(vue): can not inc-rename

iamxiaojianzheng:
can not inc-rename when volar is enabled.

After the configuration in commit was added, the inc-rename operation returned to normal.

I suspect that volar's Hybrid mode affects the rename operation.

https://github.com/vuejs/language-tools?tab=readme-ov-file#none-hybrid-modesimilar-to-takeover-mode-configuration-requires-vuelanguage-server-version-207


lua/lazyvim/plugins/extras/lang/vue.lua
```diff
@@ -19,7 +19,14 @@ return {
     "neovim/nvim-lspconfig",
     opts = {
       servers = {
-        volar = {},
+        volar = {
+          filetypes = { "vue" },
```
dpetka2001, line 23: That is already default for volar LSP server (see [here](https://github.com/neovim/nvim-lspconfig/blob/37f362ef42d1a604d332e8d3d7d47593852b4313/lua/lspconfig/server_configurations/volar.lua#L18)). So, maybe not needed?

iamxiaojianzheng: Because vtsls has dealt with ts and js, I think it may be better for volar to only deal with vue.

Since the default configuration only includes vue, it is indeed duplicated here, and it's no problem to delete it.

folke:
Can you indeed remove the redundant filetypes?
Then I can merge this. Thanks!

iamxiaojianzheng:
> Can you indeed remove the redundant filetypes?你真的可以删除多余的文件类型吗？ Then I can merge this. Thanks!然后我可以合并它。谢谢！



> Can you indeed remove the redundant filetypes?你真的可以删除多余的文件类型吗？ Then I can merge this. Thanks!然后我可以合并它。谢谢！

PR #3652, opened 2024-06-14
feat(extras): improve ruby extra by letting user chose

kevinrobayna:
## What is this PR for?

Shopify started working on its own LSP (https://github.com/Shopify/ruby-lsp) and it performs way better than Solargraph which has a lot of limitations. This paired with sorbet gives better IntelliSense when navigating the code.

This PR follows the same approach as Python and lets the user configure through vim.g options the lsp and formatter for ruby, without overriding any configuration.

One caveat though is that RubyLsp does not work very well with NeoVim < 0.10 https://github.com/Shopify/ruby-lsp/blob/main/EDITORS.md#neovim

## Checklist

### DIFF
--- a/lua/lazyvim/plugins/extras/lang/ruby.lua
+++ b/lua/lazyvim/plugins/extras/lang/ruby.lua
@@ -11,9 +11,16 @@ return {
   },
   {
     "neovim/nvim-lspconfig",
+    ---@class PluginLspOpts
     opts = {
+      ---@type lspconfig.options
       servers = {
-        solargraph = {},
+        -- disable solargraph from auto running when you open ruby files
+        solargraph = {
+          autostart = false,
+        },
+        -- ruby_lsp will be automatically installed with mason and loaded with lspconfig
+        ruby_lsp = {},
       },
     },
   },
--- a/lua/lazyvim/plugins/extras/lang/ruby.lua
+++ b/lua/lazyvim/plugins/extras/lang/ruby.lua
@@ -45,6 +45,10 @@ return {
     },
   },
   {
+    "williamboman/mason.nvim",
+    opts = { ensure_installed = { "erb-formatter", "erb-lint" } },
+  },
+  {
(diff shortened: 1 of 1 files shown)


kevinrobayna:
Removed sorbet because if your ruby project does not have it setup you'd get warnings so it's better if you add it yourself

folke:
Can you change it similar to how it's done for python?
So add a doc section to the top and a `vim.g` var to chose.
I'm fine with that flag defaulting to `rubylsp` on NEovim >= 0.10

kevinrobayna:
> Can you change it similar to how it's done for python? So add a doc section to the top and a `vim.g` var to chose. I'm fine with that flag defaulting to `rubylsp` on NEovim >= 0.10

Thank you for the idea @folke this is soo much better i took the opportunity to also make the formatter use this

lua/lazyvim/plugins/extras/lang/ruby.lua
```diff
@@ -1,3 +1,17 @@
...
+  -- LSP Server to use for Ruby.
+  -- Set to "solargraph" to use solargraph instead of ruby_lsp.
+  vim.g.lazyvim_ruby_lsp = "ruby_lsp"
+  vim.g.lazyvim_ruby_formatter = "rubocop"
+end
+
+local lsp = vim.g.lazyvim_ruby_lsp or "ruby_lsp"
+if vim.fn.has("nvim-0.10") == 0 then
+  -- ruby_lsp does not work well with Neovim < 0.10
+  lsp = vim.g.lazyvim_ruby_lsp or "solarpath"
```
dpetka2001, line 11: Shouldn't this be `solargraph`?

kevinrobayna: yes! ty for the feedback1

folke:
ty!

ahmedsayedabdelsalam:
how to make rails model validation auto completion work?

PR #3654, opened 2024-06-14
fix(vue): can not inc-rename

iamxiaojianzheng:
can not inc-rename when volar is enabled.

After the configuration in commit was added, the inc-rename operation returned to normal.

I suspect that volar's Hybrid mode affects the rename operation.

https://github.com/vuejs/language-tools?tab=readme-ov-file#none-hybrid-modesimilar-to-takeover-mode-configuration-requires-vuelanguage-server-version-207


folke:
Merged. Thanks!

PR #3655, opened 2024-06-14
fix(lang): incorporate latest changes from `rustaceanvim`

dpetka2001:
## What is this PR for?

After recent changes in `rustaceanvim` in https://github.com/mrcjkb/rustaceanvim/pull/403, `checkOnSave` is supposed to be boolean and the table that we were passing to `checkOnSave` previously should be passed to `check` instead. I left the `check` table undefined, because the options that we were passing to `checkOnSave` before are the default ones being passed to `check`. So, only if the user wants to change something should he change the values of the `check` table according to what he wants.

No
## Checklist


folke:
Great, thanks!

PR #3667, opened 2024-06-15
fix(egdy): do not add panes for not enabled neo-tree sources

andresatierf:
## What is this PR for?

When the user disables a specific neo-tree source (git_status or buffers), the panes are still created but throw an error when the user tries to open them

**NB**: This is currently not functional because `LazyVim.opts` is returning an empty table instead of the neo-tree opts. I'm hoping someone can give me a hand with this little problem

## Checklist

### DIFF
--- a/lua/lazyvim/plugins/extras/ui/edgy.lua
+++ b/lua/lazyvim/plugins/extras/ui/edgy.lua
@@ -67,24 +67,6 @@ return {
           },
           { title = "Neotest Summary", ft = "neotest-summary" },
           {
-            title = "Neo-Tree Git",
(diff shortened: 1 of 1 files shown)


folke:
Thanks!

andresatierf:
Hey @folke , like I mentioned in the description, this fix is currently not functional because `LazyVim.opts('neo-tree')` is not returning the neo-tree options, but an empty table

dpetka2001:
It should be `LazyVim.opts("neo-tree.nvim")`.

folke:
Fixed it

PR #3674, opened 2024-06-15
fix(mini.pairs): Improve the behavior of mini.pairs

iamxiaojianzheng:
When you type `"` in the place of an asterisk

- `word* -> word"*`
- `*word -> "*word`
- `"*" -> ""*`
- `*"` -> `"*`
- `"* -> ""*`
- `"12*" -> "12"*`
- `'*' -> '"*"'`

https://github.com/LazyVim/LazyVim/discussions/2248#discussioncomment-9781859

### DIFF
--- a/lua/lazyvim/plugins/coding.lua
+++ b/lua/lazyvim/plugins/coding.lua
@@ -143,7 +143,13 @@ return {
     event = "VeryLazy",
     opts = {
       -- skip autopair when next character is one of these
-      skip_next = [=[[%w%%%'%[%"%.%`%$]]=],
+      -- [ ( { ` ' "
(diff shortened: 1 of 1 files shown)


folke:
I'd rather keep what we have right now. All except of the first example you gave already work exactly like in your example.

iamxiaojianzheng:
> I'd rather keep what we have right now. All except of the first example you gave already work exactly like in your example.

I just updated to 12.16.0 and found that the following two cases did not work

- `word* -> word"*`
- `'*' -> '"*"'`

folke:
In both of those cases, I prefer how it works right now

PR #3680, opened 2024-06-16
feat(rust): more keymaps

stefanboca:
Add some extra keybinds for rust:

- `<leader>co` for `:RustLsp openDocs` 
- `<leader>ce` for `:RustLsp expandMacro`

I was also considering changing the current binding for `<leader>cR`, which conflicts with the default Rename File binding. Any suggestions for alternatives?

### DIFF
--- a/lua/lazyvim/plugins/extras/lang/rust.lua
+++ b/lua/lazyvim/plugins/extras/lang/rust.lua
@@ -44,13 +44,12 @@ return {
     ft = { "rust" },
     opts = {
       server = {
+        -- stylua: ignore
         on_attach = function(_, bufnr)
-          vim.keymap.set("n", "<leader>cR", function()
-            vim.cmd.RustLsp("codeAction")
-          end, { desc = "Code Action", buffer = bufnr })
-          vim.keymap.set("n", "<leader>dr", function()
-            vim.cmd.RustLsp("debuggables")
-          end, { desc = "Rust Debuggables", buffer = bufnr })
+          vim.keymap.set("n", "<leader>cR", function() vim.cmd.RustLsp("codeAction") end, { desc = "Code Action (Rust)", buffer = bufnr })
+          vim.keymap.set("n", "<leader>ce", function() vim.cmd.RustLsp("expandMacro") end, { desc = "Expand Macro (Rust)", buffer = bufnr })
+          vim.keymap.set("n", "<leader>co", function() vim.cmd.RustLsp("openDocs") end, { desc = "Open Docs (Rust)", buffer = bufnr })
+          vim.keymap.set("n", "<leader>dr", function() vim.cmd.RustLsp("debuggables") end, { desc = "Debuggables (Rust)", buffer = bufnr })
         end,
         default_settings = {
           -- rust-analyzer language server configuration
--- a/lua/lazyvim/plugins/extras/lang/rust.lua
+++ b/lua/lazyvim/plugins/extras/lang/rust.lua
@@ -46,6 +46,7 @@ return {
       server = {
         -- stylua: ignore
         on_attach = function(_, bufnr)
+          vim.keymap.set("n", "K", function() vim.cmd.RustLsp({"hover", "actions"}) end, { desc = "Hover (Rust)", buffer = bufnr })
           vim.keymap.set("n", "<leader>ce", function() vim.cmd.RustLsp("expandMacro") end, { desc = "Expand Macro (Rust)", buffer = bufnr })
           vim.keymap.set("n", "<leader>co", function() vim.cmd.RustLsp("openDocs") end, { desc = "Open Docs (Rust)", buffer = bufnr })
           vim.keymap.set("n", "<leader>cO", function() vim.cmd.RustLsp("openCargo") end, { desc = "Open Cargo.toml (Rust)", buffer = bufnr })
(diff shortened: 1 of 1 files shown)


dpetka2001:
Does the default codeaction (`<leader>ca`) give anything different than RustLSP("Codeaction")? If not maybe we could just skip the latter?

stefanboca:
The RustLsp code action provides the same actions as the default codeaction, but groups them by category. From the rustaceanvim README:

>  Sometimes, rust-analyzer groups code actions by category,
 which is not supported by Neovim's built-in `vim.lsp.buf.codeAction`.
 This plugin provides a command with a UI that does:
```vim
 :RustLsp codeAction
 ```
 ```lua
 vim.cmd.RustLsp('codeAction')
 ```
 > If you set the option `vim.g.rustaceanvim.tools.code_actions.ui_select_fallback`
 to `true` (defaults to `false`), it will fall back to `vim.ui.select`
 if there are no grouped code actions.


However, considering that the default codeaction has fuzzy search, the grouping feature might not needed?

**Comment by dpetka2001** [comment]
No idea, I'm not doing Rust and don't know what the preferable way would be. Maybe wait for the maintainer's feedback on this one. Maybe changing the mapping into something else so there's no conflict would be more preferable by some people. But I'm not one to tell which way is more correct.

**Comment by mirsella** [comment]
Using rust, just using the default code actions. Changing the code action keymap just for rust is not worth it, for practically nothing in return. i'm in favor for removing this extra keymap

**Comment by mirsella** [comment]
what about also adding a keymap to open cargo.toml ? (`vim.cmd.RustLsp("openCargo")`)

**Comment by stefanboca** [comment]
Here's a summary of the changes in this PR:
- Removed `<leader>cR`, previously `:RustLsp codeAction`. This conflicted with the default Rename File keybind, and wasn't very useful anyway (it only opened a different UI for code actions, see above)
- Changed `<leader>dr` to `<leader>dR` to avoid conflict with the default Toggle Repl keybind, and change its description to be consistent with the others
- Added `<leader>ce` for `:RustLsp expandMacro`
- Added `<leader>co` for `:RustLsp openDocs`
- Added `<leader>cO` for `:RustLsp openCargo`
- Added `<leader>cp` for `:RustLsp parentModule`

Please let me know if there are any issues with these changes.

**Comment by ogios** [comment]
should hover action be added?
https://github.com/mrcjkb/rustaceanvim/issues/556#issuecomment-2447859092

**Comment by stefanboca** [comment]
Would that override `Shift-K` for rust files?

**Comment by ogios** [comment]
> Would that override `Shift-K` for rust files?

yes
```lua
vim.keymap.set("n", "K", function()
  vim.cmd.RustLsp({ "hover", "actions" })
end, { desc = "Hover action (Rust)", buffer = bufnr })

-- n  K           *@<Lua 1420: ~/.config/nvim/lua/plugins/code/lang/rust.lua:40>
--                 Hover action (Rust)
```

stefanboca:
Just got back to my computer, added.

folke:
Thanks for the PR! However, this needs several changes to align with LazyVim's patterns and contributing guidelines:

## Issues:

1. **Should use LSP server keymaps pattern**: Instead of `on_attach` with manual `vim.keymap.set`, use the `keys` field in the server config. LazyVim's LSP system will automatically resolve these keymaps (see `lua/lazyvim/plugins/lsp/keymaps.lua:74-75`).

2. **Should use `<localleader>` for language-specific commands**: Following the pattern used in Haskell and R extras, Rust-specific commands should use `<localleader>` instead of `<leader>c*`:
   - `<localleader>e` - Expand Macro
   - `<localleader>o` - Open Docs
   - `<localleader>O` - Open Cargo.toml
   - `<localleader>p` - Parent Module

3. **Don't override `K`**: The `K` hover override is problematic as it changes standard LSP behavior globally.

4. **Remove custom code action**: The old `<leader>cR` for code action should be removed - Rust should use the default `<leader>ca` code action. Having a separate keymap just to see categories doesn't make sense.

5. **Debug keymap**: The change from `<leader>dr` → `<leader>dR` is fine.

## Example of proper pattern:

```lua
server = {
  keys = {
    { "<localleader>e", function() vim.cmd.RustLsp("expandMacro") end, desc = "Expand Macro" },
    { "<localleader>o", function() vim.cmd.RustLsp("openDocs") end, desc = "Open Docs" },
    { "<leader>dR", function() vim.cmd.RustLsp("debuggables") end, desc = "Debuggables" },
  }
}
```

Feel free to open a new PR following these guidelines!

―  

eihqnh:
> > Would that override `Shift-K` for rust files?
> 
> yes
> 
> ```lua
> vim.keymap.set("n", "K", function()
>   vim.cmd.RustLsp({ "hover", "actions" })
> end, { desc = "Hover action (Rust)", buffer = bufnr })
> 
> -- n  K           *@<Lua 1420: ~/.config/nvim/lua/plugins/code/lang/rust.lua:40>
> --                 Hover action (Rust)
> ```
I dont understand why this is not working for me 
but below working fine 
for people in need
```lua
{
    "neovim/nvim-lspconfig",
    opts = function(_, opts)
      opts.servers = opts.servers or {}
      opts.servers["rust-analyzer"] = opts.servers["rust-analyzer"] or {}
      opts.servers["rust-analyzer"].enabled = false --disable for rustaceanvim
      opts.servers["rust-analyzer"].keys = {
        {
          "K",
          function()
            vim.cmd.RustLsp({ "hover", "actions" })
          end,
          desc = "Hover",
        },
      }
      return opts
    end,
  },

```

PR #3684, opened 2024-06-16
feat(extra): add prisma support and config

27medkamal:
## What is this PR for?

Adding support for prisma

No

## Checklist

### DIFF
--- a/lua/lazyvim/plugins/extras/lang/prisma.lua
+++ b/lua/lazyvim/plugins/extras/lang/prisma.lua
@@ -1,7 +1,9 @@
 return {
-  recommended = {
-    ft = "prisma",
-  },
+  recommended = function()
+    return LazyVim.extras.wants({
+      ft = "prisma",
+    })
+  end,
   {
     "nvim-treesitter/nvim-treesitter",
     opts = {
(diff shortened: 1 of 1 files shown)


lua/lazyvim/plugins/extras/lang/prisma.lua
```diff
@@ -0,0 +1,19 @@
...
+    "nvim-treesitter/nvim-treesitter",
+    opts = {
+      ensure_installed = { "prisma" },
+    },
+  },
+  {
+    "neovim/nvim-lspconfig",
+    opts = {
+      servers = {
+        prismals = {},
```
Ben10164, line 15: I don't really know anything about prisma, but are there any additional configurations to the lsp that you think would be nice? If the default values are good, then thats perfect!
Here is the link https://github.com/neovim/nvim-lspconfig/blob/master/doc/server_configurations.md#prismals

Ben10164 (changes requested):

lua/lazyvim/plugins/extras/lang/prisma.lua
27medkamal: Defaults work fine in this case

folke:
Thanks!

PR #3686, opened 2024-06-16
feat(outline): free ctrl-k /ctrl+j

iamxiaojianzheng:
Allows to toggle window in outline using Ctrl-k /ctrl+j

### DIFF
--- a/lua/lazyvim/plugins/extras/editor/outline.lua
+++ b/lua/lazyvim/plugins/extras/editor/outline.lua
@@ -17,6 +17,10 @@ return {
       local opts = {
         symbols = {},
         symbol_blacklist = {},
+        keymaps = {
+          up_and_jump = "<up>",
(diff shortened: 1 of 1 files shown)


dpetka2001:
Any specific reason why this should be default instead of something that users change in their personal configuration? What about users that expect the default mapping behavior?

iamxiaojianzheng:
> Any specific reason why this should be default instead of something that users change in their personal configuration? What about users that expect the default mapping behavior?

Because the official default ctrl+j/k and up/down function is repeated, it seems more reasonable to use ctrl+j/k for switching between Windows.

dpetka2001:
Oh sorry. I misunderstood you when you said free the keymaps. You want to be able to move between windows with default LazyVim keymaps that are being used. I see now. You're totally correct. My bad.

folke:
Thanks!

PR #3690, opened 2024-06-16
feat(project): add fzf-lua keymap descriptions

iamxiaojianzheng:
image

### DIFF
--- a/lua/lazyvim/plugins/extras/util/project.lua
+++ b/lua/lazyvim/plugins/extras/util/project.lua
@@ -9,45 +9,71 @@ pick = function()
     local history = require("project_nvim.utils.history")
     local results = history.get_recent_projects()
     local actions = require("fzf-lua.actions")
+    local core = require("fzf-lua.core")
+    -- local path = require("fzf-lua.path")
+    local utils = require("fzf-lua.utils")
 
-    fzf_lua.fzf_exec(results, {
-      actions = {
-        ["default"] = {
-          function(selected)
-            fzf_lua.files({ cwd = selected[1] })
-          end,
-        },
-        ["ctrl-t"] = actions.file_tabedit,
-        ["ctrl-s"] = {
-          function(selected)
-            fzf_lua.live_grep({ cwd = selected[1] })
-          end,
-        },
-        ["ctrl-r"] = {
-          function(selected)
-            fzf_lua.oldfiles({ cwd = selected[1] })
-          end,
-        },
-        ["ctrl-w"] = {
-          function(selected)
-            local path = selected[1]
-            local ok = project.set_pwd(path)
-            if ok then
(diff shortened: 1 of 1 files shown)


folke:
I'm not sure it's a good idea to change the core actions.
You can just add a `desc = "...."` to those keymaps. Have you tried that? (I fixed that in fzf-lua recently)

iamxiaojianzheng:
> I'm not sure it's a good idea to change the core actions.我不确定更改核心操作是否是个好主意。 You can just add a `desc => I'm not sure it's a good idea to change the core actions. You can just add a `desc = "...."` to those keymaps. Have you tried that? (I fixed that in fzf-lua recently)

Like this ? It not work.
```lua
["ctrl-s"] = {
  function(selected)
    fzf_lua.live_grep({ cwd = selected[1] })
  end,
  { desc = "live_grep" },
},
```

iamxiaojianzheng:
I refer to this issue to implement this feature

https://github.com/ibhagwan/fzf-lua/pull/977

There is another way to add headers, but I think this is too rigid and does not achieve color highlighting.

```lua
fzf_opts = { ["--header"] = "custom header" ]  }
```

folke:
Do it ike this:

```lua
["ctrl-s"] = {
  fn = function(selected)
    fzf_lua.live_grep({ cwd = selected[1] })
  end,
  desc = "live_grep"
}
```

iamxiaojianzheng:
> Do it ike this:
> 
> 
> 
> ```lua
> 
> ["ctrl-s"] = {
> 
>   fn = function(selected)
> 
>     fzf_lua.live_grep({ cwd = selected[1] })
> 
>   end,
> 
>   desc = "live_grep"
> 
> }
> 
> ```

I've tried it, but it doesn't seem to work.

iamxiaojianzheng:
And I checked the code of `fzf-lua` and couldn't find the processing logic of `desc`. Did I miss something?

folke:
ok, I totally misunderstood what you were trying to do here.
My suggestions, simply add the correct help text for when you press `<f1>`.
But you seem to be setting up the header for all the actions.
Do you really think it' a good idea to have all that text there?
Most screens will cut-off the header. Isn't it enough to have the descriptions show up in fzf help?

iamxiaojianzheng:
My idea is that I hope that users can be prompted to perform those operations on the selections like `find files`, otherwise these functions may not be known to users.

Or is there any better way?

folke:
Makes sense. Thanks!

ibhagwan:
> I'm not sure it's a good idea to change the core actions.

This is a part of the code I need to refactor so it’s easier to add automatic headers for “interactive” actions, seeing users have to resort to editing the internal structure hurts my soul as programmer lol

PR #3701, opened 2024-06-17
fix(lazygit): support monorepo structure and other git providers

morland96:
## What is this PR for?
- The current implementation doesn't support multi-module/monorepo project where `<repo>/.git` can be `gitdir` alias to the main `.git` folder. This PR uses the Git CLI to resolve remote URLs to better support flexible project structure. 
- This PR also adds the support to Bitbucket and GitLab

What's the problem of reading `.git/config`?
- Not all information available there. For instance, some large projects will split a monorepo into smaller submodules, where `.git` is a file that contains the `gitdir` alias.
- There's no promise that `.git/config` has to be existed. Git supports multiple ways to store the git info outside of the default git directory like using `GIT_DIR` env variable.
- Have to do lot of reading and parsing logic.

Why `git remote -v`?
- Only contains remote info with explicit format. 
- Don't have to filter out other config info. 
- Don't have to deal with lots of weird edge cases.

No.

## Checklist

### DIFF
--- a/lua/lazyvim/util/lazygit.lua
+++ b/lua/lazyvim/util/lazygit.lua
@@ -160,13 +160,17 @@ function M.blame_line(opts)
 end
 
 function M.browse()
-  local config = require("lazy.manage.git").get_config(LazyVim.root.detectors.pattern(0, { ".git" })[1])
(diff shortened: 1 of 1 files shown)


folke:
Much better indeed. Thanks!

PR #3716, opened 2024-06-18
fix(dap): get the correct debugpy pkg path on windows

dbger:
## What is this PR for?
fix(dap): get the correct debugpy pkg path on windows.

## Checklist


folke:
Thanks!

PR #3723, opened 2024-06-18
fix(extras.python): avoid launching extra console with dap on Windows

akthe-at:
On Windows, when using dap, python.exe will launch an extra console. If you target pythonw.exe this avoids that. Explanation source: https://stackoverflow.com/questions/9705982/pythonw-exe-or-python-exe

## What is this PR for?

Modify which exe is targeted for python + dap on windows. If you use python.exe an extra console launches (eg, nvim is runnining on Wezterm, boot up dap, and as a result an external console (usually through windows terminal or built-in console terminal) will launch (dap still runs in the original window). This is avoided with pythonw.exe

No

## Checklist

### DIFF
--- a/lua/lazyvim/plugins/extras/lang/python.lua
+++ b/lua/lazyvim/plugins/extras/lang/python.lua
@@ -93,7 +93,7 @@ return {
       },
       config = function()
         if vim.fn.has("win32") == 1 then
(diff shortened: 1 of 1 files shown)


folke:
Thanks!

PR #3736, opened 2024-06-19
feat(gitsigns): navigate diff changes as well as git hunks with [h/]h

briandipalma:
## What is this PR for?

I'm used to jumping from git hunks with `[h` so I'd like to use the same mapping when navigating diffs.

No

## Checklist


folke:
Thanks!

PR #3748, opened 2024-06-20
feat(rust): add formatter for rust

viko-stamoff:
## What is this PR for?

Adding formatter for rust, using `rustfmt` that comes with `rust-analyzer` installation.
Configuration requires manual installation of `rust-analyzer` anyway (or at least present in the PATH), so adding a formatter looks like a good idea. 
Installing `rust-analyzer` comes with the `rustfmt` formatter anyway.

Missing formatter for rust 

## Checklist

## Additional Comment

I'm open to feedback from the implementation. It's my first PR contribution to OSS, so pardon me for any errors I might make. 
Thanks in advance

### DIFF
--- a/lua/lazyvim/plugins/extras/lang/rust.lua
+++ b/lua/lazyvim/plugins/extras/lang/rust.lua
@@ -114,6 +114,21 @@ return {
   },
 
   {
+    "stevearc/conform.nvim",
+    optional = true,
+    opts = {
+      formatters = {
+        rustfmt = {
+          command = "rustfmt",
+        },
+      },
+      formatters_by_ft = {
+        ["rust"] = { "rustfmt" },
+      },
+    },
+  },
+
+  {
     "nvim-neotest/neotest",
     optional = true,
     opts = {


folke:
Doesn't the lsp rust-analyzer already do this?

viko-stamoff:
> Doesn't the lsp rust-analyzer already do this?

To my knowledge, no. 
The `rust-analyzer` can pass (additional?) arguments to `rustfmt`, but it's still required as an external dependency and doesn't integrate with `conform.nvim` for formatting.

folke:
That's what I mean. There's no need for conform here.
rust-analyzer also includes a formatter = rustfmt.
Check with `:LazyFormatInfo`

amaanq:
> Doesn't the lsp rust-analyzer already do this?

yep, this isn't needed imo (unless someone doesn't want to use the lsp, not sure why they wouldn't)

viko-stamoff:
> That's what I mean. There's no need for conform here. rust-analyzer also includes a formatter = rustfmt. Check with `:LazyFormatInfo`

> > Doesn't the lsp rust-analyzer already do this?
> 
> yep, this isn't needed imo (unless someone doesn't want to use the lsp, not sure why they wouldn't)

Got it, my bad. My config had overridden the format keybinding and didn't behave properly. 
Closing the PR, as it's not needed

PR #3758, opened 2024-06-21
feat(gitsigns): new keymap for `Gitsigns blame`

dpetka2001:
Also add filetype `gitsigns.blame` to autocmd `lazyvim_close_with_q` and workaround for `+hunks` group name to show prefix correctly. If you don't like the workaround for `+hunks` group name feel free to remove it.

## What is this PR for?
Make use of the new `gitsigns` feature `Gitsigns blame`, which shows in a vertical buffer split the `blame` history of the whole buffer.

No.

## Checklist

### DIFF
--- a/lua/lazyvim/config/autocmds.lua
+++ b/lua/lazyvim/config/autocmds.lua
@@ -67,6 +67,7 @@ vim.api.nvim_create_autocmd("FileType", {
     "neotest-summary",
     "neotest-output-panel",
     "dbout",
+    "gitsigns.blame",
   },
   callback = function(event)
     vim.bo[event.buf].buflisted = false
--- a/lua/lazyvim/plugins/editor.lua
+++ b/lua/lazyvim/plugins/editor.lua
@@ -177,7 +177,7 @@ return {
         ["<leader>c"] = { name = "+code" },
         ["<leader>f"] = { name = "+file/find" },
         ["<leader>g"] = { name = "+git" },
-        ["<leader>gh"] = { name = "+hunks" },
+        ["<leader>gh"] = { name = "+hunks", ["_"] = "which_key_ignore" },
         ["<leader>q"] = { name = "+quit/session" },
         ["<leader>s"] = { name = "+search" },
         ["<leader>u"] = { name = "+ui" },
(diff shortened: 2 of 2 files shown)


dpetka2001:
Tnx @dotfrag I changed it according to your suggestion. About the window title, I'm afraid I don't understand what you're saying. Could you please explain what you mean?

dotfrag:
Thanks for the change.. a couple of more comments that popped in my head.

1. Square brackets and quotes are reduntant: `["_"]` and `_` work the same.
2. You have assigned the command to `<leader>ghB`, but the `<leader>gh` subset is for hunks, and this is an operation for the entire buffer, maybe `<leader>gB` would make more sense?

My question was related to the addition to "close with q", you added `gitsigns.blame` in the autocmd pattern. How did you get that value? I want to find the pattern for fugitive's blame, and possibly other splits/windows from plugins.

dpetka2001:
1. I just left it for conformity with the rest.
2. All actions for `gitsigns` are under `<leader>gh` even `<leader>ghR` (reset buffer) and `<leader>ghS` (stage buffer), that's why I chose this keymap as well. 

Ok now, I understand. Go with your cursor to the Fugitive window that you want and then press `:set ft?`. That will print out the filetype needed to be appended to the `lazyvim_close_with_q` autocmd.

dotfrag:
Ohh it's just the ft, I thought there was some special "win title" property for plugin windows. It worked, thanks!

dpetka2001:
Yes, you can see right [here as well](https://github.com/LazyVim/LazyVim/blob/dde4a9dcdf49719c67642d09847dbaf7f9c7a156/lua/lazyvim/config/autocmds.lua#L53-L75). The `autocmd` runs on `FileType` event and the pattern is matched against filetypes. See `:h FileType`.

dotfrag:
@dpetka2001, open a file and press `<leader>ghB`, move your cursor to any line in the blame buffer, press Enter (<cr>). A list of keybindings will popup, but none work. Not sure if a gitsigns or LazyVim issue.. either way I don't really care about this feature, just pointing it out.

dpetka2001:
Strange it seems to be working correctly for me

[Screencast 2024-06-21 20:28:01.webm](https://github.com/LazyVim/LazyVim/assets/12776461/790934f9-1b1b-4b41-b107-4523b6ec4b3f)

dotfrag:
I have probably misunderstood the keybindings logic.. what you showed works for me too. I was pressing Enter, *then* the key, which doesn't work.

dpetka2001:
@dotfrag Please see the PS in my previous edited comment.

folke:
Thanks!

PR #3765, opened 2024-06-22
feat(extra.sql): add sql options docs

arthur-mountain:
## What is this PR for?

A small contribution to enhance the documentation and make it easier to use.

While vim-dadbod and vim-dadbod-ui recommend using dotenv.vim for env file injection, we aim to provide a simpler method. Instead of implementing file parsing or relying on an external plugin (dotenv.vim), just try directly using a Lua table.

none!

## Checklist

### DIFF
--- a/lua/lazyvim/plugins/extras/lang/sql.lua
+++ b/lua/lazyvim/plugins/extras/lang/sql.lua
@@ -1,29 +1,32 @@
 if lazyvim_docs then
   -- Automatically configure connections without needing to manually input them each time.
-  vim.g.dbs = {
-    dev = "Replace with your database connection URL.",
-    staging = "Replace with your database connection URL.",
-  }
+  --    vim.g.dbs = {
+  --      dev = "Replace with your database connection URL.",
+  --      staging = "Replace with your database connection URL.",
+  --    }
   -- or
-  vim.g.dbs = {
-    { name = "dev", url = "Replace with your database connection URL." },
-    { name = "staging", url = "Replace with your database connection URL." },
-  }
+  --    vim.g.dbs = {
+  --      { name = "dev", url = "Replace with your database connection URL." },
+  --      { name = "staging", url = "Replace with your database connection URL." },
+  --    }
 
-  -- You can set your secrets directly through a Lua table. For example:
-  -- Create a secrets/db-env.lua file and return the connection table.
-  -- Or
-  -- Replace `secrets/db-env` with your own module path.
-  local ok, dbs = pcall(require, "secrests/db-env")
-  if ok then
-    vim.g.dbs = dbs
-  else
+  -- You can set your those configuration through a Lua table. For example:
+  -- Create a `.lazy.lua` file in your project and set your connections in it like this:
+  -- ```lua
+  --    vim.g.dbs = {...}
+  --
+  --    return {}
+  -- ```
+
+  local ok = pcall(dofile, vim.fn.getcwd() .. "/.lazy.lua")
+
+  if not ok then
     vim.g.dbs = {}
   end
 
   -- Alternatively, you can also use other ways to inject your environment variables.
 
-  -- After all, please make sure to protect your secrets and add them to your `.gitignore` file.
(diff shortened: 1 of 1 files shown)


lua/lazyvim/plugins/extras/lang/sql.lua
```diff
@@ -1,3 +1,31 @@
...
+  vim.g.dbs = {
+    { name = "dev", url = "Replace with your database connection URL." },
+    { name = "staging", url = "Replace with your database connection URL." },
+  }
+
+  -- You can set your secrets directly through a Lua table. For example:
+  -- Create a secrets/db-env.lua file and return the connection table.
+  -- Or
+  -- Replace `secrets/db-env` with your own module path.
+  local ok, dbs = pcall(require, "secrests/db-env")
```
folke, line 17: Maybe refer to a `.lazy.lua` file per project?

```lua
vim.g.dbs {...}

return {}
```

arthur-mountain: @folke It seems like a great idea!!! Below is the updated code. Do you think this is suitable?

```lua
if lazyvim_docs then
  -- Automatically configure connections without needing to manually input them each time.
  --    vim.g.dbs = {
  --      dev = "Replace with your database connection URL.",
  --      staging = "Replace with your database connection URL.",
  --    }
  -- or
  --    vim.g.dbs = {
  --      { name = "dev", url = "Replace with your database connection URL." },
  --      { name = "staging", url = "Replace with your database connection URL." },
  --    }

  -- You can set your those configuration through a Lua table. For example:
  -- Create a `.lazy.lua` file in your project and set your connections in it like this:
  -- ```lua
  --    vim.g.dbs = {...}
  --
  --    return {}
  -- ```

  local ok = pcall(dofile, vim.fn.getcwd() .. "/.lazy.lua")

  if not ok then
    vim.g.dbs = {}
  end

  -- Alternatively, you can also use other ways to inject your environment variables.

  -- After all, please make sure to add `.lazy` to your global `.gitignore` file.
end

```

arthur-mountain: Additionally, may I ask for your opinion!!!?

Instead of adding .lazy.lua to the .gitignore file for each project, how about automatically adding .lazy.lua to the global .gitignore file during the initialization of Lazy or LazyVim?

I'm not quite sure if the implementation below is good:
```lua
local globalGitIgnorePath = vim.env.HOME .. '/.gitignore'

if vim.uv.fs_access(globalGitIgnorePath, "R") then
  -- Check if the .gitignore file already contains .lazy.lua
  local file = io.open(globalGitIgnorePath, "r")
  local content = file:read("*all")
  file:close()

  if not content:match("%.lazy%.lua") then
    -- Add .lazy.lua to .gitignore
    local file = io.open(globalGitIgnorePath, "a")
    file:write("\n.lazy.lua\n")
    file:close()
  end
else
  -- Create .gitignore and add .lazy.lua to it
  local file = io.open(globalGitIgnorePath, "w")
  file:write(".lazy.lua\n")
  file:close()
  os.execute("chmod 644 " .. globalGitIgnorePath)
end
```

arthur-mountain: On the other hand, I also considered placing this part (db env setup) in the recipes section of the LazyVim website, rather than including it here.

What do you think~~~?

folke: You misunderstood. ` local ok = pcall(dofile, vim.fn.getcwd() .. "/.lazy.lua")` is not needed.
A `.lazy.lua` file in the project directory gets automatically merged with your specs by lazy.

arthur-mountain: Oops, indeed, I completely misunderstood. My apologies.

Let me make some adjustments. I still haven’t fully grasped the details about the 'lazy' implementation.

Thank you very much for the reminder and the explanation! I just noticed that `local_spec` was called and added `.lazy` to `spec` by using `lazy`.

PR #3768, opened 2024-06-22
feat(lazygit): improve browsing

arthur-mountain:
## What is this PR for?
A small contribution: When using SSH keys to clone a repository, \<leader\>gB cannot open the webpage. For non-HTTPS URLs, it will attempt to replace the URL using SSH configuration(readonly).

none!

## Checklist

### DIFF
--- a/lua/lazyvim/util/lazygit.lua
+++ b/lua/lazyvim/util/lazygit.lua
@@ -180,6 +180,10 @@ function M.get_url(remote)
     ret = ret:gsub(pattern[1], pattern[2])
   end
 
+  if ret:find("https://") == 1 then
+    return ret
+  end
+
   --- @param url string
   local function getRemoteURLFromSSH(url)
     local transformedUrl = ""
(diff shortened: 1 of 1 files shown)


arthur-mountain:
> Can you please tell me the use-case and give some examples on what this fixes? IT's quite a lot of code for something I don't really understand is needed


@folke 

Thank you very much for your replied!!!

I sincerely apologize for making those extensive code changes without providing an example.

Here is an example:

When cloning a project via SSH, it often looks like this: ⬇️ 
```plain
git@github.com:$org-name/$repo-name.git
```

If someone needs to manage multiple keys using SSH config, it might look like this: ⬇️ 
```plain
Host demo-test
    User git
    HostName github
    IdentityFile xxxxxxxxx
    ...other config options
```

Therefore, the original git remote alias would change to: ⬇️ 
```plain
$some-alias demo-test:$org-name/$repo-name.git
```

The prefix `demo-test` will map to the `Host` field in the SSH config and replace it with `$User@$HostName`, resulting in something like `git@github`.

PR #3786, opened 2024-06-23
fix(extras.astro): support `astro.config.ts` root

sanman1k98:
## What is this PR for?

Support [`astro.config.ts`](https://docs.astro.build/en/guides/configuring-astro/#supported-config-file-types) as a root file for the Astro extra.

N/A

## Checklist


folke:
Thanks!

dpetka2001:
This only adds to the recommended. Is this really what OP wants to do? My assumption was he wanted to change the `root_dir` for `astro` LSP server.

folke:
if so, then he should submit a PR to lspconfig indeed

PR #3789, opened 2024-06-24
feat(extras.lang.json): add json-nvim

Ben10164:
## What is this PR for?

Updates the existing JSON config to utilize [json-nvim](https://github.com/VPavliashvili/json-nvim).
This minor addition adds new features such as minifying and converting JSON files with the use of `jq`.

Since it uses `jq`, I added another dependency for mason which installs it. I know mason auto-installs lsp's if they are defined in `nvim-lspconfig`, but mason classifies `jq` as a formatter, so I manually added it as an `opts.ensure_installed`.

Nope

## Checklist


folke:
That plugin currently only has 5 stars and doesn't work properly on Windows.

Not something I'm willing to add at this point.

PR #3792, opened 2024-06-24
fix(yanky): integration fzf with yanky

zeekrs:
## What is this PR for?
integration fzf picker for yanky

## Checklist


folke:
I don't really get why this is needed?
History already works perfectly fine with fzf out of the box?

zeekrs:
Yes, YankyRingHistory works well, but it doesn't have a preview. I want it to provide a consistent experience like Telescope

folke:
This would be better to submit as a PR to the yanky repo itself.

PR #3794, opened 2024-06-24
fix(edgy): improve edgy.lua bufferline offset logic

gj86:
## What is this PR for?
Use already defined Bufferline offsets or use Edgy offset if none are found.

Also properly set offset separator position depending on offset/pane location.

No

## Checklist


folke:
Thanks!

PR #3798, opened 2024-06-24
fix(fzf): disable `preview` for `vtsls`

dpetka2001:
## What is this PR for?
Disable `preview` in `fzf-lua` Extra for `vtsls`. I tested it with Python and Typescript Extras and in Typescript pressing `<leader>ca` shows without previewer, while in Python the same shows with previewer. 

Kindly check also that this is correct and I didn't miss out anything else.

Fixes #3773

## Checklist


folke:
Awesome, thanks!

PR #3799, opened 2024-06-24
feat(c#): add support csharp-language-server

lillo42:
## What is this PR for?

Add support to [csharp-language-server](https://github.com/razzmatazz/csharp-language-server) an alternative to omnisharp server (Today some people are experiencing issues with Omnisharp Server see https://github.com/OmniSharp/omnisharp-roslyn/issues/2574 and with csharp-ls we don't have that issue) 

## Checklist

### DIFF
--- a/lua/lazyvim/plugins/extras/lang/csharp_ls.lua
+++ b/lua/lazyvim/plugins/extras/lang/csharp_ls.lua
@@ -1,0 +1,104 @@
+return {
+  recommended = function()
+    return LazyVim.extras.wants({
+      ft = { "cs", "vb" },
+      root = { "*.sln", "*.csproj", "omnisharp.json", "function.json" },
+    })
+  end,
+
+  { "Decodetalkers/csharpls-extended-lsp.nvim", lazy = true },
+
+  {
+    "nvim-treesitter/nvim-treesitter",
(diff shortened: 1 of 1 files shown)


lua/lazyvim/plugins/extras/lang/csharp_ls.lua
```diff
@@ -0,0 +1,104 @@
...
+  },
+  {
+    "williamboman/mason.nvim",
+    opts = { ensure_installed = { "csharpier", "netcoredbg", "csharp-language-server" } },
+  },
+  {
+    "neovim/nvim-lspconfig",
+    opts = {
+      servers = {
+        sharp_ls = {
```
dpetka2001, line 47:
```suggestion
        csharp_ls = {
```
I believe this should be `csharp_ls` and with this you could also remove `csharp-language-server` from `mason.ensure_installed`, since when defining the LSP server in `nvim-lspconfig` spec it will also automatically download the LSP server.

Kurren123:
@lillo42  Great PR, looking forward to having csharp-lanaguage-server due to the issues with omnisharp. The failing test mentioned not automatically installing the language server with nvim.mason but to me it seems like mason does install automatically?

PR #3801, opened 2024-06-25
Add templ lang extra

knpwrs:
## What is this PR for?

This PR adds a lang extra for [templ](https://templ.guide/) support. There is no treesitter syntax available for templ, so this installs a plain language plugin and sets up the lsp.

## Checklist

### DIFF
--- a/lua/lazyvim/plugins/extras/lang/templ.lua
+++ b/lua/lazyvim/plugins/extras/lang/templ.lua
@@ -1,0 +1,7 @@
+return {
+  "joerdav/templ.vim",
+  {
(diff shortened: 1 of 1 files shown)


dpetka2001:
Is this really worth an Extra with just 4 lines of configuration? Also, you're specifying the LSP server in `mason.ensure_installed`, but you're not setting up the LSP server in `nvim-lspconfig`. If you instead just define the LSP server in `nvim-lspconfig` spec, it sets up the LSP server and also automatically downloads it. 

Maintainer has the final say, this is just my personal opinion as a simple user.

PR #3821, opened 2024-06-26
refactor(project): use header descriptions without using `fzf-lua.actions`

dpetka2001:
## What is this PR for?
In #3690 there was a [concern expressed](https://github.com/LazyVim/LazyVim/pull/3690#issuecomment-2174024517) about changing `fzf-lua`'s core actions. This PR tries to achieve the same result by manually setting the headers in `fzf_opts["--header"]` and constructing it without use of `fzf-lua.actions`.

Feel free to disregard if you think there's no change necessary in the current code, as this is simply a refactor attempt and does not provide any new functionality.

No

## Checklist


folke:
much cleaner, thanks!

Issue #847, opened 2023-05-28
bug:  fix(neotest): support neotest adapaters that use setup() for custom setup  breaks neotest-python


frogmaster:
### Did you check docs and existing issues?

- [X] I have read all the LazyVim docs
- [X] I have searched the existing issues of LazyVim
- [X] I have searched the existing issues of plugins related to this issue

### Neovim version (nvim -v)

0.9.0

### Operating system/version

arch linux

### Describe the bug

Neotest with neotest-python stopped working for me tonight.  I know nothing of lua, but some minor changes got it to work again: extras/test/core.lua around line 60

              if adapter.setup then
                adapter.setup(config)
              else --  elseif adapter.__call then
                adapter(config)
                --else
                -- error("Adapter " .. name .. " does not support setup")
              end

### Steps To Reproduce

Try to use neotest adapter without setup function.

### Expected Behavior

adapter works

folke:
What is the spec you use?

frogmaster:
~/.config/nvim/lua/plugins/neotest.lua

     return {
      "nvim-neotest/neotest",
      dependencies = {
        "nvim-neotest/neotest-python",
      },
      opts = {
        adapters = {
          ["neotest-python"] = {
            args = { "-s", "-vv" },
            runner = "pytest",
          },
        },
      },
      -- stylua: ignore
      keys = {
        {"<F4>", function() require("neotest").run.run() end, desc="Run nearest"},
        {"<F2>", function() require("neotest").run.run(vim.fn.expand("%")) end, desc="Run file"}
      },
    }

folke:
Should be fixed now. I'll make a new release in a bit.

Let me know if it would still not be resolved.

PR #3824, opened 2024-06-27
fix(neotest): support adapter key on adapter for playwright

asfaltboy:
## What is this PR for?

Similar to https://github.com/LazyVim/LazyVim/issues/847 , only this adapter adds a setup function into the 'adapter' key in the plugin table ([source](https://github.com/thenbe/neotest-playwright/blob/f61463f5c74d53fc5975b7e086b7c6e9356d9215/lua/neotest-playwright/init.lua#L31-L33)).

It is somewhat annoying that neotest does not provide a recommended way for adapters to define their configuration function (at least not explicitly)...

Perhaps we should work around it temporarily, but reach out to either neotest or neotest-playwright authors?

I didn't see anyone complaining about playwright adapter in issues or discussions :/

## Checklist

### DIFF
--- a/lua/lazyvim/plugins/extras/test/core.lua
+++ b/lua/lazyvim/plugins/extras/test/core.lua
@@ -86,6 +86,9 @@ return {
               local meta = getmetatable(adapter)
               if adapter.setup then
                 adapter.setup(config)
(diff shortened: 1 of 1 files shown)


folke:
ty

PR #3833, opened 2024-06-27
fix(lazygit): fix browsing private git hostings

rnesytov:
## What is this PR for?

Current implementation of `Git Browse` doesn't work with onprem gitlab (and any other git hosting except gitlab, github and  bitbucket).
With this little fix, it should work with any git hosting.

No

## Checklist


folke:
ty

PR #3835, opened 2024-06-27
fix(telescope): respect `show_untracked` in user configuration

### LINKED ISSUE #3831: Setting recurse_submodules = true for git_files picker breaks <leader>ff

### Neovim version (nvim -v)

0.10.0 (release)

### Operating system/version

Ubuntu 20.04

### Describe the bug

I use the following configuration in .config/nvim/lua/plugins/telescope.lua so that the git_files picker also looks for files in submodules. 

```
return {
  {

    opts = {
      pickers = {
        find_files = { follow = true },
        git_files = { show_untracked = false, recurse_submodules = true },
      },
    },
  },
}
```
However, that breaks the \<leader\>ff picker when in a git repository, because it gets this error message from `git ls-files`:

```
[telescope.builtin.git_files]: Git does not support both --others and --recurse-submodules
```

This seems to indicate that the LazyVim picker that automatically selects find_files or git_files when in a git repository is somehow passing both options, which is surprising because the show_untracked option for telescope git_files is supposed to be false by default (show_untracked will map to use --others on the git command line).

As you can see above, I also tried to explicitly set the `show_untracked` option to false, but that does not fix the issue. Using `:Telescope find_files` works perfectly, and so does `:Telescope git_files`, so it seems to be coming from the automatic picker of LazyVim.

### Steps To Reproduce

1. Add the telescope.lua plugin configuation file as above
2. Run vim in a git repository
3. Attempt to use \<leader\>ff

### Expected Behavior

It should work just as `:Telescope find_files`.

### Repro

_No response_

dpetka2001:
## What is this PR for?
Respect the option `show_untracked` from Telescope spec in user configuration

Fixes #3831

## Checklist

### DIFF
--- a/lua/lazyvim/plugins/extras/editor/telescope.lua
+++ b/lua/lazyvim/plugins/extras/editor/telescope.lua
@@ -21,6 +21,10 @@ local picker = {
   open = function(builtin, opts)
     opts = opts or {}
     opts.follow = opts.follow ~= false
+    local tele_opts = LazyVim.opts("telescope.nvim")
+    if tele_opts.pickers and tele_opts.pickers.git_files then
+      opts.show_untracked = tele_opts.pickers.git_files.show_untracked ~= false
+    end
     if opts.cwd and opts.cwd ~= vim.uv.cwd() then
       local function open_cwd_dir()
         local action_state = require("telescope.actions.state")


dpetka2001:
@folke With all due respect. Your solution kinda enforces the users to explicitly use different mappings or change all the default ones that use LazyVim auto picker (including the one in Dashboard), which seems a bit counter-intuitive. 

Most new users will just set the option in the Telescope spec and they won't know about what's going on under the hood with LazyVim auto picker. I believe some kind of check in the Telescope picker's `open` function about setting these options would be a better approach.

folke:
Those options are only enforced when using the `auto` picker.
That picker automatically uses `files` or `git_files` where needed.
In this context it would be really weird **not** to see untracked files and for sure lead to users posting issues.

folke:
They can still use `<leader>-fg` to literally use `git_files` and that will use only their options.

dpetka2001:
It's your decision and I respect it. I just wanted to express my concerns.

dpetka2001:
> In this context it would be really weird not to see untracked files and for sure lead to users posting issues.

This would only happen if the user himself put in Telescope spec `show_untracked = false, recurse_submodules = true`, which would be safe to assume that this is what he wants. Otherwise my PR doesn't change the default LazyVim `show_untracked` value.

mux:
@folke I cannot help but share @dpetka2001's concerns here. Out of the various things we discussed, you chose the one option that least respects the user's settings. Not only would `show_untracked` still be ignored if the user sets it to false, but `recurse_submodules` will be similarly ignored. This forces the user into having a more complex configuration that will be harder to maintain, since you have to override two different keymaps to achieve the desired behavior (find files in the menu, and `<leader>ff`). For instance, if there's any change to the auto picker API in the future, this configuration will also have to be updated.

One thing that would make this a little better is if there was a way to set options to instruct the auto picker to use specific options for find_files / git_files since, at least, the user would not have to rebind two keymaps.

PR #3846, opened 2024-06-28
feat(extra): add elixirls code actions

27medkamal:
## What is this PR for?

Add some code actions for the elixir extra. These were mainly adapted from https://github.com/elixir-tools/elixir-tools.nvim/blob/main/lua/elixir/elixirls/init.lua

No

## Checklist

### DIFF
--- a/lua/lazyvim/plugins/extras/lang/elixir.lua
+++ b/lua/lazyvim/plugins/extras/lang/elixir.lua
@@ -9,7 +9,73 @@ return {
     "neovim/nvim-lspconfig",
     opts = {
       servers = {
-        elixirls = {},
+        elixirls = {
+          keys = {
+            { "<leader>cp", ":ElixirToPipe<CR>", desc = "To Pipe" },
+            { "<leader>cP", ":ElixirFromPipe<CR>", desc = "From Pipe" },
+            { "<leader>ce", mode = { "v" }, ":ElixirExpandMacro<CR>", desc = "Expand Macro" },
+          },
+          on_attach = function(client, bufnr)
+            local manipulate_pipes = function(direction)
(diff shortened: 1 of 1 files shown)


georgeguimaraes:
I don't suggest adding these to the Elixir config. People can already add these in their configs, or better yet, use elixir-tools.nvim for that.

27medkamal:
I thought about integrating elixir-tools.nvim into this, but it doesn't play well with lspconfig, and the ecosystem of `lazyvim`. You can't really use its actions without using its elixirls as far as I know. This is why I added these here.

Up to @folke at the end of the day, but even if it's not integrated into lazyvim, this PR should be a good reference for anyone wanting to put these actions in their personal config.

georgeguimaraes:
hmmm I changed my mind on this, after seeing that the [typescript config does set code actions and execute commands too](https://github.com/LazyVim/LazyVim/blob/f64bbd29fb279ba672ce1fc9a79b06ecc4237c87/lua/lazyvim/plugins/extras/lang/typescript.lua#L75-L112).

however, i'd:

1) not define the user commands, only the keymaps;
2) maybe use `LazyVim.lsp.execute` ?

I'm not used to these refactorings from Elixir LS, I'll try to use them, tks for pointing them @27medkamal

27medkamal:
@georgeguimaraes I'm open to this if you want to co-author this PR? Otherwise, I don't really have the time to refactor this at the moment and I don't really see a major issue with its current state.

georgeguimaraes:
Hi @27medkamal, I did it on PR https://github.com/LazyVim/LazyVim/pull/4148. Tks!

folke:
Closing in favor of #4148.

PR #3852, opened 2024-06-29
feat(java): enable Lombok support in jdtls

awilkins:
jdtls needs an argument passed to the JVM to load Lombok as an agent.

Mason downloads `lombok.jar`so we can use that

## What is this PR for?

Configures Lombok support in `jdtls` ; this removes a lot of red underlines in Java projects that use it.

Changes this
image

To this
image


folke:
ty

PR #3858, opened 2024-06-29
feat(cmp): disable `item.menu` for Rust filetypes

### LINKED ISSUE #3855: Rust suggestions look aesthetically bad, unreadable

### Neovim version (nvim -v)

NVIM v0.10.0

### Operating system/version

MacOS 13.6

### Describe the bug

Completion info for the Rust language look wrong compared to other languages. This happens in both cases: with the Rust Extra and without it, with rust-analyzer only. Completion window is too large and Info window is skewed. Typescript and Lua for example are ok.

Screenshot 2024-06-29 at 14 49 18

### Steps To Reproduce

Install rust-analyzer and try completion for `random` for instance.

### Expected Behavior

Readable, good aesthetics. LunarVim for reference
Screenshot 2024-06-29 at 15 38 41

### Repro

```Lua

-- set stdpaths to use .repro
end

-- bootstrap lazy
end

-- install plugins
local plugins = {
  "folke/tokyonight.nvim",
  -- add any other plugins here
}
  root = root .. "/plugins",
})

-- add anything else here
```

dpetka2001:
## What is this PR for?
Tries to improve aesthetics for `nvim-cmp` in Rust by disabling `item.menu` since it contains the `return` types that take up too much space.

Fixes #3855

## Checklist

### DIFF
--- a/lua/lazyvim/plugins/coding.lua
+++ b/lua/lazyvim/plugins/coding.lua
@@ -49,10 +49,13 @@ return {
           { name = "buffer" },
         }),
         formatting = {
-          format = function(_, item)
+          format = function(entry, item)
             local icons = LazyVim.config.icons.kinds
             if icons[item.kind] then
               item.kind = icons[item.kind] .. item.kind
+            end
+            if entry.context.filetype == "rust" then
+              item.menu = nil
             end
             return item
           end,


dpetka2001:
Before


After

osenvosem:
What if it will happen with another language?

dpetka2001:
I don't know other LSP server that does that. If you are aware of such behavior from another LSP server as well, we can add additional entries or disable `item.menu` all together. But I'm not sure how every LSP server uses `item.menu`. Maybe some of them return some other useful info. That's why I was hesitant to disable it all together and chose the more cautious approach to disable it only for Rust filetypes (since I'm not aware of other instances) for the time being. 

Maybe the maintainer has a better idea to get around this that I couldn't think of.

osenvosem:
It would be nice if maintainer with more knowledge could think about reformatting the appearance of an entry and make it more aesthetically pleasant. I copied the `format` function from the LunarVim configuration and it really looks better. They replaced menu item with completion source name which is more informative like this`vim_item.menu = source_names[entry.source.name]`.

And it looks like this in LazyVim:
Screenshot 2024-06-29 at 21 42 58

The only thing I didn't manage to do is to add borders, they aren't displayed for some reason.

dpetka2001:
That's rather subjective in my opinion. I don't like to see the `source.name`. I prefer the current way of showing only icon and `item.kind` and no `item.menu` if it gives superfluous information. But that's just a personal opinion and users should just tailor their configuration according to their preferences. But not everyone's preferences can be taken into account in a Neovim distro. It's just not possible.

folke:
Thanks!
And I also prefer the way LazyVim formats cmp items. That `(LSP)` seems kinda useless imo

AmmarAbouZor:
Hi @folke @dpetka2001 
I think the removed information here can be really important because it will shows the type that will be automatically imported. Without it we will not know if something has been imported and it get really bad when the same type name exists on different libraries like this:
Before:
Screenshot from 2024-06-30 18-42-41

After: 
Screenshot from 2024-06-30 18-43-57

In the After situation you will have no clue which type is gonna be imported

folke:
I'm fine with merging a PR that simply trims too long menu item info....

AmmarAbouZor:
Thanks for the quick response...
These long information are the only way to distinguish between different types using the same name. I don't think anyone will prefer to see two different entries with exactly the same name but each entry of them will imported a different namespace without knowing which namespace are they importing like in the after case.
I don't like the long menus entries either but removing them will hide important information for the users.

I revert that in my configuration but I think this change should be at least optional

dpetka2001:
@folke I have something in the works that looks like this at the moment


Do you know how I can make the space between `item.abbr` and `item.kind` smaller, so that it's not wasted there? Look at the space between `println` and `f Function`.

folke:
No idea tbh, but def looks better already!

dpetka2001:
It was something on my end that added the spaces. Submitted another PR.

osenvosem:
We could make room by removing entry type string `Function`, `Struct`, etc... Icons convey this information.
Also I'm not sure if it is possible, but moving menu_item information on top of the docs floating window would solve the problem.

PR #3871, opened 2024-06-30
fix(neo-tree): remove `branch`, since it obstructs updating to latest version

dpetka2001:
## What is this PR for?
`branch` was obstructing from being able to update to latest HEAD like in other plugins. This was necessary when the initial development was being done on this branch and `main` was used for the old stable release, but this is not necessary any more. Also, @folke now you can use `always_show_by_pattern`, which you might find useful to show the git tracked hidden files like you wanted but couldn't do before. 

No

## Checklist


folke:
Great, thanks!

PR #3873, opened 2024-06-30
feat(cmp): attempt for dynamic width and trimming between fields

dpetka2001:
## What is this PR for?
Attempt for dynamic width and trimming between fields.
Testing was done on my 15.6 laptop screen so maybe values could be raised in dynamic calculation? Also provides a `vim.g.cmp_fixed_width` for the users to be able to define a fixed width in their personal configuration if they'd like to.

Attempts to rectify a concern raised in #3858

## Checklist

### DIFF
--- a/lua/lazyvim/plugins/coding.lua
+++ b/lua/lazyvim/plugins/coding.lua
@@ -74,10 +74,10 @@ return {
             -- Truncate the completion entry text if it's longer than the
             -- max content width. We subtract 3 from the max content width
             -- to account for the "..." that will be appended to it.
-            if #item.abbr > (max_content_width / 2) then
+            if item.menu and #item.abbr > (max_content_width * 0.7) then
               item.abbr = vim.fn.strcharpart(item.abbr, 0, math.floor(max_content_width * 0.7) - 3) .. "..."
             end
-            if item.menu and #item.menu > (max_content_width / 2) then
+            if item.menu and #item.menu > (max_content_width * 0.7) then
               item.menu = vim.fn.strcharpart(item.menu, 0, math.floor(max_content_width * 0.7) - 3) .. "..."
             end
 
--- a/lua/lazyvim/plugins/coding.lua
+++ b/lua/lazyvim/plugins/coding.lua
@@ -54,9 +54,33 @@ return {
             if icons[item.kind] then
               item.kind = icons[item.kind] .. item.kind
             end
-            if entry.context.filetype == "rust" then
-              item.menu = nil
+
+            -- Set 'fixed_width' to false if not provided.
+            local fixed_width = vim.g.cmp_fixed_width or false
+
+            -- Set the fixed completion window width.
+            if fixed_width then
+              vim.o.pumwidth = fixed_width
             end
+
+            -- Get the width of the current window.
+            local win_width = vim.api.nvim_win_get_width(0)
+
+            -- Set the max content width based on either: 'fixed_width'
+            -- or a percentage of the window width, in this case 20%.
+            -- We subtract 10 from 'fixed_width' to leave room for 'kind' fields.
+            local max_content_width = fixed_width and fixed_width - 10 or math.floor(win_width * 0.25)
+
+            -- Truncate the completion entry text if it's longer than the
+            -- max content width. We subtract 3 from the max content width
+            -- to account for the "..." that will be appended to it.
+            if #item.abbr > (max_content_width / 2) then
+              item.abbr = vim.fn.strcharpart(item.abbr, 0, math.floor(max_content_width * 0.7) - 3) .. "..."
+            end
+            if item.menu and #item.menu > (max_content_width / 2) then
+              item.menu = vim.fn.strcharpart(item.menu, 0, math.floor(max_content_width * 0.7) - 3) .. "..."
+            end
+
             return item
           end,
         },
--- a/lua/lazyvim/plugins/coding.lua
+++ b/lua/lazyvim/plugins/coding.lua
@@ -55,14 +55,15 @@ return {
               item.kind = icons[item.kind] .. item.kind
             end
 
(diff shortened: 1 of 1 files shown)


osenvosem:
You can use the ellipsis char `…` instead of three dots.

dpetka2001:
@osenvosem Thanks for the suggestion. I made the change.

lua/lazyvim/plugins/coding.lua
```diff
@@ -54,9 +54,33 @@ return {
...
+
+            -- Truncate the completion entry text if it's longer than the
+            -- max content width. We subtract 3 from the max content width
+            -- to account for the "..." that will be appended to it.
+            if item.menu and #item.abbr > (max_content_width * 0.7) then
+              item.abbr = vim.fn.strcharpart(item.abbr, 0, math.floor(max_content_width * 0.7) - 3) .. "…"
             end
+            if item.menu and #item.menu > (max_content_width * 0.7) then
+              item.menu = vim.fn.strcharpart(item.menu, 0, math.floor(max_content_width * 0.7) - 3) .. "…"
+            end
```
folke, line 82: Maybe just keep this part? And use `vim.g.cmp_item_menu_width`? With some sane default value?
Calculating it dynamically based on the current window width or having a fixed width might not be ideal for everyone.

lua/lazyvim/plugins/coding.lua
```diff
@@ -54,9 +54,33 @@ return {
...
             end
-            if entry.context.filetype == "rust" then
-              item.menu = nil
+
+            -- Set 'fixed_width' to false if not provided.
+            local fixed_width = vim.g.cmp_fixed_width or false
+
+            -- Set the fixed completion window width.
+            if fixed_width then
+              vim.o.pumwidth = fixed_width
```
folke, line 63: I checked the cmp codebase, and afaik this doesn't do anything.
In Neovim `pumwidth` is actually a minimum width for the completion window.
So maybe it is used in cmp when using the native completion window. (which we don't)

dpetka2001: Strange, if I define `vim.g.cmp_fixed_width = 20` then it does show a much smaller window than the default one. Which is being defined in `pumwidth` for fixed size window. So, what would you like the final approach to be like? Just have a `vim.g.cmp_item_menu_width` and only trim the `item.menu` based on that value if possible?

folke:
I just looked into the cmp code and the width of a column is simply the max strdisplaywidth for each field.
So the reason that first column was wide, is because there's an entry somewhere that is that wide.

So an alternative could be a setting (with defaults):

```lua
--- max widths for the columns in the cmp completion menu
vim.g.cmp_widths = {
  abbr = 20,
  menu = 20,
 ...
}
```

20 is just a random number, not sure it actually looks good.

dpetka2001:
So, something like this @folke? I went with 30, because it looks better on my 15.6 inch laptop, but not sure a bigger value would be preferable based on bigger screens. I can't test that to see what it looks like. I also tried with 20, but `item.menu` was showing less info for what I consider the users would like to see.

lua/lazyvim/plugins/coding.lua
```diff
@@ -55,31 +55,11 @@ return {
...
-            -- Truncate the completion entry text if it's longer than the
-            -- max content width. We subtract 3 from the max content width
-            -- to account for the "..." that will be appended to it.
-            if item.menu and #item.abbr > (max_content_width * 0.7) then
-              item.abbr = vim.fn.strcharpart(item.abbr, 0, math.floor(max_content_width * 0.7) - 3) .. "…"
-            end
-            if item.menu and #item.menu > (max_content_width * 0.7) then
-              item.menu = vim.fn.strcharpart(item.menu, 0, math.floor(max_content_width * 0.7) - 3) .. "…"
-            end
+            item.abbr = vim.fn.strcharpart(item.abbr, 0, abbr_width - 3) .. "…"
```
folke, line 61: probably `-1`? Since `ellipses` is one char?

folke: And only needed when strdisplaywidth is larger than maz width

dpetka2001: What max width? I left dynamic calculation completely out of it. I thought that's what you wanted based on this comment of yours
> Calculating it dynamically based on the current window width or having a fixed width might not be ideal for everyone.

folke: I meant when > 30 in your code.
When an `abbr` is only `10` chars, there's no need to truncate it

lua/lazyvim/plugins/coding.lua
```diff
@@ -54,9 +54,13 @@ return {
...
               item.kind = icons[item.kind] .. item.kind
             end
-            if entry.context.filetype == "rust" then
-              item.menu = nil
-            end
+
+            local abbr_width = vim.g.cmp_widths and vim.g.cmp_widths.abbr or 30
+            local menu_width = vim.g.cmp_widths and vim.g.cmp_widths.menu or 30
+
+            item.abbr = vim.fn.strcharpart(item.abbr, 0, abbr_width - 1) .. "…"
```
folke, line 61: so someting like `if vim.fn.strdisplaywidth(item.abbr) > abbr_width then`....

dpetka2001: Oh, now I understand what you mean. Just a sec.

dpetka2001: I also wrapped the trimming inside an `if item.menu` because on LSPs that didn't have `item.menu` it showed another thing in place of the non-existent `item.menu` (like Lua for example).

folke:
I've refactored it a bit. If good for you, it's ready to merge!

dpetka2001:
You know best. I've no problem whatsoever. Just trying my best to keep learning from someone with so much experience :smile:

folke:
Merged, thanks!

PR #3877, opened 2024-07-01
fix(animate): avoid lagging when doing a lot of relative resizing

Old-Farmer:
## What is this PR for?

To avoid lagging when doing a lot of relative resizing.

## Checklist


Old-Farmer:
I think I have to explain about this PR.

When using animate extra, the experience of pressing and then holding <C-Up> (also <C-Down> etc.) to adjust window size is not that smooth(really bad I think) because it happens too quickly so the plugin can't calculate the relative size of window correctly. 

So I add the keymaps to avoid that case by using the mini.animate `execute_after` API to schedule the resize and make the resizing sequentially.

And I also disable the resize animate if it happens too fast when some users don't use LazyVim's keymaps to resize window.

Hope this PR can be understood correctly although the solution is not that nice.

PR #3899, opened 2024-07-03
feat(ui): use `mini.icons` instead of `nvim-wev-devicons`

folke:
## What is this PR for?

Replace the icon support with the new mini library

## Blockers

- [ ] https://github.com/echasnovski/mini.nvim/issues/1007#issuecomment-2206553024

### DIFF
--- a/lua/lazyvim/plugins/ui.lua
+++ b/lua/lazyvim/plugins/ui.lua
@@ -303,41 +303,7 @@ return {
   },
 
   -- icons
-  {
-    "echasnovski/mini.icons",
-    lazy = true,
-    opts = {},
-    init = function()
-      package.preload["nvim-web-devicons"] = function()
-        local Icons = require("mini.icons")
-        local ret = {}
-        package.loaded["nvim-web-devicons"] = ret
-        Icons.mock_nvim_web_devicons()
-
-        local function get(cat)
-          local all = {}
-          for _, name in ipairs(Icons.list(cat)) do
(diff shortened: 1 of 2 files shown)


kevinrobayna:
I tested this locally as I saw it on the Reddit post it works very nicely! and the icons are super pretty!

fpohtmeh:
Any plans to make extras for web dev icons?

dpetka2001:
It was just one line of code `{ "nvim-tree/nvim-web-devicons", lazy = true }`, so I believe users should just install that and disable `mini.icons` in their personal configuration if they wish so instead of waiting for an Extra.

fpohtmeh:
> It was just one line of code `{ "nvim-tree/nvim-web-devicons", lazy = true }`, so I believe users should just install that and disable `mini.icons` in their personal configuration if they wish so instead of waiting for an Extra.

Right, but for many plugins folke created extras to support existing users who rely on plugins which become deprecated now

dpetka2001:
I doubt there will be an Extra with just one line of code. I don't believe it's worth it. Extras are supposed to have significant configuration. At least that's what is mentioned in [Contributing](https://github.com/LazyVim/LazyVim/blob/main/CONTRIBUTING.md).

> Contributing an Extra Plugin
    The plugin should be well-known and require significant configuration.
    Simple specs containing just the plugin with some options will not be accepted.

folke:
I won't create an extra in this case. `mini.icons` provides a wrapper for `nvim-web-devicons`, so any plugin depending on that still works.

If you still want to go back, then you can do what @dpetka2001 suggested.

PR #3904, opened 2024-07-03
feat(fzf-lua): preview keymaps for git

dpetka2001:
## What is this PR for?
It seems that the preview keymaps set right now only support the builtin previewer. 

From searching through `fzf-lua` docs, I found [here](https://github.com/ibhagwan/fzf-lua/blob/3b91c1a471160bd8620bdca8f18743d954994daa/README.md?plain=1#L634-L638) that additional mappings should be set for the other previewers such as `git` (otherwise the defaults `<shift-down/up>` could be used).

I chose to use the same mappings for the `fzf` preview keymaps as they serve the same functionality. Please feel free to change if not to your liking.

Not an issue but a discussion #3895

## Checklist


folke:
ty!

PR #3908, opened 2024-07-04
feat(vue): enabled hybrid mode to avoid 2 typescript lsp running at same time

RayGuo-ergou:
## What is this PR for?

At the moment, the config for vue set hybrid mode to `false` which volar will run a typescript server under the hook. ( hybrid mode false is the takeover mode in v1, was introduced in `2.0.7` see more information here: https://github.com/vuejs/language-tools/pull/4119 ).

However, another vtsls with vue language plugin also attached to vue files, this will cause two typescript server running at the same time. It can be very easily observed with tools like `htop` volar and vtsls will have similar memory usage which is abnormal because volar should be very light by itself. This will introduce issues like duplicate diagnostics, see  https://github.com/vuejs/language-tools/issues/4159#issuecomment-2208101079

In this pull request, I set the hybrid mode to true as default because the hybrid mode is the "correct" way moving forward, thus it would be more stable. Let me know if you feel it should stay as `false` to be default.

No existing issue.

## Checklist

### DIFF
--- a/lua/lazyvim/config/options.lua
+++ b/lua/lazyvim/config/options.lua
@@ -49,11 +49,6 @@ vim.g.bigfile_size = 1024 * 1024 * 1.5 -- 1.5 MB
 -- Show the current document symbols location from Trouble in lualine
 vim.g.trouble_lualine = true
 
--- Enable hybrid mode for Vue files
--- If the hybrid mode is enabled, a sperate vtsls will run for Vue files
--- If the hybrid mode is disabled, volar will run a tsserver under the hood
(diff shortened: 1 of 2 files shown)


folke:
if hybrid mode is the way to go, then this should be configured to just that. ty

iamxiaojianzheng:
Expect to see that in the next release.

folke:
@RayGuo-ergou can you update the PR to always use hybrid mode? It does not need to be configurable.

RayGuo-ergou:
Hi @folke done!

folke:
ty!

PR #3911, opened 2024-07-04
feat(erlang): add Erlang language

georgeguimaraes:
I usually work with Elixir codebases, but sometimes I want to browse and code with Erlang codebases, so here is the config I've been using for Erlang.

### DIFF
--- a/lua/lazyvim/plugins/extras/lang/erlang.lua
+++ b/lua/lazyvim/plugins/extras/lang/erlang.lua
@@ -1,0 +1,20 @@
+return {
+  recommended = function()
+    return LazyVim.extras.wants({
+      ft = { "erlang" },
+      root = { "rebar.config", "erlang.mk" },
+    })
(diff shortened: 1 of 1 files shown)


lua/lazyvim/plugins/extras/lang/erlang.lua
```diff
@@ -0,0 +1,20 @@
+return {
+  recommended = function()
+    return LazyVim.extras.wants({
+      ft = { "erlang" },
+      root = "rebar.config",
```
folke, line 5: shouldn't root include ` 'erlang.mk'`?

georgeguimaraes: you're right. almost all the repos I've dealing with uses rebar3 now, but we should add erlang.mk. Already updated this branch

folke:
ty!

PR #3917, opened 2024-07-05
refactor(trouble): move options in keymaps for lsp and symbols to `opts`

jyuan0:
## What is this PR for?

A user won't also have to override the keymaps if they set options for `Trouble lsp` or `symbols` in `opts` (assuming they want the mode to always open with certain options).

No.

## Checklist

----

The default for `Trouble symbols` already seems to be on the right, and neither lsp/symbols auto-focuses, so I left these out.


folke:
ty

PR #3919, opened 2024-07-05
fix(project): properly close window after changing directory

dotfrag:
## What is this PR for?

In project extra, when using `<c-w>` in the project selection menu to change directory, the window used to show the command's exit code would not close. This fixes it.

N/A.

## Checklist

### DIFF
--- a/lua/lazyvim/plugins/extras/util/project.lua
+++ b/lua/lazyvim/plugins/extras/util/project.lua
@@ -62,7 +62,7 @@ pick = function()
             local path = selected[1]
             local ok = project.set_pwd(path)
             if ok then
-              vim.api.nvim_feedkeys(vim.api.nvim_replace_termcodes("<Esc>", true, false, true), "n", true)
+              vim.api.nvim_win_close(0, false)
               LazyVim.info("Change project dir to " .. path)
             end
           end,


dotfrag:
Folke, when you see this, please note that there is a small chance of this being a "me" problem. The `feed_keys` method didn't work for me to close the window, but I'm assuming it might have worked for the person who originally added this? Using `nvim_win_close` works for me 100% of times.

folke:
I think the reason fedkeys was used is to properly close fzf.
What if you change it to `vim.api.nvim_input("<esc>")` instead?

dotfrag:
Nope:

image

folke:
What if you remove that close or feedkeys completely and do the `set_pwd` in a vim.schedule?

dotfrag:
Like so?

```lua
            vim.schedule(function()
              project.set_pwd(path)
            end)
```

Still no, same as screenshot 2 (shows Process exited 0), window stays open.

folke:
allright, I'll just merge it as is. ty!

dpetka2001:
It only does so in the Dashboard window. Not when you have open another regular buffer.

dotfrag:
Yes, you are right.. maybe this was an issue related to dashboard instead?

dpetka2001:
I believe it is related to the Dashboard window not being `buflisted`. While on the Dashboard window I did `set buflisted` and the problem did not occur then. @folke any thoughts?

dotfrag:
While that might work, we still don't want dashboard to be listed though, right?

iamxiaojianzheng:
Your changes work for me. @dotfrag

dpetka2001:
Well it's just a "No Name" buffer that is only listed while the Dashboard is open. When it closes it goes away. Maybe it might also be related to how `fzf-lua` closes. Maybe it does some check of finding a valid buffer to return to when you hit `<esc>`. Or maybe that's just default Neovim behavior when closing the last buffer maybe? Just guessing.

PR #3929, opened 2024-07-05
fix(lualine): use the new ministarter file type to disable in mini.starter

farzadmf:
## What is this PR for?

`mini.starter` recently had a breaking change to set the file type to `ministarter` instead of the old `starter`; so, `lualine` is enabled in the dashboard because it's using the old file type.

No.

## Checklist


folke:
ty!

PR #3931, opened 2024-07-05
feat(icons): provide language specific icons in extras

m0lson84:
## What is this PR for?

Provide language specific file icons. The intent is to lay the foundation of this type of extension / configuration. I've added icons for file types that I interact with but this is definitely not holistic. I also went back and forth on whether the configuration should be within the core UI config or in the extras (I landed on the later). Definitely open to feedback on these changes.

Nope.

## Checklist


lua/lazyvim/plugins/ui.lua
```diff
@@ -306,7 +306,15 @@ return {
   {
     "echasnovski/mini.icons",
     lazy = true,
-    opts = {},
+    opts_extend = { "directory", "extension", "file", "filetype", "lsp", "os" },
```
dpetka2001, line 309: Sorry, naive question here. Is this really needed? I thought `opts_extend` was to extend list-like tables. Aren't the keys defined here key/value-like tables? I just want to understand and am not insinuating anything.

folke:
Thanks!

I do think it makes most sense to contribute PR's to `mini.icons`.

@echasnovski what's your take on this? What are your requirements to add new icons to `mini.icons` vs leaving it up to the user to configure?

See the diff in this PR.

I currently added this one for example:

```lua
        ["NEWS.md"] = { glyph = "󰎕", hl = "MiniIconsYellow" },
```

lua/lazyvim/plugins/ui.lua
m0lson84: Yeah, you're right I should have omited `opts_extend` in this context.

echasnovski:
> @echasnovski what's your take on this? What are your requirements to add new icons to `mini.icons` vs leaving it up to the user to configure?

I am leaning on a more cautious side of adding new built-in support for icons, i.e. preferring to suggest user configurations.

Which icon names are intended to be supported in 'mini.icons' is planned to be documented. Here is a rough draft:

- Category "filetype" - any filetype that is reasonably used in the wild. This is intended as a widest net for supporting use cases. Users then are encouraged to have a proper filetype detection set up. **Note**: this currently has some problems due to handful of manually tracked extensions which have precedence over `vim.filetype.match()` for speed reasons. See, for example, echasnovski/mini.nvim#1017. Give me couple of days to try to figure out the best compromise here.
- Categories "directory", "file", "extension" categories - popular **not language/software specific**, with few exceptions (like for Git, Make, etc.).
- Category "lsp" - only values of some `*Kind` defined in LSP specification.
- Category "os" - operating systems which have `nf-md-*` icon (which is used as proxy for "popular OS").

echasnovski:
There are now [documented criteria in each category](https://github.com/echasnovski/mini.nvim/blob/45b3540003566e8f19c4cc4b1c9d0b8605a2697c/doc/mini-icons.txt#L267) for an icon to be considered for adding to built-ins. TL;DR: basically what is listed in previous comment.

I plan to add new icon data tomorrow.

folke:
ok, great!

@m0lson84 would be good to check after tomorrow to see what icons can be removed from the PR

echasnovski:
Some of the filetypes from this PR were added to 'mini.icons'. Here is [the commit](https://github.com/echasnovski/mini.nvim/commit/12d275f4c3aeddfa12a758a69aff65090613ca34).

Some notes:
- 'tf' filetype doesn't seem to be for terraform, but for something called [TinyFugue MUD client](https://github.com/neovim/neovim/blob/5b778a64ec2e82a7d8d28c3c36698cd42a41c539/runtime/lua/vim/filetype/detect.lua#L1600-L1611). `vim.filetype.match()` detects terraform only by contents so I've added special handling of 'tf' extension.
- I could not find reasonable usages of 'gotmpl' and 'dotenv' filetypes, so they are not added.

folke:
Awesome, thanks @echasnovski!

The new `NEWS` icon however is currently the same as for changelog.
I used one of the `newspaper` icons. Maybe a better fit?

echasnovski:
> The new `NEWS` icon however is currently the same as for changelog.

This was deliberate as 'NEWS.md' and 'CHANGELOG.md' seem to be used interchangeably. But I am open to changing it to `nf-md-newspaper_variant`.

folke:
Added tests for `mini.icons` that checks that no icons are added that already exist in `mini.icons`...

folke:
@echasnovski for my plugins (and for a lot of repos using conventional commits), the `CHANGELOG.md` file is auto generated based on conventional commits.
The `NEWS.md` file is a manual file, more similar to what you'd find in `:h news`.

I noticed you didn't include any of the javascript/typescript ecosystem, while those are wildly used. Is there a reason for this? Most of those icons are the exact icons used by the related projects.

folke:
Also worth mentioning is that with this PR, fzf-lua seems broken...

No idea why though.

echasnovski:
> I noticed you didn't include any of the javascript/typescript ecosystem icons, while those are wildly used. Is there a reason for this? Most of those icons are the exact icons used by the related projects.

I think drawing the line at supporting languages/siftware only through filetypes is a healthy compromise long term. The last thing I want is to end up with tons of languages 'mini.icons' has to extensively support. There are Neovim distributions and language plugins for that :)

Thinking about it, I might have to export `MiniIcons.set()`, because setting icons now can be done only by calling `MiniIcons.setup()` (which would require careful config preservation).

>  Also worth mentioning is that with this PR, fzf-lua seems broken...

Very strange indeed.

folke:
@echasnovski something seems broken with the devicons integration and setting custom icons..

When I do `require("nvim-web-devicons").get_icons()`, I get the error below:

```
E5108: Error executing lua ...lke/.local/share/nvim/lazy/mini.icons/lua/mini/icons.lua:506: Invalid highlight name: 'MiniIconsGray'
stack traceback:
	[C]: in function 'get_hl_data'
	...lke/.local/share/nvim/lazy/mini.icons/lua/mini/icons.lua:506: in function 'get_hex'
	...lke/.local/share/nvim/lazy/mini.icons/lua/mini/icons.lua:529: in function 'make_icon_tbl'
	...lke/.local/share/nvim/lazy/mini.icons/lua/mini/icons.lua:535: in function 'make_category_tbl'
	...lke/.local/share/nvim/lazy/mini.icons/lua/mini/icons.lua:547: in function 'get_icons'
	[string ":lua"]:1: in main chunk
```

echasnovski:
It is [`MiniIconsGrey`](https://github.com/echasnovski/mini.nvim/blob/12d275f4c3aeddfa12a758a69aff65090613ca34/lua/mini/icons.lua#L139), not `MiniIconsGray`.

folke:
Right, should have spotted that :)
Fixed!

folke:
Merged. ty!

m0lson84:
@echasnovski Thanks for looking this over and bringing some of these suggestions into `mini.icons`. @folke Thanks for cleaning up my changes and fixing the broken areas 😅.

echasnovski:
> Added tests for `mini.icons` that checks that no icons are added that already exist in `mini.icons`...

I am not sure how those tests are run, but ['CODEOWNERS' is present in 'mini.icons'](https://github.com/echasnovski/mini.icons/blob/6fc6877d58635a13efca456cb025f1dfbf59786d/lua/mini/icons.lua#L814) (and is properly present in `MiniIcons.list('file')`), yet tests did not fail for [it also present in LazyVim](https://github.com/LazyVim/LazyVim/blob/aa418a2147f52fbe3225aced3a0e7638cf15ea78/lua/lazyvim/plugins/ui.lua#L312C37-L312C38).

folke:
You're right. The tests only check specs from extras. Will remove!

echasnovski:
> @echasnovski for my plugins (and for a lot of repos using conventional commits), the `CHANGELOG.md` file is auto generated based on conventional commits.
> The `NEWS.md` file is a manual file, more similar to what you'd find in `:h news`.

@folke, do you want to do the honors and choose color for 'NEWS' icon ("󰎕', a.k.a. plain 'nf-md-newspaper')?

folke:
Anything works for me :)

Another docs I frequently use btw is `TODO.md`, but not sure how common that is...

echasnovski:
There are now both 'NEWS{.md}' and 'TODO{.md}' on latest `main`.

folke:
Sweet! ty! :)

PR #3975, opened 2024-07-10
Fix #3973: Fixed pattern for autocomand lazyvim_wrap_spell

### LINKED ISSUE #3973: bug: Autocommand group lazyvim_wrap_spell is not triggering for expected file types

### Neovim version (nvim -v)

0.10.0

### Operating system/version

Fedora 40 - i3 spin

### Describe the bug

The autocommand in LazyVim/lua/lazyvim/config/autocmds.lua with group name lazyvim_wrap_spell is intended to trigger for text file types with a current pattern of 
`pattern = { "*.txt", "*.tex", "*.typ", "gitcommit", "markdown" }`

However, it is not actually setting the local wrap or spell variable to true in files of type *.txt, *.tex, or *.typ.  

This is fixed if the pattern is changed to 
` pattern = { "text", "plaintex", "typst", "gitcommit", "markdown" }`

### Steps To Reproduce

1. Create new text file with nvim test.txt
2.  Enter command  :setlocal spell?       This returns nospell.
4. Enter command  :setlocal wrap?        This returns nowrap.

### Expected Behavior

Since the file type is 'text', I expected that the autocommand would set the local option for spell and wrap to true, but instead they remain set to false.  

The autocommand does work as expected for markdown files. 

### Repro

```Lua

-- set stdpaths to use .repro
end

-- bootstrap lazy
end

-- install plugins
local plugins = {
  "folke/tokyonight.nvim",
  -- add any other plugins here
}
  root = root .. "/plugins",
})

-- add anything else here
```

Mike-Pilmer:
## Description

Autocommand group lazyvim_wrap_spell was not triggering for all of the specified file types.  

The autocommand will now trigger for types "text", "plaintex", "typst", "gitcommit", "markdown" as intended.

## Related Issue(s)

 - Fixes #3973 

## Checklist

### DIFF
--- a/lua/lazyvim/config/autocmds.lua
+++ b/lua/lazyvim/config/autocmds.lua
@@ -87,7 +87,7 @@ vim.api.nvim_create_autocmd("FileType", {
 -- wrap and check for spell in text filetypes
 vim.api.nvim_create_autocmd("FileType", {
   group = augroup("wrap_spell"),
-  pattern = { "*.txt", "*.tex", "*.typ", "gitcommit", "markdown" },
+  pattern = { "text", "plaintex", "typst", "gitcommit", "markdown" },
   callback = function()
     vim.opt_local.wrap = true
     vim.opt_local.spell = true


folke:
ty!

PR #3977, opened 2024-07-10
fix(autocmds): error occurs when bigfile size is nil

Horange321:
> I'm using neovim
When `vim.g.bigfile_size == nil`, then an error will occur: `attempt to compare nil with number`.

After this fix, the comparison will not be made, so no error will occur.

### DIFF
--- a/lua/lazyvim/config/autocmds.lua
+++ b/lua/lazyvim/config/autocmds.lua
@@ -119,7 +119,7 @@ vim.filetype.add({
   pattern = {
     [".*"] = {
       function(path, buf)
-	local bfs = vim.g.bigfile_size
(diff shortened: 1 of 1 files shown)


folke:
that option is always set

msvechla:
I updated neovim to v0.10.0 and I get the exact same error message. If I set `vim.g.bigfile_size` to some value, the error is gone

folke:
@msvechla that can only happen if you're not really using LazyVim

msvechla:
Hmm you are right, something weird was happening. I checked my config against the starter again and this line was missing:

```
 spec = {
```

Afterwards everything started working again and LazyVim also got updated. Thanks for the help!

PR #4000, opened 2024-07-12
fix(flit): use which-key preset keymap descriptions

baodrate:
## Description

The flit config overrides the descriptions provided by the which-key presets plugin for the `f`/`F`/`t`/`T` motions with an unhelpful description text (just the key itself).

Remove the `desc` argument so that which-key uses the existing description (e.g "Move before next char").

## Screenshots

| Before | After |
| ------------- | ------------- |
| Before | After  |

## Checklist


folke:
ty

PR #4006, opened 2024-07-12
fix(outline): use new object for `symbols.filter`

### LINKED ISSUE #4003: bug: settings for outline.nvim cause symbol search crash

### Neovim version (nvim -v)

0.10.0

### Operating system/version

Ubuntu 22.04.4 LTS

### Describe the bug

Before I open `outline.nvim` tab with LazyVim settings, if I hit `<leader>ss`, everything is working. After I open and settings for `outline.nvim` there are two issues:
1/ I don't see functions connected to classes (Rust)
2/ And if I hit `<leader>ss`, this error message is showed:

```
Error executing vim.schedule lua callback: vim/shared.lua:0: bad argument #1 to '?' (string expected, got boolean)
stack traceback:
	[builtin#81]: at 0x64743f732e00
	vim/shared.lua: in function 'tbl_map'
	...l/share/nvim/lazy/telescope.nvim/lua/telescope/utils.lua:143: in function 'filter_symbols'
	...nvim/lazy/telescope.nvim/lua/telescope/builtin/__lsp.lua:325: in function 'handler'
	/usr/share/nvim/runtime/lua/vim/lsp/client.lua:685: in function ''
	vim/_editor.lua: in function <vim/_editor.lua:0>
```

In order to temporary fix it, I just override all LazyVim related settings on this plugin and it works again:
```
return {
  {
    "hedyhli/outline.nvim",
    keys = { { "<leader>o", "<cmd>Outline<CR>", desc = "Toggle Outline" } },
    opts = function()
      local defaults = require("outline.config").defaults
      return defaults
    end,
  },
}
```

### Steps To Reproduce

1/ Enable Rust extra
2/ Open file
3/ Open side outline.nvim
4/ Hit <leader>ss and should show an error

### Expected Behavior

Expect outline to show all functions in class and `<leader>ss` key to work with symbol search

### Repro

```Lua

-- set stdpaths to use .repro
end

-- bootstrap lazy
end

-- install plugins
local plugins = {
  "folke/tokyonight.nvim",
  -- add any other plugins here
}
  root = root .. "/plugins",
})

-- add anything else here
```

dpetka2001:
## Description
In `outline.nvim` we have `symbols.filter = LazyVim.config.kind_filter`. `outline.nvim` adds an entry `exclude = false` by default when it's a table. This entry propagates to `LazyVim.config.kind_filter` and when using `LazyVim.config.get_kind_filter()` in Telescope `<leader>ss` that entry is there as well and causes the error in https://github.com/nvim-telescope/telescope.nvim/blob/bfcc7d5c6f12209139f175e6123a7b7de6d9c18a/lua/telescope/utils.lua#L143. Use `vim.deepcopy` to create a new object for `symbols.filter` in `outline.nvim`, so that the additional entries don't propagate to default `LazyVim.config.kind_filter`.

## Related Issue(s)
Fixes #4003

## Checklist

### DIFF
--- a/lua/lazyvim/plugins/extras/editor/outline.lua
+++ b/lua/lazyvim/plugins/extras/editor/outline.lua
@@ -16,7 +16,7 @@ return {
       local opts = {
         symbols = {
           icons = {},
-          filter = LazyVim.config.kind_filter,
+          filter = vim.deepcopy(LazyVim.config.kind_filter),
         },
         keymaps = {
           up_and_jump = "<up>",


folke:
oh damn, nice find!

dpetka2001:
Just got lucky with the initial error from the OP. 

I looked at Telescope `utils.lua` line 143 and saw that it tried to `string.lower` over the `opts.symbols`. I haven't looked at Telescope's codebase before and wasn't even sure about the error being a boolean. So, I just ended up printing `LazyVim.config.get_kind_filter()` before and after `outline.nvim` was loaded and observed the differences between the 2. That's why I saw the `exluded = false` after `outline.nvim` was loaded. 

So, I just deduced/guessed that it might be the reason why it was failing. Still a long way to go as far as my programming knowledge is concerned.

hedyhli:
I hadn't considered the `exclude` key propagating to other plugins. Thanks for fixing it!

PR #4013, opened 2024-07-13
feat(java): new java mappings format for which-key v3

lawrenceho:
## Description

Use which-key v3 format for java mappings.

`<leader>c` is removed as it is reported as a duplicate mapping in health check.

## Checklist

### DIFF
--- a/lua/lazyvim/plugins/extras/lang/java.lua
+++ b/lua/lazyvim/plugins/extras/lang/java.lua
@@ -237,7 +237,7 @@ return {
                   {
                     mode = "n",
                     buffer = args.buf,
-                    { "<leader>t", name = "+test" },
+                    { "<leader>t", group = "test" },
                     {
                       "<leader>tt",
                       function()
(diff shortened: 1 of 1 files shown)


lua/lazyvim/plugins/extras/lang/java.lua
```diff
@@ -190,30 +190,40 @@ return {
...
-              ["<leader>c"] = { name = "+code" },
-              ["<leader>cx"] = { name = "+extract" },
-              ["<leader>cxm"] = {
-                [[<ESC><CMD>lua require('jdtls').extract_method(true)<CR>]],
-                "Extract Method",
+            wk.add({
+              {
+                mode = "n",
+                buffer = args.buf,
+                { "<leader>cx", name = "+extract" },
```
folke, line 197: should be `group` instead of name and you can remove that `+`

lua/lazyvim/plugins/extras/lang/java.lua
```diff
@@ -190,30 +190,40 @@ return {
...
-              },
-              ["<leader>cxc"] = {
-                [[<ESC><CMD>lua require('jdtls').extract_constant(true)<CR>]],
-                "Extract Constant",
+            })
+            wk.add({
+              {
+                mode = "v",
+                buffer = args.buf,
+                { "<leader>cx", name = "+extract" },
```
folke, line 209: same here

folke (changes requested):

lua/lazyvim/plugins/extras/lang/java.lua
lawrenceho: I just found out about group and then I noticed that I missed these two. I tried to amend them before you saw them. :)

folke (approved):

folke:
ty!

Issue #4004, opened 2024-07-12
bug: debugging C# with omnisharp - netcoredbg.CMD exited with code: 1


Kurren123:
### Did you check docs and existing issues?

- [X] I have read all the LazyVim docs
- [X] I have searched the existing issues of LazyVim
- [X] I have searched the existing issues of plugins related to this issue

### Neovim version (nvim -v)

0.10.0

### Operating system/version

Windows 10

### Describe the bug

I can't debug C# apps using the base lazyVim + the DAP and omnisharp extras. I successfully set a breakpoint but then when doing `<leader>dc` for debug continue, it asks which configuration. I choose "Launch file", select the build dll and get this:

![image](https://github.com/user-attachments/assets/0e251b3e-66c0-47b9-b5a3-19d33ff1283f)

A blank terminal window pops up and in nvim I get this message:

```
DAP Debug adapter didn't respond. Either the adapter is slow (then wait and ignore this) or there is a problem with your adapter or `netcoredbg` configuration. Check the logs for errors (:help dap.set_log_level)
```

My `dap.log` file:

```

[ INFO ] 2024-07-12T14:25:48Z+0100 ] ...ppData/Local/nvim-data/lazy/nvim-dap/lua/dap/session.lua:904 ]	"Breakpoint unverified"	{
  id = 1,
  line = 5,
  message = "The breakpoint is pending and will be resolved when debugging starts.",
  verified = false
}
[ INFO ] 2024-07-12T14:25:48Z+0100 ] ...ppData/Local/nvim-data/lazy/nvim-dap/lua/dap/session.lua:1439 ]	"Process closed"	11712
[ INFO ] 2024-07-12T14:26:25Z+0100 ] ...ppData/Local/nvim-data/lazy/nvim-dap/lua/dap/session.lua:904 ]	"Breakpoint unverified"	{
  id = 1,
  line = 5,
  message = "The breakpoint is pending and will be resolved when debugging starts.",
  verified = false
}
[ INFO ] 2024-07-12T14:26:26Z+0100 ] ...ppData/Local/nvim-data/lazy/nvim-dap/lua/dap/session.lua:1439 ]	"Process closed"	11444
[ ERROR ] 2024-07-12T14:26:48Z+0100 ] ...ppData/Local/nvim-data/lazy/nvim-dap/lua/dap/session.lua:1466 ]	"stderr"	{
  args = { "--interpreter=vscode" },
[...]

PR #4023, opened 2024-07-13
fix(omnisharp): don't detach coreclr debug adapter

Kurren123:
## Description

Debugging c# with the Omnisharp plugin doesn't work, as described by myself in #4004. Turns out this is a known issue with how nvim-dap calls the mason netcoredbg cmd script. [The solution](https://github.com/jay-babu/mason-nvim-dap.nvim/issues/132) is to pass `detach = false` as an option to the debug adapter. I've tested and it works on my windows machine.

## Related Issue(s)

#4004

## Checklist


folke:
ty

PR #4033, opened 2024-07-14
fix(keymaps): don't use gj and gk in V-LINE and V-BLOCK modes

JelteF:
## Description

In regular visual mode gk and gj make sense as the default, but when doing block or line selects regular j and k are better defaults. When you use gk or gj in visual-line mode, nothing new gets selected if stay in the same wrapped line, so all it accomplishes is requiring more inputs to select the next line. In visual-block mode it's even worse as you generally want to select a block, but by using gk/gj on a wrapped line your selection changes horizontally instead of vertically.

## Checklist

### DIFF
--- a/lua/lazyvim/config/keymaps.lua
+++ b/lua/lazyvim/config/keymaps.lua
@@ -5,10 +5,14 @@
 local map = LazyVim.safe_keymap_set
 
 -- better up/down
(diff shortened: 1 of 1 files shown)


folke:
Thank you for this PR, but I'm not interested in adding this.

PR #4042, opened 2024-07-15
feat(lang): add Typst language support

2giosangmitom:
## Description

This pull request adds robust support for the Typst language to LazyVim, integrating several tools to enhance the Typst development experience:

1. **LSP Support**: Integrated with [tinymist](https://github.com/Myriad-Dreamin/tinymist) to provide comprehensive language server protocol support.
2. **Code Formatting**: Planned integration with [typstyle](https://github.com/Enter-tainer/typstyle) for automatic code formatting. Please note that typstyle is currently not available in the mason-registry. We are awaiting the merging of [this PR](https://github.com/mason-org/mason-registry/pull/6479) to enable this feature.
3. **Preview Support**: Added [typst-preview.nvim](https://github.com/chomosuke/typst-preview.nvim) for live preview of Typst documents.

## Checklist

### DIFF
--- a/lua/lazyvim/plugins/extras/lang/typst.lua
+++ b/lua/lazyvim/plugins/extras/lang/typst.lua
@@ -1,0 +1,53 @@
+return {
+  recommended = function()
+    return LazyVim.extras.wants({
+      ft = { "typst" },
+    })
+  end,
+
+  {
+    "nvim-treesitter/nvim-treesitter",
+    opts = {
+      ensure_installed = { "typst" },
+    },
+  },
+
+  -- Wait until this package is available in the mason-registry
+  -- {
+  --   "williamboman/mason.nvim",
+  --   opts = {
+  --     ensure_installed = { "typstyle" },
+  --   },
+  -- },
+
+  {
+    "neovim/nvim-lspconfig",
+    opts = {
+      servers = {
+        tinymist = {},
+      },
+    },
+  },
+
+  {
+    "stevearc/conform.nvim",
+    opts = {
+      formatters_by_ft = {
+        typst = { "typstyle" },
+      },
+    },
+  },
+
+  {
+    "chomosuke/typst-preview.nvim",
+    cmd = { "TypstPreview" },
+    build = function()
+      require("typst-preview").update()
+    end,
+    opts = {
+      dependencies_bin = {
+        ["typst-preview"] = "tinymist",
+      },
+    },
+  },
+}


Conarius:
Awesome already!
Though none-ls added support for typstyle about a month ago so it would be great to put a none-ls config in there (with `optional = true` of course)

lua/lazyvim/plugins/extras/lang/typst.lua
```diff
@@ -0,0 +1,64 @@
...
+      })
+    end,
+  },
+
+  {
+    "chomosuke/typst-preview.nvim",
+    cmd = { "TypstPreview" },
+    build = function()
+      require("typst-preview").update()
+    end,
```
storopoli, line 57:
```suggestion
    end,
    keys = {
      {
        "<leader>cp",
        ft = "typst",
        "<cmd>TypstPreviewToggle<cr>",
        desc = "Typst Preview",
      },
    },
```

lua/lazyvim/plugins/extras/lang/typst.lua
```diff
@@ -0,0 +1,72 @@
...
+      {
+        "<leader>cp",
+        ft = "typst",
+        "<cmd>TypstPreviewToggle<cr>",
+        desc = "Typst Preview",
+      },
+    },
+    opts = {
+      dependencies_bin = {
+        ["typst-preview"] = "tinymist",
```
stefanboca, line 68:
```suggestion
        ["tinymist"] = "tinymist",
```

stefanboca (changes requested):
mason and typst-preview now both support/use tinymist.

stefanboca:
`tinymist` now has `typstyle` formatting compiled in, not requiring `typstyle` to be installed separately.
Maybe it is preferable to enable formatting on the `tinymist` server:
```lua
{
  "neovim/nvim-lspconfig",
  opts = {
    servers = {
      tinymist = {
        settings = {
          formatterMode = "typstyle",
        },
      },
    },
  },
},
```
and prefer lsp formatting over a separate executable:
```lua
{
  "stevearc/conform.nvim",
  optional = true,
  opts = {
    formatters_by_ft = {
      typst = { "typstyle", lsp_format = "prefer" },
    },
  },
},
```

If this change is made, then `typstyle` can be removed from mason's `ensure-installed` and this PR would no longer be blocked by waiting for `typstyle` to be added to the mason registry - although I'm not sure how this interacts with `none-ls`.

2giosangmitom:
Everything is now working as expected, so I'll mark it as open for now. However, there is an issue when updating binaries for `typst-preview.nvim`. We run the update command with each plugin update, but since the plugin is lazy-loaded, the setup isn’t called. This causes `dependencies_bin.tinymist` to be `nil`, resulting in the unnecessary download of the `tinymist` binary which is not needed since we’re using `tinymist` binary provided by `mason.nvim`. Please give me suggestions.

stefanboca (changes requested):

lua/lazyvim/plugins/extras/lang/typst.lua
```diff
@@ -0,0 +1,48 @@
+return {
+  recommended = function()
+    return LazyVim.extras.wants({
+      ft = { "typst" },
+    })
+  end,
+
```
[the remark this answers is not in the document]
2giosangmitom: I think these lines are unnecessary, as `tinymist` should install automatically when added to `opts.servers` in `nvim-lspconfig`.

https://github.com/LazyVim/LazyVim/blob/a66b44a9fe0031c85a536f13f84582a6ba7c3f82/lua/lazyvim/plugins/lsp/init.lua#L222-L246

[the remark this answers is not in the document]
dpetka2001: Yes, that is unnecessary. If defined in `nvim-lspconfig` it will install automatically. That check is being made if the server exists in `all_mslp_servers` or not. If it exists it will be added into `ensure_installed` table, otherwise if it doesn't exist it will just call the setup function assuming that you installed via other means.

lua/lazyvim/plugins/extras/lang/typst.lua
```diff
@@ -0,0 +1,48 @@
...
+          },
+        },
+      },
+    },
+  },
+
+  {
+    "chomosuke/typst-preview.nvim",
+    cmd = { "TypstPreview", "TypstPreviewToggle", "TypstPreviewUpdate" },
+    build = function()
```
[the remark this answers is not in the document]
2giosangmitom: Thank you for suggestion

lua/lazyvim/plugins/extras/lang/typst.lua
```diff
@@ -0,0 +1,60 @@
...
+    opts = {
+      ensure_installed = { "typst" },
+    },
+  },
+
+  {
+    "neovim/nvim-lspconfig",
+    opts = {
+      servers = {
+        tinymist = {
```
stefanboca, line 19:
```suggestion
        tinymist = {
          single_file_support = true,
          root_dir = function()
            LazyVim.root.get()
          end,
```

I've been running into some issues with LSP not attaching if the the typst file isn't in a git repo. Enabling single file support fixes that, and works as normal otherwise. Actually, I'm not sure if `single_file_support = true` is needed because as far as I can tell LazyVim.root will always return the cwd if no other root is found.

There's more discussion about this at https://github.com/neovim/nvim-lspconfig/pull/3075

stefanboca (changes requested):

lua/lazyvim/plugins/extras/lang/typst.lua
```diff
@@ -0,0 +1,62 @@
...
+          single_file_support = true, -- Fixes LSP attachment in non-Git directories
+          root_dir = LazyVim.root.get,
```
folke, line 21: don't set this.
Also no need to enable single_file_support

lua/lazyvim/plugins/extras/lang/typst.lua
```diff
@@ -0,0 +1,62 @@
...
+      formatters_by_ft = {
+        typst = { "typstyle", lsp_format = "prefer" },
+      },
+    },
+  },
+
+  {
+    "chomosuke/typst-preview.nvim",
+    cmd = { "TypstPreview", "TypstPreviewToggle", "TypstPreviewUpdate" },
+    build = function(plugin)
```
folke, line 43: dont use plugin here, since it won't always work as what you have in mind

lua/lazyvim/plugins/extras/lang/typst.lua
```diff
@@ -0,0 +1,62 @@
...
+      local typst_preview = require("typst-preview")
```
folke, line 44: this line would already have done the setup too. That's what lazy does

lua/lazyvim/plugins/extras/lang/typst.lua
```diff
@@ -0,0 +1,62 @@
...
+      typst_preview.setup(plugin.opts) -- Setup preview during build to avoid redundant downloads
```
folke, line 45: remove this line

folke (changes requested):

lua/lazyvim/plugins/extras/lang/typst.lua
2giosangmitom: If `single_file_support` is not enabled, the LSP won't attach to the buffer when it’s not in a Git repository. See the attached screenshot for reference.

image

folke: ok, got it. enable then indeed

lua/lazyvim/plugins/extras/lang/typst.lua
2giosangmitom: Removing this line may cause the plugin to redundantly download the tinymist binary, even though we're already using tinymist via Mason. Any suggestions to prevent this?

folke:
```lua
      local typst_preview = require("typst-preview") -- this will trigger lazy to setup the plugin
      --- typst_preview.setup(plugin.opts) -- remove just this line
```

2giosangmitom: I tested, it still download `tinymist` binary.

folke: Sorry, I don't think I understand what you want to achieve here then. Can you please explain?

2giosangmitom: The typst-preview plugin requires both tinymist and websocat binaries. Since we're already using the tinymist binary downloaded by mason, we don’t need typst-preview to download it again (on Linux, we can check at `~/.local/share/nvim/typst-preview`). By setting `dependencies_bin.tinymist` to a non-nil value, we can prevent this redundant download. That's the goal here.

2giosangmitom: The `typst_preview.update()` will download the tinymist and websocat

2giosangmitom: Yes

folke: No it does not, or yes it does?

2giosangmitom: Just `require("typst-preview").update()` in `build` does not achieve what I want

folke: but why does this need to be ran during build? This is all so werid.
Again, what you have there won't always work. plugin.opts is not the (always) the table that will get passed to setup, so dont use that!

dpetka2001: What if you do
```lua
local typst_preview = require("typst-preview")
local fetch = require("typst-preview.fetch")
-- Set custom binary names
fetch.tinymist_bin_name = "tinymist"
typst_preview.update()
```
Would that work? It seems that [here](https://github.com/chomosuke/typst-preview.nvim/blob/master/lua/typst-preview/fetch.lua) is where the `update()` happens and it essentially calls `fetch()`. And the functions to get the bin's name check the top table 
```lua
local M = {
  -- Exposing this so when platform detection fails user can manually set
  -- bin_name
  tinymist_bin_name = nil,
  websocat_bin_name = nil,
}
```
to see if to get the bin's name or not.

stefanboca: From my testing, it looks like the opts are from the spec are not merged with the default opts from the plugin before/during build when build is first called during install - despite the `require("typst-preview")`. I'm not sure why that is. I managed to fix it by manually setting the required option with
```lua
require("typst-preview.config").opts.dependencies_bin["tinymist"] = "tinymist"
```
rather than calling the plugin's setup in `build`

dpetka2001: But the problem with this approach might be that if a user wants to use the downloaded binary from the plugin, he would also have to change the `build` function.

stefanboca: @folke as far as I can tell, a plugin's setup is not called during the `build` function when the plugin is first installed, even if it is `require`d. This is not the case after the first install, and the plugin is built with eg `:Lazy build <plugin>`. Is this intended behavior in `lazy.nvim`?

folke: yes, that's actually on purpose. Forgot about that. Just the path is added to the rtp, but setup is not ran, since most plugins need their build to be completed before doing setup.

Why does this need to be part of build anyway? Won't the plugin download it when needed?

stefanboca: Just checked, and the plugin **will not** download when needed, only during update. When toggling preview without a build function, the plugin prints this error:

image

I suppose the simplest solution would be to leave out a `build` and submit a FR to the plugin to download the executables as needed, and for now, whenever a user encounters the error they can run the command.

folke: And why not download/install both during build? IS the download that big?

stefanboca: The latest `tinymist` is 52MB, which is not too large but does take a bit (for me around 10s). I guess leaving the build as `require('typst-preview").update()` and having it download the binary a second time isn't too big of a deal, just slightly annoying. Also, this is only an issue when the plugin is installed, so afterwards the binary can safely be removed and it won't be downloaded again (because for subsequent builds, the plugin is configured).

stefanboca: > What if you do
> 
> ```lua
> local typst_preview = require("typst-preview")
> local fetch = require("typst-preview.fetch")
> -- Set custom binary names
> fetch.tinymist_bin_name = "tinymist"
> typst_preview.update()
> ```
> 
> Would that work? It seems that [here](https://github.com/chomosuke/typst-preview.nvim/blob/master/lua/typst-preview/fetch.lua) is where the `update()` happens and it essentially calls `fetch()`. And the functions to get the bin's name check the top table
> 
> ```lua
> local M = {
>   -- Exposing this so when platform detection fails user can manually set
>   -- bin_name
>   tinymist_bin_name = nil,
>   websocat_bin_name = nil,
> }
> ```
> 
> to see if to get the bin's name or not.

Sorry, I missed this earlier. I don't believe this will work because `tinymist_bin_name` is the binary to download from github, but is not used if `opts.dependencies_bin.tinymist` is set, which is checked [here](https://github.com/chomosuke/typst-preview.nvim/blob/06778d1b3d4d29c34f1faf80947b586f403689ba/lua/typst-preview/fetch.lua#L107)

dpetka2001: I tested and it doesn't download `tinymist`. It works because changing `require("typst-preview.fetch").tinymist_bin_name` it also changes the url of the bin in `M.bins_to_fetch()`, so it can't find the binary from the wrong url.

2giosangmitom:
Removed `build` because the plugin now downloads binaries during `setup`

Silzinc:
On a side note, maybe this could be added to the config:
```lua
{
  "folke/ts-comments.nvim",
  opts = { lang = { typst = "// %s" } },
}
```
Typst's commentstring was not defined for me without it.

2giosangmitom:
@folke please checkout this pr again

Eric-Song-Nop:
I also hope that you can register a command or shortcut for the pin main action: https://github.com/Myriad-Dreamin/tinymist/blob/b1bb3196b718da0c3df97020877a343a7b36c488/editors/neovim/README.md#working-with-multiple-file-projects This is used frequently for editing typst.

2giosangmitom:
> I also hope that you can register a command or shortcut for the pin main action: https://github.com/Myriad-Dreamin/tinymist/blob/b1bb3196b718da0c3df97020877a343a7b36c488/editors/neovim/README.md#working-with-multiple-file-projects This is used frequently for editing typst.

What advantages are achieved with this command?

Eric-Song-Nop:
It defines the entry point of the project.

lua/lazyvim/plugins/extras/lang/typst.lua
```diff
@@ -0,0 +1,61 @@
...
+    opts = {
+      dependencies_bin = {
+        tinymist = "tinymist",
+      },
+    },
+  },
+
+  {
+    "folke/ts-comments.nvim",
+    opts = { lang = { typst = "// %s" } },
```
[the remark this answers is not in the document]
Eric-Song-Nop: plz add `/* %s */` to this otherwise ts-comments cannot uncomment block comment

Myriad-Dreamin:
> It defines the entry point of the project.

We can also simulate behavior of that in vscode. We can implicitly call the pin function when the `:TypstPreview` is invoked or the preview is closed. Though I'm not knowledgeable to judge whether it is good to do that in neovim. This is pretty convenient from my personal use in vscode.

<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://github.com/user-attachments/assets/7e29b5d5-4589-4206-a5c1-561e1846b167" type="image/webp">
    img
</picture>

storopoli:
@github-actions don't stale this

ReKylee:
bump?

2giosangmitom:
I hate this stale bot :)

ChristophSchmidpeter:
@2giosangmitom This PR has 23 commits, it is unnecessarily  complex and confusing for a relatively small change. Why all the merge main into PR commits and not rebase? I think this should be rebased into one single commit (or at least few reasonable ones)

2giosangmitom:
> @2giosangmitom This PR has 23 commits, it is unnecessarily complex and confusing for a relatively small change. Why all the merge main into PR commits and not rebase? I think this should be rebased into one single commit (or at least few reasonable ones)

Sorry for the delayed response. While updating this PR, I had to sync changes from main because I ran into some errors with my local setup. Since I'm still not very confident with Git, I ended up updating the branch through the GitHub web interface, which caused all those merge commits. Apologies for the messy history and for making this look unprofessional 🙏.

folke:
ty

PR #4059, opened 2024-07-15
feat(treesitter-context): which-key toggle

stefanboca:
## Description

Add a which-key toggle for treesitter-context

## Screenshots

image

## Checklist

### DIFF
--- a/lua/lazyvim/plugins/extras/ui/treesitter-context.lua
+++ b/lua/lazyvim/plugins/extras/ui/treesitter-context.lua
@@ -1,21 +1,22 @@
 -- Show context of the current function
 return {
   "nvim-treesitter/nvim-treesitter-context",
-  event = "LazyFile",
-  opts = { mode = "cursor", max_lines = 3 },
-  keys = {
-    {
-      "<leader>ut",
-      function()
-        local tsc = require("treesitter-context")
-        tsc.toggle()
-        if LazyVim.inject.get_upvalue(tsc.toggle, "enabled") then
-          LazyVim.info("Enabled Treesitter Context", { title = "Option" })
+  event = "VeryLazy",
+  opts = function()
+    local tsc = require("treesitter-context")
+
+    LazyVim.toggle.map("<leader>ut", {
+      name = "Treesitter Context",
+      get = tsc.enabled,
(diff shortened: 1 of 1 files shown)


folke:
ty!

dotfrag:
Why was the event changed? Is it really necessary?

folke:
Missed that. Probably to make sure the mapping is always available?
However, that shouldnt be needed indeed.

dpetka2001:
It works totally fine with `LazyFile` as well. But does it really matter in the end? `LazyFile` is just an alias for `{ "BufReadPost", "BufNewFile" }` in the end. `VeryLazy` just happens after `Vim/UIEnter`.

dotfrag:
Well, technically it's loaded later. Whether it makes a significant difference that's a different question :)

folke:
Either way is fine. Not a big difference indeed

folke:
Unless that was needed to make it work properly when opening a file from the cmdline?

dpetka2001:
I tested with `LazyFile` and opening Neovim with a file argument and it still worked for me. But even Treesitter doesn't actually load on `LazyFile` even though it's defined in its `event`. It either loads on `VeryLazy` or as a `start` plugin when opening with a file argument, so `LazyFile` isn't even needed there I believe? But whatever.

folke:
LazyFile is still needed for a bunch of plugins to make it work properly when opening a file from the cmdline.

dpetka2001:
Yes, probably. I did say that I may be missing something. But I was only aware of `nvim-lspconfig`, which will miss the `FileType` event on the first file if it's loaded on `VeryLazy` while opening Neovim with a file argument and thus won't have any LSP attached. There might be other plugins that behave like that for sure.

stefanboca:
> Missed that. Probably to make sure the mapping is always available? However, that shouldnt be needed indeed.

You're correct that I changed the event to keep the mapping always available, eg. on the dashboard, and for the toggle state to always be correct. Previously the plugin was also loaded on the key `<leader>ut`, but I wasn't sure how to preserve that while keeping the code simple (I used the `mini.animate` extra as reference for the toggle mapping).

PR #4078, opened 2024-07-16
feat(R): added new which-key group for new install feature

Ben10164:
## Description

A new feature was added to `r.nvim` that added a new key bind. This pr adds that key bind (and group in case more are added) to the current which-key config.

(I also removed a comment that I made previously for increasing the width of the which-key window to fit the longer keybind descriptions, something that eventually was taken out of the previous PR long ago)

## Related Issue(s)

None

## Screenshots

Screenshot 2024-07-16 at 9 24 54 AM

## Checklist


folke:
While you're at it, can you also add:

```lua
          wk.add({
            buffer = true,
           the rest of the mapppings

folke:
Thanks!

PR #4080, opened 2024-07-16
feat(lang): add Lean 4 support

Ben10164:
## Description

Adds language support for lean, a popular proof assistant and theorem prover.

## Related Issue(s)

None

## Screenshots

Language Server

lsp

Lean Execution (infobox on the right)

goals

Keymaps

keymaps

Many More Commands

lots of commands

## Checklist


lua/lazyvim/plugins/extras/lang/lean.lua
```diff
@@ -0,0 +1,141 @@
...
+    "neovim/nvim-lspconfig",
+  },
+  {
+    "hrsh7th/nvim-cmp",
+    optional = true,
+    opts = function(_, opts)
+      opts.sources = opts.sources or {}
+      table.insert(opts.sources, { { name = "nvim_lsp" }, { name = "path" }, { name = "buffer" } })
+    end,
+  },
```
dpetka2001, line 140: I don't think these are needed, since `nvim-lspconfig` is just an empty spec and `nvim-cmp` has already these values by default in LazyVim?
```suggestion
 
```

lua/lazyvim/plugins/extras/lang/lean.lua
```diff
@@ -0,0 +1,141 @@
...
+      ft = { "lean" },
+      root = { "lean-toolchain" },
+    })
+  end,
+  {
+    "Julian/lean.nvim",
+    event = { "BufReadPre *.lean", "BufNewFile *.lean" },
+    dependencies = {
+      "nvim-lua/plenary.nvim",
+    },
```
dpetka2001, line 13:
```suggestion

```
I think this is also not needed?

Ben10164:
@dpetka2001 
I tested this locally, and I noticed that if the file is just:
```lua
return = {
  recommended...
  end,
  {
    "Julian/lean.nvim",
  ...
  }
}
```
it fails and throws `.al/share/nvim/lazy/lazy.nvim/lua/lazy/core/fragments.lua:109: attempt to call method 'find' (a nil value)`.
I also figured out that keeping either either of the two configs (i.e. the empty `nvim-lspconfig`), it works. Im pretty confused.

folke:
That `recommended` is a LazyVim thing.
It only works if there' more than one spec, or exactly one spec, but then it should be part of the spec.
Will have to fix that in lazy at some point

folke:
ty!

PR #4090, opened 2024-07-17
feat(terminal): clear search highlight when opening a terminal

dotfrag:
## Description

Often times when opening a terminal or lazygit when a search is active, words in the terminal are highlighted. In my opinion this is rarely, if not never intended. This attempts to fixes this behavior. The code of `terminal.lua` is a bit beyond me, but I believe I have added it to the right place.

## Related Issue(s)

N/A.

## Checklist


folke:
Makes sense. ty!

PR #4096, opened 2024-07-18
feat(lualine): allow for trouble_lualine to be overriden on buffer

Riobe:
## Description

Right now there is a default `vim.g.trouble_lualine` being set to `true`, and only that variable is being checked when deciding if the trouble output is being appended to the `lualine_c` for the lualine plugin. This is normally nice in code files, where you can get output like `packages/src/index.ts > myFunction` but in some filetypes, the user may not wish for this. In particular, I found if you have files with the `markdown` type that include long headers, then you can easily lose the file name by it trying to include the headers in this location.

Considering that one of the `CONTRIBUTING.md` guidelines is `Ensure all configurations are overridable by the user, using Lazy's specs.`, I figured that allowing this to be overrideable at the user's discretion could be a valuable feature.

This would allow the user to override this on file type by including an autocmd like this in their `lua/config/autocmds.lua` or equivalent:

```lua
-- disable trouble symbols in lualine in text filetypes
  group = vim.api.nvim_create_augroup("disable_trouble_lualine", { clear = true }),
  pattern = { "text", "plaintex", "typst", "gitcommit", "markdown" },
  callback = function()
    vim.b.trouble_lualine = false
  end,
})
```

## Related Issue(s)

None

## Checklist

### DIFF
--- a/lua/lazyvim/plugins/ui.lua
+++ b/lua/lazyvim/plugins/ui.lua
@@ -199,9 +199,11 @@ return {
 
       -- do not add trouble symbols if aerial is enabled
       -- And allow it to be overriden for some buffer types (see autocmds)
-      local trouble_lualine_enabled = vim.b.trouble_lualine == nil or vim.b.trouble_lualine
-      trouble_lualine_enabled = trouble_lualine_enabled and vim.g.trouble_lualine
-      if trouble_lualine_enabled and LazyVim.has("trouble.nvim") then
+      if
+        vim.g.trouble_luline_enabed ~= false
+        and vim.b.trouble_lualine_enabled ~= false
+        and LazyVim.has("trouble.nvim")
+      then
         local trouble = require("trouble")
         local symbols = trouble.statusline
           and trouble.statusline({
--- a/lua/lazyvim/plugins/ui.lua
+++ b/lua/lazyvim/plugins/ui.lua
@@ -198,7 +198,10 @@ return {
       }
 
       -- do not add trouble symbols if aerial is enabled
-      if vim.g.trouble_lualine and LazyVim.has("trouble.nvim") then
+      -- And allow it to be overriden for some buffer types (see autocmds)
+      local trouble_lualine_enabled = vim.b.trouble_lualine == nil or vim.b.trouble_lualine
+      trouble_lualine_enabled = trouble_lualine_enabled and vim.g.trouble_lualine
+      if trouble_lualine_enabled and LazyVim.has("trouble.nvim") then
         local trouble = require("trouble")
         local symbols = trouble.statusline
           and trouble.statusline({


lua/lazyvim/plugins/ui.lua
```diff
@@ -198,7 +198,10 @@ return {
       }
 
       -- do not add trouble symbols if aerial is enabled
-      if vim.g.trouble_lualine and LazyVim.has("trouble.nvim") then
+      -- And allow it to be overriden for some buffer types (see autocmds)
+      local trouble_lualine_enabled = vim.b.trouble_lualine == nil or vim.b.trouble_lualine
+      trouble_lualine_enabled = trouble_lualine_enabled and vim.g.trouble_lualine
+      if trouble_lualine_enabled and LazyVim.has("trouble.nvim") then
```
folke, line 204: Can you change this to `if vim.g.trouble_luline_enabed ~= false and vim.b.trouble_lualine_enabled ~= false and LAzyVim.has....`

Riobe:
I'm no expert on lua and I'm kinda new to this type of thing, so I kept this one small, hopefully useful, and any/all feedback is very welcome. I am working with some markdown files with long headers, and noticed that the file name went away, which was the part I really wanted. I didn't really need the symbols in a markdown file, and when I went to disable that I found I couldn't. That's was the impetus for creating this pull request.

lua/lazyvim/plugins/ui.lua
Riobe: Done. Thanks for the feedback!

folke:
ty!

lua/lazyvim/plugins/ui.lua
```diff
@@ -198,7 +198,12 @@ return {
       }
 
       -- do not add trouble symbols if aerial is enabled
-      if vim.g.trouble_lualine and LazyVim.has("trouble.nvim") then
+      -- And allow it to be overriden for some buffer types (see autocmds)
+      if
+        vim.g.trouble_luline_enabed ~= false
```
dpetka2001, line 203: There's a typo here? Should be `vim.g.trouble_lualine_enabled`.

folke: Fixed!

Aietes:
I'm not sure why, but I still get trouble signs in lualine when `vim.b.trouble_luline_enabled` is set to `false`.

folke:
Check should be an Or. Not home right now, so can't fix

Aietes:
The `and` should be fine, no? I was thinking it's not working due to when the check occurs.

dpetka2001:
Yeah I also think it's the timing. When I open Neovim with a file argument it works correctly. When opening without file argument, then the autocmds get loaded on `VeryLazy` and I think that misses the `FileType` event.

folke:
I'm fixing it.
The check whould also be `trouble_lualine`, not `trouble_lualine_enabled`.

@Riobe you succeeded in making dozens of typos in two lines of text.

folke:
Should work as expected now: https://github.com/LazyVim/LazyVim/commit/f9fdb356f2362e5ae4ef490944b1957b49dc6680

Will be completely disabled with `vim.g.trouble_lualine = false`.

Can be disabled per buffer at runtime with `vim.b.trouble_lualine = false`

Aietes:
Thank you @folke! Also thank you @Riobe for bringing up this feature, it helps a great deal with vue files using tailwindcss, as trouble signs are too crazy in this use case.

PR #4099, opened 2024-07-18
feat(editor): replace nvim-spectre with grug-far.nvim

folke:
## Description

I'm considering to replace [nvim-spectre](https://github.com/nvim-pack/nvim-spectre) with [grug-far.nvim](https://github.com/MagicDuck/grug-far.nvim).

It has a better ui and I like the workflow better.

I won't merge this right away. I'm mostly looking for feedback on whether this would be a good thing and whether the defaults option I've added seem ok.

## Screenshots

image

## Checklist

### DIFF
--- a/lua/lazyvim/config/autocmds.lua
+++ b/lua/lazyvim/config/autocmds.lua
@@ -55,6 +55,7 @@ vim.api.nvim_create_autocmd("FileType", {
   group = augroup("close_with_q"),
   pattern = {
     "PlenaryTestPopup",
+    "grug-far",
     "help",
     "lspinfo",
     "notify",
--- a/lua/lazyvim/plugins/editor.lua
+++ b/lua/lazyvim/plugins/editor.lua
@@ -128,13 +128,22 @@ return {
 
   -- search/replace in multiple files
   {
-    "nvim-pack/nvim-spectre",
-    build = false,
-    cmd = "Spectre",
-    opts = { open_cmd = "noswapfile vnew" },
-    -- stylua: ignore
+    "MagicDuck/grug-far.nvim",
+    opts = { headerMaxWidth = 80 },
+    cmd = "GrugFar",
     keys = {
-      { "<leader>sr", function() require("spectre").open() end, desc = "Replace in Files (Spectre)" },
+      {
+        "<leader>sr",
+        function()
+          require("grug-far").grug_far({
+            prefills = {
+              search = vim.fn.expand("<cword>"),
+              filesFilter = "*." .. vim.fn.expand("%:e"),
+            },
+          })
+        end,
+        desc = "Search and Replace",
+      },
     },
   },
 
--- a/lua/lazyvim/plugins/editor.lua
+++ b/lua/lazyvim/plugins/editor.lua
@@ -135,13 +135,16 @@ return {
       {
         "<leader>sr",
         function()
-          require("grug-far").grug_far({
-            prefills = {
-              search = vim.fn.expand("<cword>"),
-              filesFilter = "*." .. vim.fn.expand("%:e"),
-            },
+          local is_visual = vim.fn.mode():lower():find("v")
+          if is_visual then -- needed to make visual selection work
+            vim.cmd([[normal! v]])
+          end
+          local grug = require("grug-far");
+          (is_visual and grug.with_visual_selection or grug.grug_far)({
+            prefills = { filesFilter = "*." .. vim.fn.expand("%:e") },
           })
         end,
+        mode = { "n", "v" },
         desc = "Search and Replace",
       },
     },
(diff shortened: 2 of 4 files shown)


dpetka2001:
Haven't used it before, but to be honest rarely used spectre as well. Just tried the PR locally and it feels really good. Some additions I would maybe suggest are:
- Setting filetype `grug-far` to `lazyvim_close_with_q` just like spectre
- Maybe adding a keymap for current file like mentioned in [Cookbook](https://github.com/MagicDuck/grug-far.nvim?tab=readme-ov-file#launch-with-the-current-file-as-a-flag-which-limits-searchreplace-to-it). That would mean maybe `<leader>sr` would be better as a group name and add nested keymaps under this group (of course this is something that could be entirely done in the user personal configuration as well and not necessarily here).

folke:
I rarely use spectre as well tbh.

Good idea on those keymaps!

I've also just create a PR that adds treesitter highlighting to the search results https://github.com/MagicDuck/grug-far.nvim/pull/115

dotfrag:
I have rarely used spectre too, but mostly because the workflow wasn't to my liking. grug-far does look and feel better.

1. I think the highlighting of replacement could be better (i.e. show match vs replacement), instead of just replacement. But probably not possible if it's using `rg -r` to do the replacement.

2. I like how it's using `rg` for everything and feels more congruent with my terminal use, similar to how fzf-lua feels. Unlike spectre, which used `rg` to search and `sed` to replace.

3. It has the same problem I encountered with `fzf-lua` in the past, where my config file containing `--hyperlink-format=default` breaks the filenames. I think this flag should be added to https://github.com/MagicDuck/grug-far.nvim/blob/main/lua/grug-far/rg/blacklistedReplaceFlags.lua.

After spending a few minutes on it, I could see myself using it more often than spectre.

lua/lazyvim/config/autocmds.lua
```diff
@@ -55,6 +55,7 @@ vim.api.nvim_create_autocmd("FileType", {
   group = augroup("close_with_q"),
   pattern = {
     "PlenaryTestPopup",
+    "grug-far",
```
dpetka2001, line 58: Maybe you would like to remove `spectre_panel` if you go on with removing `spectre` as a whole. Or maybe just leave it if you plan to move it to Extras. Just a suggestion, don't really know what you plan to do in the end. And if you do decide to remove it, maybe clean up `tsplayground` as well, as that's invalid with `Inspect Tree`.

lua/lazyvim/plugins/editor.lua
```diff
@@ -135,13 +135,16 @@ return {
...
-          require("grug-far").grug_far({
-            prefills = {
-              search = vim.fn.expand("<cword>"),
-              filesFilter = "*." .. vim.fn.expand("%:e"),
-            },
+          local is_visual = vim.fn.mode():lower():find("v")
+          if is_visual then -- needed to make visual selection work
+            vim.cmd([[normal! v]])
+          end
+          local grug = require("grug-far");
```
dpetka2001, line 142: I've never seen this `;` at the end. What does it actually do? Chain statements so that they remain in same scope? Because when I removed it to see if it was a typo, lua_ls complained about `undefined global grug`.

folke: If you would remove that, then lua thinks the next line is a parameter for the function.
Would break lua syntax

kevinrobayna:
I use Spectre quite a bit and I've learned to use it but I think your proposed plugin is clearer and would make others use it better!

MagicDuck:
@dotfrag FYI, forced `--hyperlink-format=none` in https://github.com/MagicDuck/grug-far.nvim/pull/123 , so that should be fixed. Feel free to open an issue if you see a problem 😄 .

dotfrag:
@MagicDuck  thanks! I found another one for you:

```text
-M NUM, --max-columns=NUM
           When given, ripgrep will omit lines longer than this limit in bytes. Instead of printing long lines, only the number of matches in that line is printed.

           When this flag is omitted or is set to 0, then it has no effect.
```

I believe you should force it to 0, otherwise grug might replace text with the warning message for long lines.

folke:
My PR with treesitter highlighting just got merged, so I think this is pretty much ready to go.

image

folke:
If anyone has any additional suggestions, or counter reasons for merging this, then now is the time :)

MagicDuck:
@dotfrag thanks, I'll add that!

xzbdmw:
> FYI, forced --hyperlink-format=none in https://github.com/MagicDuck/grug-far.nvim/pull/123 , so that should be fixed

After tried your plugin I realized my rg was too old to have this flag, updated and it feels so fast 😄 ,maybe I was using x86 version all the time.

PR #4106, opened 2024-07-18
fix(dap): extend dap.configurations from .vscode/launch.json

shelldandy:
## Description

This PR allows the nvim-dap module to read from vscode's launch.json and add those configurations to the ones in nvim-dap.

## Screenshots

Before:

image

After:

image

## Checklist


folke:
This was removed on purpose, see https://github.com/LazyVim/LazyVim/pull/3556
@mehalter any suggestions on why op's screenshots are different when loading them?

shelldandy:
im on an m1 and m2 macbook pros and in neither of them the `launch.json` gets read if that help you guys

thanks!

mehalter:
I can take a look at this tomorrow when I'm back at a computer! I was pretty sure after looking at the code of nvim-dap and doing some tests that it tried to load automatically. But this definitely appears to not be doing that 🤔

mehalter:
Looking at the source code, I must have just been completely mistaken. I'm so sorry! This looks good!

shelldandy:
@mehalter have you had a chance to look at this? thanks!

mehalter:
Yup, take a look at my comment above

lua/lazyvim/plugins/extras/dap/core.lua
```diff
@@ -69,6 +69,11 @@ return {
       vscode.json_decode = function(str)
         return vim.json.decode(json.json_strip_comments(str))
       end
+
+      -- Extends dap.configurations with entries read from .vscode/launch.json
+      if vim.fn.filereadable(".vscode/launch.json") then
+        require("dap.ext.vscode").load_launchjs(nil, {})
```
mehalter, line 75:
```suggestion
        vscode.load_launchjs(nil, {})
```

lua/lazyvim/plugins/extras/dap/core.lua
```diff
@@ -69,6 +69,11 @@ return {
       vscode.json_decode = function(str)
         return vim.json.decode(json.json_strip_comments(str))
       end
+
+      -- Extends dap.configurations with entries read from .vscode/launch.json
+      if vim.fn.filereadable(".vscode/launch.json") then
+        vscode.load_launchjs(nil, {})
```
mehalter, line 75:
```suggestion
        vscode.load_launchjs()
```

This is sortof a nitpick. We can rock the defaults here rather than hardcoding an empty opts. This doesn't change the current behavior but if there are other defaults down the line it would pick them up

mehalter (approved):
This looks good to me! I don't have merge power, but it has my stamp of approval :shipit:

shelldandy:
@folke could you take a look here and merge this please?

folke:
ty!

kezhenxu94:
Hi, with this change, I always get duplicated entries in my `.vscode/launch.json`, any idea?

kezhenxu94:
@shelldandy is it possible that when you raised this PR you were using `mason-nvim-dap.nvim` without this https://github.com/jay-babu/mason-nvim-dap.nvim/pull/129 ? Should we revert this?

dpetka2001:
@mehalter Sorry for bothering you, but since you left your stamp of approval on this, are you sure this is the correct way? I've found https://github.com/mfussenegger/nvim-dap/issues/1270 in which mfussenegger suggests that users should remove the explicit `load_launchjs` call. Maybe the issue of OP was something else? I saw you also removed the specific call from Astronvim [here](https://github.com/AstroNvim/AstroNvim/commit/a4e5fd800a51f9d69d88e0717982682a32c08b26). Maybe this should be reverted indeed? Relevant issue https://github.com/LazyVim/LazyVim/issues/4432

Unfortunately, I'm not familiar with DAP at all, so sorry for bothering you for this.

mehalter:
@dpetka2001 yeah I think I was correct originally when I removed it previously. Then when this PR was opened to revert my change I couldn't re-verify why I removed it. That's when I added the line back into AstroNvim. Then recently I did some testing and it seems you don't need it,  but I can't find in the nvim-dap codebase where it actually calls the load function.

TL;DR: I do think I was correct originally that this manual call is not necessary. Sorry for approving this!

dpetka2001:
@mehalter No need to apologize. You're much more knowledgeable and your contributions are greatly appreciated. 

My programming knowledge is very limited (started actually learning programming earlier this year), but I believe I was somewhat able to track down where it happens in the `nvim-dap` codebase. 

In this [PR](https://github.com/mfussenegger/nvim-dap/pull/1237/files) we can see that [this](https://github.com/mfussenegger/nvim-dap/pull/1237/files#diff-b59b772740a7d213d40288d3c87a48040714405a93860817efa817d8cc2e2a94R287-R290) was added for `dap.launch.json`. Then [here](https://github.com/mfussenegger/nvim-dap/pull/1237/files#diff-b59b772740a7d213d40288d3c87a48040714405a93860817efa817d8cc2e2a94R464-R468) it traverses through the list of all the providers configured including `dap.launch.json` as far as I understand it. [M.getconfigs](https://github.com/mfussenegger/nvim-dap/pull/1237/files#diff-081bdb13da7cf1d6d78a70b667dda71413dfc96c9d03bba0c8806822631ba496R179-R192) is being used to automatically parse the `launch.json` file if it exists as it parses the lines of the file if available and then just passes them into `M._load_json` directly, which actually returns the configurations available in `launch.json` and these should be then merged into `all_configs` as can be seen [here](https://github.com/mfussenegger/nvim-dap/pull/1237/files#diff-b59b772740a7d213d40288d3c87a48040714405a93860817efa817d8cc2e2a94R464-R480). It seems that the function [M.load_launchjs](https://github.com/mfussenegger/nvim-dap/pull/1237/files#diff-081bdb13da7cf1d6d78a70b667dda71413dfc96c9d03bba0c8806822631ba496R196) is only used in the codebase for manually calling the command [DapLoadLaunchJSON](https://github.com/mfussenegger/nvim-dap/blob/90616ae6ae40053103dc66872886fc26b94c70c8/plugin/dap.lua#L27), which under the hood it also uses `M.getconfigs` to get the configuration.

Really not confident how accurate my previous assessment is due to my lack of knowledge, but I believe a PR about reverting this change might be appropriate?

folke:
I don't really use DAP often. So the line should be removed again?

dpetka2001:
I don't use either to be honest. 

I just saw #4432 and pinged Mehalter and then tried to skip through the [PR](https://github.com/mfussenegger/nvim-dap/pull/1237/files) I mentioned and tried to analyze my thoughts going through it, but am not confident I'm 100% correct about it. To me it seems that it's not needed, since the `launch.json` dap configuration seems to be integrated into dap providers as can be seen [here](https://github.com/mfussenegger/nvim-dap/pull/1237/files#diff-b59b772740a7d213d40288d3c87a48040714405a93860817efa817d8cc2e2a94R287-R290) and then traversed [here](https://github.com/mfussenegger/nvim-dap/pull/1237/files#diff-b59b772740a7d213d40288d3c87a48040714405a93860817efa817d8cc2e2a94R464-R468). The only reference of function [load_launchjs](https://github.com/mfussenegger/nvim-dap/blob/90616ae6ae40053103dc66872886fc26b94c70c8/lua/dap/ext/vscode.lua#L200-L221) I could find was [here](https://github.com/mfussenegger/nvim-dap/blob/90616ae6ae40053103dc66872886fc26b94c70c8/plugin/dap.lua#L27)  to create the command `DapLoadLaunchJSON`.

drewvlaz:
Is there a good way to disable this behavior? I currently work in a repo where we have a `launch.json`, but I have my own custom commands that I prefer and would like to skip reading the `launch.json`.

PR #4120, opened 2024-07-20
feat(core): defer clipboard because xsel and pbcopy can be slow

abeldekat:
## Description

See [this](https://github.com/LazyVim/LazyVim/discussions/4112) discussion

TLDR: 
Startup time performance is affected quite significantly when the clipboard provider is `xsel`(linux) or `pbcopy`(macos). I expect an improvement in these cases, especially on older pc's.

This PR resets `vim.opt.clipboard` after the `options` are loaded. Then, on `VeryLazy`, the setting is restored.

I also tested with `yanky`.

Relevant prints:

1. Before resetting `vim.opt.clipboard` in `init`, `vim.print(vim.opt.clipboard)` yields a table which will be captured:

```lua
--- fields
_name = "clipboard",
_value = "unnamedplus",
--- more fields
```

2. Set `vim.opt.clipboard = ""` and `vim.print(vim.opt.clipboard)`, also yields a table:

```lua
--- fields
_name = "clipboard",
_value = "",
--- more fields
```

## Checklist

### DIFF
--- a/lua/lazyvim/config/init.lua
+++ b/lua/lazyvim/config/init.lua
@@ -161,6 +161,7 @@ end
 
 ---@type LazyVimOptions
 local options
+local lazy_clipboard
 
 ---@param opts? LazyVimOptions
 function M.setup(opts)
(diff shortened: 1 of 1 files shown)


dpetka2001:
Tried it out locally and works as expected from my short-time test. Was also able to overwrite the value in `options.lua` user configuration. Also, `vim.o.clipboard` yields the value of the object table, so it's easier to check :stuck_out_tongue:

folke:
ty!

PR #4122, opened 2024-07-20
feat(indent-blankline): add which-key toggles

sjclayton:
## Description

Add which-key toggle mappings for toggling both the indention guides, and also scope highlight on a per buffer basis.

## Checklist

### DIFF
--- a/lua/lazyvim/plugins/ui.lua
+++ b/lua/lazyvim/plugins/ui.lua
@@ -225,28 +225,66 @@ return {
   {
     "lukas-reineke/indent-blankline.nvim",
     event = "LazyFile",
-    opts = {
-      indent = {
-        char = "│",
-        tab_char = "│",
-      },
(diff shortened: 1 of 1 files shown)


folke:
I've refactored the code and removed the scope toggle, since that seems a bit too much.
ty!

taketwo:
What is the reason to create the keymap inside the `opts`-generating function? Isn't `config` function the right place for such things? That is, something like:

```lua
    config = function(_, opts)
      require('ibl').setup(opts)
      LazyVim.toggle.map('<Leader>ug', {
        name = 'Indention Guides',
        get = function() return require('ibl.config').get_config(0).enabled end,
        set = function(state) require('ibl').setup_buffer(0, { enabled = state }) end,
      })
    end,
```

folke:
If you have multiple specs for the same plugin, then all `opts` will be evaluated, but only the last `config`

PR #4126, opened 2024-07-21
feat(python): default to new ruff instead of ruff_lsp

dotfrag:
## Description

Change default python ruff lsp from `ruff_lsp` to `ruff`. It is now marked as stable. I have been using it for a few days without any problems. I use python for relatively small to medium projects. Maybe someone who is using python more rigorously has better feedback.

References:

https://github.com/astral-sh/ruff-lsp (see note)
https://github.com/astral-sh/ruff/releases/tag/0.5.3
https://docs.astral.sh/ruff/editors/setup/#neovim

Also the issue in https://github.com/LazyVim/LazyVim/pull/3057 has been resolved. I tested it and I only get 1 `ruff server` process per nvim instance. The processes close correctly when nvim is closed.

## Checklist

### DIFF
--- a/lua/lazyvim/plugins/extras/lang/python.lua
+++ b/lua/lazyvim/plugins/extras/lang/python.lua
@@ -2,11 +2,12 @@ if lazyvim_docs then
   -- LSP Server to use for Python.
   -- Set to "basedpyright" to use basedpyright instead of pyright.
   vim.g.lazyvim_python_lsp = "pyright"
-  vim.g.lazyvim_python_ruff = "ruff_lsp"
+  -- Set to "ruff_lsp" to use the old LSP implementation version.
+  vim.g.lazyvim_python_ruff = "ruff"
 end
 
 local lsp = vim.g.lazyvim_python_lsp or "pyright"
-local ruff = vim.g.lazyvim_python_ruff or "ruff_lsp"
+local ruff = vim.g.lazyvim_python_ruff or "ruff"
 
 return {
   recommended = function()
@@ -44,6 +45,12 @@ return {
         },
         ruff = {
           enabled = ruff == "ruff",
+          cmd_env = { RUFF_TRACE = "messages" },
+          init_options = {
+            settings = {
+              logLevel = "error",
+            },
+          },
         },
         [ruff] = {
           keys = {
--- a/lua/lazyvim/plugins/extras/lang/python.lua
+++ b/lua/lazyvim/plugins/extras/lang/python.lua
@@ -31,28 +31,22 @@ return {
     "neovim/nvim-lspconfig",
     opts = {
       servers = {
-        pyright = {
-          enabled = lsp == "pyright",
-        },
-        basedpyright = {
-          enabled = lsp == "basedpyright",
-        },
-        [lsp] = {
-          enabled = true,
-        },
-        ruff_lsp = {
-          enabled = ruff == "ruff_lsp",
-        },
         ruff = {
-          enabled = ruff == "ruff",
           cmd_env = { RUFF_TRACE = "messages" },
           init_options = {
             settings = {
               logLevel = "error",
             },
           },
+          keys = {
+            {
+              "<leader>co",
+              LazyVim.lsp.action["source.organizeImports"],
+              desc = "Organize Imports",
+            },
+          },
         },
-        [ruff] = {
+        ruff_lsp = {
           keys = {
             {
               "<leader>co",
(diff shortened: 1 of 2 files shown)


dotfrag:
If this works out without issues, we can also get rid of the old one. Unless you are worried about breaking user's config.

dpetka2001:
I've read the Neovim setup and it states the following
> By default, Ruff will not show any logs. To enable logging in Neovim, you'll need to set the RUFF_TRACE environment variable to either messages or verbose, and use the [logLevel](https://docs.astral.sh/ruff/editors/settings/#loglevel) setting to change the log level:
> ```lua
> require('lspconfig').ruff.setup {
>  cmd_env = { RUFF_TRACE = "messages" }
>  init_options = {
>    settings = {
>      logLevel = "debug",
>    }
>  }
>}
```
Might I suggest you also incorporate these changes into this PR for `ruff` server, since if it's going to be used as default, it might make it easier for users if they run into any problems? I'm not sure about the differences between `messages` and `verbose` in `RUFF_TRACE`, so maybe you have a better understanding as to what is more appropriate.

**Comment by dotfrag** [comment]
I noticed it in the docs, but I was afraid it be spammy and didn't know if it create logs. Let me give it a go.

**Comment by dpetka2001** [comment]
I also don't know if they mean a more verbose logging information in `LspLog`. From my understanding of how I read it, it seems that it doesn't show any messages in `LspLog` at all, but of course I might be mistaken (since I haven't used it) and it shows basic info like other LSP servers and this is meant only for more verbose logging (in which case this might not be necessary).

**Comment by dotfrag** [comment]
> By default, this will write logs to stderr which will be available in Neovim's LSP client log file (:lua vim.print(vim.lsp.get_log_path())). It's also possible to divert these logs to a separate file with the logFile setting.

**Comment by dpetka2001** [comment]
Yes, but it's referring to the previous code snippet, in which you enable logging.

**Comment by dotfrag** [comment]
I tried them all. Imho, even warn is spammy for the average user. I'm debating between `warn` and `error`.

Error is the most user friendly, warn is the most dev-friendly.

I'm setting it to `error`, if Folke has a different opinion I can change it.

**Comment by dotfrag** [comment]
@dpetka2001 from main branch, do you know why we have duplicate keys instead of a single one?

https://github.com/LazyVim/LazyVim/blob/43dbe0e60d0083f38ab5113a82d9c74229895af2/lua/lazyvim/plugins/extras/lang/python.lua#L45-L56

**Comment by dpetka2001** [comment]
You applied to `[ruff]`, which means that it will also be set for `ruff-lsp`. The `[ruff]` is a key with the variable `local ruff` that is set to either `ruff_lsp` or `ruff`. It would be better if you applied it to only `ruff` I believe.

**Comment by dotfrag** [comment]
Ohh, didn't realise, thanks!

**Comment by folke** [comment]
This won't work as expected since `[ruff]` will simply override `ruff` now.
Will refactor this PR a bit.

**Comment by folke** [comment]
That should fix it

**Comment by folke** [comment]
Merged. ty!

**Comment by dotfrag** [comment]
Thanks, sorry for the trouble!

**Comment by dpetka2001** [comment]
Just out of curiosity and for me to understand better? Wouldn't the `[ruff]` and `ruff` specs get merged? And if the specs are evaluated in the order they are parsed, could it maybe have worked if we changed the order, so that `[ruff]` was first spec and `ruff` was evaluated after?

**Comment by folke** [comment]
You would basically do the same as:

```lua
{
  test = 1,
  test = 2,
}
```

`test` would randomly be `1` or `2`. Merging has no impact here, since the value will already be set on the table.

dpetka2001:
So it was random that the keys traversed to the ruff server table? Because i tried it 4 times i believe and every time the result was the same.

folke:
Again, there is no merging happening.

Take the example below:

```lua
{
  test = { foo = "bar" },
  test = { bar = "baz" },
}
```

`test` will be one of the two. Not merged.
Lua does not guarantee an order of which value would prevail either, so it's just something you should never do.

If you add the code above in a lua file, then luals will even show a diagnostic about that.

folke:
Previously it didn't matter, since all values were just `enabed = ...`.
That's also why I initially added it like that.
But now that we set additional properties, this would lead to issues.

dpetka2001:
Ok I see what you mean. The `enabled` property also didn't make it to the ruff server table. It was just the keys from `[ruff]`. But no matter how many times I tried (I tried about 7-8 now) it was always the second key that overwrote the key before. Maybe it was still random and was pure coincidence? But you're right that the additional properties in `[ruff]` (the `keys` property) didn't make it into the ruff server table if I reversed the order.

doulighan:
I believe my "format on save" stopped working with this change, I was relying on ruff for formatting rather than black. I didn't configure anything extra in lazy, just added some [ruff] rules to pyproject.toml.  Anyone know how to get it working again with the new lsp?

dotfrag:
@doulighan It works for me, are you sure it's not your config?

What do you see in `:ConformInfo`?

image

PR #4130, opened 2024-07-21
fix(editor): filesFilter expression

madflow:
## Description

With the current default filter, when opening nvim with `nvim .` I get:

```
rg: No files were searched, which means ripgrep probably applied a filter you didn't expect.
Running with --debug will show why files are being skipped.
```

I get expected results with `*.*`.

## Screenshots

grafik

## Checklist


folke:
that's completely wrong.
will add a proper fix

PR #4133, opened 2024-07-21
feat(text): Add Harper to grammar check comments and text

georgeguimaraes:
I've been testing a lot of English grammar/style checkers.

I liked https://github.com/elijah-potter/harper the best. It's simple to set up, and it adds very useful Code Actions to add words to your global (or file) dictionaries.

So, here's my contribution to add it to LazyVim. Not sure if you want to create this `text` dir but it feels to me that harper would not belong to linting, coding, or any other dir.


smarquez1:
Is there any way to enable Harper only when `set spell` is enabled?

georgeguimaraes:
> Is there any way to enable Harper only for buffers where `set spell` is enabled?

@smarquez1 I'm sorry, I'm afraid I don't know how.

futile:
> Is there any way to enable Harper only for buffers where `set spell` is enabled?

Maybe an autocmd with `OptionSet` could work? https://neovim.io/doc/user/autocmd.html#OptionSet

folke:
Thanks for the PR! However, this doesn't need to be an extra. Users can simply install `harper-ls` via Mason and it will automatically work with LazyVim's LSP setup. No additional configuration is needed.

For simple LSP servers that work out of the box, we prefer to keep LazyVim minimal and let users install them directly rather than adding extras.

―  

PR #4139, opened 2024-07-22
feat(markdown): replace `headlines.nvim` by `markdown.nvim`

georgeguimaraes:
Okay, so https://github.com/MeanderingProgrammer/markdown.nvim?tab=readme-ov-file is a new kid on the block that I've been enjoying a lot. It replaces headlines.nvim.

This new extra is a way for folks to try markdown.nvim effortlessly.

(Should I add here everything from the current markdown extra to make this extra a complete drop-in replacement?)


dotfrag:
I think we should keep markdown stuff into a single extra, i.e. decide between headlines or this. Not both.

Also, imo it's best to not change the highlight groups, and instead add them to tokyonight, so the plugin is supported for all people who might use the markdown plugin.

folke:
I've been using that plugin for a while as a replacement for headlines.nvim.
The only thing keeping me from changing it for now, is that the plugin doesn't properly deal with default highlight groups.

I still need to file an issue, or create a PR in the repo, so that it's easier for colorschemes to override the default hl groups.

folke:
This is what I'm using right now as a reference:

```lua
{
    "OXY2DEV/markview.nvim",
    enabled = true,
    opts = {
      checkboxes = { enable = false },
      links = {
        inline_links = {
          hl = "@markup.link.label.markown_inline",
          icon = " ",
          icon_hl = "@markup.link",
        },
        images = {
          hl = "@markup.link.label.markown_inline",
          icon = " ",
          icon_hl = "@markup.link",
        },
      },
      code_blocks = {
        style = "language",
        hl = "CodeBlock",
        pad_amount = 0,
      },
      list_items = {
        shift_width = 2,
        marker_minus = { text = "●", hl = "@markup.list.markdown" },
        marker_plus = { text = "●", hl = "@markup.list.markdown" },
        marker_star = { text = "●", hl = "@markup.list.markdown" },
        marker_dot = {},
      },
      inline_codes = { enable = false },
      headings = {
        heading_1 = { style = "simple", hl = "Headline1" },
        heading_2 = { style = "simple", hl = "Headline2" },
        heading_3 = { style = "simple", hl = "Headline3" },
        heading_4 = { style = "simple", hl = "Headline4" },
        heading_5 = { style = "simple", hl = "Headline5" },
        heading_6 = { style = "simple", hl = "Headline6" },
      },
    },

    ft = { "markdown", "norg", "rmd", "org" },
  }
```

It disables some of the bling by default to make the plugin a little less intrusive.

folke:
I've been working on support for tokyonight as well, but like I said, the plugin currently doesn't use default groups like I configured in the spec.

dotfrag:
I'm stealing that spec, thanks for sharing :)

It is in fact way better than headlines, and fast too.

georgeguimaraes:
It's a different plugin @folke:

yours:
https://github.com/OXY2DEV/markview.nvim

mine:
https://github.com/MeanderingProgrammer/markdown.nvim

I'll try yours as well.

folke:
Right, indeed a different plugin.
I tried the one from this PR as well, but I had issues with it. Don't remember what it was though.
Need to check again.

folke:
Found the biggest reason. `markdown.nvim` does not use `nvim_buf_attach`, which makes it slower.

georgeguimaraes:
Hmm makes sense!

georgeguimaraes:
btw, I had to add 

`
  {
    "folke/tokyonight.nvim",
    opts = {
      plugins = {
        all = true,
        headlines = true,
      },
    },
  },
`

In order to have CodeBlock and Headlines1..6 hls even if headlines.nvim disabled.

folke:
That's because I didn't push my changes in tokyonight yet.
Waiting till https://github.com/OXY2DEV/markview.nvim/issues/61 is done.

MeanderingProgrammer:
> Found the biggest reason. markdown.nvim does not use nvim_buf_attach, which makes it slower.

Hi @folke, I saw this thread and am curious about this comment. Neither plugin uses `nvim_buf_attach`.

- https://github.com/search?q=repo%3AOXY2DEV%2Fmarkview.nvim+nvim_buf_attach&type=code
- https://github.com/search?q=repo%3AMeanderingProgrammer%2Fmarkdown.nvim%20nvim_buf_attach&type=code

Both create autocommands at a buffer level, so performance should be nearly identical.

- https://github.com/search?q=repo%3AOXY2DEV%2Fmarkview.nvim+nvim_create_autocmd&type=code
- https://github.com/search?q=repo%3AMeanderingProgrammer%2Fmarkdown.nvim+nvim_create_autocmd&type=code

I imagine the idea you have here is to do some kind of minimal re-render using the `on_lines` callback, which is a solid point, created an issue on my end to look into that: https://github.com/MeanderingProgrammer/markdown.nvim/issues/84.

There is a legitimate performance hit from `markdown.nvim` to support anti-conceal behavior, hiding virtual text added by the plugin on the cursor line to enable fluid editing. This uses the `CursorMoved` event to check which marks to hide so ends up with a performance penalty. This behavior can be disabled with `anti_conceal = { enabled = false }`, and will end up behaving the same as `markview.nvim`.

I totally get an aesthetic preference for one plugin vs. the other, but I do not think either offers improved performance. Please lmk if I'm missing something.

Much love for all the work you do in this community!

folke:
@MeanderingProgrammer looks I was wrong on that indeed. Thank you for clarifying!

Will take a closer look to markdown.nvim again as well!

For optimizing performance, a combination of `nvim_buf_attach` with `nvim_set_decoration_provider` would probably be the best approach:
- `nvim_buf_attach`: any heavy lifiting per line could happen there
- `nvim_set_decoration_provider` only sets extmarks for calculated data.
    * this probably would also make your `anti-conceal` on the current line trivial? (by not setting any ephermal extmark when line is the current line)

This is just a suggestion though, I'm sure your current approach works really well already. Will try it out again!

georgeguimaraes:
another point in favor of markdown.nvim, it's super easy to add [livebook](https://livebook.dev/) filetypes (*.livemd) to it.

MeanderingProgrammer:
@folke, I am not sure if it is more efficient to redraw `ephemeral` marks on every redraw cycle or removing and re-drawing the marks every cursor movement (at least that is my interpretation of the difference), but I will try it out!

It seems like an interesting approach, and I should probably be more aware of the latency of different approaches. So far I've more or less done an eyeball test on various files as my latency benchmark.

@georgeguimaraes, I appreciate your enthusiasm and that you like my plugin! Has definitely been a fun project for me as well that's evolved in fun ways.

I would summarize the difference between the 2 as such, this is obviously biased on my familiarity:

## markdown.nvim

* Anti-conceal behavior, hiding marks on cursor line. This is not too hard a feature and could be ported to `markview.nvim`, probably a large refactor though.
* Compatible with neovim >= `0.9.0`, with `0.10.0` adding features. I will probably break this backwards compatibility at some point to simplify some code and remove different version paths.
* Supports any filetype so long as it has markdown. Also a feature you could port to `markview.nvim`.
* Older so has probably dealt with more weird edge cases.

## markview.nvim

* Built for `0.10.0` and makes large use of `inline` marks
* Many more style options that look more like full renders
* Focussed on viewing markdown files, currently not configurable I believe?
* A ton of people love the way it looks

I definitely have more familiarity with my own plugin and `markview.nvim` is doing great. Just throwing in my thoughts.

folke:
@MeanderingProgrammer I like it a lot!

The only thing I would like to change is to disable the heading icons, but that currently doesn't seem possible?

folke:
And the only thing I prefer with markview.nvim, is that the bg for codeblocks does not extend to the whole width of the screen

folke:
Just added proper support to tokyonight https://github.com/folke/tokyonight.nvim/commit/71429c97b7aeafecf333fa825a85eadb21426146

folke:
I've updated the PR to replace headlines.nvim with markdown.nvim

folke:
- [ ] https://github.com/MeanderingProgrammer/markdown.nvim/pull/86

folke:
I prefer tokyonight's inline styles for inline code blocks. I currently removed inline_code from the treesitter, queries, but would be great to have a separate config option for the hl group for inline styles instead, that defaults to the same one as for code blocks.

folke:
With those PRs, markdown.nvim is perfect :)

codingluke:
It's great to see my two favorit plugins to get aligned together, you two make my daly life just more colorful 🌈 .

kevinrobayna:
@folke I think you merged (https://github.com/folke/tokyonight.nvim/commit/71429c97b7aeafecf333fa825a85eadb21426146) support in Tokyo for https://github.com/OXY2DEV/markview.nvim rather than https://github.com/MeanderingProgrammer/markdown.nvim was that expected?

folke:
That was correct, but incorrect url. It's quite confusing 😅 I fixed it in a later commit though

folke:
Merged, ty!

dotfrag:
Seeing double chain link icons in LazyVim changelog `<leader>L`. Not sure where they are coming from? I don't see an icon in `CHANGELOG.md`.

dotfrag:
From testing the PRs.. yes sorry.

PR #4141, opened 2024-07-22
fix(dap): don't mess up DAP adapters provided by nvim-dap-python

moetayuko:
## Description

* mason-nvim-dap.nvim ships a `Python: Launch file` adapter which does the same thing as `Launch file` of nvim-dap-python, providing both doesn't make sense and confuses the users.
* mason-nvim-dap.nvim unexpectedly overrides nvim-dap-python's adapters and breaks its venv detection. As a result, user programs are always executed in mason debugpy venv rather than the activated one.

## Related Issue(s)

Fixes #3064

## Checklist


folke:
ty!

PR #4161, opened 2024-07-24
Add `name` to markdown.nvim config

se-jaeger:
## Description

The latest update uses [markdown.nvim](https://github.com/MeanderingProgrammer/markdown.nvim), which crashed for me because `render-markdown` could not be found. 

This pull request adds the `name = "render-markdown",` according to their documentation, which solves it for me. Also tried to use `main` without success.

## Checklist

### DIFF
--- a/lua/lazyvim/plugins/extras/lang/markdown.lua
+++ b/lua/lazyvim/plugins/extras/lang/markdown.lua
@@ -93,6 +93,7 @@ return {
 
   {
     "MeanderingProgrammer/markdown.nvim",
+    name = "render-markdown",
     opts = {
       file_types = { "markdown", "norg", "rmd", "org" },
       code = {


se-jaeger:
Without I get the following error:
```
Error  13:59:44 notify.error lazy.nvim Failed to run `config` for markdown.nvim
...azy/LazyVim/lua/lazyvim/plugins/extras/lang/markdown.lua:110: module 'render-markdown' not found:
	no field package.preload['render-markdown']
```

I believe the reason for this is that in `extras/lang/markdown.lua` you're requiring `render-markdown`, which does not exist. Setting `name` kind of creates and alias from  `MeanderingProgrammer/markdown.nvim` to `render-markdown`.

folke:
render-markdown does exist.
Did you install another plugin called `markdown.nvim` in your config?

PR #4164, opened 2024-07-24
feat(cmp): add `opts_extend` to make it easier for user to config `cmp`

huwqchn:
## Description

Since `lazy.nvim` a new configuration option has been added called `opts_extend`.
so add `opts_extend` make it easier for user to change `cmp`'s mapping, sources, etc.

## Related Issue(s)

no

## Checklist

- [x]  I've read the [CONTRIBUTING](https://github.com/LazyVim/LazyVim/blob/main/CONTRIBUTING.md) guidelines.


dpetka2001:
`opts_extend` is meant to be used only for list-like tables. In the case of this PR it will only work for `opts.sources` I believe.

huwqchn:
> `opts_extend` is meant to be used only for list-like tables. In the case of this PR it will only work for `opts.sources` I believe.

I think you are right.

PR #4184, opened 2024-07-25
feat(elixir): add support for livebook files (`*.livemd`)

georgeguimaraes:
Livebook is a subset of Markdown, so we can safely use markdown's treesitter for it.

### DIFF
--- a/lua/lazyvim/plugins/extras/lang/elixir.lua
+++ b/lua/lazyvim/plugins/extras/lang/elixir.lua
@@ -1,7 +1,7 @@
 return {
   recommended = function()
     return LazyVim.extras.wants({
-      ft = { "elixir", "eelixir", "heex", "surface" },
+      ft = { "elixir", "eelixir", "heex", "surface", "livebook" },
       root = "mix.exs",
     })
   end,
@@ -40,7 +40,11 @@ return {
   },
   {
     "nvim-treesitter/nvim-treesitter",
-    opts = { ensure_installed = { "elixir", "heex", "eex" } },
+    opts = function(_, opts)
+      opts.ensure_installed = opts.ensure_installed or {}
+      vim.list_extend(opts.ensure_installed, { "elixir", "heex", "eex" })
+      vim.treesitter.language.register("markdown", "livebook")
+    end,
   },
   {
     "nvim-neotest/neotest",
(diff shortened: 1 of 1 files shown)


georgeguimaraes:
If https://github.com/LazyVim/LazyVim/pull/4259 is merged, I need to update this PR changing it to use `render-markdown.nvim`

georgeguimaraes:
Updated

lua/lazyvim/plugins/extras/lang/elixir.lua
```diff
@@ -40,7 +40,10 @@ return {
   },
   {
     "nvim-treesitter/nvim-treesitter",
-    opts = { ensure_installed = { "elixir", "heex", "eex" } },
+    opts = function(_, opts)
+      opts.ensure_installed = { "elixir", "heex", "eex" }
```
dpetka2001, line 44:
```suggestion
      vim.list_extend(opts.ensure_installed, { "elixir", "heex", "eex" })
```
In this case these 3 parsers get correctly added to `nvim-treesitter` spec, because of the `opts_extend` feature, however if in the feature some other extra implements nvim-treesitter `opts` as a function then it will be overwritten. I tested with enabling this PR locally and changing the Python Extra nvim-treesitter `opts` into a function and the Elixir Extra `opts.ensure_installed` are getting overwritten because the Elixir Extra is parsed first and the Python Extra is parsed later on.

dpetka2001: Actually, my previous statement about `opts_extend` does not hold true after further investigation and experimentation. It depends entirely on the order the files are parsed (alphabetically). 

Thus Elixir Extra overwrites every `nvim-treesitter` spec that has been parsed before it and everything that comes after it will overwrite the Elixir Extra if nvim-treesitter `opts` happen to be a values function (as far as `ensure_installed` is concerned if it's not `vim.list_extend`ed). You can do `:=LazyVim.opts("nvim-treesitter").ensure_installed` to see what will be enabled. In the most simple case that only the base `nvim-treesitter` spec is present and this Extra gets enabled, the base spec will be overwritten (so `ensure_installed` will not include `bash`, `c`, `diff` etc parsers).

georgeguimaraes: Tks, I also added `opts.ensure_installed = opts.ensure_installed or {}` just as other extra files do.

I initially rebased and lost the ability to commit your suggestion, but I manually added a Co-authored-by line in my commit message and it seems it worked to also attribute it to you.

PR #4190, opened 2024-07-25
fix(fzf): decouple `defaults` from `opts` to easy switch profiles

dpetka2001:
## Description
This decouples `defaults` from `opts`, so that if users switch to another profile they can take advantage of 
the profile's default settings and prompts.

I use `default-title` either way. I just stumbled upon this when I tried the rest profiles and thought maybe there
would be users who would prefer to have the default prompts if they chose another profile.

## Related Issue(s)
No

## Checklist


farzadmf:
@dpetka2001 question: could you please tell me what profiles are in FZF? Is it something in `fzf-lua`?

dpetka2001:
Yes, this is specific to `fzf-lua` fuzzy finder. It has the [profiles listed here](https://github.com/ibhagwan/fzf-lua?tab=readme-ov-file#profiles) and each one has its own layout and prompts. Read the docs for more info.

PR #4225, opened 2024-07-30
feat(eslint): add option to disable eslint auto format

27medkamal:
## Description

This adds an option to disable eslint's autoformatting. The reasons for not wanting an auto format by eslint are:
1. It is generally not recommended not to use a linter as a formatter as highlighted by https://typescript-eslint.io/users/what-about-formatting/#:~:text=We%20recommend%20against%20using%20ESLint,dprint%2C%20or%20an%20equivalent%20instead 
2. A personal preference not to have a linter change my code which I'm sure other people share.
3. As highlighted in the link above, eslint auto format can be quite slow.

I would even argue that this should be the default, but I've left it as is in order not to break anyone's config or ruffle anyone's feathers.

## Checklist

### DIFF
--- a/lua/lazyvim/plugins/extras/linting/eslint.lua
+++ b/lua/lazyvim/plugins/extras/linting/eslint.lua
@@ -1,3 +1,10 @@
+if lazyvim_docs then
+  -- Set to false to disable auto format
+  vim.g.lazyvim_eslint_auto_format = true
+end
+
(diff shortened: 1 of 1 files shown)


adammfrank:
Please merge this. Eslint formatting causes constant timeouts for me.

adammfrank:
Is there a way I can copy this into a local config so I can turn off eslint formatting?

27medkamal:
My workaround for now is to not use the extra but rather set it up myself in my own config. Just copy the nvim-lspconfig part to your personal config and you should be good to go

PR #4231, opened 2024-07-30
fix(navic): use the same background color as lualine section_c

### LINKED ISSUE #328: bug: (minor UI issue) nvim-navic background colour different from lualine background

- [X] I have searched the exsiting issues of plugins related to this issue

### Neovim version (nvim -v)

0.9.0

### Operating system/version

Ubuntu 22.04

### Describe the bug

image

Super minor. You need to look closely. You can see that the nvim-navic colour doesn't align with the lualine colour.  
Not talking about the line below (it's my tmux statusline so ignore that please)

### Steps To Reproduce

N/A

### Expected Behavior

N/A

### Repro

```Lua
-- set stdpaths to use .repro
end

-- bootstrap lazy
end

-- install plugins
local plugins = {
  "folke/tokyonight.nvim",
  "folke/LazyVim",
  -- add any other plugins here
}
  root = root .. "/plugins",
})

-- add anything else here
```

natemccurdy:
## Description

Prior to this, when using the navic Extra, the symbol breadcrumbs shown
in lualine section_c had a background color different from the
background of lualine_c, causing it to standout and not blend in with
some colorschemes.

There were ways around this problem, but hose involved overriding a lot
of highlights. See https://github.com/LazyVim/LazyVim/issues/328

This change uses the new (as of April 2023) lualine component built in
to SmiteshP/nvim-navic, which has a `color_correction` option to fix the
problem described above. This built-in component replaces the manual
method of calling `require("nvim-navic").is_available()` and
`require("nvim-navic").get_location()`.

* https://github.com/SmiteshP/nvim-navic/commit/711e9f117a0936c9d4420a135d658d2ca53a9e99
* https://github.com/SmiteshP/nvim-navic/commit/bf587250f8e399da75b4efe12261e18d8783ecef

With `color_correction` set to "static" or "dynamic", navic uses the
lualine section's background color when building the breadcrumbs text.

* https://github.com/SmiteshP/nvim-navic/blob/bf587250f8e399da75b4efe12261e18d8783ecef/doc/navic.txt#L83-L88

I chose to set color_correction to "dynamic" to provide the best user
experience as that will update the background any time the lualine
section's background changes (e.g. when switching modes).

## Related Issue(s)

* Fixes #328

## Screenshots

Before this fix, the breadcrumb's background did not match lualine section_c's:
before

After this fix, the breadcrumb's background does match:
after

## Checklist

### DIFF
--- a/lua/lazyvim/plugins/extras/editor/navic.lua
+++ b/lua/lazyvim/plugins/extras/editor/navic.lua
@@ -30,14 +30,7 @@ return {
     optional = true,
     opts = function(_, opts)
       if not vim.g.trouble_lualine then
-        table.insert(opts.sections.lualine_c, {
-          function()
-            return require("nvim-navic").get_location()
-          end,
-          cond = function()
-            return package.loaded["nvim-navic"] and require("nvim-navic").is_available()
-          end,
-        })
+        table.insert(opts.sections.lualine_c, { "navic", color_correction = "dynamic" })
       end
     end,
   },


folke:
ty

PR #4247, opened 2024-08-02
fix(lazygit): not everyone uses lazygit

Alan-John-Byrne:
## Description

LazyVim as it stood automatically assumed that people had lazygit installed, and so set keymaps which were represented in the which-key plugin by default. When using these keymaps, they would throw errors. Just because that user didn't have the lazygit on their system.

## Related Issue(s)

Simply removing the keymaps would cause issues with the telescope plugin.

## Screenshots

image

## Checklist

- [YES ] I've read the [CONTRIBUTING](https://github.com/LazyVim/LazyVim/blob/main/CONTRIBUTING.md) guidelines.


Alan-John-Byrne:
This should make it easier for people who like to use neogit or any other git plugin without having to worry about keymaps they might not want to use.

dpetka2001:
Couldn't you just use `vim.fn.executable` in `keymaps.lua` instead of the helper util function?

Alan-John-Byrne:
> Couldn't you just use `vim.fn.executable` in `keymaps.lua` instead of the helper util function?

I'm new to neovim and lazyvim, but making good progress. I could absolutely try to do that. Initial fix was just intended to remove an inconvience of having key maps that wouldn't work in all cases.

Alan-John-Byrne:
Sorry, I made one small oversight in forgetting to also disable the two telescope keymaps for git commits and status. Plan to fix using the exact same method as before.

dpetka2001:
Why? Those are default Telescope mappings. Nothing to do with Lazygit.

Alan-John-Byrne:
They still clash with wanting to set similar keymappings when using other git plugins.

dpetka2001:
`<leader>gs` for `Telescope git_status` has nothing to do with Lazygit. That keymap just executes `git status` in a Telescope picker that also shows the diff in a Telescope previewer. That should be changed ideally in the user configuration if he wants to use the same mappings for another plugin.

dpetka2001:
Just to be clear I'm just a simple user like you and only the maintainer has the final say. So, whatever I say is just me expressing my personal opinion and should be taken as a suggestion from your part, that doesn't have to be necessarily followed. 

So, if you want to make that change in your PR, go ahead and do so and the maintainer will decide in the end after he comes back from vacation. I can totally understand your reasoning about Lazygit, but I don't really understand you wanting to conditionally enable the mappings for `git_status` and `commits`, since those are default Telescope mappings. 

In my opinion, every distro comes with some defaults. It's up to the user to change those defaults to something else. Unfortunately, not everyone's preferences can be taken into account from a distro's perspective and some customization will have to be made in the user configuration. That is my reasoning concerning the last 2 mappings that you want to change, but bear in mind that it's just a personal opinion after all.

Alan-John-Byrne:
I agree with your opinion, and stand corrected. 😌

jeffschwartz:
I agree with the OP that lazygit should not be included in the lazyvim distro by default. At best it should be optional or perhaps even provide an option to install it or fugitive instead. I see no reason why WhichKey has to be cluttered with lazygit bindings that I will never use and would much prefer that WhichKey reflect my key binding for fugitive. The whole idea for LazyVim as far as I understand it is to fundamentally provide Neovim users with a tool to allow them to tailor their configuration according to their personal needs. I feel that forcing lazygit upon all users of lazyvim breaks that understanding and diminished the utility of lazyvim greatly. Please understand that I've been a fan of this bistro since its inception and I only feel compelled to comment here because I am seeing this awesome bistro slowly experience feature-creep which if allowed to continue will surely be detrimental to all of its adopters.

jeffschwartz:
I used Lazy to upgrade to the latest version of LazVim to capture this change but I still have the keybindings for lazygit. Lazygit is not installed on my computer and I expected that since it wasn't that should have gotten rid of their keybindings. Am I forgetting to do something else or is this still an issue? And thanks again for this marvelous distro.

dpetka2001:
The changes are only in latest HEAD. There hasn't been made a new release yet. Probably tomorrow (see Folke's [comment here](https://github.com/LazyVim/LazyVim/pull/4706#issuecomment-2462442335)).

jeffschwartz:
@folke that is awesome and thanks so much.

PR #4253, opened 2024-08-04
chore(mason): Refactor ensure_installed to remove unnecessary network calls

MasterLambaster:
## Description

The main reason for this change is to avoid network calls/auto registry update on the startup. Refresh would perform registry update once per day which is not necessary.
Refresh is not needed because under the hood mason would [perform the refresh](https://github.com/williamboman/mason.nvim/blob/e2f7f9044ec30067bc11800a9e266664b88cda22/lua/mason-registry/sources/github.lua#L101
) 

Another change is using [MasonRegistry.is_installed](https://github.com/williamboman/mason.nvim/blob/e2f7f9044ec30067bc11800a9e266664b88cda22/lua/mason-registry/init.lua#L15) method that does not create `Package` instance without a need.

## Checklist

### DIFF
--- a/lua/lazyvim/plugins/lsp/init.lua
+++ b/lua/lazyvim/plugins/lsp/init.lua
@@ -286,14 +286,12 @@ return {
         end, 100)
       end)
 
-      mr.refresh(function()
-        for _, tool in ipairs(opts.ensure_installed) do
+      for _, tool in ipairs(opts.ensure_installed) do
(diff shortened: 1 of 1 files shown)


folke:
This would break if no registries are present.
It's also the recommended way https://github.com/williamboman/mason.nvim?tab=readme-ov-file#registries

PR #4261, opened 2024-08-05
fix(java): give java a default nvim-dap config

Phys09:
## Description

Currently, there are no Java configurations for nvim-dap upon enabling the Java extras.

This requires debugging configurations to be setup manually afterwards.

(This is unlike many other language extras which do have nvim-dap support "out of the box")

This pr creates an initial configuration for nvim-dap which lets it attach to a debugging process on port 5005.
(such as a spring boot application) More can be added if there are any experienced Java devs here.

## Related Issue(s)

None

## Screenshots

Error message:
image

Java remote debugging open:
image

## Checklist


Phys09:
Also, for those more experienced with configuring nvim-dap, what happens if there are 2 different configurations for 2 different ports?

Phys09:
Or is this feature something too specific (to spring boot) to be added as part of the distribution?

folke:
Makes sense. Thanks!

idelice:
> Or is this feature something too specific (to spring boot) to be added as part of the distribution?

Maybe it should accept a port option and default to whatever you gave it. Just my opinion but great work

Issue #3569, opened 2024-06-10
bug: Attempt to index field 'foldupdate_range' (a nil value)


kapral18:
### Did you check docs and existing issues?

- [X] I have read all the LazyVim docs
- [X] I have searched the existing issues of LazyVim
- [X] I have searched the existing issues of plugins related to this issue

### Neovim version (nvim -v)

0.10.0

### Operating system/version

MacOS 14.5

### Describe the bug

When lsp adding missing imports in TS files I occasionally get these

![image](https://github.com/LazyVim/LazyVim/assets/1625373/968e1817-05b3-4d70-a382-51c3d547caa2)

Trying to understand if this is a lazyvim error or neovim error

### Steps To Reproduce

Open TS file try removing and re-adding imports via lsp organize imports

### Expected Behavior

No error

folke:
Looks like a Neovim bug

tomtomjhj:
the error message says that the error happened when processing bufwritepre, which means that nvim was saving the file (and trigger auto formating). were you actually saving the file when this happened? can you reproduce this by doing some edits in insert mode and concurrently trigger auto formatting without leaving insert mode?

kapral18:
@tomtomjhj 

yes I do use mostly ctrl-c instead of esc as much as possible. Do you think it can affect this?

This happens randomly. For example this happens when I try to exit fzf-lua window when I am checking for lsp references and kill it with ctrl-c. But I noticed it mostly happens with list of new references that I haven't visited yet.

Or this happens when I try to complete the auto imports from normal mode with tsserver or typescript-tools add imports functionality with lsp code actions. And again happens with new imports mostly.

@tomtomjhj can you tell me how to trigger autoformatting without leaving insert mode?

tomtomjhj:
> yes I do use mostly ctrl-c instead of esc as much as possible. Do you think it can affect this?

yes this seems to be the root cause. leaving insert mode eith ctrl-c doesn't trigger insertleave autocmd, and this confuses treesitter fold stuff. i missed this point when i made changes to it.

the fix is simple: just add a nil check. i was initially hesitant to do this because it didnt seem necessary, but it turns out it's indeed necessary. i'll make pr soon.

s1n7ax:
Can confirm this is happening in 0.11 as well. This occurs in Typescript, Lua, Java files.

[...]

PR #4268, opened 2024-08-06
feat(ai): add `codecompanion.nvim` as alternative to `copilotchat.nvim`

georgeguimaraes:
Add https://github.com/olimorris/codecompanion.nvim as an alternative to CopilotChat.
CodeCompanion makes it easier to add your own custom prompts and agentic workflow.

I've remapped some keys to better integrate with LazyVim (especially closing with `q`).

### DIFF
--- a/lua/lazyvim/plugins/extras/coding/codecompanion.lua
+++ b/lua/lazyvim/plugins/extras/coding/codecompanion.lua
@@ -1,0 +1,47 @@
+return {
+  {
+    "olimorris/codecompanion.nvim",
+    cmd = { "CodeCompanion", "CodeCompanionActions", "CodeCompanionToggle", "CodeCompanionAdd", "CodeCompanionChat" },
+
+    opts = function()
+      local options = require("codecompanion.config")
+      local user = vim.env.USER or "User"
+
+      options.strategies.chat.roles = {
+        llm = "  CodeCompanion",
+        user = "  " .. user:sub(1, 1):upper() .. user:sub(2),
+      }
+
+      options.strategies.chat.keymaps.close.modes = {
+        n = "q",
+        i = "<C-c>",
+      }
+      options.strategies.chat.keymaps.stop.modes.n = "<C-c>"
+      options.strategies.chat.keymaps.save.modes.n = "gS"
+
+      return options
+    end,
+
+    keys = {
+      { "<leader>a", "", desc = "+ai", mode = { "n", "v" } },
+      { "<leader>aA", "<cmd>CodeCompanionActions<cr>", mode = { "n", "v" }, desc = "Prompt Actions (CodeCompanion)" },
+      { "<leader>aa", "<cmd>CodeCompanionToggle<cr>", mode = { "n", "v" }, desc = "Toggle (CodeCompanion)" },
+      { "<leader>ac", "<cmd>CodeCompanionAdd<cr>", mode = "v", desc = "Add code to CodeCompanion" },
+      { "<leader>ap", "<cmd>CodeCompanion<cr>", mode = "n", desc = "Inline prompt (CodeCompanion)" },
+    },
+  },
+
+  -- Edgy integration
+  {
+    "folke/edgy.nvim",
+    optional = true,
+    opts = function(_, opts)
+      opts.right = opts.right or {}
+      table.insert(opts.right, {
+        ft = "codecompanion",
+        title = "CodeCompanion Chat",
+        size = { width = 50 },
+      })
+    end,
+  },
+}


dpetka2001:
Just a suggestion and not 100% sure if it still applies in this case. But using `<c-c>` in Neovim can be a cause of quite some problems. You can search through this repo and Neovim as well for related issues. Not sure what the latest state of things currently is (since it's been a while i've searched about it) , but you might want to remap it to `<esc>` as a default for Lazyvim and let users who still prefer `<c-c>` do so in their personal configuration. Just some food for thought.

rwjblue:
>  using <c-c> in Neovim can be a cause of quite some problems

🙏 maybe it's fixed ([see here](https://github.com/neovim/neovim/issues/20726#event-13936634706)) (backported to 0.10 series [here](https://github.com/neovim/neovim/pull/30094), but AFAICT not released yet)

lua/lazyvim/plugins/extras/coding/codecompanion.lua
```diff
@@ -0,0 +1,47 @@
...
+        llm = "  CodeCompanion",
+        user = "  " .. user:sub(1, 1):upper() .. user:sub(2),
+      }
+
+      options.strategies.chat.keymaps.close.modes = {
+        n = "q",
+        i = "<C-c>",
+      }
+      options.strategies.chat.keymaps.stop.modes.n = "<C-c>"
+      options.strategies.chat.keymaps.save.modes.n = "gS"
```
rwjblue, line 20: `save` is no longer a valid entry in keymaps; I believe it was renamed to `send`

georgeguimaraes: tks for the heads up!!
I also ended changing save.modes.n to <CR>, I think it may be better for LazyVim users, but it's totally an opinion.

georgeguimaraes:
> Just a suggestion and not 100% sure if it still applies in this case. But using `<c-c>` in Neovim can be a cause of quite some problems. You can search through this repo and Neovim as well for related issues. Not sure what the latest state of things currently is (since it's been a while i've searched about it) , but you might want to remap it to `<esc>` as a default for Lazyvim and let users who still prefer `<c-c>` do so in their personal configuration. Just some food for thought.
> 
> PS: Some examples #3569 and #4142 that I remember mostly having read through them (there might be others as well).

changed to <esc>, makes total sense! tks @dpetka2001

lua/lazyvim/plugins/extras/coding/codecompanion.lua
```diff
@@ -0,0 +1,47 @@
...
+      return options
+    end,
+
+    keys = {
+      { "<leader>a", "", desc = "+ai", mode = { "n", "v" } },
+      { "<leader>aA", "<cmd>CodeCompanionActions<cr>", mode = { "n", "v" }, desc = "Prompt Actions (CodeCompanion)" },
+      { "<leader>aa", "<cmd>CodeCompanionToggle<cr>", mode = { "n", "v" }, desc = "Toggle (CodeCompanion)" },
+      { "<leader>ac", "<cmd>CodeCompanionAdd<cr>", mode = "v", desc = "Add code to CodeCompanion" },
+      { "<leader>ap", "<cmd>CodeCompanion<cr>", mode = "n", desc = "Inline prompt (CodeCompanion)" },
+    },
```
rwjblue, line 31: Is it possible to more closely mirror the keys for copilot-chat extra (to make it easier for folks to swap between them)? 

https://github.com/LazyVim/LazyVim/blob/3dbace941ee935c89c73fd774267043d12f57fe2/lua/lazyvim/plugins/extras/coding/copilot-chat.lua#L40-L74

For example:

- `<leader>ap` would be "Prompt Actions" (instead of current `<leader>aA`)

georgeguimaraes: Tks for the suggestions, makes a lot of sense.

lua/lazyvim/plugins/extras/coding/codecompanion.lua
```diff
@@ -0,0 +1,47 @@
...
+    "olimorris/codecompanion.nvim",
+    cmd = { "CodeCompanion", "CodeCompanionActions", "CodeCompanionToggle", "CodeCompanionAdd", "CodeCompanionChat" },
+
+    opts = function()
+      local options = require("codecompanion.config")
+      local user = vim.env.USER or "User"
+
+      options.strategies.chat.roles = {
+        llm = "  CodeCompanion",
+        user = "  " .. user:sub(1, 1):upper() .. user:sub(2),
```
mattarau, line 12: I don't think it should be forced into title case. Each person has chosen their user's name and in my case it looks weird to read it like that.
```suggestion
        user = "  " .. user,
```

mattarau (changes requested):

lua/lazyvim/plugins/extras/coding/codecompanion.lua
```diff
@@ -0,0 +1,47 @@
...
+      }
+
+      options.strategies.chat.keymaps.close.modes = {
+        n = "q",
+        i = "<C-c>",
```
dpetka2001, line 17:
```suggestion
        i = "<Esc>",
```
Just to make sure there's no `<C-c>` left. There was another [issue at kickstart](https://github.com/nvim-lua/kickstart.nvim/issues/1121) with `<C-c>` and deferring clipboard.

Relevant [comment](https://github.com/neovim/neovim/pull/29581#issuecomment-2318966355) from Neovim developer. I really hope I'm not perceived as persistent with nitpicking. This is just to avoid potential issues that might or might not occur due to `<C-c>`.

georgeguimaraes: Sure, the issue is still happening. don't worry about being persistent on it :)

georgeguimaraes:
Folks, tks for all the interest in this PR. I was using codecompanion multiple times daily.

But, to be honest, I switched last week to using https://github.com/yetone/avante.nvim, which makes applying patches more fluid with my workflow. I'm still experimenting with its config, so I haven't opened a PR yet.

This means that I'll close this PR since I won't be able to offer my support on integrating it into LazyVim and keeping the configuration up to date.

Of course, feel free to grab the code here.

ronald2wing:
Can you share your avante setup here as well? The thing about LazyVim is that we do not have to set or switch the plugins ourselves. Also, people should be able to use options.lua to change the type of plugins (in this case, it is codecompanion and avante).

Thank you for your work.

Issue #4292, opened 2024-08-11
bug: vue style commentstring is wrong


tkkcc:
### Did you check docs and existing issues?

- [X] I have read all the LazyVim docs
- [X] I have updated the plugin to the latest version before submitting this issue
- [X] I have searched the existing issues of LazyVim
- [X] I have searched the existing issues of plugins related to this issue

### Neovim version (nvim -v)

0.10.1

### Operating system/version

archlinux

### Describe the bug

comment in vue style use wrong commentstring
```vue
<style scoped>
<!-- .logo { -->
<!--   height: 6em; -->
<!--   padding: 1.5em; -->
<!--   will-change: filter; -->
<!--   transition: filter 300ms; -->
<!-- } -->
</style>

```

### Steps To Reproduce
1. `npm create vite@latest test -- --template vue-ts`
2. use fresh LazyVim, enable lang.vue, open `test/src/App.vue`, comment in style 

### Expected Behavior
inside
```vue
<style scoped>
/* .logo { */
/*   height: 6em; */
/*   padding: 1.5em; */
/*   will-change: filter; */
/*   transition: filter 300ms; */
/* } */
</style>

```
outside:
```vue
<!-- <style scoped> -->
<!-- .logo { -->
<!--   height: 6em; -->
<!--   padding: 1.5em; -->
<!--   will-change: filter; -->
<!--   transition: filter 300ms; -->
<!-- } -->
<!-- </style> -->
```

### attempt
i tried this, but outside not work
```lua
    {
      "folke/ts-comments.nvim",
      opts = {
        lang = {
          vue = {
            "<!-- %s -->",
            style_element = "/* %s */",
          },
        },
      },
    },

```

dpetka2001:
Try 
```lua
        vue = {
          "<!-- %s -->",
          style_element = "<!-- %s -->",
          raw_text = "/* %s */",
        },
```

tkkcc:
@dpetka2001 no change

tkkcc:
nice! didn't notice it's raw_text inside.

dpetka2001:
You should just press `<leader>uI` (capital `i`) for `InspectTree` and then go throughout the different nodes and see how Treesitter recognizes them. For me the above seems to be working

[Screencast 2024-08-11 18:23:16.webm](https://github.com/user-attachments/assets/928c7e94-02e5-444a-8458-5dbf89dce912)

iamxiaojianzheng:
[...]

PR #4308, opened 2024-08-14
fix(vue): ensure css treesitter installed

tkkcc:
## Description

If css treesitter is not installed(the current), <style> in vue file will use wrong commentstring, ts-comments.nvim or nvim-ts-context-commentstring both. so ensure it's installed.

## Related Issue(s)

explained: [in https://github.com/JoosepAlviste/nvim-ts-context-commentstring/issues/117](https://github.com/JoosepAlviste/nvim-ts-context-commentstring/issues/117#issuecomment-2287909324)
close https://github.com/LazyVim/LazyVim/issues/4292

### DIFF
--- a/lua/lazyvim/plugins/extras/lang/astro.lua
+++ b/lua/lazyvim/plugins/extras/lang/astro.lua
@@ -17,7 +17,7 @@ return {
 
   {
     "nvim-treesitter/nvim-treesitter",
-    opts = { ensure_installed = { "astro" } },
+    opts = { ensure_installed = { "astro", "css" } },
   },
 
   -- LSP Servers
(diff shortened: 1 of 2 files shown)


dpetka2001:
Hi, could you also add the `css` treesitter parser for `lang.astro`, so that we don't open multiple PRs? Otherwise the `<style>` tag doesn't get correct highlighting. Maybe the `commentstring` also applies there as well (haven't tested thoroughly, since I don't do web dev, sorry). Would be great if you could test it, since you seem to have more experience.

tkkcc:
tested and yes it has same issue if lang.astro enabled.

RayGuo-ergou:
If we want to do this for the users, I feel we can also add `scss`, otherwise people may confuse why `css` has highlight but `scss` does not.

2giosangmitom:
Consider adding `scss` parser as it encounters similar issue.

folke:
ty!

PR #4314, opened 2024-08-16
fix(project.nvim): correctly delete projects

dpetka2001:
## Description
Unfortunately, the upstream Github repo hasn't been active for over a year. This overwrites the `history.delete_project` function to correctly be able to delete projects. Because of `ipairs` being used, when you delete the first key from the table the `for` loop iteration will just stop, since there will be a gap between the available indices. 
> will iterate over the pairs (`1,t[1]`), (`2,t[2]`), ..., up to the first integer key absent from the table.

There's also a [PR](https://github.com/ahmedkhalf/project.nvim/pull/106/files) for this upstream, but like I mentioned the Github repo hasn't been active for over a year.

Feel free to disregard, since this is obviously something that should be rather fixed upstream under normal circumstances. I just created this PR, since it's such a basic functionality with little change in LazyVim's codebase, that I thought it might be acceptable.

## Related Issue(s)
No, rather a discussion #4310

## Checklist

### DIFF
--- a/lua/lazyvim/plugins/extras/util/project.lua
+++ b/lua/lazyvim/plugins/extras/util/project.lua
@@ -91,6 +91,15 @@ return {
     event = "VeryLazy",
     config = function(_, opts)
       require("project_nvim").setup(opts)
+      local history = require("project_nvim.utils.history")
+      history.delete_project = function(project)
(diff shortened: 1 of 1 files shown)


folke:
ty

PR #4315, opened 2024-08-16
feat(type): better hinting diagnostics option

aarnphm:
## Description

`nvim_open_win()` in fact support the following:

```lua
float = {"char", ...} | {{"char", "hlgroup"}, ...}
```

I have only tested this on 0.10+ and nightly build, so I'm not entirely sure about backward compat.

## Checklist


folke:
Fix it upstream instead?

PR #4332, opened 2024-08-20
feat(root): provide `vim.g.root_lsp_ignore` to ignore LSP servers

### LINKED ISSUE #4330: feature: Add `ignore_lsp` option to `lazyvim.util.root`
### Did you check the docs?

- [X] This is not a request for a new or existing extra (see above)

Hi, first of all thanks for the great project. I have a small issue with root dir detection: When no other LSPs are enabled, the lsp detector will load the root dir from copilot LSP which is always set to the home folder.

Provide a `ignore_lsp` configuration option similarly to https://github.com/ahmedkhalf/project.nvim?tab=readme-ov-file#%EF%B8%8F-configuration

Since project.nvim already provides the same functionality maybe you could leverage that instead of using a custom implementation? Just an idea.

### Additional context

_No response_

dpetka2001:
## Description
This provides `vim.g.root_lsp_ignore` for users to customize which LSP servers should be taken into account when evaluating the LSP `root_dir`.

## Related Issue(s)
Closes #4330

## Checklist


dpetka2001:
Should we maybe add corresponding comments in `options.lua`, so that it also appears on LazyVim web docs and make it easier for users to become aware of this option?

amaanq:
+1 for this, clangd messes up the root dir if the file it picks up on is not at the "real" root dir :expressionless:

folke:
@dpetka2001 yeah, makes sense to add this to options.lua and also add copilot by default, since this is the main culprit?

dpetka2001:
Added the additions requested in `options.lua`. I don't use Copilot, so I'm guessing the LSP server is called `copilot`? Please confirm about the correct wording of the Copilot LSP server and make changes if needed.

Gioyik:
this is the config for copilot:

```
{
  "nvim-cmp",
  dependencies = {
    {
      "zbirenbaum/copilot-cmp",
      dependencies = "copilot.lua",
      opts = {},
      config = function(_, opts)
        local copilot_cmp = require("copilot_cmp")
        copilot_cmp.setup(opts)
        -- attach cmp source whenever copilot attaches
        -- fixes lazy-loading issues with the copilot cmp source
        LazyVim.lsp.on_attach(function(client)
          copilot_cmp._on_insert_enter({})
        end, "copilot")
      end,
    },
  },
  ---@param opts cmp.ConfigSchema
  opts = function(_, opts)
    table.insert(opts.sources, 1, {
      name = "copilot",
      group_index = 1,
      priority = 100,
    })
  end,
}
```

it looks in fact the LSP server is called "copilot"

dpetka2001:
Thank you for that @Gioyik. I didn't think to look into the Copilot Extra to see its configuration.

folke:
ty!

PR #4342, opened 2024-08-22
feat(extras): custum the path of packages

### LINKED ISSUE #4339: bug: `astro-language-server` must install by `mason.nvim` if I want Lazyvim don't warn for me

- [X] I have updated the plugin to the latest version before submitting this issue

### Neovim version (nvim -v)

NVIM v0.10.1 Build type: Release LuaJIT 2.1.1713773202 Run "nvim -V1 -v" for more info

### Operating system/version

NixOS 24.11.20240809.5e0ca22 (Vicuna) x86_64

### Describe the bug

I'm using nixos, so I disabled the `mason.nvim` plugin, and use system package to install LSP, DAP, Linter and Formatter, it works normal.

When I enabled `lang.astro`, and install `pkgs.astro-language-server` package, LazyVim warn me as soon as I start neovim:
warning
So, I put it in my nvim config.
```lua
    {
      "neovim/nvim-lspconfig",
      opts = function(_, opts)
        for i, v in
          ipairs(opts.servers.vtsls.settings.vtsls.tsserver.globalPlugins)
        do
          if v.name == "@astrojs/ts-plugin" then
            opts.servers.vtsls.settings.vtsls.tsserver.globalPlugins[i].location =
              vim.env.HOME .. "node_modules/@astro/ts-plugin"
          end
        end
      end,
    },
```
And execute:
```bash
npm i @astro/ts-plugin
```

After that, astro-language-server works, but LazyVim also warn me as soon as I start neovim.

### Steps To Reproduce

1. Disable `mason.nvim` plugin
2. Enable `lang.astro` lazyvim extra
3. Install `astro-language-server` package provide by system package manager
4. Start neovim

### Expected Behavior

LazyVim will not warn for it to me, because I already installed `astro-language-server` package by system package manager.
图片

### Repro

```lua

require("lazy.minit").repro({
  spec = {
    { "williamboman/mason.nvim", enabled = false },
    { import = "lazyvim.plugins.extras.lang.astro" },
    {
      "neovim/nvim-lspconfig",
      opts = function(_, opts)
        for i, v in
          ipairs(opts.servers.vtsls.settings.vtsls.tsserver.globalPlugins)
        do
          if v.name == "@astrojs/ts-plugin" then
            opts.servers.vtsls.settings.vtsls.tsserver.globalPlugins[i].location =
              --[[ path of the @astro/ts-plugin ]]
          end
        end
      end,
    },
  },
})
```

Buer-Nahida:
## Description

if I enabled the `lang.astro` lazyvim extra and disabled `mason.nvim` then vtsls will not found `@astro/ts-plugin` plugin, and LazyVim will raise a warning.

image

We can do to make vtsls found `@astro/ts-plugin` this plugin, but LazyVim will also warn of this.

```lua
return {
  "neovim/nvim-lspconfig",
  opts = function(_, opts)
    for i, v in
      ipairs(opts.servers.vtsls.settings.vtsls.tsserver.globalPlugins)
    do
      if v.name == "@astrojs/ts-plugin" then
        opts.servers.vtsls.settings.vtsls.tsserver.globalPlugins[i].location =
          vim.env.HOME .. "node_modules/@astro/ts-plugin"
      end
    end
  end,
}
```

This PR added some option for LazyVim, users can set it to set path to package, then will work normal and LazyVim will not raise this warning.

So many extras has this bug, so I changed six files of lazyvim extra to fix it.

## Screenshots

***No screenshot***

## Checklist


dpetka2001:
In my personal opinion, this is unnecessary noise. You mentioned in your issue that your problem was that the Warning message kept appearing even after you made it work in your personal configuration. Couldn't you just introduce into `opts.warn` a global variable that takes precedence in the user configuration in order to just disable the message like I mentioned in the issue as well? It's way less intrusive. 

Providing a configuration for additional custom location is out of the scope of your issue. If users don't want to use Mason, then they make adjustments in their personal configuration. You just want a way to disable the Warning.

Buer-Nahida:
Ummmmmm, LazyVim warn me `Mason package path not found`, and this package is a tsserver globalplugin of vtsls.

So tssever of vtsls it isn't found necessary `@astro/ts-plugin` globalplugin, I think it's an error, just this error difficult to detect, this error will make users' completion list of ts file missing someting.

dpetka2001:
Didn't you say in the issue you created that you got it working and only the warning kept appearing? So, just disable that warning. Did you read what i suggested in the issue?

Buer-Nahida:
I think I should't link that issue to this PR. um, I will close that issue and rewrite information about this PR.

dpetka2001:
https://github.com/LazyVim/LazyVim/blob/12818a6cb499456f4903c5d8e68af43753ebc869/lua/lazyvim/plugins/extras/lang/astro.lua#L34-L45

I believe instead of all the changes you made, you can just replace line 40 with 
```lua
location = vim.g.get_pkg and vim.g.get_pkg.astro or LazyVim.get_pkg_path("astro-language-server", "/node_modules/@astrojs/ts-plugin"),
```
and then the user can just do in his `options.lua` 
```lua
vim.g.get_pkg = { astro = "custom_path_node_modules" }
```
I believe that 1 line change of code is preferable. Could you test it out to see if it works?

Buer-Nahida:
Yes, it works.

Buer-Nahida:
OK, I will add it.

Buer-Nahida:
um, I think I should use `Lazyvim.config.pkg_path` to instead of `vim.g.pkg_path`, because it's some options of LazyVim, not of vim.

dpetka2001:
It's the same option in Vim and Neovim as well. You can see [here](https://github.com/LazyVim/LazyVim/blob/12818a6cb499456f4903c5d8e68af43753ebc869/lua/lazyvim/config/options.lua#L2-L51) other options being defined like this as well. It's being used for folds, root_spec etc. It's more concise like this in my opinion. `vim.g` is just to define a global variable/dictionary and is used the same in Vim and Neovim.

Buer-Nahida:
Yes, you have a point there.

dpetka2001:
An even better solution would be to change the `LazyVim.get_pkg_path` function directly from this https://github.com/LazyVim/LazyVim/blob/12818a6cb499456f4903c5d8e68af43753ebc869/lua/lazyvim/util/init.lua#L250-L263
to the following
```lua
function M.get_pkg_path(pkg, path, opts)
  if vim.g.pkg_path and vim.g.pkg_path[pkg] then
    local ret = vim.g.pkg_path[pkg]
    return ret
  else
    pcall(require, "mason") -- make sure Mason is loaded. Will fail when generating docs
    local root = vim.env.MASON or (vim.fn.stdpath("data") .. "/mason")
    opts = opts or {}
    opts.warn = opts.warn == nil and true or opts.warn
    path = path or ""
    local ret = root .. "/packages/" .. pkg .. "/" .. path
    if opts.warn and not vim.loop.fs_stat(ret) and not require("lazy.core.config").headless() then
      M.warn(
        ("Mason package path not found for **%s**:\n- `%s`\nYou may need to force update the package."):format(
          pkg,
          path
        )
      )
    end
    return ret
  end
end
```
That way you don't have to put in every file the `vim.g.pkg_path` and you could just leave the 
```lua
if lazyvim_docs then
  -- If you don't use Mason to install the `angular-language-server` you can use the following to custom path for `@angular/language-server`.
  vim.g.pkg_path = {
    ["@angular/language-server"] = "path_to_this_node_package",
  }
end
```
at the top of the files so that the options appear in the LazyVim docs as well. Just make sure to use in the `vim.g.pkg_path` the names as they appear in Mason packages, so that they would be compatible. 

For example for Astro, it would have to be `vim.g.pkg_path["astro-language-server"] = "path_to_node_modules/@astro/ts-plugin"` instead of `vim.g.pkg_path["@astro/ts-plugin"]` like you currently have. Same for `angular-language-server` and the rest.

Buer-Nahida:
Maybe manually specify package name it's better than this, if we do, some users will write some wrong configurtion beacuse this name, for example:
```lua
vim.g.pkg_path = {
  ["astro-language-server"] = vim.env.HOME .. "/node_modules/astro-lanugage-server",
}
```
And, if an extra need use two files and they are all in the same package, only path is different, what should we do?

dpetka2001:
Ok i see. I don't do any web dev, so i'll leave that decision up to you, since you seem to have more experience.

folke:
This increases unneeded complexity. Use mason, or just do `LazyVim.get_pkg_path = function(...)` in your own config

PR #4375, opened 2024-08-30
fix(snippet): don't set navigation mappings in nightly

MariaSolOs:
## Description

Since [this commit in Neovim](https://github.com/neovim/neovim/pull/27339), snippet navigation mappings are set by default to `<tab>` and `<s-tab>`, and hence don't need to be redefined here. Moreover, the core implementation makes sure that those snippet navigation mappings are active only when snippet expansion is happening, and if other user bindings are using those keys, they're restored when snippet navigation ends.

## Checklist


folke:
Awesome, thanks!

PR #4379, opened 2024-08-31
Fix filename/directory highlighting in lualine pretty_path truncation

daviesjamie:
## Description

This is a small fix to the path truncating in `LazyVim.lualine.pretty_path()` to make the filename highlighting actually apply to only the filename, rather than the filename and all path elements after the inserted ellipsis (`…`).

This keeps it consistent with the behaviour exhibited when the path isn't truncated, and is what I think most users would expect the behaviour to be.

## Screenshots

Before this fix:

image

```
   nvim/…/nvim/lua/plugins/lazyvim/ui.lua
   └──┬──┘└──────────────┬──────────────┘
directory_hl        filename_hl
```

After this fix:

image

```
nvim/…/nvim/lua/plugins/lazyvim/ui.lua
└──────────────┬───────────────┘└─┬──┘
         directory_hl        filename_hl
```

## Checklist


dpetka2001:
Thanks for catching that. Should not have concatenated the rest of the parts into a single string. Didn't think about how highlighting would show.

folke:
ty

PR #4406, opened 2024-09-06
feat(lang): add ts_ls to lsp config

arthur-mountain:
## Description

In this [PR](https://github.com/neovim/nvim-lspconfig/pull/3232), tsserver is renamed to ts_ls, but it hasn't been released yet. 

Therefore, ts_ls is added to the LSP configuration in case the user updates their plugins locally.

## Related Issue(s)

N/A!

## Screenshots

N/A!

## Checklist

### DIFF
--- a/lua/lazyvim/plugins/extras/lang/typescript.lua
+++ b/lua/lazyvim/plugins/extras/lang/typescript.lua
@@ -19,7 +19,12 @@ return {
     opts = {
       -- make sure mason installs the server
       servers = {
+        --- @deprecated -- tsserver renamed to ts_ls but not yet released, so keep this for now
+        --- the proper approach is to check the nvim-lspconfig release version when it's released to determine the server name dynamically
         tsserver = {
+          enabled = false,
+        },
+        ts_ls = {
           enabled = false,
         },
         vtsls = {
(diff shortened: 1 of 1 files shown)


Conarius:
I would just rename tsserver to ts_ls instead of adding it on top as it has been renamed not added.

arthur-mountain:
> I would just rename tsserver to ts_ls instead of adding it on top as it has been renamed not added.

It's also okay to directly rename `tsserver` to `ts_ls`. 

The main concern here is that removing `tsserver` might affect users who haven't updated `nvim-lspconfig` yet.

WDYT~~~?

Conarius:
> It's also okay to directly rename `tsserver` to `ts_ls`. 
> 
> The main concern here is that if `tsserver` is removed, it might affect users who haven't updated `nvim-lspconfig`, and they may end up running both `tsserver` and `vtsls` if `tsserver` is not correctly disabled.
> 
> WDYT~~~?

It is a breaking change in nvim-lspconfig already so if anyone updates that plugin, they'll have to update their configs so it wouldn't be much of a concern ig

Conarius:
> There might be users that pin their plugins to stable versions only. That would cause conflict if LazyVim doesn't take both into consideration. Last stable release of `nvim-lspconfig` is from May 21. Whereas LazyVim has usually more frequent releases.

Ah, forgot that. Then it's alright as is

PR #4426, opened 2024-09-10
[lazyvim.plugins.extras.lang.typescript.lua] treesitter - ensure_installed typescript

cedi:
## Description

When enabling language extras, most other extras automatically install the treesitter modules for syntax highlighting for the language.
The typescript module was missing this `"nvim-treesitter/nvim-treesitter", opts = { ensure_installed = { "typescript", "javascript" } },` config.

## Checklist


dpetka2001:
Not needed, since they're already included in the default treesitter base spec as can be seen [here](https://github.com/LazyVim/LazyVim/blob/3dbace941ee935c89c73fd774267043d12f57fe2/lua/lazyvim/plugins/treesitter.lua#L46-L61).

PR #4427, opened 2024-09-10
feat(keybinds): add a "find files based on current word/WORD" command

awilkins:
## Description

Adds a "find file" that contains the word under the cursor, because I found myself wanting this a lot.

For both `fzf-lua` and `telescope`

Not sure if `cWORD` is also a useful pick.

## Checklist


dpetka2001:
Isn't this the same as opening `Find Files` and just typing the word (and the point of fuzzy finding is not typing the words exactly)? Nice that you like it, but I believe it's a bit too specific to a user's preferences in my personal opinion. Nevertheless, I'm just a simple user and maintainer has the final say.

awilkins:
> Isn't this the same as opening Find Files and just typing the word

Well, you don't have to type the word, so it's faster.

And my current project has a LOT of files with very similar difficult to type names (basically : source code files in a large enterprise project on an old operating system that still has 8.3 all-caps file naming). Being able to position your cursor and whack `<space> f w` is an order of magnitude easier than typing cryptic file names made of short mnemonics and digits that don't parse into natural language words, and also means that finding related files (which tend to have the same prefix and differ by one char at the end) is a backspace away.

I know I can add it to my local profile, but it seemed like a generally useful feature and there was this gap in the keymap, so ...

kezhenxu94:
> Isn't this the same as opening `Find Files` and just typing the word (and the point of fuzzy finding is not typing the words exactly)? Nice that you like it, but I believe it's a bit too specific to a user's preferences in my personal opinion. Nevertheless, I'm just a simple user and maintainer has the final say.

We have `<leader>sw` to find/grep word under cursor, I’m +1 to have `<leader>fw` to find file with name under cursor.

folke:
Not interested in this. Thank you anyway

PR #4428, opened 2024-09-11
fix(lualine): weird path truncation when not under cwd/root

patnr:
## Description

It does not make sense to truncate unless root *actually* in path.

Also, should check if `opts.relative == "root"`
otherwise the substitution can be surprising
(since user expects substitution only wrt. cwd)

## Related Issue(s)

N/A

## Checklist

### DIFF
--- a/lua/lazyvim/util/lualine.lua
+++ b/lua/lazyvim/util/lualine.lua
@@ -96,7 +96,7 @@ function M.pretty_path(opts)
 
     if opts.relative == "cwd" and path:find(cwd, 1, true) == 1 then
       path = path:sub(#cwd + 2)
-    else
+    elseif opts.relative == "root" and path:find(root, 1, true) == 1 then
       path = path:sub(#root + 2)
     end
 


dpetka2001:
I agree about the check for root existing in path, but not with the condition about `opts.relative = "root"`. Without it, it always shows the first part in the `root` directory component, whereas with `opts.relative = "root"` it doesn't show it. 

For example if `root` is `~/projects/plugins/LazyVim` and `cwd = ~/projects/plugins/LazyVim/lua/lazyvim/plugins` if I open a file (for example `util/lualine.lua`) then in lualine it will be printed `lua/.../util/lualine.lua` without the check for `opts.relative = "root"`. With the check `opts.relative = "root"` that you're suggesting in lualine it will be printed `/.../util/lualine.lua`.

This is just my own personal preference as a simple user but I prefer how it's printed without the check for `opts.relative = "root"` because it also shows the first part of the path relative to the `root` dir.

willnorris:
bump.  still an issue I ran into as well and ended up filing a duplicate PR to fix.

folke:
I don't see what this fixes.
If I open a file that's not in my cwd, then I want to see the file like it is currently implemented.

root will always be part of the path, so that's why there was no check for that

willnorris:
@folke did you see the description and screenshots in #4499, which was a dupe of this.  Does that help explain what this fixes?

folke:
Just checked and I can't reproduce that.
With the exact cmd op used, I get a properly formatted path...

folke:
Added an extra check just to be sure

PR #4436, opened 2024-09-14
feat(extras): lang julia

storopoli:
## Description

Adds [Julia](https://julialang.org/) as one the extras/lang with sensible and minimalist defaults.

## Related Issue(s)

None, :)

## Screenshots

None :)

## Checklist


mehalter (commented):
I would also recommend adding in something like `cmp-latex-symbols` for unicode completion since that is very widely used in Julia code and also importing the TOML lang extra when installing julia support since that's a core part of configuring Julia packages and environments.

These are just some recommendations from putting together the Julia extension for [AstroCommunity for AstroNvim](https://github.com/AstroNvim/astrocommunity/blob/main/lua/astrocommunity/pack/julia) :)

mehalter:
I did actually go through the VS Code julia extension source code a little bit more compared against just the `LanguageServer.jl` source code and there are a few other settings that I didn't realized differed as well. I updated the AstroCommunity spec if you are interested. It's a shame that they don't just make the defaults the ones you would actually want in the editor 😅

storopoli:
> I would also recommend adding in something like `cmp-latex-symbols` for unicode completion since that is very widely used in Julia code and also importing the TOML lang extra when installing julia support since that's a core part of configuring Julia packages and environments.
> 
> These are just some recommendations from putting together the Julia extension for [AstroCommunity for AstroNvim](https://github.com/AstroNvim/astrocommunity/blob/main/lua/astrocommunity/pack/julia) :)

Done them and squashed to a single commit, thanks!

mehalter (commented):
These look great settings wise. I would still recommend adding latex completion source to `cmp` as well as the importing of the TOML pack simply because that's everything needed.  By adding the latex completion you basically make the extra a true full replacement for [juia-vim](https://github.com/JuliaEditorSupport/julia-vim) which includes latex to unicode expansion as well as the what the Julia VS Code extension provides.

PR #4438, opened 2024-09-14
feat(extras): lang typst

storopoli:
## Description

Adds [`typst`](https://typst.app) support with preview.

## Checklist

### DIFF
--- a/lua/lazyvim/plugins/extras/lang/typst.lua
+++ b/lua/lazyvim/plugins/extras/lang/typst.lua
@@ -1,0 +1,35 @@
+return {
+  recommended = function()
+    return LazyVim.extras.wants({
+      ft = { "typst" },
+    })
+  end,
+  {
+    "nvim-treesitter/nvim-treesitter",
+    opts = { ensure_installed = { "typst" } },
+  },
+  {
+    "neovim/nvim-lspconfig",
+    opts = {
+      servers = {
+        typst_lsp = {},
+      },
+    },
+  },
+  -- Typst preview
+  {
+    "chomosuke/typst-preview.nvim",
+    ft = "typst",
+    build = function()
+      require("typst-preview").update()
+    end,
+    keys = {
+      {
+        "<leader>cp",
+        ft = "typst",
+        "<cmd>TypstPreviewToggle<cr>",
+        desc = "Typst Preview",
+      },
+    },
+  },
+}


lua/lazyvim/plugins/extras/lang/typst.lua
```diff
@@ -0,0 +1,36 @@
...
+      },
+    },
+  },
+  -- Typst preview
+  {
+    "chomosuke/typst-preview.nvim",
+    ft = "typst",
+    version = "0.3.*",
+    build = function()
+      require("typst-preview").update()
```
storopoli, line 25: I don't know if I should do this here
```suggestion
      require("lazy").load({ plugins = { "typst-preview.nvim" } }).update()
```

or
```suggestion
      require("lazy").load({ plugins = { "typst-preview" } }).update()
```

Please advise

dpetka2001: Why not just `require("typst-preview").update()` as you already have it? Did you encounter any problems?

lua/lazyvim/plugins/extras/lang/typst.lua
```diff
@@ -0,0 +1,36 @@
...
+      servers = {
+        typst_lsp = {},
+      },
+    },
+  },
+  -- Typst preview
+  {
+    "chomosuke/typst-preview.nvim",
+    ft = "typst",
+    version = "0.3.*",
```
dpetka2001, line 23: This will lock the plugin version to `0.3.*`. If there are any updates to the plugin version it won't get updated. Any reason this is needed? All plugins in LazyVim (except if I'm forgetting something) install the latest HEAD.

storopoli: Following the instructions in https://github.com/chomosuke/typst-preview.nvim

storopoli: But please, I am open to suggestions

lua/lazyvim/plugins/extras/lang/typst.lua
storopoli: No, but the `markdown-preview` stuff uses the suggested approaches.
Just trying to know which is more "LazyVim idiomatic".

lua/lazyvim/plugins/extras/lang/typst.lua
dpetka2001: My suggestion as just a simple user would be to not include `version` at all.

lua/lazyvim/plugins/extras/lang/typst.lua
dpetka2001: `markdown-preview` uses this approach because the plugin doesn't have a Lua `setup` function and the plugin is lazy-loaded and the command `mkdp#util#install` is a vim command that is not available if the plugin is not loaded. I don't believe this is needed here and you should just leave it as you originally have, unless you're experiencing any problems. Requiring the module should also load the plugin and call the `update` function if I'm not mistaken.

lua/lazyvim/plugins/extras/lang/typst.lua
storopoli: Done in 2c3103a04aabaf316fcb80e01a03baddcbb5e35e

dpetka2001:
@storopoli As I'm just a simple user like you, I don't believe there's a point in asking me to review this. I can tell you my recommendations and am happy to answer any questions you might have to the best of my ability and knowledge, but that's it. The maintainer will be the one reviewing this PR and has the final say.

2giosangmitom:
Duplicate PR. Please check out #4042.

storopoli:
Closing in favor of #4042.

PR #4448, opened 2024-09-16
feat(extras): add biome formatter

svirschevskiy:
## Description
This PR adds biome as a conform/null-ls formatter.

When using biome lsp formatting directly, the syntax highlighting is flickering. However, it works great when formatting is configured with conform.

To avoid conflicts with Prettier, it is recommended to set
`vim.g.lazyvim_prettier_needs_config = true` In this case, both prettier and biome formatters could be activated simultaneously. 

## Checklist

### DIFF
--- a/lua/lazyvim/plugins/extras/formatting/biome.lua
+++ b/lua/lazyvim/plugins/extras/formatting/biome.lua
@@ -1,0 +1,59 @@
+---@diagnostic disable: inject-field
+if lazyvim_docs then
+  -- Enable this option to avoid conflicts with Prettier.
+  vim.g.lazyvim_prettier_needs_config = true
+end
+
+-- https://biomejs.dev/internals/language-support/
+local supported = {
+  "astro",
+  "css",
+  "graphql",
+  -- "html",
+  "javascript",
+  "javascriptreact",
+  "json",
+  "jsonc",
+  -- "markdown",
+  "svelte",
+  "typescript",
+  "typescriptreact",
+  "vue",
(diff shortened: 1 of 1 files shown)


lua/lazyvim/plugins/extras/formatting/biome.lua
```diff
@@ -0,0 +1,61 @@
...
+        opts.formatters_by_ft[ft] = opts.formatters_by_ft[ft] or {}
+        table.insert(opts.formatters_by_ft[ft], "biome")
+      end
+
+      opts.formatters = opts.formatters or {}
+      opts.formatters.biome = {
+        condition = function(_, ctx)
+          return vim.fs.find({ "biome.json", "biome.jsonc" }, { path = ctx.filename, upward = true })[1]
+        end,
+      }
```
exsesx, line 47: You could also use this: https://github.com/stevearc/conform.nvim/blob/6a28c90082a67f62a323ab90d988e000a718a8c7/lua/conform/formatters/biome.lua#L11-L14

And just set:

```lua
opts.formatters.biome = {
  require_cwd = true,
}
```

exsesx:
Thank you for the pull request! I’m looking forward to it being merged 👏 

We transitioned from using Prettier and ESLint to Biome for my main project and noticed a significant performance improvement. In our CI, checking for formatting and linting used to take 2-3 minutes, but with Biome, it only takes 5 seconds using the `biome check` command.

folke:
ty!

PR #4471, opened 2024-09-20
fix(extras): .env* files shouldn't warn about unused vars

minusfive:
## Description

As `.env` files are configured as `filetype: "sh"` through the Dot Files extra, `shellcheck` linting kicks in which causes very noisy rendering, full of `[VAR] appears unused. Verify use (or export if used externally). [SC2034]`.

This PR changes their `filetype` to `conf` instead. Pros: significantly less noisy. Cons: loses some syntax highlighting / formatting features. I think it's an "OK" immediate compromise, but it'd perhaps be better to configure `shellcheck` to specifically ignore `SC2034` for `.env*` files by default. Alas, not much time to dig into it nowadays.

## Related Issue(s)

No issue I could find, but related discussion https://github.com/LazyVim/LazyVim/discussions/4027

### Before
Screenshot 2024-09-20 at 09 39 28

### After
Screenshot 2024-09-20 at 09 40 26

## Checklist

### DIFF
--- a/lua/lazyvim/plugins/extras/util/dot.lua
+++ b/lua/lazyvim/plugins/extras/util/dot.lua
@@ -32,7 +32,12 @@ return {
       end
 
       vim.filetype.add({
-        extension = { rasi = "rasi", rofi = "rasi", wofi = "rasi" },
+        extension = {
+          env = "conf",
+          rasi = "rasi",
+          rofi = "rasi",
+          wofi = "rasi",
+        },
         filename = {
           ["vifmrc"] = "vim",
         },
@@ -41,7 +46,7 @@ return {
           [".*/mako/config"] = "dosini",
           [".*/kitty/.+%.conf"] = "bash",
           [".*/hypr/.+%.conf"] = "hyprlang",
-          ["%.env%.[%w_.-]+"] = "sh",
+          ["%.env%.[%w_.-]+"] = "conf",
         },
       })
 


minusfive:
@dpetka2001 I looked into it, read shellcheck docs before this PR, but unfortunately shellcheck doesn't yet offer a way to globally configure conditionally enabling/disabling rules (e.g. just for `.env` files). It's all or nothing. So it'd have to be done at the LSP / linter level in LazyVim regardless. That's the path I started looking into but unfortunately don't have the time to dive in.

One could tell users to manually disable rules per file on all their `.env` files, but that is telling users to pollute their project files with tool-specific code, which isn't great. Lots of people use `.env` files in all sorts of projects/editors, not all of them use/care about shellcheck specifics.

I'd love to spend the time to sort out the LSP/lint config here, and perhaps will in the future. In the meantime, though, I'm running the changes on this PR on my own config because without them `.env` files _**feel broken**_. With these changes they no longer feel broken. And yes, you lose some functionality, but I'm not sure `.env` files are complex enough for that to be a big deal while the rest is sorted.

dpetka2001:
Vanilla Neovim itself recognizes `.env` files as `ft = sh`. LazyVim just extends the default behavior to further include additional patterns `.env.whatever`. What you want can also be achieved by just disabling LSP diagnostics for the `.env` files (since you associate them with `.conf` filetype either way). A simple autocmd like the following should work I believe
```lua
vim.api.nvim_create_autocmd("BufReadPost", {
  pattern = "*.env",
  desc = "Disable diagnostic for .env files",
  callback = function(ev)
    vim.diagnostic.enable(false, { bufnr = ev.buf })
  end,
})
```

minusfive:
@dpetka2001 nice, thanks! Will try that out later.

folke:
The dot extra does **not** configure the `sh` filetype, it only does so for `.env.*` files.
Neovim sets the `sh` filetype, so we follow the same.

PR #4477, opened 2024-09-22
feat(rust): update rustaceanvim

edevil:
The recommended version of rustaceanvim has changed to 5.x, which brings in a lot of improvements.

## Description

Version 5.x of rustaceanvim has a lot of improvements over 4.x, and is the recommended version according to its maintainer.

## Checklist

### DIFF
--- a/lua/lazyvim/plugins/extras/lang/rust.lua
+++ b/lua/lazyvim/plugins/extras/lang/rust.lua
@@ -42,7 +42,7 @@ return {
 
   {
     "mrcjkb/rustaceanvim",
-    version = "^4", -- Recommended
+    version = vim.fn.has("nvim-0.10.0") == 0 and "^4" or false,
     ft = { "rust" },
     opts = {
       server = {


dpetka2001:
Is there a specific reason we lock `rustaceanvim` to stable releases? All other plugins update to latest HEAD. This also incurs additional attention for whenever the plugin has a major release.

edevil:
I suspect the reason is the fact that the rustaceanvim README recommends specifying a major version. Currently it's at 5.

lua/lazyvim/plugins/extras/lang/rust.lua
```diff
@@ -42,7 +42,7 @@ return {
 
   {
     "mrcjkb/rustaceanvim",
-    version = "^4", -- Recommended
+    version = "^5", -- Recommended
```
stefanboca, line 45: It's possible that rustaceanvim was pinned to 4.x because 5.x requires neovim>=0.10.0, while LazyVim currently requires only neovim>=0.9.0. It's probably better to keep neovim<0.10.0 on 4.x, using something like

```lua
version = vim.fn.has("nvim-0.10.0") == 1 and "^5" or "^4"
```

dpetka2001: I believe it should be better in the check for 0.10 to define the version as `false` to always update to latest HEAD.

stefanboca: In that case the condition would become
```lua
version = vim.fn.has("nvim-0.10.0") == 0 and "^4" or false
```
I agree that for neovim>=0.10 rustaceanvim shouldn't be pinned to a specific major version.

edevil: Done.

lua/lazyvim/plugins/extras/lang/rust.lua
```diff
@@ -42,7 +42,7 @@ return {
 
   {
     "mrcjkb/rustaceanvim",
-    version = "^4", -- Recommended
+    version = vim.fn.has("nvim-0.10.0") == 0 and "^4" or false,
```
IvanProg00, line 45: Maybe is better to use `vim.version()`.

```suggestion
    version = (vim.version().major == 0 and vim.version().minor == 9) and "^4" or false,
```

edevil: Agree.

dpetka2001: `vim.fn.has` is more consise and used throughout many plugins to check for features. It's more of a convention to use it.

edevil: I see that this is a subjective thing, so I'll use the convention @folke tells me to. :)

dpetka2001: https://github.com/search?q=repo%3ALazyVim%2FLazyVim%20vim.version&type=code

https://github.com/search?q=repo%3ALazyVim%2FLazyVim%20vim.fn.has&type=code

The only usage of `vim.version` is to construct `0_9` as a string to be used in the compat import for Neovim versions 0.9+

IvanProg00: If the user has Neovim version `0.9.5`, it means he will have the latest version of Rustacean Vim (not `^4`).

edevil: `vim.fn.has("nvim-0.10.0")` returns 0 unless the user is on 0.10 or above.

dpetka2001: Where do you base this off? I just tested Neovim 0.9.5 and `vim.fn.has("nvim-0.10")` returns `0`.

edevil: Exactly, that's what I said. :) Then it will use 4.

dpetka2001: I also tested this PR locally with Neovim 0.9.5 and `rustaceanvim` is 4.26.1

dpetka2001: @edevil Sorry my previous comment wasn't towards you, rather towards @IvanProg00. I'm just mentioning it, so that there will be no confusion.

folke:
ty!

PR #4479, opened 2024-09-22
Migrate away from deprecated `get_target_window()` in mini-files

torinmr:
## Description

This method is deprecated in the most recent version of mini-files: https://github.com/echasnovski/mini.files/blob/a3a9cce82115a69dba161ac45bda16f4e606f73b/lua/mini/files.lua#L1060-L1068, which causes a warning message to be displayed every time the `<leader>wv` and similar commands are used from the mini file browser.

I've tested this change locally and it gets rid of the warning message.

## Screenshots
Warning message without this fix:
image

## Checklist


folke:
ty!

PR #4489, opened 2024-09-25
enhance pasting and deleting text from buffer

melsiir:
## Description

by nature if yanked text and selected other  text and pasted it over it   neovim will  automatically replace the yanked  text with the selected text.
This mapping will disable this behavior 

## Checklist

- [ x ] I've read the [CONTRIBUTING](https://github.com/LazyVim/LazyVim/blob/main/CONTRIBUTING.md) guidelines.


dpetka2001:
I don't believe that's good default. There's already `:h v_P` to achieve this behavior. Not all ppl want this by default. It should belong in user's personal configuration in my opinion.

PR #4499, opened 2024-09-26
fix(lualine): only trim root from path if present

willnorris:
Ensure that path actually begins with root before trimming it off the front. Otherwise, the path gets mangled.

## Description

`Util.lualine.pretty_path()`  checks if that path should be relative to `cwd` and trims the value of `cwd` off the front of path if present.  If not, then it just blindly trims the length of `root` off of the front of path, without checking to see if the path is prefixed with `root`. This results in paths getting mangled.

For example, if I run `cd /tmp && vi ~/.config/nvim/lua/plugins/init.lua` (so that I'm editing a file outside of cwd), then lualine.pretty_path() sets a path of `ns/init.lua`:

2024-09-26T11:41:18,006345055-07:00

With this fix, it leaves the path alone, and collapses path elements to the configured length:

2024-09-26T11:41:40,978182873-07:00

## Checklist

### DIFF
--- a/lua/lazyvim/util/lualine.lua
+++ b/lua/lazyvim/util/lualine.lua
@@ -96,7 +96,7 @@ function M.pretty_path(opts)
 
     if opts.relative == "cwd" and path:find(cwd, 1, true) == 1 then
       path = path:sub(#cwd + 2)
-    else
+    elseif path:find(root, 1, true) == 1 then
       path = path:sub(#root + 2)
     end
 


dpetka2001:
Duplicate of #4428.

willnorris:
> Duplicate of #4428.

hah, yep!    Didn't see that one.  I'll close this in favor of the existing PR

PR #4518, opened 2024-10-01
fix(extras): rename `lspconfig.server_configurations` to `lspconfig.configs`

sethtrain:
## Description

https://github.com/neovim/nvim-lspconfig/commit/bedb2a0df105f68a624a49b867f269b6d55a2c89 introduces a change to the Lua module structure for `nvim-lspconfig` that broke `tailwindcss`.  Tried to also address all the other usages of `lspconfig.server_configurations` within the ecosystem.

## Screenshots

mason-lspconfig nvim

## Checklist

### DIFF
--- a/lua/lazyvim/plugins/extras/lang/tailwind.lua
+++ b/lua/lazyvim/plugins/extras/lang/tailwind.lua
@@ -28,7 +28,7 @@ return {
       },
       setup = {
         tailwindcss = function(_, opts)
-          local tw = require("lspconfig.server_configurations.tailwindcss")
+          local tw = require("lspconfig.configs.tailwindcss")
           opts.filetypes = opts.filetypes or {}
 
           -- Add default filetypes
--- a/lua/lazyvim/plugins/extras/lang/java.lua
+++ b/lua/lazyvim/plugins/extras/lang/java.lua
@@ -1,4 +1,4 @@
--- This is the same as in lspconfig.server_configurations.jdtls, but avoids
+-- This is the same as in lspconfig.configs.jdtls, but avoids
 -- needing to require that when this module loads.
 local java_filetypes = { "java" }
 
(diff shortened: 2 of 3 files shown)


hampusolsen:
Was just about to create a PR for this -- happy to see there was already a fix waiting to be merged! 🥇

lua/lazyvim/plugins/extras/lang/tailwind.lua
```diff
@@ -28,7 +28,7 @@ return {
       },
       setup = {
         tailwindcss = function(_, opts)
-          local tw = require("lspconfig.server_configurations.tailwindcss")
+          local tw = require("lspconfig.configs.tailwindcss")
```
justinmk, line 31: Even before https://github.com/neovim/nvim-lspconfig/pull/3330, the documented way to use configs from nvim-lspconfig has always been:

```suggestion
          local tw = require'lspconfig'.tailwindcss
```

so I wasn't expecting https://github.com/neovim/nvim-lspconfig/pull/3330 to break anything.

I will add a shim to avoid breakage, but was there a reason to not use `require'lspconfig'.tailwindcss` (it would have avoided the breakage) ?

dpetka2001: My guess would be to get access to the `default_config.filetypes` table which is used later on. It's not for setting up the LSP server. Maybe `require'lspconfig'.tailwindcss.document_config` could be used instead? That way later on `tw.default_config.filetypes` will be accessible. Maybe there's a better way?

Andreasgdp (approved):

folke:
ty!

lua/lazyvim/plugins/extras/lang/tailwind.lua
folke: Indeed. `default_config` is not available when getting it with `require'lspconfig'.tailwindcss`

justinmk: Ok. I added `config_def` for that purpose in https://github.com/neovim/nvim-lspconfig/pull/3335 , and that's now documented.

PR #4531, opened 2024-10-02
feat(options): add `localoptions` to `opt.sessionoptions`

linkarzu:
## Description

  
  I take markdown notes in multiple languages (English and Spanish), but my langspell settings are not remembered if I switch to something other than the config in `opt.spelllang` in the `options.lua` file, which I have set to `English`
  
  By adding `localoptions` to  `opt.sessionoptions` these settings are persisted so that I can change the language to `english` , `spanish` or `en,es` on a buffer and when I re-open the buffer it will pick up that information so that I don't have to be setting the spellang again
  
  See [this video](https://youtu.be/uLFAMYFmpkE) for reference

## Checklist

### DIFF
--- a/lua/lazyvim/config/options.lua
+++ b/lua/lazyvim/config/options.lua
@@ -89,7 +89,8 @@ opt.pumblend = 10 -- Popup blend
 opt.pumheight = 10 -- Maximum number of entries in a popup
 opt.relativenumber = true -- Relative line numbers
 opt.scrolloff = 4 -- Lines of context
-opt.sessionoptions = { "buffers", "curdir", "tabpages", "winsize", "help", "globals", "skiprtp", "folds" }
+opt.sessionoptions =
+  { "buffers", "curdir", "tabpages", "winsize", "help", "globals", "skiprtp", "folds", "localoptions" }
 opt.shiftround = true -- Round indent
 opt.shiftwidth = 2 -- Size of an indent
 opt.shortmess:append({ W = true, I = true, c = true, C = true })


dpetka2001:
`spelllang` can take a list of comma-separated languages. Wouldn't it be more prudent to change that in your personal configuration instead of this?

linkarzu:
> `spelllang` can take a list of comma-separated languages. Wouldn't it be more prudent to change that in your personal configuration instead of this?

I set spelllang in some buffers to `english`, other ones to `spanish` and some others to both `english and spanish`, are you referring to setting this option `vim.opt.spelllang = { "en" }` to something else in my `options.lua` file?

linkarzu:
But just to clarify, I'm not touching spelllang in this PR, it's related directly to adding `localoptions` to `sessionoptions` so that the language I set on each buffer is persisted. Not sure if that helps or clarifies things, but let me know.

dpetka2001:
Yes, i'm talking about `vim.opt.spelllang` in your `options.lua` file. You can set more than one language. Read the help. You're only adding `localoptions`, so that the `spelllang` setting persists across sessions. I believe `localoptions` might have too much unnecessary info and should belong in the user personal configuration if someone wants that. Not be a LazyVim default.

But i'm just a simple user like you expressing only my personal opinion and the maintainer has the final say.

linkarzu:
I appreciate your response, you've helped me a lot in the past with Neovim questions. 
I have localoptions set in my local config, and I wasn't planning on submitting a PR until a user in the video suggested it.

I think it would help users that want to take notes in more than 1 language and not set their spellang to a static value, as I can have it set to English, but in some buffers I want to set it to Spanish, and I would like that language information to be remembered for that buffer.

It's ok either way, the video explains how to set it up in case the maintainer thinks this is not a good idea

folke:
This is a bad idea since it also sets buffer-local mappings, which can lead to unexpected behavior.

dpetka2001:
Now I remember what I did and didn't like this option. I had tried it in the past and somewhere along the lines I was experimenting with `foldexpr` and `foldmethod`. I had set `set foldmethod=manual` and `set fde=` to try manual folds and after restarting Neovim the option persisted obviously due to `localoptions` and the default settings LazyVim had for folding were messed up. 

Of course there might be other ways as well where you can get unexpected behavior. 

@linkarzu If you also mention this in a video which probably quite a descent number of newcomers to Neovim might watch, please be sure to mention that `localoptions` could lead to unexpected behavior like the example I mentioned above (which at first might leave them baffled as to what is going on and why default LazyVim settings don't work) and users should use it if they really understand what they're doing.

linkarzu:
Appreciate both responses, and thanks for the folding examples @dpetka2001 Ill try to keep this in mind when things break and I cannot explain what's going on.
By any chance, does any of you know if there's a better way to **persist** the `spelllang` on a per buffer basis without adding `localoptions` to `sessionoptions`?

dpetka2001:
Have you tried passing multiple languages to `spelllang`? The help file has examples. Something like
```lua
vim.opt.spelllang:append("es") -- to append to the default `en` that LazyVim has
-- or
vim.opt.spelllang = { "gr", "es" }  -- if you want an entirely new `spelllang` without the English language
```
should do it in your case I believe. Although I don't use spelling and am not 100% sure I'm right.

linkarzu:
Yep, by looking at my `options.lua` file, I once in the past set it to both `en,es`, but then I switched it bach to `en` a seen here below

```lua
-- Most of my files have 2 languages, spanish and english, so even if I set the
-- language to spanish, I always add some words in English to my documents, so
-- it's annoying to be adding those to the spanish dictionary
-- vim.opt.spelllang = { "en,es" }

-- I mainly type in english, if I set it to both above, files in English get a
-- bit confused and recognize words in spanish, just for spanish files I need to
-- set it to both
vim.opt.spelllang = { "en" }
```

So basically, I just set the default to `en` because if set to both languages all the time, the documents that I just have in 100% english would get confused and start recognizing words in Spanish, I don't remember quite well, I should have documented an example in a better way because now I forgot. 

So I change the language on a "per buffer" basis with keymaps, but maybe this is not the best approach?

```lua
-- Keymap to switch spelling language to Spanish lamw25wmal
vim.keymap.set("n", "<leader>msls", function()
  vim.opt.spelllang = "es"
  vim.cmd("echo 'Spell language set to Spanish'")
end, { desc = "[P]Spelling language Spanish" })

-- Keymap to switch spelling language to both spanish and english lamw25wmal
vim.keymap.set("n", "<leader>mslb", function()
  vim.opt.spelllang = "en,es"
  vim.cmd("echo 'Spell language set to Spanish and English'")
end, { desc = "[P]Spelling language Spanish and English" })
```

Now I'm thinking of trying again setting `vim.opt.spelllang = { "en,es" }` and documenting better why I didn't like it, or why It didn't work for me

PR #4557, opened 2024-10-08
refactor(cmp): add `optional` where necessary

dpetka2001:
## Description
Makes it easier to disable it for users who want to try other completion engines, without it being pulled back by some Extra.

## Related Issue(s)
None

## Checklist

### DIFF
--- a/lua/lazyvim/plugins/coding.lua
+++ b/lua/lazyvim/plugins/coding.lua
@@ -85,6 +85,7 @@ return {
   -- snippets
   {
     "nvim-cmp",
+    optional = true,
     dependencies = {
       {
         "garymjr/nvim-snippets",
@@ -209,6 +210,7 @@ return {
   -- Add lazydev source to cmp
   {
     "hrsh7th/nvim-cmp",
+    optional = true,
     opts = function(_, opts)
       table.insert(opts.sources, { name = "lazydev", group_index = 0 })
     end,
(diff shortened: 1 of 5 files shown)


idelice:
Shouldn't this PR also apply optional capabilities for nvim-jdtls in some way?

https://github.com/LazyVim/LazyVim/blob/13a4a84e3485a36e64055365665a45dc82b6bf71/lua/lazyvim/plugins/extras/lang/java.lua#L169

dpetka2001:
`LazyVim.has("cmp-nvim-lsp")` checks if `cmp-nvim-lsp` exists. If it exists it adds to capabilities, otherwise it's nil.

idelice:
> `LazyVim.has("cmp-nvim-lsp")` checks if `cmp-nvim-lsp` exists. If it exists it adds to capabilities, otherwise it's nil.

I know, but what if you want to add other capabilities?

dpetka2001:
I don't do Java, so can't test this. My guess would be to also expose capabilities in `opts` and then in the local `config` do something like what is being done [here](https://github.com/LazyVim/LazyVim/blob/13a4a84e3485a36e64055365665a45dc82b6bf71/lua/lazyvim/plugins/lsp/init.lua#L187-L193) to merge all the capabilities together. 

One who does Java should make an appropriate PR after having it tested thoroughly.

folke:
ty!

PR #4587, opened 2024-10-13
feat(extras): expose `prios` to users for customization

### LINKED ISSUE #4584: feature: customize priority of user extras
### Did you check the docs?

As I see, there is currently no way to change priority of user extras (plugins from "plugins.extras" directory) as it is done with some builtin extras; all user extras are given the same priority of 50.

Possible solution is to expose `prios` table from https://github.com/LazyVim/LazyVim/blob/main/lua/lazyvim/plugins/xtras.lua as a property on `LazyVim` global object so that user can modify it.

I don't see how it can be done right now.

### Additional context

_No response_

dpetka2001:
## Description
Offer to users capability to customize priorities of the Extras.

## Related Issue(s)
Closes #4584

## Checklist

### DIFF
--- a/lua/lazyvim/plugins/xtras.lua
+++ b/lua/lazyvim/plugins/xtras.lua
@@ -10,6 +10,10 @@ local prios = {
   ["lazyvim.plugins.extras.editor.outline"] = 100,
 }
 
+if vim.g.xtras_prios then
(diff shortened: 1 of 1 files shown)


folke:
ty!

PR #4603, opened 2024-10-17
docs(dap): document `continue` to be used also for `run`

LorenzoBettini:
The `dap.continue` is meant to be used also to start the program if no debugging session is active.

The documentation states:

> `continue()` is the main entry-point for users to start debugging an application.

With this change, it should be clearer to the user that `continue` is the way to start a program. Otherwise, one could think to use `Run with args` (which is misleading). At least for me, it was, until I read the dap documentation.

## Checklist


folke:
ty

PR #4605, opened 2024-10-17
chore(extras/fzf-lua): specify mini.icons dependency

skoch13:
Despite LazyVim using mini.icons by default, `fzf-lua` requires you to specify the icons dependency to use them.

## Checklist


dpetka2001:
No? The dependency is specifically to install the plugin. I can see icons just fine in fzf-lua that I'm using. Did you encounter some problem?

skoch13:
@dpetka2001 I did, but now I think it might be a glitch within `fzf-lua` as once I reopened neovim the problem disappears even without specifying the dependency 🤔

dpetka2001:
The dependency is not really needed, since `mini.icons` is already installed by LazyVim.

PR #4333, opened 2024-08-20
feat(lazygit): add support for `ssh://host[:port]/url` pattern in git browse

consoleaf:
## Description

Hi! This adds support for the `ssh://host[:port]` pattern in Git Browse (`<leader>gB`)

## Related Issue(s)

No related issues

## Checklist

- [x] I've read the [CONTRIBUTING](https://github.com/LazyVim/LazyVim/blob/main/CONTRIBUTING.md) guidelines.

PR #3941, opened 2024-07-06
fix(lazygit): improve git browse

med8bra:
## Description
Improves git browse command by handling different types of remotes, and allows user to extend to other git hosts.

## Related Issue(s)
Fixes #3886

## Checklist
- [x] I've read the [CONTRIBUTING](https://github.com/LazyVim/LazyVim/blob/main/CONTRIBUTING.md) guidelines.

PR #4631, opened 2024-10-23
feat(lazygit): add user extensible capability for `M.remote_patterns`

dpetka2001:
## Description
I'm not sure if this is something you want. I've noticed quite a few people would like to extend `M.remote_patterns` to accommodate their own use case (#4333 and [this comment](https://github.com/LazyVim/LazyVim/pull/3941#issuecomment-2396569469)). Instead of opening new PRs, this provides a way for users to add new patterns in their personal configuration.

Again, I don't know if that's something you want. Maybe you would prefer PRs to be created for this purpose, so feel free to disregard this.

## Related Issue(s)
None

## Checklist


folke:
can be configured now with snacks. ty

PR #4638, opened 2024-10-23
fix(autocmds): change mapping for `lazyvim_close_with_q`

dpetka2001:
## Description
Currently `checkhealth` buffers stay visible on `bufferline` when you close them with `q`. Use `:bd` instead to delete the buffer from bufferlist. `vim.schedule` is needed because `LspInfo` adds its own mapping to close the window (see [here](https://github.com/neovim/nvim-lspconfig/blob/edd9591199d1c78c0cb20514231f7f936f9412a2/lua/lspconfig/health.lua#L328)), so we need to overwrite it.

## Related Issue(s)
None

## Checklist

### DIFF
--- a/lua/lazyvim/config/autocmds.lua
+++ b/lua/lazyvim/config/autocmds.lua
@@ -72,11 +72,13 @@ vim.api.nvim_create_autocmd("FileType", {
   },
   callback = function(event)
     vim.bo[event.buf].buflisted = false
-    vim.keymap.set("n", "q", "<cmd>close<cr>", {
-      buffer = event.buf,
-      silent = true,
-      desc = "Quit buffer",
-    })
+    vim.schedule(function()
+      vim.keymap.set("n", "q", "<cmd>:bd<cr>", {
+        buffer = event.buf,
+        silent = true,
+        desc = "Quit buffer",
+      })
+    end)
   end,
 })
 


folke:
ty!

folke:
I had to change the close command because when closing help buffers, an empty window remainded.

Let me know if you'd still experience issues

dpetka2001:
Strange I had tested for help buffers as well as I use them very often and I hadn't observed this (maybe you use edgy and you noticed that as I tested without edgy, but don't know if that would be the reason). Either way with your new change `checkhealth` buffers still behave as they should and that was the main reason I created this PR, so goal accomplished :smile:

dpetka2001:
@folke I found this doesn't work well with `notify` windows (and possible other floating windows?). A window remains after closing the buffer. I had to modify the code inside the callback to
```lua
    vim.schedule(function()
      if vim.bo[event.buf].filetype == "notify" then
        vim.keymap.set("n", "q", "<cmd>close<cr>", { buffer = event.buf, silent = true, desc = "Quit buffer" })
      else
        vim.keymap.set("n", "q", function()
          vim.api.nvim_buf_delete(event.buf, { force = true })
        end, {
          buffer = event.buf,
          silent = true,
          desc = "Quit buffer",
        })
      end
    end)
```
Is this the correct way to do this? If so I could make a PR. Maybe you have a better idea as to how to deal with this?

folke:
My bad, `close` is always needed. Added an extra buf delete now.

tbh, I couldn't reproduce your original issue with the healthcheck buffers, but what was indeed not correct is that the buffer was hidden instead of deleted.

Hopefully this fixes your original issue

dpetka2001:
Yes I don't see the original issue with your latest changes, so it works perfectly fine.

Thanks!! :smile:

PR #4655, opened 2024-10-26
feat(toggle): add `silent` field to toggle to customize behavior of notifications

dpetka2001:
## Description
Allow user to customize the behavior of notifications pop-ups to be able to hide them or not.

## Related Issue(s)
A comment https://github.com/LazyVim/LazyVim/discussions/4232#discussioncomment-11061942

## Checklist

### DIFF
--- a/lua/lazyvim/util/toggle.lua
+++ b/lua/lazyvim/util/toggle.lua
@@ -8,6 +8,7 @@ local M = {}
 ---@field color_disabled? string
 ---@field get fun():boolean
 ---@field set fun(state:boolean)
+---@field silent? boolean
 
 ---@class lazyvim.Toggle.wrap: lazyvim.Toggle
 ---@operator call:boolean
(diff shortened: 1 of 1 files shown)


ficcdaf:
This seems like a really small fix that will have a big impact. Considering only a few lines are changed and the option is user overridable, I don't see this having any kind of knockon effect and I really hope it gets accepted.

ficcdaf (approved):
It seems good that if `silent` is not supplied, it will be treated as false, which will be compatible with everyone's existing configs. I wonder if maybe the default value of silent should be set to false explicitly? I am not too familiar with Lua conventions though, so I'm unsure if this is best practice :)

I think this is a very small fix, with big positive ramifications in terms of user experience. I don't see this having any sort of negative knock-on effect or breaking anything.

dpetka2001:
>  I wonder if maybe the default value of silent should be set to false explicitly?

I don't believe that's necessary, since this field is only meant to change the current default behavior. Setting it to `false` explicitly would mean to change the current LazyVim toggles that exist. I don't see any drawback in the current approach I chose, except if I'm missing something. The `nil` value is a `falsy` value, so should be the same.

ficcdaf:
> > I wonder if maybe the default value of silent should be set to false explicitly?
> 
> I don't believe that's necessary, since this field is only meant to change the current default behavior. Setting it to `false` explicitly would mean to change the current LazyVim toggles that exist. I don't see any drawback in the current approach I chose, except if I'm missing something. The `nil` value is a `falsy` value, so should be the same.

You are right, thanks for clearing that up. I'm still learning about Lua. I don't have any other concerns with this PR, and I definitely support it being merged after the maintainer is back from vacation.

Issue #238, opened 2023-02-11
feature: Add option to disable all LazyVim keymaps


MaazSiddiqi:
### Did you check the docs?

- [X] I have read all the LazyVim docs

### Is your feature request related to a problem? Please describe.

It's really frustrating when reassigning keymaps, and you want to map say `<leader>f`, but somewhere there's a map for `<leader>fr`. Of course, you don't initially know an `fr` key bind exists, so whenever you do `<leader>f`, nvim is pending for another keystroke before confirming the call...I think. So you have to hunt for any `<leader>f{char}` keymaps and unbind them, or even worse, have to reassign them. 

I love how LazyVim ships with a great set of default bind. However, as someone who's built strong preferences for their keymaps, it can make the process of reassigning binds frustrating and tedious. 

### Describe the solution you'd like

I would love it if there was a way to just not load any of the default keymaps at all. Give the user the option to fully reconfigure their key binds from scratch, or let them start off with a much smaller subset of key binds, and provide the rest as opt-ins. This would be similar to how we approach `lazyvim.plugins.extras.lang.typescript` but instead of language extensions, we have keymap extensions. 

As someone migrating with a preconfigured list of keymaps, this would simplify the process for me a lot more. We can enable it through the main `lazy.lua` file:
```lua
require("lazy").setup({
  spec = {
    -- add LazyVim and import its plugins
    { "LazyVim/LazyVim", import = "lazyvim.plugins", keys = false },
```

Keymap extensions aren't necessary, but it would be great if LazyVim had a way to disable importing its default keymaps. Then I can use the same principle of adding keymaps to my plugins inside spec files `keymaps.lua` and they'd just get merged like normal. 

### Describe alternatives you've considered

Going into each separate plugin and disabling keys that are conflicting or making neovim wait. 

### Additional context

Sorry if the issue is too wordy, I'm not too used to writing issues yet. Also, I've tried adding `keys = function () return {} end` to the lazyVim spec, but keymaps were still getting loaded. If this is already implemented, am I doing it the wrong way?

folke:
Made this configurable. See https://www.lazyvim.org/configuration#default-settings

Nohac:
@folke Will this new setting also disable default keymaps for telescope and lsp? If not, I think it would be great to have, or at least add separate options for those.

[...]

PR #4660, opened 2024-10-28
feat(keymaps): enable modifying and disabling keymaps

dunix241:
Hello!

Firstly thank you, guys, for this amazing work, I utmostly admire all of your hard work and talent.

Secondly, sorry for this rather big PR I have no other ways. But please give me a chance to explain it.

## Problem statements

- The predefined features are highly functional, but sometimes users may want to modify the key mapping to a feature to fit their needs. They could disable the predefined keymap, redefine the feature based on our configuration, and create their own key mapping to that feature, which creates redundant code and thus slower performance.

```lua
{
  "foo/bar",
  keys = function()
    return {
      { new_key, redefined_feature },
    }
  end
}
```

- By redefining a feature based on our configuration, when that feature is updated, users will have to be aware and update their keymaps correspondingly. While in case keymaps are centralized and override-able, users don't have to worry about it.
- Also, sometimes they just want to modify or disable a single keymap without doing any other configuration thus overriding the plugin seems to be a lot of extra work.

## Solutions

To override or disable a keymap, users shall do something like this:

```lua
-- place this before lazy setup
vim.g.lazyvim_keymaps = {
  window_left = "keys1",
  window_lower = "keys2",
  size_increase_height = "keys3"
}
require("lazy").setup({
  --...
})
```

Additionally, users can group keys into tables for better organization if desired:

```lua
-- place this before lazy setup
vim.g.lazyvim_keymaps = {
  {
    window_left = "keys1",
    window_lower = "keys2"
  },
  {
    size_increase_height = "keys3"
  }
}
require("lazy").setup({
  --...
})
```

Or

```lua
-- place this before lazy setup
vim.g.lazyvim_keymaps = {
  keymaps = {
    window = {
      window_left = "keys1",
      window_lower = "keys2"
    },
    resize_window = {
      size_increase_height = "keys3"
    }
  },
}
require("lazy").setup({
  --...
})
```

Note that the list of keymaps can be quite long, so it is better to separate it into different tables for better organization.

- To override a keymap users can set the key to a new keymap.

- To disable a keymap user can set the key to an empty string - `""`

  - Currently only able to disable previously set keymaps with `vim.keymap.set()` by switching to `LazyVim.keymap_set` instead. But if this feature can be accepted, I can create a PR beforehand to [lazy.nvim](https://github.com/folke/lazy.nvim/blob/cf8ecc2c5e4332760431a33534240b0cbc6680ab/lua/lazy/core/handler/keys.lua#L116) to add this functionality to lazy plugin's `keys` table by extending the `add()` and `set()` function to include not setting keymap if `lfs` is nil or empty (I have tried, and it worked as expected).

Benefits:

- Allow users to modify or disable keymaps.
- Keep all the keymaps in one place for ease of maintenance. (Users can extend `vim.g.lazyvim_keymaps` to include their own keymaps and use it as a centralized keymap configuration)

```lua
local keymaps = vim.g.lazyvim_keymaps.user_added_plugin
return {
  "foo/user-added-plugin",
  keys = {
    [keymaps.feature1] = "feature1"
  }
}
```

- Unify keys to similar features. For example, `fzf-lua` and telescope use the same keys for their keymaps thus, when users switch to `fzf-lua` from `telescope` they don't have to override `fzf-lua` keymaps. This is especially useful for setting `lsp` as well, as if I use different languages and I want to change the keys for `Organize Imports` I just need to set it once without having to do it for all the languages.

This feature is included in any modern IDEs or text editors, and it is also simple to implement. I believe this feature is wished by many, and I will be extremely happy and grateful if this feature is accepted. Thank you.

## Related Issue(s)
- #238

## Checklist

- [x] I've tested it with all the plugins' keys set to empty strings (with the above-mentioned modifications to `lazy.nvim`)


dpetka2001:
Really not a fan of this. Unnecessary complexity for the sake of having centralized keymaps. Personally I prefer to have whatever keymaps bound to corresponding plugin spec and do whatever changes there. 

But this is just my personal opinion as a simple user like you.

dpetka2001:
The slower startup time doesn't make sense to me as it's the `lazy.nvim` handlers that would have to get parsed that might add up whatever miniscule amount of startup time. In your case the `keys` handlers still have to get parsed, so I don't really see the benefit. But I guess if `vim.g.lazyvim_keymaps` does get parsed before `lazy.nvim` setup you're only merging to the default LazyVim plugin specs and that way you don't have to define additional handlers in the user personal configuration that would have to also get parsed (only the initial instantiation of `vim.g.lazyvim_keymaps` would add additional time). Maybe you have a point there. I'm not sure, but it does sound logical.

> Other than that, I found myself loving a feature but I had to constantly copy and paste the config to override it without adding any functionality which is kind of frustrating.

You're clearly doing something wrong, as the the default config doesn't have to be copied over to the user's personal configuration. Only the relative parts that you want to change have to make it into the user's configuration. 

About the CI test pipeline, unfortunately I'm not familiar with Github actions myself, so can't help you there. Better wait for maintainer's input when he's back from vacation to give you a proper answer.

dpetka2001:
For the stylua fail it seems that in the file `telescope.lua` it formats it to 
```lua
mappings = {
    276  |+          n = k.flash.normal == "" and {} or { [k.flash.normal] = flash },
    277  |+          i = k.flash.insert == "" and {} or { [k.flash.insert] = flash },
    278  |+        },
```
instead of it being in one line. Maybe you could try it like that and see if that fixes the stylua failure.

dunix241:
Thank you. I got this one. Already committed but I'm looking to fix the test pipeline to push them all at once.

lua/lazyvim/keymaps.lua
```diff
@@ -0,0 +1,673 @@
...
+      find_in_files = "<leader>/",
+      go_to_symbol = "<leader>ss",
+    },
+  },
+}
+
+---@return LazyVimKeymaps
+M.get_keymaps = function()
+  if keymaps == nil then
+    keymaps = vim.tbl_deep_extend("force", {}, M.default_keymaps, vim.g.lazyvim_keymaps)
```
dpetka2001, line 667:
```suggestion
    keymaps = vim.tbl_deep_extend("force", {}, M.default_keymaps, vim.g.lazyvim_keymaps or {})
```
Here you need a check for empty table in case `vim.g.lazyvim_keymaps` is not defined. I tried your PR locally and Neovim crashed with error about `vim.tbl_deep_extend` 2nd argument expecting a table but got `nil`.

dpetka2001: You can test by running in your local forked repo at the root of the repo `./scripts/test`.

dunix241: Thank you, I really appreciate it, already passed locally. But I remembered I had included this don't know why and when I removed it 😆

dunix241:
> You're clearly doing something wrong, as the the default config doesn't have to be copied over to the user's personal configuration. Only the relative parts that you want to change have to make it into the user's configuration.

Btw, may I ask how you would replace the keymaps? Is there a better way? Do you modify the `keys` list or something? I'm just curious. 

I just normally do something like this:

```lua
{
  "foo/bar",
  keys = function()
    return {
      { new_key, redefined_feature },
    }
  end
}
```

dpetka2001:
By using a `values` function in `keys` which explicitly returns a table, you're overwriting all the LazyVim default keymaps set by `keys` initially. That's why you have to define all the mappings again even if you only want to change one. I usually do 
```lua
`keys = {
  { "key_to_disable" = false },
  { "key_to_change" = "new_feature" },
  { "key_to_add" = "something" },
}
```
So, I just pass it a table instead of a function (this will get merged with the LazyVim defaults). Personally I haven't met the need to use a `values` function in `keys` yet. Only use it in `opts`.

dunix241:
Hey @dpetka2001,

Would you mind taking a look at the new commit? I realized a strict structure brings a headache if we want to separate the keymaps into different tables, so I want to allow a more dynamic structure.

dpetka2001:
I like it more with the less strict structure, but I don't think there's a need to group mappings by tables. It adds unnecessary complexity. The function `flattenTable` shouldn't be needed. Users can use comments to separate logically their mappings.

Also, I don't feel comfortable being asked for review as I'm just a simple user like you. Just ping me with my handle and I'll leave my personal opinion when needed.

Maintainer will be doing the review when he comes back and is available.

dunix241:
Understood thank you. As for grouping mappings its for this use case: a plugin keymaps table are kept in another file and users want to include those in the overrides table, so they can just require that list and place its in the overrides table. But I agree we can leave this to users to merge the tables so we don't need to flatten them. Anyways, thanks for your words.

folke:
Not a fan of this. This overly complicates a lot of things, with little benefit in return.
Closing...

PR #4664, opened 2024-10-29
fix(eslint): use correct option name to select workingDirectory

l3d00m:
## Description

The eslint default config used a wrong option name. It has to be `workingDirectory` instead of the plural that was currently used. Reference here: https://github.com/neovim/nvim-lspconfig/blob/master/doc/configs.md#eslint

There might be some implications of fixing this, as it changes the default option (which is `location` instead of `auto`), that has been implicitly used until now.

Documentation of that option here: https://marketplace.visualstudio.com/items?itemName=dbaeumer.vscode-eslint#settings-options

## Related Issue(s)

fixes #4663

## Checklist

### DIFF
--- a/lua/lazyvim/plugins/extras/linting/eslint.lua
+++ b/lua/lazyvim/plugins/extras/linting/eslint.lua
@@ -8,7 +8,7 @@ return {
         eslint = {
           settings = {
             -- helps eslint find the eslintrc when it's placed in a subfolder instead of the cwd root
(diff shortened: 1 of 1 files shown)


dpetka2001:
The Microsoft link that you quoted mentions `eslint.workingDirectories` though. Maybe `nvim-lspconfig` is the one that doesn't have it right? I don't use eslint, but maybe you would like to double-check before proceeding with this PR?

PR #4666, opened 2024-10-29
fix(lazygit): file history when cwd is outside the repo

r-darwish:
## Description

When using `<leader>gf` Lazygit is launched in the current directory rather than the root directory.

## Checklist


folke:
ty!

PR #4667, opened 2024-10-30
fix(extras): lazy-load telescope-terraform plugins

maddawik:
## Description

In the current terraform extra, both `cappyzawa/telescope-terraform.nvim` and
`ANGkeith/telescope-terraform-doc.nvim` are dependencies of `telescope.nvim`.
This should be reversed so that the telescope extensions only load when needed
(I added a `ft` trigger which I think makes the most sense), and not as a
dependency of `telescope.nvim`.

## Related Issue(s)

No related issues, just fixed this when the loading time of
`telescope-terraform-doc.nvim` in-particular was annoying (+~30ms startup).

## Checklist

### DIFF
--- a/lua/lazyvim/plugins/extras/lang/terraform.lua
+++ b/lua/lazyvim/plugins/extras/lang/terraform.lua
@@ -58,9 +58,10 @@ return {
   {
     "nvim-telescope/telescope.nvim",
     optional = true,
-    dependencies = {
+    specs = {
       {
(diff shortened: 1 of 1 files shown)


maddawik:
> That;s not correct. In the original spec, change `dependencies` to `specs` and also add those `ft` options. That should do what you want.

TIL about `specs`! Updated and re-based on `13.0.0` - tested it locally too 👍🏻 Thanks @folke !

folke:
ty!

PR #4673, opened 2024-10-30
fix(java): make `Run with Args` work for Java

### LINKED ISSUE #4672: bug: "Run with Args" does not work for Java
After installing Java and Dap extras, the command "Run with Args" does not work as expected: after inserting the arguments to pass to the Java program, you get

```
Debug adapter didn't respond. Either the adapter is slow (then wait and ignore this) or there is a problem with your adapter or `java` configuration. Check the logs for errors (:help dap.set_log_level)
```

As discussed here https://github.com/LazyVim/LazyVim/discussions/4604, this happens because the Java Dap adapter expects the `config.args' to be a string, while the default implementation in dap core, (see https://github.com/LazyVim/LazyVim/blob/v12.43.0/lua/lazyvim/plugins/extras/dap/core.lua#L34 and https://github.com/LazyVim/LazyVim/blob/v12.43.0/lua/lazyvim/plugins/extras/dap/core.lua#L2) passes the `config.args` as a table.

Note, increasing dap log level does not show anything, simply the Java Dap keeps hanging.

My proposal would be to redefine the "Run with Args" keybinding in Java to properly provide `config.args' as a string.

I'll submit a PR ASAP.

LorenzoBettini:
## Description

When Java and Dap extras are installed, the command "Run with Args" does not work as expected: after inserting the arguments to pass to the Java program, you get

```
Debug adapter didn't respond. Either the adapter is slow (then wait and ignore this) or there is a problem with your adapter or `java` configuration. Check the logs for errors (:help dap.set_log_level)
```

As discussed here https://github.com/LazyVim/LazyVim/discussions/4604, this happens because the Java Dap adapter expects the `config.args' to be a string, while the default implementation in dap core, (see https://github.com/LazyVim/LazyVim/blob/v12.43.0/lua/lazyvim/plugins/extras/dap/core.lua#L34 and https://github.com/LazyVim/LazyVim/blob/v12.43.0/lua/lazyvim/plugins/extras/dap/core.lua#L2) passes the `config.args` as a table.

My proposal would be to redefine the "Run with Args" keybinding in Java to properly provide `config.args' as a string.

## Related Issue(s)

Fixes #4672

## Checklist

### DIFF
--- a/lua/lazyvim/plugins/extras/lang/java.lua
+++ b/lua/lazyvim/plugins/extras/lang/java.lua
@@ -15,6 +15,18 @@ local function extend_or_override(config, custom, ...)
   return config
 end
 
+---@param config {args?:string|fun():string?}
+local function get_args(config)
+  local args = config.args and (config.args .. " ") or ""
+  config = vim.deepcopy(config)
+  ---@cast args string
+  config.args = function()
+    local new_args = vim.fn.input("Run with args: ", args) --[[@as string]]
+    return new_args --[[@as string]]
+  end
+  return config
+end
+
 return {
   recommended = function()
     return LazyVim.extras.wants({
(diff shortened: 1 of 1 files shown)


folke:
I'd rather have this changed in `dap.core`. You can probably check on `config.type` for jdtls and return a string in that case.

LorenzoBettini:
> I'd rather have this changed in `dap.core`. You can probably check on `config.type` for jdtls and return a string in that case.

@folke that was indeed my possible alternative solution but I thought doing that check in dap would not be appropriate ;)

I'll try to update or create another PR soon.

Thanks for the feedback.

PR #4680, opened 2024-11-01
feat(extras): blink

folke:
## Description

Extra to use [blink.cmp](https://github.com/Saghen/blink.cmp) instead of **nvim-cmp**.

## Todo

- [x] tokyonight suport
- [x] basic integration
- [ ] check / update all cmp sources
- [ ] copilot and others integration 
- [x] native lazydev source

## Limitations

There's no copilot source, so instead when enabling both blink and copilot:
- blink ghost text is disabled
- copilot suggestions are enabled
- use `<tab>` to navigate snippets and accept copilot completions 

## Related Issue(s)

- https://github.com/LazyVim/LazyVim/discussions/4679


folke:
Merging this is as. Can be refined later.

PR #4684, opened 2024-11-01
feat(extras): enable crates.nvim in-process lsp server

stefanboca:
## Description

Rather than manually handling hover and completions through nvim-cmp and nvim-lspconfig, enable the crates.nvim in-process lsp server. This also allows crates.nvim to provide code actions.

## Related Issue(s)

This change also removes the direct dependency on nvim-cmp, which should help with #4680

## Checklist


folke:
ty!

PR #4686, opened 2024-11-02
fix(java): make `Run with Args` work for Java in Dap

### LINKED ISSUE #4672: bug: "Run with Args" does not work for Java

LorenzoBettini:
## Description

When Java and Dap extras are installed, the command "Run with Args" does not work as expected: after inserting the arguments to pass to the Java program, you get

```
Debug adapter didn't respond. Either the adapter is slow (then wait and ignore this) or there is a problem with your adapter or `java` configuration. Check the logs for errors (:help dap.set_log_level)
```

As discussed here https://github.com/LazyVim/LazyVim/discussions/4604, this happens because the Java Dap adapter expects the `config.args' to be a string, while the default implementation in dap core, (see https://github.com/LazyVim/LazyVim/blob/v12.43.0/lua/lazyvim/plugins/extras/dap/core.lua#L34 and https://github.com/LazyVim/LazyVim/blob/v12.43.0/lua/lazyvim/plugins/extras/dap/core.lua#L2) passes the `config.args` as a table.

Alternative to https://github.com/LazyVim/LazyVim/pull/4673
as suggested in https://github.com/LazyVim/LazyVim/pull/4673#issuecomment-2452925460

## Related Issue(s)

Fixes #4672

## Checklist

### DIFF
--- a/lua/lazyvim/plugins/extras/dap/core.lua
+++ b/lua/lazyvim/plugins/extras/dap/core.lua
@@ -1,5 +1,20 @@
+---@param config {args?:string|fun():string?}
+local function get_args_as_string(config)
+  local args = type(config.args) == "function" and (config.args() or "") or config.args or ""
+  ---@cast args string
+  if string.len(args) > 0 then
+    args = args .. " "
+  end
+  config = vim.deepcopy(config)
+  config.args = function()
+    local new_args = vim.fn.input("Run with args: ", args) --[[@as string]]
+    return new_args --[[@as string]]
+  end
+  return config
+end
+
 ---@param config {args?:string[]|fun():string[]?}
-local function get_args(config)
+local function get_args_as_table(config)
   local args = type(config.args) == "function" and (config.args() or {}) or config.args or {}
   config = vim.deepcopy(config)
   ---@cast args string[]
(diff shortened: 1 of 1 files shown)


LorenzoBettini:
@folke as you suggested here https://github.com/LazyVim/LazyVim/pull/4673#issuecomment-2452925460 I implemented the check in `dap.core` directly.

I hope I did the right things in Lua, which I don't know much.

I tested this solution with a few Java projects, both with the configurations created automatically by Java Dap and with custom `.vscode/launch.json` configurations, possibly with existing `args`; everything works.

folke:
Why not just chaning the return statement in the `config.args` method depending on the `config.type`?
I don't really like duplicating code for no reason, that's why I suggested to make the changes in core.dap as opposed to having two different `get_arsg` functions.

LorenzoBettini:
@folke it's not just a matter of returning a string or a table: an existing Java configuration would have args as a string and the original code would throw an error when it prepares the string for the dialogue, because it would treat it as a table.

folke:
Can you check https://github.com/LazyVim/LazyVim/pull/4689 ?

LorenzoBettini:
> Can you check #4689 ?

@folke I added my review; thanks! :)

PR #4689, opened 2024-11-03
fix(java): jdtls run with args

folke:
## Description

Makes run with args work for jdtls

## Related Issue(s)
- Closes #4686
- Closes #4673

### DIFF
--- a/lua/lazyvim/plugins/extras/dap/core.lua
+++ b/lua/lazyvim/plugins/extras/dap/core.lua
@@ -1,11 +1,17 @@
----@param config {args?:string[]|fun():string[]?}
+---@param config {type?:string, args?:string[]|fun():string[]?}
 local function get_args(config)
-  local args = type(config.args) == "function" and (config.args() or {}) or config.args or {}
+  local args = type(config.args) == "function" and (config.args() or {}) or config.args or {} --[[@as string[] | string ]]
+  local args_str = type(args) == "table" and table.concat(args, " ") or args --[[@as string]]
+
   config = vim.deepcopy(config)
   ---@cast args string[]
   config.args = function()
-    local new_args = vim.fn.input("Run with args: ", table.concat(args, " ")) --[[@as string]]
-    return vim.split(vim.fn.expand(new_args) --[[@as string]], " ")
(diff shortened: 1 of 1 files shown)


lua/lazyvim/plugins/extras/dap/core.lua
```diff
@@ -1,11 +1,17 @@
...
   ---@cast args string[]
   config.args = function()
-    local new_args = vim.fn.input("Run with args: ", table.concat(args, " ")) --[[@as string]]
-    return vim.split(vim.fn.expand(new_args) --[[@as string]], " ")
+    local new_args = vim.fn.expand(vim.fn.input("Run with args: ", args_str))
+    if config.type and config.type == "java" then
+      ---@diagnostic disable-next-line: return-type-mismatch
+      return new_args
+    end
+    return vim.split(new_args, " ")
```
LorenzoBettini, line 14: Before this line, as you did for the previous return statement, you might want to add

```lua
---@diagnostic disable-next-line: param-type-mismatch
```

To avoid the warning

LorenzoBettini (approved):
Besides the warning that can be disabled as said in the comment, there's a small unwanted behavior: if the existing args is a string and contains some arguments, the dialog shows the arguments and places the cursor right after the last existing argument, without a space (as it happens for the table case, and as I happened in my PRs). But that's just nitpicking, so, please feel to ignore that, I can live with that. ;)

I tested your PR and it works both for existing arguments and no arguments in Java and it works!

As for me, you can go on merging.

Many thanks for taking care of fixing the issue :)

folke:
Thanks for the review!

LorenzoBettini:
@folke This didn't make it into "v12.44.0", did it? I see you also released "v12.44.1" but I don't see that in https://github.com/LazyVim/LazyVim/releases

spacealex69:
> @folke This didn't make it into "v12.44.0", did it? I see you also released "v12.44.1" but I don't see that in https://github.com/LazyVim/LazyVim/releases

@LorenzoBettini  Probably will ship with https://github.com/LazyVim/LazyVim/pull/4694

PR #4693, opened 2024-11-03
fix(lazyfile): account for treesitter disable highlights

### LINKED ISSUE #4687: bug: Inconsistent `nvim-treesitter` Initialization Behavior

### Neovim version (nvim -v)

NVIM v0.10.2

### Operating system/version

ArchLinux

### Describe the bug

Hello,

I've noticed an inconsistency in the time/Event when `nvim-treesitter` initializes when opening `.tex` files.

- When I simply open Neovim and then open a .tex file (e.g., by typing :edit main.tex), the nvim-treesitter plugin initializes with the VeryLazy event.
- However, if I directly open a .tex file from the command line (e.g., nvim main.tex), the nvim-treesitter plugin initializes with the Start event.

### Steps To Reproduce

1. Prepare a minimal tex file `main.tex` like 
```latex
\documentclass{article}
\begin{document}
text $\alpha $ 
\end{document}
```
2. Install LazyVim and run `:TSInstall latex`
3. try The first way `nvim` then `:e main.tex` and `:Lazy` (I saw `VeryLazy` for `nvim-treesitter`), and the second way run `nvim main.tex` see `:Lazy` (I saw `nvim-treesitter` start at `Start` and vimtex give warning `Syntax highlighting is controlled by Treesitter!`)
4. If enable extra `lang.tex` , then the second way can't will raise a warning from vimtex

### Expected Behavior

If I do it in the second way the vimtex warning show up, the conceal for the latex symbol don't show up. What I notice is that `nvim-treesitter` plugin do not initialize consistently. Maybe that is not the reason. I hope consistent behavior for concealing, regardless of how I open the file.

### Repro

require("lazy.minit").repro({
  spec = {
    -- add any other plugins here
-- not sure how to produce it with `nvim -u repro.lua` I just paste the config here
  {
    "lervag/vimtex",
    lazy = true, -- lazy-loading will disable inverse search
    ft={"tex"},
    config = function()
      vim.g.vimtex_mappings_disable = { ["n"] = { "K" } } -- disable `K` as it conflicts with LSP hover
      vim.g.vimtex_quickfix_method = vim.fn.executable("pplatex") == 1 and "pplatex" or "latexlog"
    end,
    keys = {
      { "<localLeader>l", "", desc = "+vimtex" },
    },
    dependencies = {
      {
        opts = {
          ensure_installed = { "latex", "bibtex" },
          highlight = {
            enable = true,
            disable = { "latex" },
            additional_vim_regex_highlighting = { "latex" }
          },
        },
      },
    },
  },

  -- Correctly setup lspconfig for LaTeX 🚀
  {
    "neovim/nvim-lspconfig",
    optional = true,
    opts = {
      servers = {
        texlab = {
          keys = {
            { "<Leader>K", "<plug>(vimtex-doc-package)", desc = "Vimtex Docs", silent = true },
          },
        },
      },
    },
  },

  },
})

dpetka2001:
## Description
In `util.plugin.lazyfile` the function always starts `vim.treesitter.start()` for treesitter highlighting, but this causes problems with the `tex` Extra which expects the highlighting for `latex` to be disabled.

Solution: In `tex` Extra define a global `vim.g.treesitter_disable_highlight` table that will include `latex` and then in `lazyfile` check if `lang` exists in the aforementioned table, so that the protected call to `vim.treesitter.start` won't execute and instead we get legacy syntax.

## Related Issue(s)
Fixes #4687.

## Checklist

### DIFF
--- a/lua/lazyvim/plugins/extras/lang/tex.lua
+++ b/lua/lazyvim/plugins/extras/lang/tex.lua
@@ -9,6 +9,13 @@ return {
   -- Add BibTeX/LaTeX to treesitter
   {
     "nvim-treesitter/nvim-treesitter",
+    init = function()
+      if
+        vim.fn.argc() > 0 and vim.fn.isdirectory(vim.fn.argv(0) --[[@as string]]) ~= 1
+      then
+        vim.g.treesitter_disable_highlight = { "latex" }
+      end
+    end,
     opts = function(_, opts)
       opts.highlight = opts.highlight or {}
       if type(opts.ensure_installed) == "table" then
--- a/lua/lazyvim/plugins/extras/lang/tex.lua
+++ b/lua/lazyvim/plugins/extras/lang/tex.lua
@@ -13,7 +13,7 @@ return {
       if
         vim.fn.argc() > 0 and vim.fn.isdirectory(vim.fn.argv(0) --[[@as string]]) ~= 1
       then
-        vim.g.treesitter_disable_highlight = { "latex" }
+        vim.g.treesitter_syntax_legacy = { "latex" }
       end
     end,
     opts = function(_, opts)
--- a/lua/lazyvim/util/plugin.lua
+++ b/lua/lazyvim/util/plugin.lua
@@ -86,10 +86,10 @@ function M.lazy_file()
       if ft then
         -- Add treesitter highlights and fallback to syntax
         local lang = vim.treesitter.language.get_lang(ft)
-        if vim.tbl_contains(vim.g.treesitter_disable_highlight or {}, lang) then
-          lang = nil
-        end
         if not (lang and pcall(vim.treesitter.start, event.buf, lang)) then
+          vim.bo[event.buf].syntax = ft
+        end
+        if vim.tbl_contains(vim.g.treesitter_syntax_legacy or {}, lang) then
           vim.bo[event.buf].syntax = ft
         end
 
--- a/lua/lazyvim/util/plugin.lua
+++ b/lua/lazyvim/util/plugin.lua
@@ -86,6 +86,9 @@ function M.lazy_file()
       if ft then
         -- Add treesitter highlights and fallback to syntax
         local lang = vim.treesitter.language.get_lang(ft)
+        if vim.tbl_contains(vim.g.treesitter_disable_highlight or {}, lang) then
+          lang = nil
+        end
         if not (lang and pcall(vim.treesitter.start, event.buf, lang)) then
           vim.bo[event.buf].syntax = ft
         end


ChasingZenith:
Thank you for your contribution.  I am new to `lazy.nvim` and I don't quite get the code. But if I understand the description right. You will disable `treesitter` for tex. However, that may not be expected sometimes.

Some lsp need treesitter, and `vimtex` needs syntax. Sometimes we need either and sometimes both.

I think we should 
1. use `enable` and `disable` to control whether `treesitter` is on.
2. and use [additional_vim_regex_highlighting in nvim-treesitter configuration](https://github.com/nvim-treesitter/nvim-treesitter?tab=readme-ov-file#highlight) to control whether `syntax` should be turned on.

In the case both are needed, we need to `:set syntax=tex` AFTER calling `lua vim.treesitter.start()` as stated in [the doc](http://neovim.io/doc/user/treesitter.html#vim.treesitter.start()).

Can you also check whether [function `nvim-treesitter.highlight.attach` in nvim-treesitter](https://github.com/nvim-treesitter/nvim-treesitter/blob/01a92da4389b5346b4049c5b42256c9d84526f04/lua/nvim-treesitter/highlight.lua#L18) is useful? I think the function can give expected result, but I don't know how to call it in LazyVim.

dpetka2001:
In this specific case of `lazyfile`, `nvim-treesitter` is not used so we don't have access to its fields yet as it hasn't loaded. Instead `vim.treesitter.start()` invokes directly the Neovim builtin Treesitter to enable highlighting for the buffer. 

I'm not that familiar with Treesitter's internals myself, so let's just better wait for maintainer's input. There might indeed be a better resolution for this.

dpetka2001:
We could change instead of `vim.g.treesitter_disable_highlight` to `vim.g.treesitter_legacy_syntax` and make the check on that and if `lang` is present in the table then do additionally `syntax = "on"`. Would that be a better solution?

folke:
That `additional_vim_regex...` is pure evil and leads to tons of issues, so definitely a no go on that.
It's either treesitter or syntax, but never both at the same time.

What should indeed be added to the lazy_file function is to disable the lazy_file behavior for some treesitter langs.

This method only pre-starts treesitter before loading all plugins, so if it doesnt do this for some langs, no big deal.

Either way, I'm working on a new plugin that contains a collection of tiny QOL plugins based on LazyVim functionality. 
LazyVim will also use the new plugin instead for those things and the lazy file thing is part of that. I've also already included a fix where you can easily configure treesitter langs to be excluded.

More info soon.

dpetka2001:
Closing in favor of the new implementation that @folke is working on, which will definitely be better as my knowledge is lacking in many sections.

Issue #4704, opened 2024-11-04
bug: eslint cannot be disabled with `vim.g.root_lsp_ignore`


guillaumeLamanda:
### Did you check docs and existing issues?

- [x] I have read all the LazyVim docs
- [x] I have updated the plugin to the latest version before submitting this issue
- [x] I have searched the existing issues of LazyVim
- [x] I have searched the existing issues of plugins related to this issue

### Neovim version (nvim -v)

NVIM v0.10.1

### Operating system/version

MacOS sequolia

### Describe the bug

I would like to disable eslint lsp per project using the newly introduced `vim.g.root_lsp_ignore`. (https://github.com/LazyVim/LazyVim/pull/4332)

I guess, because it's a formatter that being registered, it's is not checked by the actual implementation.

### Steps To Reproduce

1. add the `eslint` extra
2. go to a project containing a `js` or `ts` file
3. create a `.nvim.lua` file and add `vim.g.root_lsp_ignore = { "eslint" }
4. go in the ts file and run `LspInfo`

### Expected Behavior

`eslint` to not be part of active lsp 

### Repro
```lua
vim.env.LAZY_STDPATH = ".repro"
load(vim.fn.system("curl -s https://raw.githubusercontent.com/folke/lazy.nvim/main/bootstrap.lua"))()

require("lazy.minit").repro({
  spec = {
    { "LazyVim/LazyVim", import = "lazyvim.plugins" },
    -- add any other plugins here
	{ import = "lazyvim.plugins.extras.linting.eslint" },
  },
})

vim.g.root_lsp_ignore = { "eslint" }
```

PR #4705, opened 2024-11-04
fix(eslint): check if disabled before register

guillaumeLamanda:
## Description

This PR checks if eslint has been disabled by `vim.g.root_lsp_ignore`

## Related Issue(s)

<https://github.com/LazyVim/LazyVim/issues/4704>

## Checklist


folke:
That;s not what that setting is for.

You can disable an lsp server by setting enabled= false for the lspconfig opts of that server

Issue #4702, opened 2024-11-04
feature(blink): add support for AI completion extras


folke:
### Did you check the docs?

- [x] I have read all the LazyVim docs
- [x] This is not a request for a new or existing extra (see above)

### Is your feature request related to a problem? Please describe.

Right now, only the copilot extra has been updated to work with blink.

With blink, we no longer show AI suggestions as completions, since I couldn't get that to work with `blink-compat`, but we use `<tab>` to accept a suggestion.

### Describe the solution you'd like

Implement a similar approach for the other AI extras.

### Describe alternatives you've considered

none

Conarius:
There's also blink-cmp-compat afaik to make everything for nvim-cmp compatible with blink, so you can replace nvim-cmp with blink entirely.

OneOfOne:
I can confirm it's not working for copilot, only thing keeping me from switching.

dpetka2001:
@folke I tried to implement for `codeium.nvim` but couldn't make it work. I did
```lua
      keymap = {
        ["<Tab>"] = {
          function(cmp)
            if cmp.is_in_snippet() then
              return cmp.accept()
            elseif require("codeium.virtual_text").get_current_completion_item() then
              LazyVim.create_undo()
              require("codeium.virtual_text").accept()
              return true
            else
              return cmp.select_and_accept()
            end
          end,
          "snippet_forward",
          "fallback",
        },
      },
```
but it doesn't complete anything. I added a `print` just after `LazyVim.create_undo` and it gets printed. So, the condition about a completion item existing is true. I even cloned `codeium.nvim` and added a `print` inside `codeium.virtual_text.accept()` and that gets printed as well (so the function gets called) but I can't exactly understand why it's not working. I'm not familiar with either codebase and my lack of knowledge makes it harder for me to understand.

On the other hand if I add a keymap into `keymaps.lua` like the following
```lua
vim.keymap.set("i", "<Tab>", function()
  return require("codeium.virtual_text").accept()
end, { silent = true, expr = true, nowait = true })
```
It works correctly. I also tried the same in `keys` section of the plugin and works correctly there as well.

[...]

PR #4716, opened 2024-11-05
fix(copilot-chat): setup cmp conditionally

joshmedeski:
## Description

Check for cmp before setting up copilot chat.

## Related Issue(s)

For https://github.com/LazyVim/LazyVim/issues/4702

I would like to set up the cmp compatability later if possible, but this is a quick patch to allow copilot chat to work without cmp (for now).

## Checklist


folke:
ty

PR #4720, opened 2024-11-07
chore(telescope): Use gmake instead of make on non-linux system

lakako:
## Description

On non-Linux systems(eg freebsd/netbsd), make is not gmake. Therefore, gmake must be used on non-Linux systems.

## Related Issue(s)

none

## Screenshots

N/A

## Checklist


folke:
This doesn't make sense. `make` exists for those systems as well.
Your implementation woul've also tried using `gmake` on Windows btw

PR #4734, opened 2024-11-08
fix(extras): fix git recommended ft

kevintraver:
Fix git extras recommended filetype to use `LazyVim.extras.wants`. This fixes an issue where the `ft` table does not get read correctly.


folke:
Where did you see that fail? It works as it should for me.
recommended can be a function that returns a boolean, or a table that gets passed to wants

kevintraver:
It was not working with with [telescope-lazy-plugins.nvim](https://github.com/polirritmico/telescope-lazy-plugins.nvim):

https://github.com/polirritmico/telescope-lazy-plugins.nvim/issues/16

It was fixed and is working now.

PR #4749, opened 2024-11-09
feat(extras): add Dart language

stanislav-zeman:
## Description

This PR add Dart language support to **extras** section.

## Testing

I checked that:
- the LSP correctly loads on Dart files and provides suggestions and highlights errors
- `treesitter` parses the Dart language tree and highlights the code
- `conform` correctly formats Dart files using the built-in formatter
- `neotest` is capable of running Dart tests

## Checklist

### DIFF
--- a/lua/lazyvim/plugins/extras/lang/dart.lua
+++ b/lua/lazyvim/plugins/extras/lang/dart.lua
@@ -1,0 +1,45 @@
+return {
+  recommended = function()
+    return LazyVim.extras.wants({
+      ft = "dart",
+      root = { "pubspec.yaml" },
+    })
+  end,
+  {
+    "neovim/nvim-lspconfig",
+    opts = {
+      servers = {
+        dartls = {},
+      },
+    },
+  },
+  {
+    "nvim-treesitter/nvim-treesitter",
+    opts = { ensure_installed = { "dart" } },
+  },
+  {
+    "stevearc/conform.nvim",
+    opts = {
+      formatters = {
+        dart_format = {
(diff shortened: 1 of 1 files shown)


Conarius:
Maybe flutter-tools.nvim might be a good addition for it

stanislav-zeman:
Yeah, I am well aware of [flutter-tools.nvim](https://github.com/nvim-flutter/flutter-tools.nvim) and I got to admit, I personally use it myself. However, I decided not to add it as I wanted to keep the config minimal with original tooling and I felt that `flutter-tools` does way too many things. 

Additionally, if I had imported `flutter-tools` I would have to drop the `nvim-lspconfig` as `flutter-tools` [does not depend on it](https://github.com/nvim-flutter/flutter-tools.nvim?tab=readme-ov-file#warning) and configuring both might have some unwanted consequences.

I also somewhat felt like it adding `flutter-tools` contradicts with the CONTRIBUTING guidelines (LSP wrapper packages) but I might have misinterpreted it.

Additionally, for example, for Go, there exists similar package [ray-x/go](https://github.com/ray-x/go.nvim) that does basically the same things as `flutter-tools` for Dart/Flutter but for Go. This package is not defined in Go extras config and is also kept pretty minimal instead.

That was my reasoning for not adding `flutter-tools`. If more people would like to add `flutter-tools`, we definitely could add it but it might require more research into the package and its dependencies.

PS: I also tried adding `dap` adapter but I could not make it to work. That could be an another possible addition.

PR #4752, opened 2024-11-10
feat(ai): better completion/suggestions of AI engines

### LINKED ISSUE #4702: feature(blink): add support for AI completion extras

folke:
## Description

The whole completion / snippets / AI is very tricky:
- multiple snippet engines
- native snippets on > 0.11 set their own keymaps, but not on 0.10
- multiple completion engines, like `nvim-cmp` and `blink.cmp`
- multiple ai completion engines that have a different API
- user's preference of showing ai suggestions as completion or not
- none of the ai completion engines currently set undo points, which is bad

Solution:
- [x] added `LazyVim.cmp.actions`, where snippet engines and ai engines can register their action.
- [x] an action returns `true` if it succeeded, or `false|nil` otherwise
- [x] in a completion engine, we then try running multiple actions and use the fallback if needed
- [x] so `<tab>` runs `{"snippet_forward", "ai_accept", "fallback"}`
- [x] added `vim.g.ai_cmp`. When `true` we try to integrate the AI source in the completion engine.
- [x] when `false`, `<tab>` should be used to insert the AI suggestion
- [x] when `false`, the completion engine's ghost text is disabled
- [x] luasnip support for blink (only works with blink `main`)
- [x] create undo points when accepting AI suggestions 

## Test Matrix

| completion   | snippets     | ai          | ai_cmp | tested? |
|--------------|--------------|-------------|--------|---------|
| nvim-cmp     | native       | copilot     | true   | ✅      |
| nvim-cmp     | native       | copilot     | false  | ✅      |
| nvim-cmp     | native       | codeium     | true   | ✅      |
| nvim-cmp     | native       | codeium     | false  | ✅      |
| nvim-cmp     | luasnip      | copilot     | true   | ✅      |
| nvim-cmp     | luasnip      | copilot     | false  | ✅      |
| nvim-cmp     | luasnip      | codeium     | true   | ✅      |
| nvim-cmp     | luasnip      | codeium     | false  | ✅      |
| blink.cmp    | native       | copilot     | true   | ✅      |
| blink.cmp    | native       | copilot     | false  | ✅      |
| blink.cmp    | native       | codeium     | true   | ✅      |
| blink.cmp    | native       | codeium     | false  | ✅      |
| blink.cmp    | luasnip      | copilot     | true   | ✅      |
| blink.cmp    | luasnip      | copilot     | false  | ✅      |
| blink.cmp    | luasnip      | codeium     | true   | ✅      |
| blink.cmp    | luasnip      | codeium     | false  | ✅      |

## Related Issue(s)

- [ ] Closes #4702

## Checklist

### DIFF
--- a/lua/lazyvim/plugins/extras/ai/codeium.lua
+++ b/lua/lazyvim/plugins/extras/ai/codeium.lua
@@ -1,17 +1,31 @@
 return {
 
+  -- codeium
+  {
+    "Exafunction/codeium.nvim",
+    cmd = "Codeium",
+    build = ":Codeium Auth",
+    opts = {
+      enable_cmp_source = vim.g.ai_cmp,
+      virtual_text = {
+        enabled = not vim.g.ai_cmp,
+        accept_fallback = vim.g.ai_suggest_accept,
+        key_bindings = {
+          accept = vim.g.ai_suggest_accept,
+          accept_word = vim.g.ai_suggest_accept_word,
+          accept_line = vim.g.ai_suggest_accept_line,
+          next = vim.g.ai_suggest_next,
+          prev = vim.g.ai_suggest_prev,
+          clear = vim.g.ai_suggest_clear,
+        },
+      },
+    },
+  },
   -- codeium cmp source
   {
     "nvim-cmp",
-    dependencies = {
-      -- codeium
-      {
-        "Exafunction/codeium.nvim",
-        cmd = "Codeium",
-        build = ":Codeium Auth",
-        opts = {},
-      },
-    },
+    optional = true,
+    dependencies = { "codeium.nvim" },
     ---@param opts cmp.ConfigSchema
     opts = function(_, opts)
       table.insert(opts.sources, 1, {
@@ -29,5 +43,19 @@ return {
     opts = function(_, opts)
       table.insert(opts.sections.lualine_x, 2, LazyVim.lualine.cmp_source("codeium"))
     end,
+  },
+
+  {
+    "saghen/blink.cmp",
+    optional = true,
+    opts = {
+      sources = {
+        compat = vim.g.ai_cmp and { "codeium" } or nil,
+      },
+    },
+    dependencies = {
+      "codeium.nvim",
+      vim.g.ai_cmp and { "saghen/blink.compat" } or {},
+    },
   },
 }
--- a/lua/lazyvim/plugins/extras/ai/codeium.lua
+++ b/lua/lazyvim/plugins/extras/ai/codeium.lua
@@ -9,24 +9,34 @@ return {
       enable_cmp_source = vim.g.ai_cmp,
       virtual_text = {
         enabled = not vim.g.ai_cmp,
-        accept_fallback = vim.g.ai_suggest_accept,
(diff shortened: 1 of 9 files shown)


lua/lazyvim/plugins/extras/ai/codeium.lua
```diff
@@ -30,4 +44,18 @@ return {
...
+    "saghen/blink.cmp",
+    optional = true,
+    opts = {
+      sources = {
+        compat = vim.g.ai_cmp and { "codeium" } or nil,
+      },
+    },
+    dependencies = {
+      "codeium.nvim",
+      vim.g.ai_cmp and { "saghen/blink.compat" } or {},
```
dpetka2001, line 58: @folke Please take a look at [this comment](https://github.com/Saghen/blink.compat/pull/6#discussion_r1828128222). The maintainer of `blink-compat` says that it should not be in `blink.cmp`'s dependencies. And when I tested it, I also put it as a plugin spec on its own, not a dependency. Not sure if it actually affects anything, just thought I should let you know about this.

folke: It doesn't really matter since it would be loaded directly anyway.
`blink-compat` is a library, so adding it there doesn't really load anything.

dpetka2001: Whatever you think is best. Just wanted to make you aware of this and let you be the judge, since you know way better compared to my novice knowledge.

lua/lazyvim/plugins/extras/ai/copilot.lua
```diff
@@ -69,42 +83,38 @@ return {
             copilot_cmp._on_insert_enter({})
           end, "copilot")
         end,
+        specs = {
+          {
+            "zbirenbaum/copilot.lua",
+            opts = { suggestion = { enabled = false } },
+          },
```
dpetka2001, line 90: Is there a reason for this, since it's also defined in the `dependencies` on line 75? From what I can understand `specs` just don't load the plugin spec, but it's already listed in `dependencies`. So, I'm just wondering about the reason of this spec existing in both places.

folke: The code changed quite a bit in the meantime, so this is probably no longer relevant :)

folke:
The PR should be ready.
It's been a real pain in handling all the border cases...

dpetka2001:
I tested the `codeium` part of this PR along with `blink.cmp` and from a first glance everything seems to be working as expected.

nadunindunil:
Hi @folke 
I don't get auto-suggestion anymore after updating the lazyVim to the latest version.
Can this be a reason? I didn't configure anything, I was using existing setup that using nvim-cmp.
Now I dont't get copilot suggestions 🤔 
Please let me know a way to fix this, Thank you.

PR #4801, opened 2024-11-15
feat(format): adds mode for autoformat to apply either to the file or …

aubrich:
## Description

This adds two modes for the autoformat:
* `file` : apply the format function to the buffer file
* `changes`: applies the format on each git hunk to format only the modified lines

## Checklist

- [X ] I've read the [CONTRIBUTING](https://github.com/LazyVim/LazyVim/blob/main/CONTRIBUTING.md) guidelines.

### DIFF
--- a/lua/lazyvim/config/options.lua
+++ b/lua/lazyvim/config/options.lua
@@ -4,6 +4,13 @@ vim.g.maplocalleader = "\\"
 
 -- LazyVim auto format
 vim.g.autoformat = true
+-- LazyVim auto format mode (files | changes)
+-- if set to changes only format actual changes based on git hunks
+-- otherwise format all the file
+vim.g.formatmode = "file"
+--- LazyVim auto format changes only mode will be ignored
+--- for these filetypes and fallback to file
+vim.g.format_range_exclude_ft = { "lua" }
 
 -- Snacks animations
 -- Set to `false` to globally disable all snacks animations
--- a/lua/lazyvim/config/options.lua
+++ b/lua/lazyvim/config/options.lua
@@ -4,7 +4,7 @@ vim.g.maplocalleader = "\\"
 
 -- LazyVim auto format
 vim.g.autoformat = true
--- LazyVim auto format mode (files | changes)
+-- LazyVim auto format mode (file | changes)
 -- if set to changes only format actual changes based on git hunks
 -- otherwise format all the file
 vim.g.formatmode = "file"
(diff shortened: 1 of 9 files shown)


lua/lazyvim/config/options.lua
```diff
@@ -4,6 +4,13 @@ vim.g.maplocalleader = "\\"
 
 -- LazyVim auto format
 vim.g.autoformat = true
+-- LazyVim auto format mode (files | changes)
+-- if set to changes only format actual changes based on git hunks
+-- otherwise format all the file
+vim.g.formatmode = "file"
+--- LazyVim auto format changes only mode will be ignored
+--- for these filetypes and fallback to file
+vim.g.format_range_exclude_ft = { "lua" }
```
dpetka2001, line 13: Do these need to be set specifically, since in `/util/format.lua` you do 
```lua
local gmode = (vim.g.formatmode == nil and "file") or vim.g.formatmode
```
and 
```lua
if vim.tbl_contains(vim.g.format_range_exclude_ft or M.format_range_exclude_ft, vim.bo[buf].filetype) then
    bmode = "forced file"
end
```
respectively? Also for the second snippet maybe
```lua
vim.tbl_contains(LazyVim.dedup(vim.list_extend(vim.g.format_range_exclude_ft or {}, M.format_range_exclude_ft)), vim.bo[buf].filetype)
```
would be more appropriate, so that it neither contains duplicate filetypes nor the `vim.g.format_range_exclude_ft` overwrites the `M.format_range_exclude_ft` in case `vim.g.format_range_exclude_ft` does not contain the `lua` filetype. Whereas if you use `vim.list_extend` it will just extend the filetype patterns. That goes for both `M.register` and `M.format` functions.

dpetka2001: Yes, you are right about the overwritable thing. I didn't think it through. What I proposed would just ensure that the ft `lua` would always be included, which might be not desirable by some users. Feel free to mark this as resolved.

lua/lazyvim/config/options.lua
```diff
@@ -4,6 +4,13 @@ vim.g.maplocalleader = "\\"
 
 -- LazyVim auto format
 vim.g.autoformat = true
+-- LazyVim auto format mode (files | changes)
```
dpetka2001, line 7:
```suggestion
-- LazyVim auto format mode (file | changes)
```
Just a small typo so users don't get confused.

folke:
The only thing that's blocking is me having time to go over this PR :)
I'm busy with a bunch of other things I want to finish first.

folke:
I' not interested in adding this. Would make more sense to add this to conform.nvim directly.

gmnz:
Pity, I was looking forward to this feature. @folke Any plans on making it possible to use extras from external repos? So people could offer their own extras for LazyVim from their own repos, which would be, of course, oficially unsupported and somehow manually added to the config, not through :LazyExtras.

PR #4802, opened 2024-11-15
feat(copilot-chat): remove deprecated options

deathbeam:
- Diagnostic help was deprecated and now just proxies to prompt actions (by default diagnostics are auto included in all selections so this feature was no longer necessary)
- Selecton is now visual || buffer by default so the custom selection config is no longer needed too


folke:
ty!

PR #4804, opened 2024-11-15
feat(cmp): replace nvim-cmp with magazine.nvim

OneOfOne:
## Description

Use iguanacucumber/magazine.nvim instead of hrsh7th/nvim-cmp for better performance

- [X ] I've read the [CONTRIBUTING](https://github.com/LazyVim/LazyVim/blob/main/CONTRIBUTING.md) guidelines.


OneOfOne:
@folke ideas?

OneOfOne:
@folke i can rebase if you're interested in this.

igorgue:
@OneOfOne I wonder if there's a different way to do this, since from the look of it any custom configuration to `"hrsh7th/nvim-cmp"` from end users would be lost, I wonder if this can be fixed by changing https://github.com/LazyVim/LazyVim/blob/main/lua/lazyvim/plugins/xtras.lua and add it with a higher priority?

OneOfOne:
Setting name=nvim-cmp fixes that issue, at least for me. 

But I'm open to suggestions. 

I have been using that setup for a few months now, I'm still having issues with blink sorting so I keep going back to it.

dpetka2001:
Why would users' custom configuration be lost? When I tried it, I just configured it exactly like `nvim-cmp`. It's `nvim-cmp` itself under the hood for the most part if I understand correctly (I say "for the most part" because I don't know if anything new has been provided by the new maintainer and I don't mean just backlog from PRs that `nvim-cmp` had accumulated over time, so I'm ignorant in that regard), just a different name.

And with defining `name = nvim-cmp` it should be just plug and play by `lazy.nvim`. Just try it out in your personal configuration.

folke:
This should be an extra instead of overwriting all refs to nvim-cmp.
Haven't tested this, but doing something like:

```lua
{ "hrsh7th/nvim-cmp", url="magazine url"}
```

And the same for the other changes, should work.

igorgue:
Yeah I think this works for me:

```lua
return {
  {
    "hrsh7th/nvim-cmp",
    url = "iguanacucumber/magazine.nvim",
    dependencies = {
      { "iguanacucumber/mag-cmdline", name = "cmp-cmdline" },
      { "iguanacucumber/mag-nvim-lsp", name = "cmp-nvim-lsp", opts = {} },
      { "iguanacucumber/mag-nvim-lua", name = "cmp-nvim-lua" },
      { "iguanacucumber/mag-buffer", name = "cmp-buffer" },
      { "iguanacucumber/mag-cmdline", name = "cmp-cmdline" },
      { "https://codeberg.org/FelipeLema/cmp-async-path", name = "cmp-path" },
    },
  },
}
```

iguanacucumber:
Hi, magazine.nvim author here, if you want to use magazine.nvim, please pull from the [releases](https://github.com/iguanacucumber/magazine.nvim/releases) to have a more stable experience since it's a "beta" nvim-cmp

So the config should look like this:
```lua
return {
  {
    "hrsh7th/nvim-cmp",
    url = "iguanacucumber/magazine.nvim",
    version = "*",
    dependencies = {
      { "iguanacucumber/mag-cmdline", name = "cmp-cmdline", version = "*" },
      { "iguanacucumber/mag-nvim-lsp", name = "cmp-nvim-lsp", opts = {}, version = "*" },
      { "iguanacucumber/mag-nvim-lua", name = "cmp-nvim-lua", version = "*" },
      { "iguanacucumber/mag-buffer", name = "cmp-buffer", version = "*" },
      { "https://codeberg.org/FelipeLema/cmp-async-path", name = "cmp-path" },
    },
  },
}
```

OneOfOne:
I have since moved to blink, however if @folke is interested, I can rebase it against master.

80avin:
Looks like https://github.com/iguanacucumber/magazine.nvim is archived now.
Is this PR still valid ?

PR #4811, opened 2024-11-16
fix(keymaps): fix which-key description of `<c-_>`

FedericoStra:
## Description

Since `<C-/>` and `<C-_>` are mapped to the same command, I think it's better to use the same description.
Moreover, I don't see how `ignore_which_key` is appropriate here, since the description never shows up automatically (because it's a single character).
Putting the new description however improves the result shown in the keymaps with `<space>sk` and allows to discover it by searching for "terminal".

## Checklist

### DIFF
--- a/lua/lazyvim/config/keymaps.lua
+++ b/lua/lazyvim/config/keymaps.lua
@@ -159,11 +159,11 @@ map("n", "<leader>L", function() LazyVim.news.changelog() end, { desc = "LazyVim
 map("n", "<leader>fT", function() Snacks.terminal() end, { desc = "Terminal (cwd)" })
 map("n", "<leader>ft", function() Snacks.terminal(nil, { cwd = LazyVim.root() }) end, { desc = "Terminal (Root Dir)" })
 map("n", "<c-/>",      function() Snacks.terminal(nil, { cwd = LazyVim.root() }) end, { desc = "Terminal (Root Dir)" })
-map("n", "<c-_>",      function() Snacks.terminal(nil, { cwd = LazyVim.root() }) end, { desc = "which_key_ignore" })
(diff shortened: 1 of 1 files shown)


FedericoStra:
Also, now that I think of it, I don't see the point of having both `<c-_>` and `<c-/>` mapped to the same command. Maybe one of the two should open the terminal in the current working directory instead of the project root, reflecting the difference between `<leader>ft` and `<leader>fT`?

folke:
You misunderstand what this is for.
Some terminals trigger `<c-_>` when you actually type `<c-/>`.
That's why we have two mappings both doing the same thing.
That's also why that one mapping is hidden.
It's effectively the same mapping.

dpetka2001:
Just open a tmux session and in insert mode do `<C-v><C-/>` and you'll see that it interprets it as `^_` which is equivalent to 
`<C-_>`. Other terminal emulators have the same behavior. 

Users won't know that they're pressing `<C-_>`. So, searching for the mapping in `<leader>sk` will give them the correct mapping, that is they will still press `<C-/>` on their keyboard, it's just that their terminal emulator will interpret it as `<C-_>`.

FedericoStra:
Oh I see, well then sorry for having disturbed you unnecessarily.
I'm not expert about all these weird behaviours of terminals and I thought it was a leftover typo or something.

As a matter of fact, now I quickly experimented mapping
```vim
nmap <c-_> :split<cr>
nmap <c-/> :vert split<cr>
```
in `neovim` running in a few terminals and:

- `neovide` distinguishes between them;
- `gnome-terminal` and `terminator` do the same as `<c-_>` with both;
- `alacritty`, `kitty` and `foot` do no receive either of the commands.

This is hell...

PR #4818, opened 2024-11-17
fix(keymaps): `gitbrowse` also in Visual mode

dpetka2001:
## Description
If https://github.com/folke/snacks.nvim/pull/89 gets accepted, then we should also update the keymap for Visual mode

## Related Issue(s)
None

## Checklist


folke:
thanks!

PR #4822, opened 2024-11-18
feat(copilot-chat): remove call to nvim-cmp integration

deathbeam:
nvim-cmp integration was removed in favour of custom autocomplete (it was pointless trying to support all the new completion plugins when its 15 lines to implement something plugin specific)

### DIFF
--- a/lua/lazyvim/plugins/extras/ai/copilot-chat.lua
+++ b/lua/lazyvim/plugins/extras/ai/copilot-chat.lua
@@ -66,9 +66,6 @@ return {
     },
     config = function(_, opts)
       local chat = require("CopilotChat")
-      if pcall(require, "cmp") then
-        require("CopilotChat.integrations.cmp").setup()
-      end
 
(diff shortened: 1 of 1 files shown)


folke:
Awesome, thanks!

jsantos42:
Hello! With this update, the [Intellij-like mapping](https://github.com/hrsh7th/nvim-cmp/wiki/Example-mappings) of the auto complete is broken. I can no longer use tabs to select the autocomplete options, be it on copilot or just from the lsp. How can I restore that?

deathbeam:
> Hello! With this update, the [Intellij-like mapping](https://github.com/hrsh7th/nvim-cmp/wiki/Example-mappings) of the auto complete is broken. I can no longer use tabs to select the autocomplete options, be it on copilot or just from the lsp. How can I restore that?

It do not uses nvim-cmp anymore so nvim-cmp settings no longer affect the chat completion either

PR #4832, opened 2024-11-18
feat(snacks): use `snacks.dashboard` as the default dashboard. moved `dashboard-nvim` to extras

folke:
## Description

Snacks has a new dashboard plugin that will be LazyVim's default.

Check the docs at https://github.com/folke/snacks.nvim/blob/main/docs/dashboard.md

## Checklist

### DIFF
--- a/lua/lazyvim/plugins/ui.lua
+++ b/lua/lazyvim/plugins/ui.lua
@@ -307,68 +307,31 @@ return {
   { "MunifTanjim/nui.nvim", lazy = true },
 
   {
-    "nvimdev/dashboard-nvim",
-    lazy = false, -- As https://github.com/nvimdev/dashboard-nvim/pull/450, dashboard-nvim shouldn't be lazy-loaded to properly handle stdin.
-    opts = function()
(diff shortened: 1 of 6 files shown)


lua/lazyvim/plugins/ui.lua
```diff
@@ -307,68 +307,31 @@ return {
...
+          ---@param keys snacks.dashboard.Item[]
+          keys = function(keys)
+            -- add LazyExtra before Lazy
+            for k, key in ipairs(keys) do
+              if key.action == ":Lazy" then
+                key.key = "l" -- we don't have multiple panes, so `l` is free
+                table.insert(keys, k, { icon = " ", desc = "Lazy Extras", action = ":LazyExtras", key = "x" })
+                break
+              end
+            end
```
dpetka2001, line 331: I noticed something with this change. If the user has also a `keys` function in his personal configuration, then he would also have to include the changes here in his personal configuration as well, because the `keys` function only allows the user to modify/extend the default keys.

PR #4854, opened 2024-11-20
feat(noice): change keymap to <leader>n instead of <leader>sn

uthmanmoh:
Not sure why `noice` uses `<leader>sn` instead of just `<leader>n`. I don't think there is anything else currently mapped to `<leader>n` and I think it makes sense `n` for "noice" or "notifications"


folke:
I want to keep that free for user mappings.

PR #4870, opened 2024-11-22
feat(snacks): add keymap to copy url from gitbrowse

benelan:
## Description

Adds a `<leader>gY` keymap that copies the URL from `Snacks.gitbrowse()` to the clipboard. 

This is a feature in other git browse plugins like `vim-rhubarb`, so other people will probably be looking for this when migrating like I did.

Let me know if you'd prefer a different key combination, thanks!

## Related Issue(s)

N/A

## Screenshots

N/A

## Checklist

### DIFF
--- a/lua/lazyvim/config/keymaps.lua
+++ b/lua/lazyvim/config/keymaps.lua
@@ -144,7 +144,10 @@ if vim.fn.executable("lazygit") == 1 then
 end
 
 map("n", "<leader>gb", function() Snacks.git.blame_line() end, { desc = "Git Blame Line" })
-map({ "n", "x" }, "<leader>gB", function() Snacks.gitbrowse() end, { desc = "Git Browse" })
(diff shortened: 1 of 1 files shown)


folke:
ty!

dgmora:
❤️

I usually had `gy` to copy the url without specifying the line (I mostly share whole files). 
It could also be two mappings `gy` for the current file vs `gY` for the current line

skoch13:
Shouldn't both `git` and `gitbrowse` modules be explicitly enabled to make this work?

folke:
no, since they don't require any setup

skoch13:
I think we might need a separate `copy` method for the `gitbrowse` module, as there is currently no way to turn off the notification produced by the code [here](https://github.com/skoch13/snacks.nvim/blob/6b98aa11d31227081f780d6321ea7dfd97f1da59/lua/snacks/gitbrowse.lua?plain=1#L187).

folke:
added option `opts.notify = false`

PR #4891, opened 2024-11-26
fix(sql): only use cmp if it is available

jpfender:
## Description

The blink and sql extras are currently conflicting because the sql extra has a hard dependency on nvim-cmp, which is disabled by the blink extra. Introducing a check for cmp in the sql extra resolves this.

## Checklist


folke:
ty

PR #4896, opened 2024-11-27
feat(rest): add more keymap and lualine

ngavinsir:
## Description

- add more keymap to kulala:
  - search requests
  - set selected env
- add kulala lualine

## Screenshots
1. New keymaps and lualine:
  image

  
## Checklist

### DIFF
--- a/lua/lazyvim/plugins/extras/util/rest.lua
+++ b/lua/lazyvim/plugins/extras/util/rest.lua
@@ -5,34 +5,48 @@ vim.filetype.add({
 })
 return {
   {
-    "mistweaverco/kulala.nvim",
-    ft = "http",
-    keys = {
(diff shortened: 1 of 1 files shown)


alsmnn:
Currently all kulala shortcuts are only active if a `http`-file is already open. Is this intended behaviour?
I could imagine, that atleast the `scratchpad` command should be available even if you are not in a http-file.
My suggestion would be to use:
```
{ "<leader>R", desc = "+Rest"},
{ "<leader>Rb", "<cmd>lua require('kulala').scratchpad()<cr>", desc = "Open scratchpad"},
```
And maybe somebody has a great idea for a icon in the which-key popup for the REST-section

folke:
there was another PR merged that conflictrs wit this one and also added more keymaps

PR #4900, opened 2024-11-27
fix(sql): fix autocomplete, it was using the wrong plugin name on the `LazyVim.has` check

igorgue:
## Description

`LazyVim.has` was called with the wrong plugin name.

## Checklist


folke:
ty

Issue #4763, opened 2024-11-11
bug: Lualine pretty_path root and cwd on Windows


Kayzels:
### Did you check docs and existing issues?

- [x] I have read all the LazyVim docs
- [x] I have updated the plugin to the latest version before submitting this issue
- [x] I have searched the existing issues of LazyVim
- [x] I have searched the existing issues of plugins related to this issue

### Neovim version (nvim -v)

NVIM v0.11.0-dev-1103+g04274a4173 Build type: RelWithDebInfo LuaJIT 2.1.1727870382

### Operating system/version

Windows 10

### Describe the bug

The pretty path function for Lualine doesn't work on Windows for relative paths, it always works directly from the root of the drive instead.

It seems to be related to this:
```lua
local root = LazyVim.root.get({ normalize = true })
local cwd = LazyVim.root.cwd()

if opts.relative == "cwd" and path:find(cwd, 1, true) == 1 then
    path = path:sub(#cwd + 2)
elseif path:find(root, 1, true) == 1 then
    path = path:sub(#root + 2)
end
```

The slashes for cwd and root are the wrong direction compared to the slashes on Windows, and so the find never works:
Path is `C:\Users\Kyzan\.windots\nvim\lua\functions\test.lua`
Root (normalised) is `C:/Users/Kyzan/.windots/nvim`
Cwd is `C:/Users/Kyzan/.windots/nvim`

And so this never finds a match, even though it should.

And so the Lualine file path component shows the full path from the drive letter down, instead of the shortened one.

### Steps To Reproduce

1. Open a file on Windows that is in a file path that is relative, where the root should be used.
2. Check the way that Lualine displays the file path when using LazyVim's `pretty_print`.

### Expected Behavior

I expect the path to be shortened so that it shows just the relative ones, either relative to the cwd or root.

So if path is `C:\Users\Kyzan\.windots\nvim\lua\functions\test.lua` and cwd is `C:/Users/Kyzan/.windots/nvim` (using the forms they're in above), I should have `lua\functions\test.lua` displaying.

At the moment, it shows `C:\...\functions\test.lua`

### Repro

```lua
vim.env.LAZY_STDPATH = ".repro"
load(vim.fn.system("curl -s https://raw.githubusercontent.com/folke/lazy.nvim/main/bootstrap.lua"))()

require("lazy.minit").repro({
  spec = {
    { "LazyVim/LazyVim", import = "lazyvim.plugins" },
    -- no other plugins needed
  },
})
[...]

PR #4911, opened 2024-11-29
fix(lualine): pretty path fix for mixed case paths on Windows

stormherz:
## Description

Here's an example of the issue:

```lua
local pretty_path = require('lazyvim.util').lualine.pretty_path({ relative = 'cwd' })

-- just a stub
local component = {
  create_hl = function() end,
  format_hl = function() return '' end,
  get_default_hl = function() return '' end
}

local cwd = vim.fn.getcwd()
vim.print('cwd: ' .. cwd)
vim.print('pretty path: ' .. pretty_path(component))

-- results if `cd d:\tmp` was called previously
-- cwd: d:\tmp
-- pretty path: d:\tmp\pretty_path_issue.lua
--
-- results if `cd D:\tmp` was called previously
-- cwd: D:\tmp
-- pretty path: pretty_path_issue.lua
```

Depending on the initial path of the `cd` we either get a pretty path or we don't =)

I'm not sure if this should be fixed in the neovim itself (considering windows paths as case-insensitive), but I would assume that would take a lot longer to land there, if it's even considered a needed change

## Related Issue(s)

Somewhat related to #4763, where I've left a comment, but decided to look a bit deeper

## Checklist


folke:
ty

PR #4927, opened 2024-12-01
Mini.files windowpicker

Aietes:
## Description

Add the ability in mini.files extra to pick an existing window when the nvim-window-picker plugin is installed. The mini.files extra already offers the ability to open in a new split, but often the splits already exist. In that case it makes sense to be able to pick the window. A similar approach is used by neotree.

## Checklist

### DIFF
--- a/lua/lazyvim/plugins/extras/editor/mini-files.lua
+++ b/lua/lazyvim/plugins/extras/editor/mini-files.lua
@@ -108,41 +108,38 @@ return {
       end,
     })
 
-    local plugin = require("lazy.core.config").plugins["nvim-window-picker"]
-    if plugin then
-      --- Opens the file in a picked window.
-      --- @param close boolean Whether to close the mini.files window after opening the file.
-      local open_in_window = function(close)
-        close = close or false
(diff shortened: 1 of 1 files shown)


folke:
Thanks for the PR! However, we're not interested in adding this as it adds a dependency on the `nvim-window-picker` plugin.

LazyVim already has built-in window picking functionality via `Snacks.picker.util.pick_win()` which you can use in your own configuration without adding extra dependencies. If you want this functionality, you can easily integrate it yourself:

```lua
-- In your mini.files config
local picked_window_id = Snacks.picker.util.pick_win()
if picked_window_id then
  mini_files.set_target_window(picked_window_id)
  mini_files.go_in({ close_on_file = close })
end
```

We prefer to keep LazyVim minimal and avoid adding optional plugin dependencies when equivalent functionality already exists in the core setup.

―  

Aietes:
This was before Snacks was fully materializing. I'm a happy Snacks user, and it does exactly what this PR was trying to solve. Thank you @folke for Snack.nvim, it's amazing!

PR #4940, opened 2024-12-02
fix(extras.treesitter-context): change event back to `LazyFile`

stefanboca:
## Description

Load treesitter-context on `LazyFile` instead of `VeryLazy`. IMHO it didn't make sense to load the plugin earlier just so that the toggle is available, especially when the toggle won't actually have any effect anything until a file is opened.

~Previously, treesitter-context was loaded on `VeryLazy` and its toggle map in `opt`, which also `require`d it. Now, mapping happens in config, after treesitter-context is setup (also how toggle is handled in other extras, eg for render-markdown).~

## Checklist

### DIFF
--- a/lua/lazyvim/plugins/extras/ui/treesitter-context.lua
+++ b/lua/lazyvim/plugins/extras/ui/treesitter-context.lua
@@ -1,7 +1,7 @@
 -- Show context of the current function
 return {
   "nvim-treesitter/nvim-treesitter-context",
-  event = "VeryLazy",
+  event = "LazyFile",
   opts = function()
     local tsc = require("treesitter-context")
     Snacks.toggle({


dpetka2001:
I don't understand this change. Why does it matter if it was required in `opts`? `opts` are passed into `setup` either way and whatever `require` happens after the plugin is loaded, no? Just asking out of pure curiosity to understand better in case I'm misunderstanding something.

stefanboca:
You're right that there's no functional difference with respect to whether the mapping happens in opts or setup. I didn't consider that because opts is a function, it wouldn't get called until the plugin is configured. However, IMHO this change makes it easier to understand exactly *when* exactly the toggle is mapped - it's clear that this happens when the plugin is setup, as opposed to when the spec is parsed (as I initially thought might happen). In that sense I guess I mislabeled this as a fix when it could have been more accurately considered a refactor. This is also very similar to how [`render-markdown`](https://github.com/LazyVim/LazyVim/blob/efc95dc6aaed3d09ea4eaefe21f668f24da2ae2e/lua/lazyvim/plugins/extras/lang/markdown.lua#L109-L125) is configured.

The other change here is that I changed the event back to `LazyFile`, from #4059 when I changed the event it to `VeryLazy`. I've thought about this some more and IMO it didn't make sense to load the plugin earlier just so that the toggle is available, especially when the toggle won't actually change anything until a file is opened.

dpetka2001:
Moving the `Snacks.toggle` mapping into `config` means that if the user for whatever reason wants to configure the plugin via the `config` function, he will have to redefine the `Snacks.toggle` mapping himself again. Whereas if the mapping is in `opts` and the user still does
```lua
config = function(_, opts)
  require("treesitter-context").setup(opts)
  -- whatever else he wants here
end
```
the `Snacks.toggle` mapping will remain. In my personal opinion I believe that `config` should generally be avoided in Neovim distros and leave it as a last measure to the user to change if he knows what he's doing. `render-markdown` has also the same "disadvantage" in my opinion, but it's out of scope with regards to this PR.

stefanboca:
That's a very good point that I didn't consider. I'll remove the changes in opts and config, and keep only the change to the event.

dpetka2001:
Please do keep in mind that I'm just a regular user like you (and not an experienced programmer) and everything I say is just a suggestion from a user's perspective. But in my eyes, in the time I've been using LazyVim, is that `config` is used mostly to define specific behavior that doesn't have to do with user's settings really (like treesitter having to `require("nvim-treesitter.configs").setup(opts)` or the recent changes you did in the Extra `blink.cmp` with the wrappers or something else). And although `Snacks.toggle` mappings are not explicitly user settings, it would really be a hassle having to redefine them again. 

I really hope I didn't come off as obnoxious and didn't offend you (I apologize if I gave any such vibes) as I myself explicitly use `opts` myself and it wouldn't matter to me, I just tried to think from a user's perspective and in my personal experience with LazyVim (which of course I certainly may be misunderstanding some things due to my lack of knowledge).

stefanboca:
You didn’t come off as obnoxious at all, and I really appreciate your feedback and suggestions!

PR #4942, opened 2024-12-03
feat(extras.ai): blink.cmp integration and kind overrides for menu drawing

stefanboca:
## Description

blink.cmp integration for codeium and copilot, and correct menu drawing for codeium, copilot, and supermaven.

I've simplified the blink.cmp config a bit for these extras (especially for copilot, which was extremely nested) by only including the blink.cmp spec if vim.g.ai_cmp is true.

Multiple AI extras can now be enabled at the same time with blink.cmp.

blink.cmp ghost text is now always enabled. Although some ai plugins always display virtual text, at worst it overlaps with blink's ghost text and is not noticable.

Lastly, I can't test copilot because I don't have a subscription, nor do I want to sign up for one, but it should work just as well as the others.

## Screenshots

With Codeium:
image

Multiple extras enabled at the same time:
image

## Checklist

### DIFF
--- a/lua/lazyvim/plugins/extras/ai/codeium.lua
+++ b/lua/lazyvim/plugins/extras/ai/codeium.lua
@@ -66,18 +66,7 @@ return {
       completion = {
         menu = {
           draw = {
-            components = {
-              kind = {
-                text = function(ctx)
-                  return ctx.source_name == "codeium" and "Codeium" or ctx.kind
-                end,
-                highlight = function(ctx)
-                  return ctx.source_name == "codeium" and "BlinkCmpKindCodeium"
-                    or require("blink.cmp.completion.windows.render.tailwind").get_hl(ctx)
-                    or ("BlinkCmpKind" .. ctx.kind)
-                end,
-              },
-            },
+            override_kind_by_source_name = { codeium = "Codeium" },
           },
         },
       },
--- a/lua/lazyvim/plugins/extras/ai/codeium.lua
+++ b/lua/lazyvim/plugins/extras/ai/codeium.lua
@@ -63,13 +63,7 @@ return {
       sources = {
         compat = { "codeium" },
       },
-      completion = {
-        menu = {
-          draw = {
-            override_kind_by_source_name = { codeium = "Codeium" },
-          },
-        },
-      },
+      providers = { codeium = { kind = "Codeium" } },
     },
   } or nil,
 }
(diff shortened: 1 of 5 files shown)


stefanboca:
I'm going to make some changes to the implementation, so I'm marking this as a draft until those are done.

folke:
Always enabling `ghost_text` makes no sense when `ai_cmp = false`. It's super annoying if ghost text is enabled in that case.

I reworked your PR to update the symbol kinds of the AI sources without the need to change any draw components.

Let me know what you think

stefanboca:
I misunderstand how `ghost_text` was disabled. Makes sense the way it is now.

I do prefer adding a `kind` in the providers config to what I did before. Unfortunately I believe that the draw functions still need to be overwritten, because

stefanboca:
I misunderstood when `ghost_text` was disabled. LGTM

I much prefer adding a `kind` to the source in the provider spec, but unfortunately `blink.cmp` currently treats `item.kind` as an integer and resolves it using its internal CompletionItemKind table - see [here](https://github.com/Saghen/blink.cmp/blob/f4e53f2ac7a3d8c3ef47be0dffa97dca637bf696/lua/blink/cmp/completion/windows/render/context.lua#L40) - so overwriting item.kind by itself is not enough and `kind` is still `Unknown`, see screenshot below. Adding custom kinds to blink is [not planned](https://github.com/Saghen/blink.cmp/pull/253#issuecomment-2454984622), but there may be some alternative in the future.
Until something like that is implemented, I believe the options are to override the draw functions (which could fallback to the user-provided ones, if they exist) or mutate `blink.cmp.types.CompletionItemKind`, which seems a bit more difficult. WDYT?

image

I realize the default draw style doesn't contain the `kind`. Here's what I use, and what is visible in the screenshot above:
```lua
draw = {
  columns = { { "kind_icon" }, { "label", "label_description", gap = 1 }, { "kind" } },
}
```

folke:
Are you sure you tested this PR with my changes? I also update that kind table, and it all works as expected on my end.

folke:
Example:

image

folke:
What AI source are you using? Maybe I added the wrong kind for that source.

folke:
You're probably using an old blink.cmp version, since in the latest version it should be `Field` instead of `Unknown` for an unknown symbol kind.

folke:
Great, just found that issue as well :)

folke:
Yes indeed. Good idea!

folke:
ty!

PR #4951, opened 2024-12-04
feat(ui): add toggle to show or hide bufferline

bitbirddev:
## Description

Adds a Toggle to show/hide the Bufferline. Keymap tbd.

### DIFF
--- a/lua/lazyvim/plugins/ui.lua
+++ b/lua/lazyvim/plugins/ui.lua
@@ -46,6 +46,22 @@ return {
     },
     config = function(_, opts)
       require("bufferline").setup(opts)
+
+      -- Add Keymap to Show/Hide Bufferline
+      Snacks.toggle({
+        name = "Bufferline",
+        get = function()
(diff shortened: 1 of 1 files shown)


folke:
Refactored the code and added this in `keymaps.lua` instead. ty!

bitbirddev:
@folke thank you!

bitbirddev:
btw it says "showtabline" in whichkey which looks a bit off since all other Options start with an uppercase letter. I think "Tabline" woud be a better name.
image


looks like `desc="Tabline"` isn't working here https://github.com/LazyVim/LazyVim/commit/8232651c4a4a80da34f8b6c1742b04144ae4b959

`Snacks.toggle.option("showtabline", {off = 0, on = vim.o.showtabline > 0 and vim.o.showtabline or 2, desc = "Tabline"}):map("<leader>uA")`

Same is true for conceallevel. 

Maybe this is just personal preference.

folke:
good catch. just fixed this on main.

bitbirddev:
thanks again! :)

PR #4983, opened 2024-12-10
feat(vscode): add vscode-specific keymaps and sync undo/redo with vscode

dgokcin:
## Summary
This pull request introduces several new keymaps specifically for VSCode when using LazyVim. These changes aim to enhance the integration between VSCode and LazyVim by adding keymaps for, tab navigation, and syncing nvim undo/redo actions with vscode undo/redo.

## Changes
- Synced undo/redo lists with VSCode using `VSCodeNotify`: (check https://github.com/vscode-neovim/vscode-neovim/issues/1139 for more details)
  - `u` for undo
  - `<C-r>` for redo
- Enabled navigation of VSCode tabs similar to LazyVim buffers:
  - `<S-h>` to go to the previous editor
  - `<S-l>` to go to the next editor

## Additional Notes
These changes are intended to improve the user experience for those who use LazyVim within VSCode by providing more intuitive and consistent keybindings. Please test these keymaps to ensure they work as expected in your VSCode setup.

### DIFF
--- a/lua/lazyvim/plugins/extras/vscode.lua
+++ b/lua/lazyvim/plugins/extras/vscode.lua
@@ -33,9 +33,18 @@ end
 vim.api.nvim_create_autocmd("User", {
   pattern = "LazyVimKeymapsDefaults",
   callback = function()
+    -- VSCode-specific keymaps for search and navigation
     vim.keymap.set("n", "<leader><space>", "<cmd>Find<cr>")
     vim.keymap.set("n", "<leader>/", [[<cmd>lua require('vscode').action('workbench.action.findInFiles')<cr>]])
     vim.keymap.set("n", "<leader>ss", [[<cmd>lua require('vscode').action('workbench.action.gotoSymbol')<cr>]])
+
+    -- Keep undo/redo lists in sync with VsCode
+    vim.keymap.set("n", "u", "<Cmd>call VSCodeNotify('undo')<CR>")
+    vim.keymap.set("n", "<C-r>", "<Cmd>call VSCodeNotify('redo')<CR>")
+
+    -- Navigate VSCode tabs like lazyvim buffers
+    vim.keymap.set("n", "<S-h>", "<Cmd>call VSCodeNotify('workbench.action.previousEditor')<CR>")
+    vim.keymap.set("n", "<S-l>", "<Cmd>call VSCodeNotify('workbench.action.nextEditor')<CR>")
   end,
 })
 


folke:
ty!

theol0403:
Vscode-neovim maintainer here, this (the undo redo mapping) should not have been merged, it creates more problems than it 
solves. For the longest time I was trying to figure out why undo is broken on my machine, didn't notice this slipped in.

The fundamental issue is that vscode's undo points are based on a time-based heuristic, wheras neovim's undo points are based on atomic actions (delete, insert, etc). Vscode doesn't understand the context of what the user is doing, it just receives a stream of edits from neovim. 

Ultimately this results in multiple unrelated edits being undone with a single press of the u key, resulting in very confusing behavior. 

Fixed in https://github.com/LazyVim/LazyVim/pull/5957

dgokcin:
Hello @theol0403 

Thanks for the explanation. I've actually been using these vscode-specific undo/redo mappings (u for undo and <C-r> for redo) for a while now, and personally, I've found the experience to be an improvement over the default undo/redo behavior in vscode-neovim before these mappings. 

Tried to reproduce your example in #5957 and it seems to be working fine for me.

theol0403:
@dgokcin vscode doesn't know when to insert undo points. if multiple edits are made in RAPID sequence (your screencap is too slow), it doesn't matter if they were produced by separate nvim edits, vscode will consider them to be one undo point. 

This is more a problem on slower computers/larger buffers. Basically vscode is unaware of the semantics of the neovim undo points and will group them together if they are in rapid succession. 

Why do you need these in the first place? Buffer desyncs are extremely rare. 

If we could natively support this, we would make it a default on the plugin side.

dgokcin:
@theol0403 You're absolutely right — my addition should be reverted (#5957). cc @folke

I was trying to address the undo/redo sync issues mentioned in [vscode-neovim#1139](https://github.com/vscode-neovim/vscode-neovim/issues/1139) by mapping Neovim's undo/redo to VSCode's undo/redo system. However, my approach fundamentally changes how Neovim handles undo/redo operations.

I probably didn’t notice the side effects because I have auto-save enabled in VSCode and tend to make slow, deliberate edits. That said, do you have any suggestions for a cleaner way to handle the "dirty buffer" issue — where a file remains marked as modified even after undoing all changes?

I remember losing some changes while undoing and then discovering the undo/redo remappings. I also looked through the PR and comments in [vscode-neovim#247](https://github.com/vscode-neovim/vscode-neovim/issues/247), but I’m still unsure if there’s a consistent solution to improve undo/redo behavior without remapping.

theol0403:
@dgokcin the dirty buffer issue has already been solved by the PR you linked. What issue are you trying to solve? PS please don't use AI agents on gh issues

PR #4987, opened 2024-12-10
chore(luasnip): add default user snippet location

vdegenne:
## Description

Loads snippets in user directories if it exists.

## Checklist

### DIFF
--- a/lua/lazyvim/plugins/extras/coding/luasnip.lua
+++ b/lua/lazyvim/plugins/extras/coding/luasnip.lua
@@ -14,6 +14,7 @@ return {
         "rafamadriz/friendly-snippets",
         config = function()
           require("luasnip.loaders.from_vscode").lazy_load()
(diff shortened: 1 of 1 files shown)


dpetka2001:
This should belong in the user personal configuration in my opinion. When I used Luasnip for example I had mine in `/lua/snippets`. I don't believe a default directory which is purely based in personal preference should be defined by default. 

If you still want to go through with this, I believe a better choice would be to use a global variable to store the path and a check be made if the variable exists to `require` the path into Luasnip. This way users can customize the path to their liking instead of it being hardcoded.

vdegenne:
This sure ensures a default, users are still free to override it, or even extend it, because from what I understand `require("luasnip.loaders.from_vscode").lazy_load` can be called multiple times to load multiple locations separately.

folke:
I do agree with this change. I did not configure anything special and it's already being used in my case.
It's the default apparently for both `nvim-snippets` and `blink.cmp`.
Can you fix that stylua issue?

folke:
ty!

PR #4991, opened 2024-12-11
fix(java): only use mason-registry if mason installed

ugur-a:
## Description
The current config breaks when mason is not installed (e.g. for Nix users). This PR makes the mason-related functionality (which, as of now, consists only of finding `lombok`) dependent on mason being installed. As noted in the commit, this does make it impossible to use lombok without mason, but I'm unaware of any other method of finding it (which is why mason-registry was used in the first place, I assume)

Please note that I'm not very experienced with Lua and the code might be not the most idiomatic one.

## Checklist


folke:
ty

PR #4999, opened 2024-12-12
fix(blink): set kind to int, not string, if overriding

stefanboca:
## Description

when sorting completion results, blink compares kinds, which can result in an error if the kind is a string

## Checklist


folke:
ty!

PR #5009, opened 2024-12-12
fix(java): avoid using mason if not present

iniw:
## Description

This patch to `extras.lang.java` guards uses of `mason-registry` in the jdtls setup behind a `LazyVim.has("mason.nvim")` check, since mason may not always be present (e.g: nix systems)

## Checklist


iniw:
What a coincidence! A PR was opened yesterday regarding this issue: #5009. The code is better so I'll cose this one.

PR #5021, opened 2024-12-13
feat(rust): exclude bad directories from rust-analyzer

iniw:
## Description

rust-analyzer is unfortunately not smart enough to use a folder's .gitignore (or similar) when looking for rust files to analyze, meaning that when it finds a large enough folder (e.g: .git, .direnv) it completely dies going through it searching for rust files.

This commit adds a filter of common folders that are big enough to potentially choke rust-analyzer. The list could be amended in the future.

## Checklist


lua/lazyvim/plugins/extras/lang/rust.lua
```diff
@@ -92,6 +92,18 @@ return {
...
+            files = {
+              excludeDirs = {
+                ".direnv",
+                ".git",
+                ".github",
+                ".gitlab",
+                "bin",
+                "node_modules",
+                "target",
+                "venv",
```
amaanq, line 104: `.venv` maybe too?

iniw: Sure, added.

PR #5037, opened 2024-12-13
style(blink): use block vs. icon for colors

minusfive:
## Description

Replace use of icon for `Color` kind with block characters for `blink.cmp` autocomplete items to make swatches more usable. This also makes the `blink.cmp` built-in Tailwind integration better match `nvim-cmp`.

**Note:** I didn't want to replace the default `LazyVim.config.icons.kinds.Color` in case it's used by other plugins (e.g. context), but didn't actually find any real uses, so perhaps that could be done instead? Figured best to leave those untouched as they can be more generically used as actual icons, and are not always guaranteed to have custom highlights applied. 

## Related Issue(s)

None.

### Old

Screenshot 2024-12-13 at 12 18 47

### New

Screenshot 2024-12-13 at 12 18 13

## Checklist


folke:
like it. ty!

PR #5044, opened 2024-12-13
fix(blink): `draw.treesitter` expects table

dpetka2001:
## Description
If you would like to enable more sources feel free. I just went with the recommended setting in blink's default configuration. Although on their repo the have it to empty table, but I assumed since you had it to `true` you would at least want the `lsp` source.

## Checklist

### DIFF
--- a/lua/lazyvim/plugins/extras/coding/blink.lua
+++ b/lua/lazyvim/plugins/extras/coding/blink.lua
@@ -53,7 +53,7 @@ return {
         },
         menu = {
           draw = {
-            treesitter = true,
+            treesitter = { "lsp" },
           },
         },
         documentation = {


minusfive (approved):

folke:
Merged and will fix for latest version.
blink is still making major changes to its config, so we need to make sure the changes we make in `opts` match `main` and then in that if statement alter the opts to make it work on latest version.

ty!

dpetka2001:
You're right. Sorry I didn't think of that. I'm just using blink nightly version and didn't think about the stable release. Just tried to fix the error that was in front of me. Will take better consideration until the next blink stable release is out.

folke:
No worries.
Those issues with ai were caused by me removing that compat section.
I had even added a note to remove after 0.7.6 was released, but removed it today, since I thought that was released already 😅

PR #5063, opened 2024-12-14
feat(r): add blink support for cmp-r

EraPartner:
## Description

The [blink.cmp](https://github.com/Saghen/blink.cmp) plugin is great but [cmp-r](https://github.com/R-nvim/cmp-r) support hasn't been added yet, which is present for [nvim-cmp](https://github.com/hrsh7th/nvim-cmp). This pull request implements this support.

## Checklist

### DIFF
--- a/lua/lazyvim/plugins/extras/lang/r.lua
+++ b/lua/lazyvim/plugins/extras/lang/r.lua
@@ -53,15 +53,13 @@ return {
     dependencies = {
       "R-nvim/cmp-r",
     },
-    opts = function(_, opts)
-      table.insert(opts.sources.compat or {}, "cmp_r")
-
-      opts.sources.providers["cmp-r"] = vim.tbl_deep_extend(
-        "force",
-        opts.sources.providers["cmp-r"] or {},
-        { name = "cmp_r", module = "blink.compat.source" }
-      )
-    end,
+    opts = {
+      sources = {
+        compat = {
+          cmp_r = { name = "cmp_r", module = "blink.compat.source" },
+        },
+      },
+    },
   },
   {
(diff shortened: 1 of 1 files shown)


lua/lazyvim/plugins/extras/lang/r.lua
```diff
@@ -47,6 +47,22 @@ return {
...
+    opts = function(_, opts)
+      table.insert(opts.sources.compat or {}, "cmp_r")
+
+      opts.sources.providers["cmp-r"] = vim.tbl_deep_extend(
+        "force",
+        opts.sources.providers["cmp-r"] or {},
+        { name = "cmp_r", module = "blink.compat.source" }
+      )
+    end,
+  },
```
dpetka2001, line 65: Is this really needed? I think that is what is done when you just define the source name in `opts.sources.compat` either way.

folke: furthermore, `opts.sources.compat` is in list_extend, so you can just specify the option instead of a function.

lua/lazyvim/plugins/extras/lang/r.lua
```diff
@@ -53,15 +53,13 @@ return {
...
-        "force",
-        opts.sources.providers["cmp-r"] or {},
-        { name = "cmp_r", module = "blink.compat.source" }
-      )
-    end,
+    opts = {
+      sources = {
+        compat = {
+          cmp_r = { name = "cmp_r", module = "blink.compat.source" },
+        },
```
dpetka2001, line 60:
```suggestion
        compat = { "cmp_r" },
```
Just this should be enough.

PR #5069, opened 2024-12-14
docs(NEWS): fix instruction to stay on nvim-cmp

Frederick888:
## Description

Enabling the nvim-cmp extra does not disable the blink extra. Since these two extras disable each other, users are actually left with no completion engine.

## Related Issue(s)

N/A

## Checklist

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -11,7 +11,7 @@ With these changes, default **LazyVim** is now just `34` plugins.
 - [fzf-lua](https://github.com/ibhagwan/fzf-lua) as a replacement for [telescope.nvim](https://github.com/nvim-telescope/telescope.nvim)
   - to use **telescope.nvim** instead, enable the `editor.telescope` extra
 - [blink.cmp](https://github.com/saghen/blink.cmp) as a replacement for [nvim-cmp](https://github.com/hrsh7th/nvim-cmp)
-  - to use **nvim-cmp** instead, enable the `coding.nvim-cmp` extra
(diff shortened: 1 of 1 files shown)


folke:
That's just not correct. You don't need to set that global. Enabling the `nvim-cmp` extra is enough.

Frederick888:
@folke Hmmm... Yeah `cmp_engine()` checks if a user has enabled the nvim-cmp extra.

However the check seems to only work if I use `lazyvim.json`. I usually use `config/lazy.lua` + `{ import = '...' }` to manage my extras, which somehow does not work with this check. It seems `require("lazy.core.config").spec.modules` during startup here is always `{ "lazyvim.plugins" }` only.

I added some logs to `cmp_engine`:

```lua
function M.cmp_engine()
  vim.notify('checking cmp_engine')
  vim.g.lazyvim_cmp = vim.g.lazyvim_cmp or "auto"
  local result = ''
  if vim.g.lazyvim_cmp == "auto" then
    result = LazyVim.has_extra("coding.nvim-cmp") and "nvim-cmp" or "blink.cmp"
    vim.notify('lazyvim_cmp = ' .. result)
    return result
  end
  result = vim.g.lazyvim_cmp
  vim.notify('lazyvim_cmp = ' .. result)
  return result
end
```

And I could see notifications: 'checking cmp_engine', 'lazyvim_cmp = blink.cmp'.

After neovim finishes loading, `require('lazyvim.util').cmp_engine()` then returns `nvim-cmp` no matter where I add the nvim-cmp extra to.

folke:
If you don't use LAzyExtras then you need to use that global.

PR #5099, opened 2024-12-15
fix(blink): allow user to overwrite `<Tab>` mapping

### LINKED ISSUE #5095: bug: unable to override keymap

### Neovim version (nvim -v)

0.10.2

### Operating system/version

MacOS

### Describe the bug

The following change makes it hard (maybe impossible) to override the `Tab` keymap in `opts` as it will be overwritten.

```lua
      -- fix super-tab completion
      if opts.keymap.preset == "super-tab" then
        opts.keymap["<Tab>"] = {
          function(cmp)
            if cmp.snippet_active() then
              return cmp.accept()
            else
              return cmp.select_and_accept()
            end
          end,
          LazyVim.cmp.map({ "snippet_forward", "ai_accept" }),
          "fallback",
        }
      end
```

### Steps To Reproduce

Set your own keymap for Tab, using the `super-tab` preset, it will be ignored.

### Expected Behavior

Should be able to provide my own keymap

### Repro

```lua

require("lazy.minit").repro({
  spec = {
    -- add any other plugins here
  },
})
```

dpetka2001:
## Description
Users can't overwrite `<Tab>` mapping with recent [commit](https://github.com/LazyVim/LazyVim/commit/413566af591e9152f156944bff35c89d5d973148)

## Related Issue(s)
Fixes #5095

## Checklist

### DIFF
--- a/lua/lazyvim/plugins/extras/coding/blink.lua
+++ b/lua/lazyvim/plugins/extras/coding/blink.lua
@@ -102,7 +102,7 @@ return {
       end
 
       -- fix super-tab completion
-      if opts.keymap.preset == "super-tab" then
+      if opts.keymap.preset == "super-tab" and not opts.keymap["<Tab>"] then
         opts.keymap["<Tab>"] = {
           function(cmp)
             if cmp.snippet_active() then


dpetka2001:
I also noticed that with this [recent change](https://github.com/LazyVim/LazyVim/commit/15c81fdbb839f0c657fe4c077114475e82f423fe) pressing `<esc>` on command line actually executes it. Using something like 
```lua
map({ "i", "n" }, "<esc>", function()
  vim.cmd("noh")
  LazyVim.cmp.actions.snippet_stop()
end, { desc = "Escape and Clear hlsearch" })
```
I believe is better. But maybe this also indicates something on `Snacks.util.on_key` that you might want to rectify. Not sure what the correct approach for this would be, that's why I didn't make a PR about it and preferred to let you know instead.

folke:
This change will never set `<tab>`, since we're always setting `<tab>` in options.
Will see for another solution.

folke:
And ty for that escape! Didn't think that though :)

Updated it to:

```lua
-- Clear search and stop snippet on escape
map({ "i", "n", "s" }, "<esc>", function()
  vim.cmd("noh")
  LazyVim.cmp.actions.snippet_stop()
  return "<esc>"
end, { expr = true, desc = "Escape and Clear hlsearch" })
```

dpetka2001:
You're right. I thought that would do the trick, but didn't test it locally. Will check better next time.

PR #5101, opened 2024-12-15
feat(fzf): bind ctrl+u

r-darwish:
## Description

Add the ctrl+u binding to discard the entire line in fzf


pkazmier:
This conflicts with the half-page up binding five lines up from your change.

dpetka2001:
Yeah this is already default in fzf-lua and was changed in LazyVim. So, either `<c-d>/<c-u>` should be changed for example to `<alt-d>/<alt-u>` or what OP wants should be put in the user's personal configuration if they want the fzf-lua default behavior. 

Not sure what the best solution is. Up to the maintainer to decide.

pkazmier:
I like lazy’s settings as the default preview up/down of fzf-Lua requires me taking hands off main part of keyboard. Admittedly, I was confused at first, but I just use ctrl-w (how often is my search more than a word or two?) which works fine enough.

mokevnin:
> This conflicts with the half-page up binding five lines up from your change.

Removing a line using c-u is the default behavior not only for fzf-lua but for any terminal as well. So, for people like me, who use it constantly, changing the mapping significantly affects DX.

dpetka2001:
You can always change that in your personal configuration if you like. Not all defaults will appease everyone. `<c-u>` is widely used in Neovim for scrolling up. And this PR should not add the mapping you added. Instead it should just remove the `<c-u>` five lines above, so that there isn't duplicated and unneeded code. In the end the maintainer will decide.

mokevnin:
> You can always change that in your personal configuration if you like. Not all defaults will appease everyone. `<c-u>` is widely used in Neovim for scrolling up. And this PR should not add the mapping you added. Instead it should just remove the `<c-u>` five lines above, so that there isn't duplicated and unneeded code. In the end the maintainer will decide.

I agree. I use C-U a lot for scrolling, but there are two cases: a page with a lot of text and a terminal-like line.

dpetka2001:
It's not up to me to decide. I'm just a regular user like you. Maintainer has the final say. Either way I don't mind that much. I'll just adjust my personal configuration or not.

But again. Regarding this PR. You should remove your own mapping and also remove the mapping 5 lines above. Then it will be just the default from fzf-lua and there won't be any duplicated code for no good reason or am I wrong?

mokevnin:
While we're chatting. I've been wanting to talk about this for a long time, so a little essay from me :) (I hope maintainers will find it at least interesting to think of)

I really like LazyVim and appreciate the contributions of everyone involved in its development. The only thing that frustrates me occasionally is the rather loose handling of defaults. Such combinations are a significant change that should not be done just like that, even with a version change. Any combinations that are standard for the system as a whole, not just the editor, should not be changed from my point of view because for people who spend a lot of time in different terminals and editors, such key changes break years (in my case decades) of developed habits. I realize that everything can be changed, and sometimes you have to, but I would like it not to be about standard things. I'd like it to be part of the editor's philosophy.

dpetka2001:
Different people different preferences?? I mean you surely get that. Like I said I don't mind with this change, but defaults mean nothing to me except something that I will change if I don't like it. And this is just my personal opinion as a simple user and human being. Maybe we don't agree on that and we'll just have to agree to disagree and that's perfectly fine. I don't believe I made an issue about defaults (I hope my memory is right about this, but maybe I could be mistaken about my memory), because in my mind that's just someone else's preference and I can simply change it if I don't like it. 

By the way I just realised that I kept telling you about the needed changes in this PR and just now saw that you're not the OP of this PR. Sorry about that.

mokevnin:
Yeah, I was trying to restore default mapping and ended up here :) I just wanted to explain why, for some people like me, it is not just a regular mapping change and what they feel about it.

genmzy:
> You can always change that in your personal configuration if you like. Not all defaults will appease everyone. `<c-u>` is widely used in Neovim for scrolling up. And this PR should not add the mapping you added. Instead it should just remove the `<c-u>` five lines above, so that there isn't duplicated and unneeded code. In the end the maintainer will decide.
How to disable the <c-u> in personal configuration???

PR #5127, opened 2024-12-16
fix(blink): <Tab> should respect manual selection

minusfive:
## Description

When the `completion.list.selection` option is set to `"manual"` the custom `<Tab>` keymap doesn't respect it and still selects and accepts the first option.

Simply ensuring `cmp.accept()` is triggered when this option is set (instead of `cmp.select_and_accept()`) works because `blink` internally checks whether a completion item is selected, and if it isn't it exits without executing anything.

## Related Issue(s)

none

## Screenshots

none

## Checklist

### DIFF
--- a/lua/lazyvim/plugins/extras/coding/blink.lua
+++ b/lua/lazyvim/plugins/extras/coding/blink.lua
@@ -110,7 +110,7 @@ return {
         if opts.keymap.preset == "super-tab" then -- super-tab
           opts.keymap["<Tab>"] = {
             function(cmp)
-              if cmp.snippet_active() then
+              if cmp.snippet_active() or opts.completion.list.selection == "manual" then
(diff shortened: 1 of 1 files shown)


folke:
I don't see why this is needed?
The only thing that LazyVim code does is make sure the tab key also does the ai stuff when super-tab is enabled.
The rest of the keymap is identical to blinks supertab preset: https://github.com/Saghen/blink.cmp/blob/main/lua/blink/cmp/keymap/presets.lua#L23

If you feel like something is wrong there, then it should be fixed in the blink repo, not here

minusfive:
@folke sure thing, will submit an upstream PR (do think there's a bug). In that case, what do you think about doing something like this #5130, so upstream fixes to that function can propagate to LazyVim?

PR #5129, opened 2024-12-16
feat(treesitter): add ensure_installed_map

binhtran432k:
## Description

This update introduces a simple way for users to enable or disable parsers through the configuration. It also streamlines the process of adding additional Tree-sitter parsers without relying on `list_extend`. This approach can also be applied to Mason.

## Checklist

### DIFF
--- a/lua/lazyvim/plugins/treesitter.lua
+++ b/lua/lazyvim/plugins/treesitter.lua
@@ -38,31 +38,32 @@ return {
     opts = {
       highlight = { enable = true },
       indent = { enable = true },
-      ensure_installed = {
-        "bash",
-        "c",
-        "diff",
-        "html",
-        "javascript",
-        "jsdoc",
-        "json",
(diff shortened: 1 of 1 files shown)


dpetka2001:
You don't have to use `list_extend` to add additional Treesitter parsers. `lazy.nvim`'s `opts.extend = { "ensure_installed" }` takes care of that.

The only thing you would need to do to overwrite Mason's `ensure_installed` completely from LazyVim's defaults is to define `opts` as a value function in your personal configuration.

folke:
Yeah, I don't see why this is needed. Thank you anyway

binhtran432k:
Thank @folke, I think it is still useful when user want to disable parser they do not want to ensure installed from their config. This is similar to Mason.

folke:
But they can, with opts as a function like @dpetka2001 mentioned

PR #5138, opened 2024-12-17
feat(picker): use `<c-cr>` to resume picker's session

igorgue:
## Description

In Telescope uses `<c-cr>` (maybe some other keymap should be used?) to restore the Telescope's session, in case of a session not found call `LazyVim.picker("files")`.

I find this really useful when searching, since it restores the list to the position it was before, this is basically my killer feature from Telescope.

In `FzfLua` it uses `FzfLua resume`, but doesn't do the toggle the `Telescope` method does, nor it puts you back to the previous position of the cursor.

### DIFF
--- a/lua/lazyvim/plugins/extras/ai/codeium.lua
+++ b/lua/lazyvim/plugins/extras/ai/codeium.lua
@@ -4,6 +4,7 @@ return {
   {
     "Exafunction/codeium.nvim",
     cmd = "Codeium",
+    event = "InsertEnter",
     build = ":Codeium Auth",
     opts = {
       enable_cmp_source = vim.g.ai_cmp,


jyuan0:
> Would love to do this on fzf, but I'm not sure how.

I think `:FzfLua resume` is the equivalent.

dpetka2001:
`:FzfLua resume` has a limitation of not being able to restore the position of the cursor in the entry that was in the last search. Basically it just re-spawns a different process with the last query and reloads the window. Maintainer has admitted that it's a limitation with how it communicates with the external fzf binary. I'm not sure if something's changed since when I read about it (1-2 months ago or 3 I don't really remember for certain). 

That's the only thing that's kind of a bummer but not a big deal. But you can press `<M-esc>` to hide the window instead and then use `:FzfLua resume` to resume your search and the cursor will be in the entry that it was before the window was hidden, although I don't find it personally that much useful as I usually want to perform an action when doing a search and that will reset the cursor position if you do a resume later.

jyuan0:
> :FzfLua resume can not restore the position of the cursor in the entry that was in the last search. Basically it just re-spawns a different process with the last query and reloads the window.

I never noticed that, but I also don't use this command often. Thanks for the info @dpetka2001!

igorgue:
Updated the PR to use `FzfLua resume`, it's not as nice but it works.

igorgue:
@jyuan0 In my setup the ergonomic of having it on `<c-cr>` is very nice, makes me use restore all the time.

lua/lazyvim/plugins/extras/editor/fzf.lua
```diff
@@ -208,6 +208,7 @@ return {
     keys = {
       { "<c-j>", "<c-j>", ft = "fzf", mode = "t", nowait = true },
       { "<c-k>", "<c-k>", ft = "fzf", mode = "t", nowait = true },
+      { "<c-cr>", "<cmd>FzfLua resume<cr>", desc = "Resume Picker" },
```
dpetka2001, line 211: That is already mapped to `<leader>sR`. I don't believe there's a need for that.

igorgue: @dpetka2001 then maybe I should use that keymap for telescope...

dpetka2001: I didn't actually test what your code for Telescope does and how it might be different than `Telescope resume` that is also mapped to `<leader>sR`, so maybe keep it as it is?? Wait also for maintainer's input just to be sure.

PR #5141, opened 2024-12-17
refactor(blink): color swatch icon upstreamed

minusfive:
## Description

I've upstreamed the block char vs. icon replacement for Tailwind colors so we shouldn't need this here anymore. I've replaced it by extending the default blink icons instead, just in case they introduce an icon which isn't yet covered by Lazy's, it would still show up (unlikely, but better, no?).

## Related Issue(s)

none

## Screenshots

none

## Checklist


folke:
ty!

PR #5170, opened 2024-12-18
fix(go): update go.lua to eliminate fieldalignment from analyses

cjon256:
## Description

Removed a setting from the go lang plugin. The setting now just causes an error:
LSP[gopls] Invalid settings: setting option "analyses": this setting is deprecated, use "the 'fieldalignment' analyzer was removed in gopls/v0.17.0; instead, hover over struct fields to see size/offset information (https://go.dev/issue/66861)" instead

## Checklist

Note: submitted similar request as f96aac6 but was rejected by CI/CD for naming reasons. Changing the name did not help. Retrying.

### DIFF
--- a/lua/lazyvim/plugins/extras/lang/go.lua
+++ b/lua/lazyvim/plugins/extras/lang/go.lua
@@ -37,7 +37,6 @@ return {
                 rangeVariableTypes = true,
               },
               analyses = {
-                fieldalignment = true,
                 nilness = true,
                 unusedparams = true,
                 unusedwrite = true,


cjon256:
The issue related to the removal of this option from gopls is https://github.com/golang/go/issues/67762 

That the warning message goes to the wrong issue is mentioned in that issue.

seflue:
Stumbled over the same issue and fixed it locally in the same way. The problem with this error is, that it actually breaks a couple of features interacting with the language server (e.g. fuzzy searching for go symbols). Merging this PR is heavily appreciated.

itsrobel:
> Stumbled over the same issue and fixed it locally in the same way. The problem with this error is, that it actually breaks a couple of features interacting with the language server (e.g. fuzzy searching for go symbols). Merging this PR is heavily appreciated.

what did you do locally to fix it, it's throwing the error for me every time I open a go file and it's annoying

cjon256:
Someone asked via email: "what did you do locally to fix it, it's throwing the error for me every time I open a go file and it's annoying"

To which I would say they should just look at the files tab above. You remove a line ("fieldalignment = true," line 40) from lua/lazyvim/plugins/extras/lang/go.lua Where that is on your system depends on your setup, but is should be wherever nvim keeps your files. For me that is ~/.local/share, but your milage may vary.

Cullen

itsrobel:
> Someone asked via email: "what did you do locally to fix it, it's throwing the error for me every time I open a go file and it's annoying"
> 
> To which I would say they should just look at the files tab above. You remove a line ("fieldalignment = true," line 40) from lua/lazyvim/plugins/extras/lang/go.lua Where that is on your system depends on your setup, but is should be wherever nvim keeps your files. For me that is ~/.local/share, but your milage may vary.
> 
> Cullen

That was me, thank you, I realized it was a silly question and fixed it shortly after, deleting the reply.

cjon256:
This change should still be considered for acceptance. It does break peoples' workflows.

folke:
ty

PR #5177, opened 2024-12-18
fix(fzf): fzf.lua compatible hidden scrollbar

oozzal:
## Description

`--no-scrollbar` option is not supported by [fzf-lua](https://github.com/ibhagwan/fzf-lua)
The recommended way is to set `winopts.preview.scrollbar=false`

## Checklist

### DIFF
--- a/lua/lazyvim/plugins/extras/editor/fzf.lua
+++ b/lua/lazyvim/plugins/extras/editor/fzf.lua
@@ -89,9 +89,6 @@ return {
       return {
         "default-title",
         fzf_colors = true,
-        fzf_opts = {
-          ["--no-scrollbar"] = true,
-        },
         defaults = {
           -- formatter = "path.filename_first",
           formatter = "path.dirname_first",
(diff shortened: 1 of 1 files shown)


dpetka2001:
That is not correct. The option depends on the version of `fzf` you have installed. Latest versions of `fzf` do support it. So if a user doesn't install latest version of fzf, they should just disable the option in their personal configuration.

folke:
I think OP is correct. I quickly checked the fzf-lua code and as far as I can tell using this option instead will make sure it's only used for compatible fzf versions.

nfurfaro:
I would love to see this land! 
I can't use file/buffer switcher, color scheme switcher, etc...

folke:
ty!

ibhagwan:
The 2 options are fundamentally different, fzf’s `--no-scrollbar` is for the main window and if using native previewers (git, bat, etc) while `winopts.preview.scrollbar` is for the builtin neovim previewer (the neovim win preview, the default):
IMG_2817

From `man fzf` (v0.36 and above):
```man
       --scrollbar=CHAR1[CHAR2]
              Use the given character to render scrollbar. (default: 'â' or
              ':' depending on --no-unicode). The optional CHAR2 is used to
              render scrollbar of the preview window.


       --no-scrollbar
              Do not display scrollbar. A synonym for --scrollbar=''
```

folke:
ok, thanks. In that case, I just reverted this PR.

PR #5182, opened 2024-12-18
fix(codeium): doesnt automatically load when `vim.g.ai_cmp` is set to false

manuuurino:
## Description

I think the title of the PR says it all. Before the user has to manually load/trigger the plugin via the cmd (`:Codeium Auth`) or by requiring.

PS: The copilot extra also has this event.

~~PPS: Ive also noticed that supermaven doesnt have this event setup, perhaps it suffers to the same issue when this option is disabled?~~

Edit:
For the PPS, have tested supermaven, i can confirm, it loads properly.

## Checklist

### DIFF
--- a/lua/lazyvim/plugins/extras/ai/codeium.lua
+++ b/lua/lazyvim/plugins/extras/ai/codeium.lua
@@ -4,6 +4,7 @@ return {
   {
     "Exafunction/codeium.nvim",
     cmd = "Codeium",
+    event = "InsertEnter",
     build = ":Codeium Auth",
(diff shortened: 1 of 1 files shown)


folke:
ty!

dpetka2001:
Shouldn't instead `vim.g.ai_cmp` be moved inside blink's `opts` and that way it would always load as a dependency of either nvim-cmp or blink?

manuuurino:
Havent thought about it, but tbh i found this approach to be cleaner

PR #5183, opened 2024-12-18
perf(supermaven): add lazy-loading

manuuurino:
## Description

As title, somehow these events have been forgotten, ive checked the git history of it aswell. 

## Checklist

### DIFF
--- a/lua/lazyvim/plugins/extras/ai/supermaven.lua
+++ b/lua/lazyvim/plugins/extras/ai/supermaven.lua
@@ -3,16 +3,8 @@ return {
     "supermaven-inc/supermaven-nvim",
     event = "InsertEnter",
     cmd = {
-      "SupermavenStart",
-      "SupermavenStop",
-      "SupermavenRestart",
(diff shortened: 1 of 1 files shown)


folke:
This doesn't really make sense, since creating those lazy loaded commands also has a cost.
Better to just load on `VeryLazy` instead if those commands are really that important.

dpetka2001:
The only useful commands are `SupermavenUseFree/Pro`. The loading should be taken care of `nvim-cmp/blink` as their dependency. However, `vim.g.ai_cmp` should be moved inside blink's `opts`, so that the blink spec is still responsible for loading supermaven regardless of `vim.g.ai_cmp` in my personal opinion.

dpetka2001:
Never mind just keep `InsertEnter` like copilot and maybe just the 2 commands? Or `VeryLazy` indeed.

manuuurino:
Just updated it, ive decided to keep these two commands, so its steup is similar to its others like copilot and codeium

folke:
ty

PR #5194, opened 2024-12-19
fix(ai): mappings for accept when `ai_cmp=false`

dpetka2001:
## Description
Define `accept` mappings when `ai_cmp = false` (I chose the default for Copilot for every plugin as `<Tab>` is used by LazyVim) and also for Codeium only create the `nvim-cmp` source if `ai_cmp = true`.

## Related Issue(s)
None

## Checklist

### DIFF
--- a/lua/lazyvim/plugins/extras/ai/codeium.lua
+++ b/lua/lazyvim/plugins/extras/ai/codeium.lua
@@ -11,9 +11,7 @@ return {
       virtual_text = {
         enabled = not vim.g.ai_cmp,
         key_bindings = {
-          accept = false, -- handled by nvim-cmp / blink.cmp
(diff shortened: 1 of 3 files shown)


folke:
I think you misunderstand how `ai_cmp` works? 
When `false`, which is what I'm using, then you can use `<tab>` to accept ai completions. Those specific mappings should always be false for each AI provider, since LazyVim does the `<tab>` mapping and or completions though nvim-cmp/blink.

dpetka2001:
Ohhh, that's how it is, then I definitely misunderstood. I thought they were just not defined. Sorry for unnecessary noise, will close. I just tried it with `ai_cmp=false` and indeed is as you say.

PR #5175, opened 2024-12-18
feat(colorscheme): add fzf integration to catppuccin

kezhenxu94:
## Checklist

- [ ] I've read the [CONTRIBUTING](https://github.com/LazyVim/LazyVim/blob/main/CONTRIBUTING.md) guidelines.

PR #5203, opened 2024-12-19
feat(lualine): add fzf extensions.

milanglacier:
## Description

Similar to #5175, add `fzf` integration with other plugins.

Note that similar to #5175, integration with `fzf` is included unconditionally. I think we don't need to check if the users choose `fzf` as their default picker to conditionally add the fzf integration with lualine.

## Screenshots

** Before **:

Screenshot 2024-12-19 at 17 58 30

** After **:

Screenshot 2024-12-19 at 17 59 28

### DIFF
--- a/lua/lazyvim/plugins/ui.lua
+++ b/lua/lazyvim/plugins/ui.lua
@@ -159,7 +159,7 @@ return {
             end,
           },
         },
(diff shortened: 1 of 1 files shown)


folke:
ty

PR #5211, opened 2024-12-20
fix(luasnip): correct name to not show duplicate snippets

### LINKED ISSUE #5210: every snippet has it's duplicate for some reason.
### Discussed in https://github.com/LazyVim/LazyVim/discussions/5142

<div type='discussions-op-text'>

<sup>Originally posted by **milkpuff** December 17, 2024</sup>
 I enable lazyextra `coding.blink` and `coding.luasnip`:
图片
If disable `coding.luasnip`, snippets show once, but some snippet like `classi` raise error:
`/usr/share/nvim/runtime/lua/vim/lsp/_snippet_grammar.lua:177: snippet parsing failed`</div>

I also have this issue. I use luasnip with pretty basic setup.
here is [my LazyVim config](https://github.com/monoira/.dotfiles/tree/main/nvim/.config/nvim)

Image

dpetka2001:
## Description
Using `compat = { "luasnip" }` causes the following `sources.providers.luasnip.name = "luasnip"`. Instead manually defining `sources.defaults` and `sources.providers.luasnip.name = "Luasnip"` seems to solve the problem of duplicate snippets appearing with blink+Luasnip enabled. 

Not sure why the `name` matters. I just saw in `blink.cmp` that they also have a source for Luasnip (although it only supports the latest stable version of Luasnip) and in their documentation they register the source with `name = Luasnip`. 

On the other side `blink.compat` docs say to use the same name as you would use with `nvim-cmp`, so in this case it should be `name = "luasnip"`.

@folke I would appreciate it if you could solve some light as to why this happens and also verify if my proposed solution doesn't have any side-effects. Testing locally, this change got rid of duplicate snippets for me.

## Related Issue(s)
Fixes #5210

## Checklist

### DIFF
--- a/lua/lazyvim/plugins/extras/coding/luasnip.lua
+++ b/lua/lazyvim/plugins/extras/coding/luasnip.lua
@@ -72,7 +72,15 @@ return {
       { "saadparwaiz1/cmp_luasnip" },
     },
     opts = {
-      sources = { compat = { "luasnip" } },
+      sources = {
+        default = { "luasnip" },
+        providers = {
+          luasnip = {
+            name = "Luasnip",
+            module = "blink.compat.source",
+          },
+        },
+      },
       snippets = {
         expand = function(snippet)
           require("luasnip").lsp_expand(snippet)


dpetka2001:
My guess is that blink's internal Luasnip source also is defined as `luasnip` in `sources.defaults`. So, maybe because we also use `luasnip` through `blink.compat` layer in `sources.defaults`, it still enables the internal blink source?

dpetka2001:
I tested again and if we just change `compat = { "luasnip" }` to `compat = { "Luasnip "}` it works without duplicate snippets. I believe the blink's internal source is also enabled with `luasnip`. 

With `compat = { "Luasnip" }` it means that the following will be equivalent
```lua
sources.defaults = { "Luasnip" },
sources.providers = {
  Luasnip = { name = "Luasnip", module = "blink.compat.source" }
}
```
so it will differentiate from blink's internal source which I believe is enabled by setting `sources.defaults = { "luasnip" }` as well.

folke:
If blink has a luasnip source, shouldn't we just delete the compat one then?

dpetka2001:
I don't have any problem with it. It just mentions in blink's docs that it's only compatible with Luasnip latest stable release 2.3 [here](https://github.com/Saghen/blink.cmp?tab=readme-ov-file#luasnip). But I still get duplicate snippets from my testing. Only when using the compat with the changes in this PR, I don't get duplicate snippets. 

I've no idea why this happens. Could you also test from your side and see what you get? Whatever you decide is best, I obviously don't mind. And I don't use Luasnip to be honest. Nor do I know where the problem lies in the end.

folke:
Just configured it to enable blink's luasnip source instead.
I don't get any duplicates.
Well I do with `for` for example, but that's not a real duplicate.
One is from luasnip, the other is from the lsp.

dpetka2001:
Tnx for looking into this :smile:

dpetka2001:
@folke Sorry for pinging you, but I think I might have a lead as to why duplicate snippets appear with Luasnip. As per [blink's docs](https://github.com/Saghen/blink.cmp?tab=readme-ov-file#luasnip) the sources should be 
```lua
sources = {
      default = { 'lsp', 'path', 'luasnip', 'buffer' },
    },
```
But we already have `snippets` from the default blink setup. So, I believe that `snippets` should be removed from `opts.sources.default` when Luasnip is enabled. And Luasnip's setup will also take care of `friendly-snippets` and the custom search path `vim.fn.stdpath('config') .. '/snippets'`.

Whenever you have some time to look into it after you come back.

folke:
oh that makes sense. Feel free to provide a PR.
luasnip is loaded after blink, so in luasnip we can change it to a function, add luasnip and remove snippets.

PR #5212, opened 2024-12-20
fix(blink): remove compatibility with 0.7.6 release

igorgue:
## Description

With the version update 5 mins ago, I got an error with this setting.

I'm not sure if removal is the correct way to fix this but it does runs well on my local.

### DIFF
--- a/lua/lazyvim/plugins/extras/coding/blink.lua
+++ b/lua/lazyvim/plugins/extras/coding/blink.lua
@@ -118,17 +118,6 @@ return {
         end
       end
 
-      ---  NOTE: compat with latest version. Currenlty 0.7.6
-      if not vim.g.lazyvim_blink_main then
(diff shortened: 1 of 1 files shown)


dpetka2001:
Yes, this was only needed for as long as 0.7.6 was the latest stable release. It shouldn't be needed any more now that 0.8.0 is the latest stable release.

dpetka2001:
By the way do you use Luasnip? If yes, do you observe duplicate snippets?

folke:
ty

surmish:
Can you author a release for this?

igorgue:
> By the way do you use Luasnip? If yes, do you observe duplicate snippets?

Not really, I saw this behavior and decided to just use the builtin snippets and remove Luasnip's extra.

PR #5223, opened 2024-12-21
fix(lualine): fix Illegal Character issue with fzf.

milanglacier:
HACK: see lualine.nvim issue #1201, Illegal character from fzf could
cause lualine to hang. To prevent this, we disable lualine from
retrieving the content of the currently selected item in fzf, avoiding
potential invalid characters.


folke:
this should be fixed in lualine instead of disabling here

PR #5237, opened 2024-12-23
feat(bufferline): add keymap to rename current tab

HalibutGitWiz:
## Description

Adds a keymap to rename current tab, using a command from [bufferline.nvim](https://github.com/akinsho/bufferline.nvim)

This command requires an argument (the new name), so the keymap opens a command prompt for the user to type the name. For some reason, using `<cmd>BufferLineTabRename ` syntax outputs an error because of the lack of a `<cr>` at the end, whereas `:BufferLineTabRename ` allows the partial command.

I'm unsure about how clear it is for the user that they have to type a name after the command in the prompt. But I don't see a way to improve the UX on this.

## Checklist


n-crespo:
could consider something like this: 

```lua
          vim.ui.input({ prompt = "Rename tab to: " }, function(input)
            if input then
              vim.cmd("BufferLineTabRename " .. input)
            end
          end)
```

folke:
not interested in adding this

PR #5214, opened 2024-12-20
fix(luasnip): correctly remove duplicate snippets

dpetka2001:
## Description
Remove `snippets` from `opts.sources.default` when Luasnip is enabled to not get duplicate snippets.

## Related Issue(s)
Fixes #5210

## Checklist

- [x] I've read the [CONTRIBUTING](https://github.com/LazyVim/LazyVim/blob/main/CONTRIBUTING.md) guidelines.

PR #5274, opened 2024-12-29
feat(snippets): an extra for mini.snippets

abeldekat:
## Description

Mini.snippets, a plugin to manage and expand snippets, is currently in [beta](https://github.com/echasnovski/mini.nvim/issues/1428).

I think the plugin is a very interesting alternative to `luasnip`.

The following `lazydoc` will be added:

```lua
if lazyvim_docs then
  -- Set to `false` to prevent "non-lsp snippets"" from appearing inside completion windows
  -- Motivation: Less clutter in completion windows and a more direct usage of snippits
  vim.g.lazyvim_mini_snippets_in_completion = true

  -- Using default mini.snippets, completion suggestions might appear directly after inserting a snippet
  -- This extra prevents that from happening.
  -- Motivation: A better display of the current snippet.
  -- Also, those completions do not appear when using luasnip or native snippets
  --
  -- Set to `false` to enable completion suggestions directly after inserting a snippet
  vim.g.lazyvim_mini_snippets_override_expand_insert = true

  -- NOTE: Please also read:
  -- https://github.com/echasnovski/mini.nvim/blob/main/readmes/mini-snippets.md#expand
  -- :h MiniSnippets-session

  -- Example override for your own config:
  --[[
  return {
    {
      "echasnovski/mini.snippets",
      opts = function(_, opts)
        -- By default, for opts.snippets, the extra for mini.snippets only adds gen_loader.from_lang()
        -- This provides a sensible quickstart, integrating with friendly-snippets
        -- and your own language-specific snippets
        --
        -- In order to change opts.snippets, replace the entire table inside your own opts

        local snippets, config_path = require("mini.snippets"), vim.fn.stdpath("config")

        opts.snippets = { -- override opts.snippets provided by extra...
          -- Load custom file with global snippets first (order matters)
          snippets.gen_loader.from_file(config_path .. "/snippets/global.json"),

          -- Load snippets based on current language by reading files from
          -- "snippets/" subdirectories from 'runtimepath' directories.
          snippets.gen_loader.from_lang(), -- this is the default in the extra...
        }
      end,
    },
  }
--]]
end

```

It's possible to use the code from this PR directly in your own configuration:

1. Copy the code to lua/extras/coding/mini_snippets.lua in your config folder
2. Restart nvim and enable the "user" extra for `mini.snippets`
3. Restart nvim
 
## Checklist


lua/lazyvim/plugins/extras/coding/mini-snippets.lua
```diff
@@ -0,0 +1,159 @@
...
+
+--]]
+
+local function expand(args)
+  ---@diagnostic disable-next-line: undefined-global
+  local insert = MiniSnippets.config.expand.insert or MiniSnippets.default_insert
+  insert({ body = args.body }) -- insert at cursor
+end
+
+local snippet_select_for_cmp = function(snippets, insert, opts)
```
echasnovski, line 52: `select` step [takes only two `snippets` and `insert` arguments](https://github.com/echasnovski/mini.nvim/blob/733afb94ed88d32b839a644dc137604691e46032/doc/mini-snippets.txt#L626-L631). The third `opts` argument is only present in `default_select()` to be able to adjust its behavior inside custom `expand.select`.

lua/lazyvim/plugins/extras/coding/mini-snippets.lua
```diff
@@ -0,0 +1,159 @@
...
+
+Note that there is no completion source. 
+Other snippets(custom or from friendly-snippets) are handled by pressing <c-j> in insert mode.
+
+In default LazyVim, neither cmp nor blink define the default mappings used by mini.snippets:
+- "<c-j>" to expand
+- "<c-l>" to jump next (dynamically created when in snippet context)
+- "<c-h>" to jump previous (dynamically created when in snippet context)
+
+It's difficult to have jump_next or jump_previous working in all cases when mapped to "<tab>"/"<s-tab>".
```
echasnovski, line 15: The best first approximation is [present in 'mini.snippets' docs](https://github.com/echasnovski/mini.nvim/blob/733afb94ed88d32b839a644dc137604691e46032/doc/mini-snippets.txt#L407). But yes, it is indeed somewhat difficult to make a flexible mapping for a distro.

lua/lazyvim/plugins/extras/coding/mini-snippets.lua
```diff
@@ -0,0 +1,159 @@
...
+          "`mini.snippets`:",
+          "Don't override jump_next or jump_previous.",
+          "This does not work correctly in all cases.",
+        }, { title = "LazyVim" })
+      end
+      require("mini.snippets").setup(opts)
+    end,
+  },
+
+  -- nvim-cmp integration: Use mini.snippets to expand snippets from lsp
```
echasnovski, line 124: I'd also assume that at least *some* people would want to have 'nvim-cmp' suggest snippets that 'mini.snippets' prepares/matches in current context. The array of snippets can be computed [like so](https://github.com/echasnovski/mini.nvim/blob/733afb94ed88d32b839a644dc137604691e46032/doc/mini-snippets.txt#L478-L484). When to do it and shaping into correct data structure is expected to be done on the "adapter's" side.

lua/lazyvim/plugins/extras/coding/mini-snippets.lua
abeldekat: Thanks! Done in commit `6b5e0e3063c9`

lua/lazyvim/plugins/extras/coding/mini-snippets.lua
abeldekat: I tried the extra in its current state with the following:

```lua
mappings = {
  -- keep the default expand, a global mapping
  jump_next = "<tab>",
  jump_prev = "<s-tab>"
```

No luck with `blink.cmp` however. As there are no sources yet I did not investigate any further.

lua/lazyvim/plugins/extras/coding/mini-snippets.lua
abeldekat: > I'd also assume that at least some people would want to have 'nvim-cmp' suggest ...

I am sure of that, also regarding `blink.cmp`...)
Although, maybe some people see the benefit of separating non-lsp snippets from the completion engine. I do at least. 

1. Less clutter in the available completions
2. A more purposeful and direct approach of using snippets in the code. Typing `ll<c-j>` for example is a very fast way to define a `lua` variable.

This extra provides a quick start for `mini.snippets` in its current state. When sources for `nvim-cmp` or `blink.cmp` become available, `mini.snippets` would likely be connected to the completion engine in the same manner as seen in the [luasnip](https://github.com/LazyVim/LazyVim/blob/d0c366e4d861b848bdc710696d5311dca2c6d540/lua/lazyvim/plugins/extras/coding/luasnip.lua#L31) extra. 
Also, when the snippets are offered by the completion engine, there could be a way to disable `vim.ui.select` from triggering on `expand`?

For now, I thought it best to only implement the "expand" part, and let other parts be completely handled by `mini.snippets`. 

Do you agree? Maybe you have a suggestion to modify the comment at the start of the `extra` to be more clear.

Is it correct that you are considering an `lsp` solution to expose `mini.snippets` to the completion engine? In that case, separate sources for `blink.cmp` or `nvim-cmp` might not be needed. One would lose the ability to prioritize `sources` though.

echasnovski: > Also, when the snippets are offered by the completion engine, there could be a way to disable `vim.ui.select` from triggering on `expand`?

There are four steps for users/config/distro to check that, so probably possible.

> For now, I thought it best to only implement the "expand" part, and let other parts be completely handled by `mini.snippets`.
> 
> Do you agree? Maybe you have a suggestion to modify the comment at the start of the `extra` to be more clear.

I don't have many opinions here, to be honest. Just that there certainly are people who prefer to have snippets as completion suggestions even if they don't come from LSP server. I am not one of those people. What's interesting for me is to make sure that 'mini.snippets' has all it needs to export for this to be implemented.

> Is it correct that you are considering an `lsp` solution to expose `mini.snippets` to the completion engine?

Yes, this is one of the next goals together with 'mini.completion' supporting LSP snippets.

abeldekat: > I don't have many opinions here, to be honest. Just that there certainly are people who prefer to have snippets as completion suggestions even if they don't come from LSP server. I am not one of those people. What's interesting for me is to make sure that 'mini.snippets' has all it needs to export for this to be implemented.

I'm very sorry for not understanding the statement above completely... Would you like me to assign the `draft` status to this PR, perhaps until those sources have been written?
For `nvim-cmp`, something similar as `cmp_ luasnip`(written by the author of `luasnip`). For `blink.cmp` there is already an [open](https://github.com/Saghen/blink.cmp/issues/741) issue.

echasnovski: > I'm very sorry for not understanding the statement above completely... Would you like me to assign the `draft` status to this PR, perhaps until those sources have been written?
> For `nvim-cmp`, something similar as `cmp_ luasnip`(written by the author of `luasnip`). For `blink.cmp` there is already an [open](https://github.com/Saghen/blink.cmp/issues/741) issue.

Please don't. It was a drive-by comment describing what should be possible and might be valuable for some people. I think having only 'mini.snippets' `insert` functionality used as snippet engine for other completion engines is very valuable in itself. Thanks for making this :+1:

abeldekat: > Thanks for making this

It's my pleasure!
Could you have a look at the latest commit(a0885fc3f1f7)? I changed the comment in the first block a bit to hopefully express the current state of the extra more clearly.

echasnovski: > Could you have a look at the latest commit([a0885fc](https://github.com/LazyVim/LazyVim/commit/a0885fc3f1f76ef7dc00c669f882347bada26c44))? I changed the comment in the first block a bit to hopefully express the current state of the extra more clearly.

Sure, looks good. I don't think it *has* to be this detailed and formal for Folke, but having extra clarity for him to trim down from might be indeed useful.

abeldekat: Thanks for the review!

dpetka2001:
Couldn't this be consolidated in one file and differentiate the behavior with a global variable or something? The `ai` Extras, for example, use `vim.g.ai_cmp` to have completion via sources in `nvim-cmp`/`blink.cmp` or via ghost text. I believe something similar could be done here, no??

abeldekat:
> Couldn't this be consolidated...

Thanks @dpetka2001!

I think you are right and will change accordingly. Initially, I thought the code would be messy with too many "if" statements.

dpetka2001:
That's the only thing I can comment on really, since I'm not familiar with either `mini.snippets` internals or snippets in general. You even created a source for `nvim-cmp`. Great job on that :smile: 

I'll trust your judgement and Folke's as the final reviewer and will definitely be looking forward to testing this once it's done. 

One thing I noticed when I tested with your initial commit (the standalone version at the time), is that when you're in a snippet session auto-pairs don't work. Do you happen to know if maybe in the future this will be supported? I don't use it right now and haven't really tested it in length, so am reluctant creating an issue about potential features when I don't really use the plugin yet. 

Also, happy new year and hope you and your family are healthy :smiley:

echasnovski:
> One thing I noticed when I tested with your initial commit (the standalone version at the time), is that when you're in a snippet session auto-pairs don't work.

Autopairs from 'mini.pairs' work as expected when inside 'mini.snippets' session. At least in my setup. So this is either something in LazyVim or your config in particular.

abeldekat:
Thanks @dpetka2001, and a happy new year!

I am having lots of fun creating this extra and the `cmp_mini_snippets`. Thank you for trying it out. Hoping you'll use lots of snippets in 2025...).

I also did not notice problems with autopairs in stock `LazyVim` using native mini.pairs.

@echasnovski, I was a bit hesitant to say this given the situation in your country, but I do like to wish you the best for 2025.

dpetka2001:
@echasnovski Also best wishes from me for 2025.

I should have been more clear. When testing a `function` snippet for example then yes auto-pairs works as expected. When using the `lua_ls` snippet placeholders when auto-completing, for example `vim.api.nvim_create_autocmd(event, opts)`, then typing `"` when you are on the event placeholder only puts one `"` at the place of the placeholder. If I do `Backspace` and then type `"` again then I get the normal auto-pairs output `""`.

echasnovski:
> @echasnovski Also best wishes from me for 2025.

:pray: And likewise to both of you!

> I should have been more clear. When testing a `function` snippet for example then yes auto-pairs works as expected. When using the `lua_ls` snippet placeholders when auto-completing, for example `vim.api.nvim_create_autocmd(event, opts)`, then typing `"` when you are on the event placeholder only puts one `"` at the place of the placeholder. If I do `Backspace` and then type `"` again then I get the normal auto-pairs output `""`.
> 
> I tested this with fresh LazyVim installation to ensure no personal configuration is taken into account. I'm not saying this is `mini.snippets` fault. It might as well be something in default LazyVim configuration.

Here is how it looks with 'mini.snippets' and 'mini.pairs' for me:

https://github.com/user-attachments/assets/350ee92a-663a-42ec-9d35-31decd24e9ee

dpetka2001:
thanks @echasnovski. I looked into LazyVim default configuration and it seems to be because of [this line](https://github.com/LazyVim/LazyVim/blob/d0c366e4d861b848bdc710696d5311dca2c6d540/lua/lazyvim/plugins/coding.lua#L22). Commenting this out, does indeed have the same behavior as in your video.

dpetka2001:
Hey @abeldekat why doesn't `LazyVim.cmp.actions.snippet_stop` return anything? This makes the `esc` mapping which also stops the snippet session [here](https://github.com/LazyVim/LazyVim/blob/d0c366e4d861b848bdc710696d5311dca2c6d540/lua/lazyvim/config/keymaps.lua#L51) not work with regards to stopping the snippet session and you have to explicitly press `<C-c>`. 

I tried locally the following 
```lua
LazyVim.cmp.actions.snippet_stop = function()
        vim.schedule(function()
          MiniSnippets.session.stop()
        end)
      end
```
and it seems to be working. Could you also check? Without `vim.schedule` I was getting an error. Maybe there's some other side-effect that I'm not able to think of (not familiar with `mini.snippets` at all like I've already mentioned)?

abeldekat:
This is by design.

:h MiniSnippets, line 296:

- Stop session manually by pressing <C-c> or make it stop automatically:
  if final tabstop is current either make a text edit or exit to Normal mode.
  If snippet doesn't explicitly define final tabstop, it is added at the end
  of the snippet.

You can also explicitly stop with "<c-c>" in insert mode

abeldekat:
> The important thing to understand here is that "stop snippet session" and "exit into Normal mode" are independent actions (if current tabstop is not final)

The intention is that the `<esc>` mapping does *not* stop the session. With `<esc>` one can do intermediate normal mode edits, before going back to insert mode and still be in the snippet session.

See also: https://github.com/echasnovski/mini.nvim/issues/1428#issuecomment-2564302617

Yes, it can be done, but I think it's important to keep the config for the plugin close to how it's intended.

dpetka2001:
> The intention is that the <esc> mapping does not stop the session. With <esc> one can do intermediate normal mode edits, before going back to insert mode and still be in the snippet session.

Ohhh, ok I see what you mean. Admittedly, I hadn't thought about this workflow. Now that you explained to me, I can see how this can be desirable. I never thought doing intermediate normal edits and still remain in snippet session. 

Thank you for taking the time to explain this and I hope I didn't sound too presumptuous. I just failed to think of the workflow that you mentioned.

abeldekat:
> ... and I hope I didn't sound too presumptuous

Not at all. I very much appreciate your help here. I was in the same boat, not understanding the workflow.

abeldekat:
@dpetka2001 

Many thanks for the vim.g suggestion! I think the code is ready to merge. I also took the opportunity to remove "restrictions". The extra is implemented to operate as close as possible to `mini.snippets` design.

If you have the time, could you do a quick comparison using the `luasnip` extra? I am especially curious if there are perceivable differences in performance when opening the menu, selecting a snippet( including the docs of the snippet) and expanding the snippet.

echasnovski:
> The intention is that the `<esc>` mapping does _not_ stop the session. With `<esc>` one can do intermediate normal mode edits, before going back to insert mode and still be in the snippet session.
> ...
> Ohhh, ok I see what you mean. Admittedly, I hadn't thought about this workflow. Now that you explained to me, I can see how this can be desirable. I never thought doing intermediate normal edits and still remain in snippet session.

Just to be clear, the behavior of not stopping snippet session(s) on `<Esc>` (or exit to Normal mode in general) is the *default* choice of 'mini.snippets'. Users/distros are of course free to do whatever they want. There are couple of caveats:
- `MiniSnippets.session.stop()` stops only single active session. If it was nested, then it will resume previous session. To force stop all sessions, use [this approach](https://github.com/echasnovski/mini.nvim/blob/782969a40d2870b1d7cd7c9db8f965d8aa9705ef/doc/mini-snippets.txt#L1137-L1139).
- For a more general "stop on Normal mode exit" approach see [this comment](https://github.com/echasnovski/mini.nvim/issues/1428#issuecomment-2563668092).
- Inside expression mapping `MiniSnippets.session.stop` indeed needs to be called with `vim.schedule`. The reason is quite technical with TL;DR like "modifying buffer text is not allowed inside expression mappings and `session.stop()` can't avoid 'modifying' text although it is not visible".

dpetka2001:
@abeldekat I just did a minimal test between the 2 side by side on 2 different Neovim profiles (each activated with one of the 2 snippet engines) and I didn't observe significant differences other than the visual selection between the 2 and the fact that pressing `jump(1)/jump(-1)` in `mini.snippets` cycles through the tabstops, whereas in Luasnip when it reaches the last tabstop, it doesn't cycle but inserts a tab character (probably because Luasnip stops the snippet session even if the last tabstop is empty whereas `mini.snippets` doesn't stop the session if the last tabstop is empty unless manually stopped)?? Menu, selecting snippet, docs and expanding look the same with the minor difference that I mentioned. But keep in mind that I have little experience using snippets and in programming as well and so my feedback should be taken as equivalent to a novice hobbyist who learns programming. I'm sure other people with more experience would provide better substantial feedback.

@echasnovski Thank you very much for explaining. Glad to know about the nested snippet sessions and how to correctly stop them. Like I've said already, I wasn't aware of the intentional behavior and the intended workflow until @abeldekat explained it to me (I have little experience with snippets and programming myself). And thanks for also explaining why the `vim.schedule` is actually needed in case someone would want to implement that behavior.

abeldekat:
> Just to be clear, the behavior of not stopping snippet session(s) on <Esc> (or exit to Normal mode in general) is the default choice of 'mini.snippets'. Users/distros are of course free to do whatever they want. There are couple of caveats:

I think it's important to stick with the default choices as much as possible. The plugin is an alternative to `luasnip`. However, it's not *equal* to `luasnip`. 

That said, both @dpetka2001 and I, initially did not grasp the intend of `<esc>`. I do expect `LazyVim` users to ask questions in that regard...).

Users, however unfortunate, might not read `:h MiniSnippets` that closely. I would like to add for example this [link](https://github.com/echasnovski/mini.nvim/blob/main/readmes/mini-snippets.md#expand) to the `LazyVim` docs. That section provides easy-to-grasp info:

> Using snippets is done via what is called "expanding". It goes like this:
 
I would suggest to also add the intended usage of `<esc>` to that section.

echasnovski:
> I didn't observe significant differences other than the visual selection between the 2

It is [described here](https://github.com/echasnovski/mini.nvim/blob/782969a40d2870b1d7cd7c9db8f965d8aa9705ef/doc/mini-snippets.txt#L96-L98) and is a bit bigger difference than just visualization.

> the fact that pressing `jump(1)/jump(-1)` in `mini.snippets` cycles through the tabstops, whereas in Luasnip when it reaches the last tabstop, it doesn't cycle but inserts a tab character (probably because Luasnip stops the snippet session even if the last tabstop is empty whereas `mini.snippets` doesn't stop the session if the last tabstop is empty unless manually stopped)??

It is [described here](https://github.com/echasnovski/mini.nvim/blob/782969a40d2870b1d7cd7c9db8f965d8aa9705ef/doc/mini-snippets.txt#L99-L102). Manually stopping also works but [autostopping is usually enough](https://github.com/echasnovski/mini.nvim/blob/782969a40d2870b1d7cd7c9db8f965d8aa9705ef/doc/mini-snippets.txt#L1025-L1033), i.e. exit to Normal mode or type another text when final tabstop is current.

There are [other differences](https://github.com/echasnovski/mini.nvim/blob/782969a40d2870b1d7cd7c9db8f965d8aa9705ef/doc/mini-snippets.txt#L81-L95), of course.

> > Using snippets is done via what is called "expanding". It goes like this:
> 
> I would suggest to also add the intended usage of `<esc>` to that section.

Both README and `:h MiniSnippets-overview` link to `:h MiniSnippets-session` which [has it mentioned](https://github.com/echasnovski/mini.nvim/blob/782969a40d2870b1d7cd7c9db8f965d8aa9705ef/doc/mini-snippets.txt#L1031-L1033). This is also the similar behavior to the session after `vim.snippet.expand()` (in which `<Esc>` doesn't end the session), so with time should be more expected.

abeldekat:
> Both README and :h MiniSnippets-overview link to :h MiniSnippets-session which [has it mentioned](https://github.com/echasnovski/mini.nvim/blob/782969a40d2870b1d7cd7c9db8f965d8aa9705ef/doc/mini-snippets.txt#L1031-L1033). This is also the similar behavior to the session after vim.snippet.expand() (in which <Esc> doesn't end the session), so with time should be more expected.

Thanks @echasnovski!

I intend to modify `lazydoc` like this:

```lua
if lazyvim_docs then
  -- set to `false` to prevent "non-lsp snippets"" from appearing inside completion popups
  -- motivation: less clutter in completion windows and a more direct usage of snippits
  vim.g.lazyvim_mini_snippets_in_cmp = true

  -- NOTE: Blink has an open issue to address mini.snippets integration: #741
  -- For now, blink ignores vim.g.lazyvim_mini_snippets_in_cmp, operating as if it where false.

  -- NOTE: Please read:
  -- https://github.com/echasnovski/mini.nvim/blob/main/readmes/mini-snippets.md#expand
  -- :h MiniSnippets-session
end

```

lua/lazyvim/plugins/extras/coding/mini-snippets.lua
```diff
@@ -138,12 +138,17 @@ return {
...
       snippet_select = snippet_select_for_blink
 
-      opts.sources.default = vim.tbl_filter(function(source) -- rm builtin snippets
-        if source == "snippets" then
-          return false
-        end
-        return true
-      end, opts.sources.default)
+      if snippets_in_cmp then
+        opts.sources.default = { "mini_snippets" }
```
dpetka2001, line 142: Won't this remove the rest of the sources like `lazydev` etc? I had done something similar for Luasnip PR and [this](https://github.com/LazyVim/LazyVim/pull/5214#discussion_r1894312158) was Folke's comment. Shouldn't we just do something like 
```lua
if snippets_in_cmp then
  table.insert(opts.sources.default, "mini_snippets")
  opts.sources.default = vim.tbl_filter(....)   -- remove blink's "snippets" source
end
```

abeldekat: Thanks @dpetka2001!

Perhaps returning the favour, in your LuaSnip [PR](https://github.com/LazyVim/LazyVim/pull/5214), below the comment "Luasnip sources for blink": I think that the `saghen/blink.cmp` block should be optional...

dpetka2001: Ohh great catch @abeldekat. Totally missed that :smile: 

Will fix.

abeldekat:
The blink PR landed in `main`. 

The setup for snippets now contains a [preset](https://github.com/Saghen/blink.cmp/blob/main/docs/configuration/snippets.md#snippets-) field.
Extra `luasnip` perhaps also needs adjusting.

abeldekat:
We need to wait for this [PR](https://github.com/Saghen/blink.cmp/pull/951)

abeldekat:
Ready for review...

For completeness I would like to mention this [issue](https://github.com/hrsh7th/nvim-cmp/issues/2119) in `nvim-cmp`.

abeldekat:
When using `mini.snippet` with `nvim-cmp`, completions can be suggested after initial snippet insert. This "clutters" the UI.

<details>

<summary>Video</summary>


https://github.com/user-attachments/assets/82e27030-edb8-40b6-935b-6c9c416c56dc


</details>

I found a solution using `require("cmp.config").set_onetime({ sources = {} })` and intend to commit the following:

<details>

<summary>mini_snippets.lua</summary>

```lua
if lazyvim_docs then
  -- Set to `false` to prevent "non-lsp snippets"" from appearing inside completion windows
  -- Motivation: Less clutter in completion windows and a more direct usage of snippits
  vim.g.lazyvim_mini_snippets_in_completion = true

  -- Using default mini.snippets, completion suggestions might appear directly after inserting a snippet
  -- This extra prevents that from happening.
  -- Motivation: A better display of the current snippet.
  -- Also, those completions do not appear when using luasnip or native snippets
  --
  -- Set to `false` to enable completion suggestions directly after inserting a snippet
  vim.g.lazyvim_mini_snippets_expand_insert_override = true

  -- NOTE: Please also read:
  -- https://github.com/echasnovski/mini.nvim/blob/main/readmes/mini-snippets.md#expand
  -- :h MiniSnippets-session
end

local include_in_completion = vim.g.lazyvim_mini_snippets_in_completion == nil
  or vim.g.lazyvim_mini_snippets_in_completion
local override_expand_insert = vim.g.lazyvim_mini_snippets_expand_insert_override == nil
  or vim.g.lazyvim_mini_snippets_expand_insert_override

--[[
Example override for your own config:
return {
  {
    "echasnovski/mini.snippets",
    opts = function(_, opts)
      -- By default, for opts.snippets, the extra for mini.snippets only adds gen_loader.from_lang()
      -- This provides a sensible quickstart, integrating with friendly-snippets
      -- and your own language-specific snippets
      --
      -- In order to change opts.snippets, replace the entire table inside your own opts

      local snippets, config_path = require("mini.snippets"), vim.fn.stdpath("config")

      opts.snippets = { -- override opts.snippets provided by extra...
        -- Load custom file with global snippets first (order matters)
        snippets.gen_loader.from_file(config_path .. "/snippets/global.json"),

        -- Load snippets based on current language by reading files from
        -- "snippets/" subdirectories from 'runtimepath' directories.
        snippets.gen_loader.from_lang(), -- this is the default in the extra...
      }
    end,
  },
}
--]]

local function expand_from_lsp(snippet)
  local insert = MiniSnippets.config.expand.insert or MiniSnippets.default_insert
  insert({ body = snippet })
end

local function jump(direction)
  local is_active = MiniSnippets.session.get(false) ~= nil
  if is_active then
    MiniSnippets.session.jump(direction)
    return true
  end
end

---@type fun(snippets, insert) | nil
local expand_select_override = nil

---@type fun(snippet) | nil
local expand_insert_override = nil

return {
  -- disable builtin snippet support:
  { "garymjr/nvim-snippets", optional = true, enabled = false },
  -- disable luasnip:
  { "L3MON4D3/LuaSnip", optional = true, enabled = false },

  -- add mini.snippets
  desc = "mini.snippets(beta), a plugin to manage and expand snippets (alternative for luasnip)",
  {
    "echasnovski/mini.snippets",
    event = "InsertEnter", -- don't depend on other plugins to load...
    dependencies = "rafamadriz/friendly-snippets",
    opts = function()
      ---@diagnostic disable-next-line: duplicate-set-field
      LazyVim.cmp.actions.snippet_stop = function() end -- by design, <esc> should not stop the session!
      ---@diagnostic disable-next-line: duplicate-set-field
      LazyVim.cmp.actions.snippet_forward = function()
        return jump("next")
      end

      local mini_snippets = require("mini.snippets")
      return {
        snippets = { mini_snippets.gen_loader.from_lang() },
        expand = {
          select = function(snippets, insert)
            if expand_select_override then
              -- Close completion window on snippet select - vim.ui.select
              -- Needed to remove virtual text for fzf-lua and telescope, but not for mini.pick...
              expand_select_override(snippets, insert)
            else
              MiniSnippets.default_select(snippets, insert)
            end
          end,
          insert = function(snippet)
            if override_expand_insert and expand_insert_override then
              -- Prevent completion suggestions directly after snippet insert
              expand_insert_override(snippet)
            else
              MiniSnippets.default_insert(snippet)
            end
          end,
        },
      }
    end,
  },

  -- nvim-cmp integration
  {
    "hrsh7th/nvim-cmp",
    optional = true,
    dependencies = include_in_completion and { "abeldekat/cmp-mini-snippets" } or nil,
    opts = function(_, opts)
      if override_expand_insert then
        expand_insert_override = function(snippet)
          MiniSnippets.default_insert(snippet)
          require("cmp.config").set_onetime({ sources = {} })
        end
      end

      -- stylua: ignore
      opts.snippet = { expand = function(args) expand_from_lsp(args.body) end }

      -- Return early
      if include_in_completion then
        table.insert(opts.sources, { name = "mini_snippets" })
        return
      end

      -- Standalone --
      local function close_cmp()
        local cmp = require("cmp")
        -- stylua: ignore
        if cmp.visible() then cmp.close() end
      end
      expand_select_override = function(snippets, insert)
        close_cmp()
        MiniSnippets.default_select(snippets, insert)
      end
    end,
    -- stylua: ignore
    -- counterpart to <tab> defined in cmp.mappings
    keys = include_in_completion and { { "<s-tab>", function() jump("prev") end, mode = "i" } } or nil,
  },

  -- blink.cmp integration
  {
    "saghen/blink.cmp",
    optional = true,
    opts = function(_, opts)
      -- Return early
      if include_in_completion then
        opts.snippets = { preset = "mini_snippets" }
        return
      end

      -- Standalone --
      local function close_cmp()
        require("blink.cmp").cancel()
      end
      expand_select_override = function(snippets, insert)
        -- Schedule, otherwise blink's virtual text is not removed on vim.ui.select
        close_cmp()
        vim.schedule(function()
          MiniSnippets.default_select(snippets, insert)
        end)
      end
      --
      -- Blink performs a require on blink.cmp.sources.snippets.default
      -- By removing the source, the default engine will not be used
      opts.sources.default = vim.tbl_filter(function(source)
        return source ~= "snippets"
      end, opts.sources.default)
      opts.snippets = { -- need to repeat blink's preset here
        expand = expand_from_lsp,
        active = function()
          return MiniSnippets.session.get(false) ~= nil
        end,
        jump = function(direction)
          jump(direction == -1 and "prev" or "next")
        end,
      }
    end,
  },
}
```

</details>

Shall I proceed?
Effectively, when selecting any function snippet provided by the `lsp`, nvim-cmp will use that [onetime](https://github.com/hrsh7th/nvim-cmp/blob/8c82d0bd31299dbff7f8e780f5e06d2283de9678/lua/cmp/config.lua#L90), merging the onetime config with the global config. On my old laptop I don't notice a performance hit. Still, it's something to consider.

In this old [issue](https://github.com/hrsh7th/nvim-cmp/issues/865#issuecomment-1082034684) @hrsh7th suggests using `onetime`

folke:
Yep, all good for me :)

Still need to checkout `mini.snippets`.
Personally I'm pretty happy with `vim.snippet` and there's not really anything that I feel like missing.

folke:
btw, unrelated, but if you feel adventurous, I'm looking for testers for https://github.com/LazyVim/LazyVim/pull/5368
docs are still pretty much missing at this point though.

echasnovski:
> Personally I'm pretty happy with `vim.snippet` and there's not really anything that I feel like missing.

There are quite a few things missing at the moment: handling nested tabstops (`${2:$1}`) and nested session are big ones.

> btw, unrelated, but if you feel adventurous, I'm looking for testers for #5368 docs are still pretty much missing at this point though.

Now you're just trying to reimplement 'mini.nvim' :(

abeldekat:
> Personally I'm pretty happy with vim.snippet and there's not really anything that I feel like missing.

To me, the fact that `mini.snippets` does not use select mode is the biggest improvement in workflow and `ui` experience. Using `native snippets` or `luasnip`, when I type `vim.tbl_filter` and accept the `lsp` snippet, select mode is triggered. As a result diagnostics appear with warnings and errors I am about to resolve, as I am still in the process of using the snippet. Also, the code is dimmed.

I did not even know the editing power that comes with using `lsp` snippets before diving into `mini.snippets`. By default, the tabstops are visible. So now I `<c-h><c-l>` all the time...)

Also, nested sessions are powerful

> btw, unrelated, but if you feel adventurous, I'm looking for testers for https://github.com/LazyVim/LazyVim/pull/5368

Sure! I will participate after I finished beta testing `mini.snippets`...)

@echasnovski, what do you think about [this](https://github.com/LazyVim/LazyVim/pull/5274#issuecomment-2585185291) comment?

echasnovski:
> @echasnovski, what do you think about [this](https://github.com/LazyVim/LazyVim/pull/5274#issuecomment-2585185291) comment?

I haven't looked at it deeply (maybe yet) because repro setup and steps feels a bit too complicated. On the surface it still looks like 'nvim-cmp' / 'cmp-nvim-lsp' issue, so might be worth fixing it there.

abeldekat:
> On the surface it still looks like 'nvim-cmp' / 'cmp-nvim-lsp' issue, so might be worth fixing it there.

Not 100% sure, but:
When `luasnip` expands a snippet, mode changes to `select`. As such, there are no completions until mode changes to `insert`, which is the desired behavior.
I do agree that this could also be fixed in `nvim-cmp`, especially since the `blink` engine already seems to behave as expected. I have no idea how complicated such a fix will be.

In the meantime, if you don't object, I intend to commit the code including the `onetime` fix.

> because repro setup and steps feels a bit too complicated...

Not sure here. Do you mean:

> It's possible to use the code from this PR directly in your own configuration:
> Copy the code to lua/extras/coding/mini_snippets.lua in your config folder
> Restart nvim and enable the "user" extra for mini.snippets
> Restart nvim

echasnovski:
> In the meantime, if you don't object, I intend to commit the code including the `onetime` fix.

Of course not. If it fixes the issue without obviously bad side effects, then it should be fine. It is not up to me to make this decision, though :)

> Not sure here. Do you mean:

There is just a lot of code which I personally prefer to read in full first before trying to reproduce it locally. I might get to it eventually.

folke:
@echasnovski re:nested snippets
Right, LazyVim actually **fixes** LSP snippets so that `vim.snippet` can handle them.
This also includes nested snippets, so that's why I didn't really have any issues with this I guess :)
Either way, I'm going to look into `mini.snippet`.

And regarding the picker, it's been a lot of fun to build.
I was able to make it quite a bit faster than both telescope with fzf_native and FzfLua.

It supports the full fzf search syntax, with some extended field searches.
Uses treesitter highlights where it makes sense.
....

echasnovski:
> And regarding the picker, it's been a lot of fun to build.
> I was able to make it quite a bit faster than both telescope with fzf_native and FzfLua.

Noice! Well, that's not particularly hard to do for Telescope (even with fzf_native). But I guess it depends on what you mean by "faster" and how it is measured.

folke:
It's faster in every way.
The big difference as well is that the snacks pickers always shows matched / sorted results real-time.
So while the timings below are already considerable faster, the actual result list conversion is much / much faster.

For find files (using `fd`) in my home directory (1.7 million files)

- loading all files (finder + initial match)
  - snacks: 2.7s
  - telescope: 8.5s

Same, but with prefilled query ("lua" as test query)
- snacks: 4.5s (and important here is that the results are update live, so the list converges typically way faster than the total time)
- telescope: 9.4s

Now without prefilled, so all files and then pasting `lua` in the input box:
- snacks: 3s, but sorted results show realtime, and the list typically converges before it processed all files (so before the 3s mark)
- telescope: 5.5s (does not show live results before this)

folke:
As for fzf-lua (so basically fzf), snacks is faster for loading results.
Filtering is slightly faster.
Fzf uses a similar algorithm for real-time results converging.

echasnovski:
Regarding matching speed, it is not only about the raw numbers, but what is actually done behind the scenes. In extreme example, comparing complex matching algorithm accounting for 10+ parameters and exact substring matching is not really correct. I don't doubt that 'snacks.picker' has good fuzzy matching results, just that this should be accounted when interpreting benchmarks. Optimizing code goes only so far, the real speed optimization comes from actually doing less.

------

I've given picker from 'snacks.nvim' a quick go. Seems snappy overall indeed. Real time feedback about loading and matching is pretty (spinner and numbers update), just not something I find useful. There is very rare actual use of partial results visualization, especially when order of them can change. Actually *matching* during loading would indeed be faster, but not sure if 'snacks.picker' does it.

I was more curious about how it compares to 'mini.pick' (obviously). Using [this repo](https://github.com/Yggdroot/a_large_project) (as I always do for "big project" use case testing):
- Loading speed is indeed visibly faster in 'snacks.picker', but that is a known 'mini.pick' trade-off (there is an extra `vim.fn.tolower` applied to every item for faster matching ignoring case). 1M+ items to select from is usually an outlier than a common case. On 80K items ('torvalds/linux') the difference is barely noticeable.
- Matching speed (from pressing the key to fully finished matching) visually looks faster in 'mini.pick'. I think because its matching algorithm has more room for optimization, as it does slightly less during matching (here is [the description](https://github.com/echasnovski/mini.nvim/blob/89c9f7bb00ef0b90d28533038763469e06ea7f29/doc/mini-pick.txt#L869-L876) if you are curious).

One early alpha/beta testing feedback is that 'snacks.picker' doesn't work well with `set foldmethod=indent` (all items are behind folds).

------

Anyway, I am happy to discuss this elsewhere. This PR is about 'mini.snippets' :)

folke:
Just had a look at the `mini.pick` code and your way of sorting is pretty neat!
I calculate scores and keep a minheap of topk items since I allow arbitrary comparators. However I can still probably do something similar to what you do to further optimize it.

I think the biggest difference with the matcher is the fuzzy matching. The more advanced search syntax probably causes little overhead.

My fuzzy algo is roughly:
- initial match (forward pass)
- get all possible end positions
- loop over each end positions and do a backward pass to tighten the results
Looking at your code, my code actually gets that certain example correct that you have there in your comments.

But thinking about it, the same comment about the optimal match not being that relevant for your scoring sorting is for the snacks picker scoring probably also the case. Need to further look into this. 

And thanks for spotting that bug with the folds. Will fix.

lua/lazyvim/plugins/extras/coding/mini-snippets.lua
```diff
@@ -0,0 +1,187 @@
...
+
+  -- add mini.snippets
+  desc = "mini.snippets(beta), a plugin to manage and expand snippets (alternative for luasnip)",
+  {
+    "echasnovski/mini.snippets",
+    event = "InsertEnter", -- don't depend on other plugins to load...
+    dependencies = "rafamadriz/friendly-snippets",
+    opts = function()
+      ---@diagnostic disable-next-line: duplicate-set-field
+      LazyVim.cmp.actions.snippet_stop = function() end -- by design, <esc> should not stop the session!
```
folke, line 84: I still think it would be a good idea to set `config.mappings.stop = "<esc>"` so that you can get a similar (but better) behavior of using `<esc>` for stopping session(s) and finally exiting to normal mode.
It's also simple and visible enough for users to change that setting if they want.

Otherwise we're going to get lots of issues/questions regarding stopping of sessions I'm afraid...

echasnovski:
> * loop over each end positions and do a backward pass to tighten the results
>       Looking at your code, my code actually gets that certain example correct that you have there in your comments.

Yes, I know what you mean. Initial implementation also had a backward pass. But besides being not relevant to the final "score" I vaguely remember some issues with it. There might have been some sneaky counter example that required more complex logic for 100% result *or* simply that making it work with multibyte characters is tough (as simple `string.reverse` is not enough here).

lua/lazyvim/plugins/extras/coding/mini-snippets.lua
echasnovski: That's up to you, but having `<Esc>` as "stop session" mapping basically removes the ability to temporarily exit into Normal mode for quick edit. Not stopping on `<Esc>` is both useful *and* the behavior of built-in `vim.snippet`.

folke: ok, then lets leave it as is.
@abeldekat but can you add the settings (commented) to the config with a proper description?
Then at least it's in the docs and users that read the docs (0.001%) can change it themselves.

folke:
For sure, even my impl doesn't always get the most optimal result.
Initially I had an implementation with linear programming to get the proper result, but that was waaaaay too slow.
Might consider removing the backward pass. Didn't think about unicode stuff. That might indeed cause issues right now...

lua/lazyvim/plugins/extras/coding/mini-snippets.lua
abeldekat: Yes, will do.

abeldekat: See [this](https://github.com/LazyVim/LazyVim/pull/5274/commits/ba6364a6caabd34456915c2ba8ea183d914e2daa) commit.

I also added the example to override the config inside `lazyvim_docs`, and renamed vim.g.lazyvim_mini_snippets_expand_insert_override to vim.g.lazyvim_mini_snippets_override_expand_insert

The opts:

```lua

    opts = function()
      ---@diagnostic disable-next-line: duplicate-set-field
      LazyVim.cmp.actions.snippet_stop = function() end -- by design, <esc> should not stop the session!
      ---@diagnostic disable-next-line: duplicate-set-field
      LazyVim.cmp.actions.snippet_forward = function()
        return jump("next")
      end

      local mini_snippets = require("mini.snippets")
      return {
        snippets = { mini_snippets.gen_loader.from_lang() },

        -- Following the behavior of vim.snippets,
        -- the intended usage of <esc> is to be able to temporarily exit into normal mode for quick edits.
        --
        -- If you'd rather stop the snippet on <esc>, activate the line below in your own config:
        -- mappings = { stop = "<esc>" }, -- <c-c> by default, see :h MiniSnippets-session

        expand = {
          select = function(snippets, insert)
            -- Close completion window on snippet select - vim.ui.select
            -- Needed to remove virtual text for fzf-lua and telescope, but not for mini.pick...
            local select = expand_select_override or MiniSnippets.default_select
            select(snippets, insert)
          end,
          insert = function(snippet)
            -- Prevent completion suggestions directly after snippet insert
            local insert = expand_insert_override or MiniSnippets.default_insert
            insert(snippet)
          end,
        },
      }
    end,


```

abeldekat:
I updated the code because [this](https://github.com/hrsh7th/nvim-cmp/pull/2126) PR in `nvim-cmp` has been merged.

folke:
Merged, ty!

abeldekat:
Thanks for the merge! Do note that it will not work for `blink` just yet, because [this](https://github.com/Saghen/blink.cmp/pull/951) PR is still only on `main`.

abeldekat:
`blink.cmp` is now on `0.11`, which includes the PR needed to work with mini.snippets

PR #5275, opened 2024-12-30
feat(chezmoi): enhance fzf-lua chezmoi picker and add snacks.dasbhoard entry

vandalt:
## Description

This PR enhances the fzf-lua picker for chezmoi by using the built-in `files` picker and by opening the selected file with `ChezmoiEdit` and `vimcmd_entry` to parse the entry. This simplifies the code and adds icons and file preview.

The PR also replaces the config entry in snacks.dashboard with the chezmoi picker, similar to what was already implemented for dashboard-nvim

## Screenshots

Current picker:

image

New picker:

image

## Checklist

### DIFF
--- a/lua/lazyvim/plugins/extras/util/chezmoi.lua
+++ b/lua/lazyvim/plugins/extras/util/chezmoi.lua
@@ -55,6 +55,7 @@ return {
   },
   {
     "xvzc/chezmoi.nvim",
+    cmd = { "ChezmoiEdit" },
     keys = {
       {
         "<leader>sz",
--- a/lua/lazyvim/plugins/extras/util/chezmoi.lua
+++ b/lua/lazyvim/plugins/extras/util/chezmoi.lua
@@ -110,6 +110,27 @@ return {
       table.insert(opts.config.center, 5, projects)
     end,
   },
+  {
+    "folke/snacks.nvim",
+    optional = true,
+    opts = function(_, opts)
+      local chezmoi_entry = {
+        icon = " ",
+        key = "c",
+        desc = "Config",
+        action = pick_chezmoi,
+      }
+      local config_index
+      for i = #opts.dashboard.preset.keys, 1, -1 do
+        if opts.dashboard.preset.keys[i].key == "c" then
+          table.remove(opts.dashboard.preset.keys, i)
+          config_index = i
+          break
+        end
+      end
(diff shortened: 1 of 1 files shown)


qw457812:
@vandalt Hi there, thanks for your codes. Didn't know how to enable previewer via fzf-lua before.

In my case, the previewer only says "Unable to stat file xxx" when I enter nvim from ~/.config (just not home dir).

After I add `--path-style=absolute` in cmd, it works as your screenshots. Here are my [codes](https://github.com/qw457812/dotfiles/blob/003c8824789744a109abb227dc7613ef9f4a07e6/dot_config/nvim/lua/plugins/util/chezmoi.lua#L44) in case you interested.

Thanks again!

vandalt:
Hi,

The code you linked to is slightly different from the one in this PR. I think adding `cwd = os.getenv("HOME")` to your `opts` table from `vimcmd_entry` would fix the issue you are describing.

Regarding this PR, I would tend to prefer changing `cwd` over adding `--path-style=absolute` to stay consistent with the Telescope picker.

lalvarezt:
Hi,
wouldn't <leader>fz make more sense now, it would be functionally the same as <leader>fc (for config files) but for chezmoi files instead

vandalt:
I just rebased on main now that  #5429 has been merged!

> wouldn't <leader>fz make more sense now

This PR only aims to add functionality/features for `snacks.dashboard` and `fzf-lua` with chezmoi, so I think changing the mapping for the picker would be out of scope. That being said  I don't have a strong opinion on what the default mapping should be.

folke:
ty!

ReKylee:
Hi, idk if this has been merged into the latest releases yet but using snacks dashboard I still can't activate the chezmoi file picker, it uses the default config picker instead.
image

dpetka2001:
@ReKylee It hasn't yet. The last release is from [2025-02-15](https://github.com/LazyVim/LazyVim/releases/tag/v14.14.0) while this was merged on Feb 21 2025. You'll have to wait until the next stable release or just follow LazyVim HEAD as per the [docs](https://www.lazyvim.org/configuration/lazy.nvim) if you'd like.

t40mas:
Thank you for your work!
Unfortunately I noticed strange behavior.

Open config from dashboard results in this:
image

After change to exclude hidden files:
image

I can't find any alias or something that would override any command.

PR #5288, opened 2025-01-01
Add an option to prevent LSP keymaps from being applied to specified filetypes

garymjr:
## Description

Currently LSP keymaps overwrite all other keymaps due to how they're loaded. This change adds the ability to prevent LSP keymaps from being applied to specific buffers of specified filetypes

## Checklist


folke:
Thanks for the PR! However, this is not the right approach. 

If you need to prevent specific keymaps from being applied, you should handle it in your own configuration by overriding specific keymaps or using the existing LSP keymap system. The LSP keymaps can already be customized per-server through the `keys` field in server configs.

Adding a global filetype ignore list adds unnecessary complexity to the core system. If there's a specific use case where LSP keymaps are conflicting, it should be addressed differently (e.g., fixing the conflicting keymaps or making them conditional).

―  

PR #5319, opened 2025-01-06
fix(nlua): debugger not responding

willheryanto:
## Description

Reference: https://github.com/jbyuki/one-small-step-for-vimkind/issues/45#issuecomment-2125749906

Fix nlua debug adapter not responding, which for some reason because of `fzf-lua`.

## Screenshots

No screenshot instead I'm providing how to reproduce with assumption `dap.nlua` extra is enabled. Steps:

1. Launch the nlua `:lua require'osv'.launch({port=8086})`

2. Attach from any neovim instance to the debugger

   1. Run the debugger
      - Can do `<leader>dc`; or
      - Can do `:lua require'dap'.continue()`
   2. Choose `Attach to running Neovim instance (port = 8086)`

3. Prior to this changes, debugger will got timeout like the following

   ```text
   Debug adapter didn't respond. Either the adapter is slow (then wait and ignore this) or there is a problem with your adapter or `nlua` configuration. Check the logs for errors (:help dap.set_log_level)
   ```

## Checklist

### DIFF
--- a/lua/lazyvim/plugins/extras/editor/fzf.lua
+++ b/lua/lazyvim/plugins/extras/editor/fzf.lua
@@ -45,8 +45,9 @@ return {
     "ibhagwan/fzf-lua",
     cmd = "FzfLua",
     opts = function(_, opts)
-      local config = require("fzf-lua.config")
-      local actions = require("fzf-lua.actions")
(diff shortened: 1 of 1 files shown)


folke:
wow, good find. ty!

PR #5397, opened 2025-01-14
Don't make AI plugins dependencies of completion plugins

danielwe:
## Description

It ought to be possible to enable completion without always enabling AI plugins. For example, some files have content that should never be sent to a remove server, but can still benefit from LSPs and other local completion sources.

Currently, this is not possible because AI plugins are added as dependencies of completion plugins, and dependencies can't be overridden by the user. This PR removes AI plugins as direct or indirect dependencies of completion plugins.

This makes no practical difference for people who stick to the default config, since AI plugins are configured to load on `event = "InsertEnter"`. However, it's now possible to override this with `event = false` and add whatever hook the user prefers instead.

NB! I only use codeium.nvim & blink.cmp myself, so I haven't tested the other combinations. I don't know how nvim-cmp feels about being configured with sources that aren't loaded (blink.cmp seems perfectly happy with this). The copilot & nvim-cmp config is especially intricate. Appreciate expert feedback and volunteer testers!

## Related Issue(s)

Discussion: https://github.com/LazyVim/LazyVim/discussions/5396

## Checklist


danielwe:
Would still love to have this merged or hear some feedback

danielwe:
still interested in this

danielwe:
let’s keep this open until maintainers have had a chance to comment

folke:
I don;t get it. Just don't install the AI plugin? Of mark it as disabled?

danielwe:
The use case is that I want to have the AI plugin installed and available, but not always loaded. For example, I might want it to load on a custom `:ActivateAI` command. The pretext here was realizing that it had been active while I was editing a rather sensitive file. However, that doesn't mean I want to disable it for all the other work I do.

folke:
in your own config, you can add an `enabled()` function to the specs you want to load not always.
no other changes needed

PR #5432, opened 2025-01-19
feat(copilot-chat): add Snacks picker integration for copilot-chat

tku137:
Adds support for using Snacks picker when picking prompt actions.

## Description

This adds support to use snacks.nvim picker to choose prompt actions in CopilotChat.nvim including prompt preview. 

Will need to wait till https://github.com/CopilotC-Nvim/CopilotChat.nvim/pull/720 is merged!
Edit: its merged now.

## Screenshots

image

## Checklist


tku137:
https://github.com/CopilotC-Nvim/CopilotChat.nvim/pull/720 is merged, this can now be reviewed.

qw457812:
Hi @tku137 ,

I think `pcall(require, "snacks")` is always true. Maybe `LazyVim.pick.want() == "snacks"` or `LazyVim.pick.picker.name == "snacks"` should be used instead.

tku137:
> Hi @tku137 ,
> 
> I think `pcall(require, "snacks")` is always true. Maybe `LazyVim.pick.want() == "snacks"` or `LazyVim.pick.picker.name == "snacks"` should be used instead.

Good call, thanks! I refactored the code to dynamically use the globally configured picker using `LazyVim.pick.want()`, since this should reflect what the user actually configured based on configuration, while `LazyVim.pick.picker.name` could be something else in early stages?

qw457812:
Thanks for accepting the suggestion. I noticed a bug, the LazyVim picker is `"fzf"`, but the CopilotChat integrations is `"fzflua"`.

I think picker verification can be omitted since LazyVim uses fzf as default picker [here](https://github.com/LazyVim/LazyVim/blob/83bf6360a1f28a3fc1afe31ae300247fc01c7a90/lua/lazyvim/util/pick.lua#L51-L51) and installs it [here](
https://github.com/LazyVim/LazyVim/blob/83bf6360a1f28a3fc1afe31ae300247fc01c7a90/lua/lazyvim/plugins/editor.lua#L388-L393).

tku137:
Again, thanks for the suggestion, good catch! Sorry for not paying attention here. 

However, I am not sure about omitting the verification completely. I agree that [this](https://github.com/LazyVim/LazyVim/blob/83bf6360a1f28a3fc1afe31ae300247fc01c7a90/lua/lazyvim/util/pick.lua#L51-L51) always returns `fzf` if neither `snacks` nor `telescope` are explicitly set via `vim.g.lazyvim_picker`. However, what would happen if a user does this for example to use mini.pick or whatever?

```
  {
    import = "lazyvim.plugins.extras.editor.fzf",
    enabled = false,
  },
```

As far as I understand, calling `require("CopilotChat.integrations.fzf").pick(items)` would fail in that case, because fzf is _not_ installed?

tku137:
If we just assume that fzf is always available, this is my suggestion:

```lua
local picker = LazyVim.pick.want()

-- Map LazyVim picker names to CopilotChat integration modules
local integration = ({
  telescope = "telescope",
  snacks = "snacks",
})[picker] or "fzflua"

require("CopilotChat.integrations." .. integration).pick(items)
```

dpetka2001:
I personally like your current implementation with explicit mappings that doesn't make any assumptions. You never know what users might do in their personal configuration. But that's just the opinion of a simple user like you.

qw457812:
Agree with you guys. Sorry for the noice and thanks for explaining!

Praczet:
Hi, 
I am just curious, would this work?
```lua
 local snacks_ok = pcall(require, "snacks.picker")
```

tku137:
Yes, but that would not mean that it is the active picker. `pcall ` will evaluate to `true` to every module that is installed, so if the user has telescope and fzf-lua and snacks picker installed, we still dont know what the user has configured as the active picker. `LazyVim.pick.want()` is the way :)

folke:
I did a big rework today and had actually removed that `want` function so this PR could no longer work.

Just fixed it in https://github.com/LazyVim/LazyVim/commit/2a608f00d47bb6679a27a313fb0404e4d3a2196c

PR #5440, opened 2025-01-20
fix(extras.project): open projects picker from dashboard with snacks backend

Silzinc:
## Description

Clicking on `p` on the dashboard, the projects picker opens, even with snacks' picker. This previously did nothing.

## Checklist


folke:
That would open the sacks project picker if you activate the projects.nvim extra?
Doesn't really make sense to me.
Those are two different plugins managing projects differently.

Silzinc:
Ok I see where I was wrong. Would it be possible to add the p shortcut on the dashboard with snacks though ? Currently, it does not exist for this picker.

PR #5443, opened 2025-01-21
feat(project): add Snacks picker

### LINKED ISSUE #5408: bug: snacks.picker breaks dashboard project feature

### Neovim version (nvim -v)

NVIM v0.10.2

### Operating system/version

MacOS 15.2

### Describe the bug

When the snacks.picker extra is enabled and the uitl.project extra is also enabled, the dashboard project functionality stops working (that is to say, pressing "p", does nothing). 

LIkewise its a little awkward that with snacks.picker the project search shortcut moved from <leader>f -> p to <leader>q -> p 

### Steps To Reproduce

Fresh install of lazyvim
enable utils.project extra
enabled snacks.picker extra
quit, restart to open dashboard, press p 

### Expected Behavior

The project picker/search (via snacks.picker) is opened

### Repro

```lua

```

dpetka2001:
## Description
Adds Snacks picker for `util.project` Extra.

One thing I noticed is that when pressing `<CR>` for the `confirm` action, it opens the `files` picker correctly, but there's a slight afterimage of the previous picker closing. I even tried to launch the picker with `vim.schedule` but it was the same behavior. If you have any idea as to why this happens and you know how to fix it, then please go ahead and make any necessary changes.

If you think this PR isn't needed because of Snacks own project picker, feel free to close this obviously.

## Related Issue(s)
Fixes #5408

## Checklist

### DIFF
--- a/lua/lazyvim/plugins/extras/util/project.lua
+++ b/lua/lazyvim/plugins/extras/util/project.lua
@@ -79,6 +79,121 @@ pick = function()
     }
 
     fzf_lua.fzf_exec(results, opts)
+  elseif LazyVim.pick.picker.name == "snacks" then
+    local project = require("project_nvim.project")
+    local history = require("project_nvim.utils.history")
+    local results = history.get_recent_projects()
+    ---@type snacks.picker.Config
+    Snacks.picker({
+      layout = {
+        layout = {
+          box = "horizontal",
+          width = 0.8,
+          height = 0.8,
+          {
+            box = "vertical",
+            border = "rounded",
+            title = "Project picker",
+            { win = "input", height = 1, border = "bottom" },
+            { win = "list", border = "none" },
+          },
+        },
+      },
+      win = {
+        input = {
+          keys = {
+            ["<c-s>"] = { "live_grep", mode = { "i", "n" } },
+            ["<c-r>"] = { "oldfiles", mode = { "i", "n" } },
+            ["<c-t>"] = { "tabedit", mode = { "i", "n" } },
+            ["<c-w>"] = { "change_dir", mode = { "i", "n" } },
+            ["<c-d>"] = { "delete_project", mode = { "i", "n" } },
+          },
+        },
+      },
+      finder = function()
+        local items = {}
+        for _, item in ipairs(results) do
+          items[#items + 1] = {
+            file = item,
+            text = item,
+          }
+        end
+        return items
+      end,
+      format = function(item, _)
+        local file = item.file
+        local ret = {}
+        local a = Snacks.picker.util.align
+        local icon, icon_hl = Snacks.util.icon(file.ft, "directory")
+        ret[#ret + 1] = { a(icon, 3), icon_hl }
+        ret[#ret + 1] = { " " }
+        ret[#ret + 1] = { a(file, 20) }
+
+        return ret
+      end,
+      actions = {
+        confirm = function(_, item)
+          ---@diagnostic disable-next-line: missing-fields
+          Snacks.picker.pick("files", {
+            cwd = item.file,
+            on_show = function()
+              vim.schedule(function()
+                vim.api.nvim_feedkeys(vim.api.nvim_replace_termcodes("i", true, false, true), "n", false)
+              end)
+            end,
+          })
+        end,
+        tabedit = function(_, item)
+          vim.cmd("tabedit")
+          ---@diagnostic disable-next-line: missing-fields
+          Snacks.picker.pick("files", {
+            cwd = item.file,
+            on_show = function()
+              vim.schedule(function()
+                vim.api.nvim_feedkeys(vim.api.nvim_replace_termcodes("i", true, false, true), "n", false)
+              end)
+            end,
+          })
+        end,
+        live_grep = function(_, item)
+          Snacks.picker.pick("grep", {
+            cwd = item.file,
+            on_show = function()
+              vim.schedule(function()
+                vim.api.nvim_feedkeys(vim.api.nvim_replace_termcodes("i", true, false, true), "n", false)
+              end)
+            end,
+          })
+        end,
+        oldfiles = function(_, item)
+          ---@diagnostic disable-next-line: missing-fields
+          Snacks.picker.pick("recent", {
+            filter = { cwd = item.file },
+            on_show = function()
+              vim.schedule(function()
+                vim.api.nvim_feedkeys(vim.api.nvim_replace_termcodes("i", true, false, true), "n", false)
+              end)
+            end,
+          })
+        end,
+        change_dir = function(picker, item)
+          local path = item.file
+          picker:close()
+          local ok = project.set_pwd(path)
+          if ok then
+            LazyVim.info("Change project dir to " .. path)
+          end
+        end,
+        delete_project = function(_, item)
+          local path = item.file
+          local choice = vim.fn.confirm("Delete '" .. path .. "' project? ", "&Yes\n&No")
+          if choice == 1 then
+            history.delete_project({ value = path })
+          end
+          pick()
+        end,
+      },
+    })
   end
 end
 
@@ -180,5 +295,8 @@ return {
         key = "p",
       })
     end,
+    keys = {
+      { "<leader>fp", pick, desc = "Projects" },
+    },
   },
 }


folke:
I just pushed a bunch of changes to the snacks projects picker. See here: https://github.com/folke/snacks.nvim/issues/871#issuecomment-2629434228

So you can simplify this PR a lot. Just start `Snacks.picker.projects({projects = ...})` and add the folders from projects.nvim.

Is projects.nvim something you use yourself? If so, do you think there's anything missing in the snacks projects picker?

dpetka2001:
No, I don't use `project.nvim`. I was using the very basics of `Snacks.picker.projects()` (with changing the default `confirm` action and adding a `grep` action also, because I wasn't accustomed for `confirm` action to immediately load the session and change the current working directory when I had other buffers open. In that regard maybe a better workflow would be to use a separate Neovim tab and open `Snacks.picker.projects()` there and load the session and only change `cwd` for the current tab using `tcd`. 

I just created this because of the other 2 issues and 2 PRs that tried to add `Snacks.picker.projects()` for the dashboard keymap. You closed those with the reason being they are 2 different plugins and I totally agree with that. 

I don't want to change this PR as I still think they are 2 different things. This provides just a custom picker for the Extra just like what was done with fzf-lua.

I don't really mind with disregarding this PR and I even mentioned in one of the issues that I'd be also in favor of removing this Extra (it being unmaintained). So, no opposition from me with regards to that, but of course other people might have different opinion and I can totally understand that.

I will personally look into the new changes you introduced into `Snacks.picker.projects` and don't bother with this Extra in my personal configuration. This PR is just something for users who would like to use the Extra instead.

folke:
tcd (and lcd etc) is a good idea. But I will need to probably change snacks to use vim's cwd. I use vim.uv.cwd everywhere.
Need to investigate.

dpetka2001:
Also, another minor nitpick unrelated to projects. In explorer, could we apply the same logic to `copy` action as the `move` action for selecting items? 

Currently a user has to be aware that the starting path is the item's current directory (instead of `cwd`) and do `../../some_other_dir/another_dir/file.lua`. Whereas when moving files with selecting them, you can go to input and search for the directory you want and then press `m` to move them there. The second approach that is currently applied in `move` action is much faster than actually having to write the whole relative path you want to copy the file. And also I haven't found a way to copy multiple files (but I might be doing something wrong). But looking at `copyfile` it seems to only support copying a single file. So, maybe we would have to rename the files manually but with only `vim.fn.rename` instead of `Snacks.rename` so as to not trigger the additional LSP stuff that `Snacks.rename` does (which I'd assume that it doesn't make sense to apply when copying files or maybe I'm wrong)?

folke:
Just added that. When there is no selection, it works like before (but also works for directories now).

If you have a selection, it's similar to move, so go to another file/directory and press `c`.

Would be great if you could do some testing as well, since copy files is harder than you would think it is :)

dpetka2001:
Yes, I did some testing in my personal configuration with copying/moving files around and it seems to be working as expected. I've been trying to make a slight modification in the `copy` action. If possible, I would like to restore the initial state of the tree before I did the copy. Because when I search for a directory to copy to the state of the tree changes. Is there a way to do what I want? I've tried to do `state:open(picker.init_opts.cwd)` after `state:update()` but that doesn't do anything.

folke:
You mean you don't want the newly copied directory to be opened?

dpetka2001:
I mean that I only want it to open for me to choose it and copy the files there, but after the copying operation is finished, I usually don't want that directory to be focused. I have to press `<BS>` until I get to the previous state that the explorer was before copying. Or close and open the explorer again to start anew. 

I actually think it's useful for the copied directory is opened but I can't see the rest because it's focused. 
Example: I open explorer in my personal configuration and I can see the root tree opened in `~/.config/nvim` and it shows the file of the buffer opened under `~/.config/nvim/lua/plugins/` directory. But if I copy 2-3 files from `~/.config/nvim/lua/plugins/` directory to another directory under `~/.config/nvim/dpetka2001/util/` directory, then the state of the tree is with that directory as the root of the tree instead of `~/.config/nvim` where I had opened Neovim and was the initial state of the tree.

I hope I didn't confuse you.

folke:
The directory should not get focused after a copy. That's the part I don't understand.
a copy doesn't focus the directory.
What are the exact steps to reproduce this?

dpetka2001:
Just select a file with `<tab>`, press `/` to focus input field and do a search for a directory and then `c` to copy it there. Here's also a screencast of what happens with a fresh LazyVim installation with no user configuration at all

[Screencast_2025-02-03-00-53-16.webm](https://github.com/user-attachments/assets/f9b72994-8303-47fa-be04-4d87fa56bb02)

dpetka2001:
Wait. I'm pressing `<Cr>` when I do the search. Does that change the `explorer_focus` to the copied directory? 

So, press `i`/`/` to focus input, do a search like `config` shown in the screencast and press `<CR>`. And then press `<M-w>` to focus list window and `c` to copy. Is the pressing of `<CR>` after the search what causes this?

folke:
Ha, yes indeed. From search, pressing a directory focusses it, since they are all open anyway and can't be closed from search.

dpetka2001:
I think it's [this code here](https://github.com/folke/snacks.nvim/blob/00905642eb42a9aaa9ac2d55623882fd31815102/lua/snacks/picker/source/explorer.lua#L611-L617). Is there a reason for setting the `cwd` to the new `item.file`? Were you trying to solve some issue opened in the Snacks repo?

dpetka2001:
I just tried to comment out `state:set_cwd(item.file)` and see how it behaves and didn't like it. So, I think that the current behavior is desirable except maybe if there was a way to revert to the previous state of the tree after the copying process.

But really this is very minor and I can just press `<BS>` to go up dirs or if it might be too nested from the root directory I just close and open the explorer again. So, it's not a big deal.

folke:
But why do you press `<cr>`? Just don't press that? :)

dpetka2001:
Will do just that. No worries.

dpetka2001:
Just tried to test this again with all the latest changes that you made in snacks. One thing I noticed, is that the pickers invoked via actions start in normal mode again. 

This used to be fixed and you can also see that in the second commit I made to remove the `feedkeys` for pressing `i` to get into insert mode because it had been fixed upstream. I guess some of the latest changes introduced that again. This also happens with the `Snacks.projects` picker.

folke:
Right, I indeed had to remove my hack since multiple pickers can now be open (explorer etc).

Will see for another work-around. Really annoying behavior of Neovim that they always stopinsert when leaving a prompt buffer. The issue is that if you start a new picker (new prompt buffer), then that startinsert gets cancelled.

folke:
Hey @dpetka2001! Thanks so much for all your amazing contributions to LazyVim - your work is always high quality and much appreciated! 🙏

After reviewing this PR and the discussion, I've decided not to merge it for a few reasons:

## Main concerns:

1. **Snacks already has project functionality** - Most LazyVim users will use the built-in `Snacks.picker.projects()` rather than installing the `util.project` extra. This makes the extra less relevant for the majority of users.

2. **project.nvim appears to be abandoned** - The last commit was over a year ago, and there are many open issues/PRs without maintainer response. We generally try to avoid extras for unmaintained plugins.

3. **Better location for this implementation** - If a Snacks picker integration is needed for project.nvim users, it would be better implemented in the project.nvim plugin itself (or a fork) rather than in LazyVim. This way all project.nvim users benefit, not just LazyVim users.

## What users can do:

Users who prefer project.nvim over Snacks projects can easily integrate the Snacks picker themselves using your code as a reference - it's well-written and would make a great example in their own configs!

Again, I really appreciate the effort you put into this and all your contributions. You're a valuable member of the community! 

―  

folke:
Claude :)

dpetka2001:
Yep, no problem with me. Don't use it anyway. 

I would also suggest you consider removing the `util.project` Extra as a whole. In my personal opinion it will only create unnecessary issues from users who use it in the future and will expect that it's properly configured, since it's included in LazyVim.

folke:
I agree, but seems to be used a lot https://github.com/search?q=lazyvim.plugins.extras.util.project&type=code

PR #5459, opened 2025-01-23
fix(keymaps): remove unneeded and ambiguous `<leader>w` keymap

### LINKED ISSUE #5456: bug: Potentially ambiguous keymaps in the windows group (`<leader>w`), leading to wrong command being executed

### Neovim version (nvim -v)

0.10.3

### Operating system/version

NixOS, MacOS

### Describe the bug

See also the relevant [discussion](https://github.com/LazyVim/LazyVim/discussions/4157), from which the video was taken from.

CleanShot 2024-07-24 at 08 34 26

When interacting with a newly-opened window, quickly pressing the `<leader>w` group of keybinds may lead to the wrong key sequence being executed, producing errors or just not doing anything.

As stated in the discussion linked above, I suspect it has something to do with how the keymaps are defined:

https://github.com/LazyVim/LazyVim/blob/7fb320700ad0162be33ce816f2ba6f9dda41883f/lua/lazyvim/config/keymaps.lua#L184-L190

The combination `<leader>wd` to delete a window is ambiguous, it could execute either `<C-W>c`, defined on line 188, which is the one we want, or it could fall through the general "Windows" case, defined on line 185, which wil execute `<C-W>d`. That keymap is actually `Show diagnostics under the cursor`, hence the `E388: Couldn't find definition` error that we see being thrown in the video. The same error is thrown when executing `<C-W>d` manually, supporting this theory.

This may be the case for other keymaps as well.

It may have something to with the "timeout" family of options `timeout`, `timeoutlen` and `ttimeout`, which in my config are `true`, `300` and `false`, respectively. Though I'm pretty sure `which-key`'s timeout is independent from nvim. The fact that the issue is kind of hard to reproduce may also be an indicator of this, since the speed with which you press the key sequence seems to be what causes it.

### Steps To Reproduce

1. Split a window.
2. Press `<leader>wd` very quickly, with little to no timeout between each key.
3. Error `E388: Couldn't find definition` may get thrown instead of the window being closed.

### Expected Behavior

1. Split a window.
2. Press `<leader>wd` very quickly, with little to no timeout between each key.
3. Window closes.

### Repro

```lua

```

iniw:
## Description

The `<leader>w` mapping, described as "Windows", was causing potentially timeout-related bugs with many other mappings that also started with `<leader>w`.

The problems caused by this are described in detail in issue #5456 and in discussion #4157.

@dpetka2001 Suggested removing it as an experiment, which I tried and found to work.

## Checklist

### DIFF
--- a/lua/lazyvim/config/keymaps.lua
+++ b/lua/lazyvim/config/keymaps.lua
@@ -182,7 +182,6 @@ map("t", "<C-/>", "<cmd>close<cr>", { desc = "Hide Terminal" })
 map("t", "<c-_>", "<cmd>close<cr>", { desc = "which_key_ignore" })
 
 -- windows
-map("n", "<leader>w", "<c-w>", { desc = "Windows", remap = true })
 map("n", "<leader>-", "<C-W>s", { desc = "Split Window Below", remap = true })
 map("n", "<leader>|", "<C-W>v", { desc = "Split Window Right", remap = true })
 map("n", "<leader>wd", "<C-W>c", { desc = "Delete Window", remap = true })


dpetka2001:
Just for clarification. I'm also not sure if that mapping is needed or not. Maybe I'm missing something here. I just thought that in the `which-key` spec which the same mapping is also defined to proxy the `<c-w>` mappings, maybe this one was not needed after all. Obviously maintainer knows better.

iniw:
> Just for clarification. I'm also not sure if that mapping is needed or not. Maybe I'm missing something here. I just thought that in the `which-key` spec which the same mapping is also defined to proxy the `<c-w>` mappings, maybe this one was not needed after all. Obviously maintainer knows better.

The commit that introduced this was [#bab54406](https://github.com/LazyVim/LazyVim/commit/bab54406), which doesn't go into too much detail on why the keymap is needed, but all the keybinds that were removed ("Other Window", "Split Window Below", etc.) still work even without `<leader>w` being mapped, presumably because of `which-key`, as you said.

dpetka2001:
No idea as I was not able to reproduce when pressing they keymap really fast. It would always close the window. I only observed another discrepancy which I also mentioned in the issue you created.

iniw:
> No idea as I was not able to reproduce when pressing they keymap really fast. It would always close the window. I only observed another discrepancy which I also mentioned in the issue you created.

I assum they're both caused by the `<leader>w` keymap. Could you checkout this PR and see if it fixes your issue too?

iniw:
> But I have also lowered which-key's delay to 50 myself. Don't know if that made any difference.

That does sound relevant.

minusfive:
Yeah, this solves the flakiness for me too.

iniw:
@folke Hi! Could you take a look at this PR now that you're back from vacation? It solves an annoying issue with LazyVim :)

folke:
The reason I added the regular keymap is for users that disabled which-key, but it indeed causes issues,
so better not to remap it. ty!

PR #5461, opened 2025-01-24
chore(which-key): remove unnecessary icons

minusfive:
## Description

These are already handled by which-key exactly the same way, no need to re-declare them here. Removing them from here has no effect.

## Related Issue(s)

N/A

## Screenshots

Screenshot 2025-01-23 at 21 21 14

## Checklist


folke:
ty!

PR #5470, opened 2025-01-25
fix(luasnip): schedule `jump` to be compatible with blink, nvim-cmp still works after change

### LINKED ISSUE #5469: bug: LuaSnip extra and blink.cmp don't work together

### Neovim version (nvim -v)

NVIM v0.10.3

### Operating system/version

Arch Linux with kernel version 6.12.10-arch1-1

### Describe the bug

When LuaSnip extra is enabled,  sometimes <Tab> doesn't jump to the next item in a snippet, instead, nothing will happen happened for a few presses, and then literal <Tab> appears.

### Steps To Reproduce

1. Run neovim with `repro.lua`, and enable LuaSnip extra
2. Open a Lua file
3. Type `for` and use the `for~` snippet
4. Change the default parameter by typing and press <Tab>
5. The first three items work fine, the <Tab> after doesn't go into the loop

### Expected Behavior

It should go into the loop.

### Repro

```lua

require("lazy.minit").repro({
  spec = {
    -- add any other plugins here
  },
})
```

dpetka2001:
## Description
Schedule `jump` in Luasnip `snippet_forward`, so that it doesn't get stuck before the last placeholder.

## Related Issue(s)
Fixes #5469

## Checklist

### DIFF
--- a/lua/lazyvim/plugins/extras/coding/luasnip.lua
+++ b/lua/lazyvim/plugins/extras/coding/luasnip.lua
@@ -30,7 +30,9 @@ return {
     opts = function()
       LazyVim.cmp.actions.snippet_forward = function()
         if require("luasnip").jumpable(1) then
-          require("luasnip").jump(1)
+          vim.schedule(function()
+            require("luasnip").jump(1)
+          end)
           return true
         end
       end


folke:
ty!

PR #5479, opened 2025-01-27
feat(general): add support for Ctrl + Backspace to delete a word

iambasantarai:
## Description

I've added a keymap to make Vim more accessible to newcomers by introducing support for deleting a word using `Ctrl + Backspace` in insert mode. My intent with this mapping was to replicate the behavior of `Ctrl + Backspace`, which is specifically used in many editors (e.g., VS Code, Sublime, IntelliJ) to delete entire words. This behavior is common in other editors and enhances the onboarding experience for users transitioning to Vim.

The keymap maps `<C-BS>` to Vim's built-in `<C-w>` command, which is already familiar to experienced Vim/Neovim users for deleting the previous word.

Even if this functionality already exists, I believe someone will find it helpful ;). 
Both `<C-BS>` and `<C-H>` are included to ensure compatibility, as key interpretation can vary across terminals, but either keymap can be removed accordingly based on the terminal's behavior.
```lua
map("i", "<C-BS>", "<C-w>", { desc = "Delete the previous word in insert mode" })
map("i", "<C-H>", "<C-w>", { desc = "Delete the previous word in insert mode" })
```

### More on key interpretation

We can check our terminal's interpretation by running `cat -v` in our terminal and pressing `Ctrl + Backspace`. 
> For example, If we see `^H` (`^` represents `Ctrl`), our terminal supports `<C-H>` version. In this case, we can remove the `<C-BS>` keymap.

## Checklist

### DIFF
--- a/lua/lazyvim/config/keymaps.lua
+++ b/lua/lazyvim/config/keymaps.lua
@@ -16,6 +16,10 @@ map("n", "<C-j>", "<C-w>j", { desc = "Go to Lower Window", remap = true })
 map("n", "<C-k>", "<C-w>k", { desc = "Go to Upper Window", remap = true })
 map("n", "<C-l>", "<C-w>l", { desc = "Go to Right Window", remap = true })
 
+-- Delete an entire word in insert mode with <ctrl> backspace
+map("i", "<C-BS>", "<C-w>", { desc = "Delete entire word in insert mode" })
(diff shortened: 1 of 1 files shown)


folke:
not interested in adding this, since like you say you can use `<c-w>`

PR #5447, opened 2025-01-22
feat(snacks.picker): added <leader>su to search undotree

minusfive:
## Description

As it reads on the title ;)

## Related Issue(s)

None.

## Screenshots

Screenshot 2025-01-22 at 14 02 30
Screenshot 2025-01-22 at 14 03 04

## Checklist

- [x] I've read the [CONTRIBUTING](https://github.com/LazyVim/LazyVim/blob/main/CONTRIBUTING.md) guidelines.

PR #5514, opened 2025-02-03
feat: LazyExtra for Quarto

jmbuhr:
## Description

Adds a `LazyExtra` for the Quarto computational notebook system (https://quarto.org/).

The `quarto` extra adds the `quarto-nvim` plugin (https://github.com/quarto-dev/quarto-nvim), which associates the quarto filetype with the markdown treesitter grammar for highlighting, provides a command to preview quarto  notebooks and enables LSP features in quarto code cells via the `otter.nvim` plugin (https://github.com/jmbuhr/otter.nvim) .

The `otter` extra is an optional extra to add an additional keyboard shortcut to activate `otter.nvim` to use the same LSP features in other documents that include injected code.

## Related Issue(s)

https://github.com/LazyVim/LazyVim/discussions/1367

## Screenshots

image

## Checklist


dpetka2001:
Just a simple user here with a couple recommendations. 
- Languages Extras also require a `recommended` field, please see other `lang` Extras (for example [python Extra](https://github.com/LazyVim/LazyVim/blob/fb256f2b688cb7ac9875f704fe6c00f27efc2354/lua/lazyvim/plugins/extras/lang/python.lua#L13-L25))
- It's not a good idea to use `wk.add` to add mappings, since it makes it difficult for users to overwrite them. See also #5447 as of late for example and the maintainer's response. This will also eliminate the use of `config` function, since it's mostly use for the `which-key` mapping as far as I can see.

jmbuhr:
Perfect, thanks! I was wondering about policies regarding keymappings.

Jacobus-afk:
<leader>qp might also not logically make sense for how lazyvim's keymaps are assigned currently (<leader>q is used for quit/session commands)

But I'm also just a normal user 😅

NaCl-Salt-12:
Is this PR dead or discontinued?

jmbuhr:
ready from my end

folke:
not worth it to put into (2?) extras. there's zero to no config in here. Check the contribution guidelines for more info

PR #5517, opened 2025-02-03
feat(extras): add extra for zoxide tabpage navigation

Demianeen:
## Description

I am not a fan of using tmux or relying on terminal tabs. So, I tried to move completely from depending terminal tabs and decided to try to use nvim tabpages for navigation between multiple projects and it worked quite well for me. So I wanted to share

The idea is that you can have multiple projects open in different tabpages in nvim each having it's own cd. Each tabpage is automatically named after it's cd:

https://github.com/user-attachments/assets/bf373e99-9d5c-481e-95cc-ff7aa559b015

There is also a bit longer [demo](https://share.cleanshot.com/HCrYgq1D)(~40s), but it feels less convoluted

### Example (My usecase)

Whenever I work I will have naturally my work dir open. Whenever I have some small fix or tweak in my mind for neovim or my nix config I used to open it in the same tabdir directly. But then my work tabline quickly became bloated and I needed to delete each buffer 1 by 1. But with tabpage I can just do what I need in new tab, <leader><tab>d, and continue with my day. Or even leave it open if I potentially will do something else there. + it can be handy to have some other project easily accessible when I want to copy something from it.

Neovim has easy way to navigate between different tabs, so it all feels very smooth,  `1gt` and you are in first tabpage, 3gt - in third.

telegram-cloud-photo-size-2-5276083580872811553-y

## Note

It currently has issue that if you exit from any other cwd rather than main one, the session wouldn't be saved in the main cwd but rather [in tabpage cwd](https://github.com/folke/persistence.nvim/issues/88).

I am also not sure how to structure autocmds properly in extras. I have seen that lazyvim autocmds have groups but most extras that I looked don't use groups. Does lazyvim uses them in extras?

## Related Issue(s)

N/A

## Checklist

EDIT: updated info after adding some commits


Demianeen:
.

Demianeen:
.

n-crespo:
This is cool~ (but is somewhat of a specific use case). I did something similar for my own workflow with snacks.picker:


```lua
    {
      "<leader>F",
      function()
        if vim.fn.executable("zoxide") ~= 1 then
          return vim.notify("Zoxide is not installed", vim.log.levels.WARN)
        end
        Snacks.picker.zoxide({
          title = "Load Session at...",
          -- load session at directory
          confirm = function(picker, item)
            picker:close()
            vim.fn.chdir(item._path)
            local session = Snacks.dashboard.sections.session()
            if session then
              vim.cmd(session.action:sub(2))
              vim.notify("Loading Session at: " .. vim.fn.fnamemodify(item._path, ":~"), "info")
            end
          end,
        })
      end,
      desc = "Load Session",
    },
    {
      "<leader>j",
      function()
        if vim.fn.executable("zoxide") ~= 1 then
          return vim.notify("Zoxide is not installed", vim.log.levels.WARN)
        end
        Snacks.picker.zoxide({
          title = "Jump picker to...",
          confirm = function(picker, item)
            Snacks.picker.files({
              cwd = item._path,
              title = vim.fn.fnamemodify(item._path, ":~"),
              hidden = true,
            })
            picker:close()
          end,
        })
      end,
      desc = "Jump with Zoxide",
    },
```

Demianeen:
@n-crespo yes, it is a bit of a specific use case. I think it can be useful for people who have 3-4 projects often at the same time. Then it can be useful

I am curious of your usage of the keymaps above, how do you use them? From what I see they open other dir/save in the same tabpage

Demianeen:
Unstale!

folke:
not interested in adding this

PR #5536, opened 2025-02-05
fix(rust): install bacon-ls

matadaniel:
mason-lspconfig.nvim does not have a mapping for bacon_ls

## Checklist

I've never written a line of rust code.


folke:
this will alraady be automatically installed

PR #5537, opened 2025-02-05
fix(rust): let rustaceanvim handle codelldb discovery

matadaniel:
rustaceanvim handles more cases

## Checklist

I've never written a line of rust code. I would appreciate it if people who have experience with rust could verify that the extra behaves as expected.

### DIFF
--- a/lua/lazyvim/plugins/extras/lang/rust.lua
+++ b/lua/lazyvim/plugins/extras/lang/rust.lua
@@ -110,18 +110,6 @@ return {
       },
     },
     config = function(_, opts)
-      if LazyVim.has("mason.nvim") then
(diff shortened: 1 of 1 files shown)


matadaniel:
rustaceanvim now also handles that case.

I think `LazyVim.get_pkg_path` could've been used here instead of mason's `get_install_path`, which is being removed. If that's the case then the `php` and `java` extras should probably be refactored to use `LazyVim.get_pkg_path`. I don't use those languages either. Somebody who does should verify the change and submit a PR.

phucisstupid:
any update?

phucisstupid:
no

folke:
outdated and probably no longer needed

PR #5540, opened 2025-02-06
feat(refactoring): fallback to using vim ui select for refactoring.nvim

rbhanot4739:
## Description

Fallback to using `require("refactoring").select_refactor()`  if neither of `telescope` or `fzf-lua` is installed. Since `select_refactor()` uses `vim.ui.select` internally, it can use `snacks.picker` if it is enabled.
## Checklist


folke:
ty!

PR #5570, opened 2025-02-10
fix(project): ensure keymap works with snacks

hiaw:
## Description

Currently the project keymap (\<leader\>fp) while using snacks is not working. This would fix it.

## Checklist


folke:
And that would open the snacks picker, not nvim-projects,
which makes no sense.

Just disable nvim-projects if you want to use the snacks picker.

PR #5594, opened 2025-02-13
feat(extras): setup neogen and mini.snippets integration

rkaminsk:
## Description

Configure neogen and mini.snippets integration.

See also #5593.

## Checklist

### DIFF
--- a/lua/lazyvim/plugins/extras/coding/neogen.lua
+++ b/lua/lazyvim/plugins/extras/coding/neogen.lua
@@ -17,6 +17,7 @@ return {
 
     local map = {
       ["LuaSnip"] = "luasnip",
+      ["mini.snippets"] = "mini",
       ["nvim-snippy"] = "snippy",
       ["vim-vsnip"] = "vsnip",
     }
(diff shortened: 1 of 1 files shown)


rkaminsk:
Please don't merge. This does not work. Apparently, requiring `mini.snippets` is not enough for neogen to work.

rkaminsk:
Using
```lua
dependencies = LazyVim.has("mini.snippets") and { "mini.snippets" } or {},
```
works.

dpetka2001:
Well, the `mini.snippets` docs mention
> Important: don't forget to call require('mini.snippets').setup() with non-empty snippets to have snippets to match from.

Now, I think we could get the `mini.snippets` opts with `LazyVim.opts("mini.snippets")` and manually call the setup, but the thing is that the `mini.snippets` has 2 other configuration to change some `LazyVim.cmp.actions` which I don't think will propagate to `LazyVim.opts`. 

My advice would be to use the ternary check in `dependencies`.

folke:
ty

PR #5602, opened 2025-02-14
feat(keymaps): enhance buffer navigation and deletion commands

MaJinjie:
In LazyVim, the original buffer navigation keybindings (e.g., <S-h> and <S-l>) only allow switching buffers one by one, making it impossible to directly jump to a specific buffer by count. This results in the swap buffer always being the adjacent buffer. To support jumping to a specific buffer by count and using the current buffer as the swap buffer, I proposed this PR.

## Description
- Updated `<S-h>`, `<S-l>`, `[b`, and `]b` mappings to support `v:count1` for navigating buffers with a count, improving flexibility.
- Modified `<leader>bo` to delete buffers not in the current tab layout, providing better control over buffer cleanup.
- Added `<leader>bO` as a new mapping to delete other buffers, complementing the existing buffer deletion commands.
- Improved descriptions for clarity and consistency across all buffer-related mappings.

## Checklist


dpetka2001:
I personally have `<leader>bb` mapped to `BufferLinePick` and also use the picker buffer if I have too many buffers open. 

Both seem like a better personal workflow compared to having to count to which buffer you want to jump to with `vim.v.count1`.

MaJinjie:
When working with a small number of buffers, count-based navigation (v:count1) makes it easy to determine the target buffer's position relative to the current buffer. 
However, with a larger number of buffers, both count-based navigation and <leader>bb mapped to BufferLinePick require first ensuring that the buffer is visible on the tabline and its orientation. In such cases, directly using snacks.picker.buffers is more convenient. 
Of course, having a shortcut to jump to the pinned buffer would be even better.

phucisstupid:
don't stale

folke:
not interested in adding this...

Issue #5601, opened 2025-02-14
bug: angular root directory not found


fegyi001:
### Did you check docs and existing issues?

- [x] I have read all the LazyVim docs
- [x] I have updated the plugin to the latest version before submitting this issue
- [x] I have searched the existing issues of LazyVim
- [x] I have searched the existing issues of plugins related to this issue

### Neovim version (nvim -v)

0.10.4

### Operating system/version

MacOS 15.3.1

### Describe the bug

My neovim config: https://github.com/fegyi001/nvim

I have an Nx angular monorepo. I wish to use the angular language server, added via Lazy Extras.
But when I check the LspInfo for an opened angular typescript file, I see that the root is not found.

<img width="779" alt="Image" src="https://github.com/user-attachments/assets/80ae2ce4-6d55-49ac-ade9-efc287af70ed" />

In my project I have an nx.json file, and it should be enough for the angular extras:
https://github.com/LazyVim/LazyVim/blob/main/lua/lazyvim/plugins/extras/lang/angular.lua

<img width="661" alt="Image" src="https://github.com/user-attachments/assets/9f45664b-b17c-49d0-abe0-20c80fefa033" />

Here is the output of my LspInfo:
```
==============================================================================
lspconfig: require("lspconfig.health").check()

LSP configs active in this session (globally) ~
- Configured servers: docker_compose_language_service, html, vtsls, taplo, tailwindcss, dockerls, helm_ls, jsonls, yamlls, marksman, ansiblels, emmet_ls, lua_ls, eslint, angularls
- OK Deprecated servers: (none)

LSP configs active in this buffer (bufnr: 9) ~
- Language client log: ~/.local/state/nvim/lsp.log
- Detected filetype: `typescript`
- 4 client(s) attached to this buffer
- Client: `eslint` (id: 1, bufnr: [9])
  root directory:    ~/code/nvr/nvr-frontend/libs/modules/ver/
  filetypes:         javascript, javascriptreact, javascript.jsx, typescript, typescriptreact, typescript.tsx, vue, svelte, astro
  cmd:               ~/.local/share/nvim/mason/bin/vscode-eslint-language-server --stdio
  version:           `?` (Failed to get version) Tried:
  `/Users/padanyig/.local/share/nvim/mason/bin/vscode-eslint-language-server --version`
  `/Users/padanyig/.local/share/nvim/mason/bin/vscode-eslint-language-server -version`
[...]

PR #5611, opened 2025-02-15
fix(angular): make angularls work for nx workspaces as well

fegyi001:
## Description
Nx workspaces are not supported out-of-the-box since the root dir is not found (in case of Nx workspaces there is no angular.json, but rather an nx.json in the root dir). With this modification the LSP works as intended.

## Related Issue(s)
https://github.com/LazyVim/LazyVim/issues/5601 

## Checklist

### DIFF
--- a/lua/lazyvim/plugins/extras/lang/angular.lua
+++ b/lua/lazyvim/plugins/extras/lang/angular.lua
@@ -31,7 +31,13 @@ return {
     "neovim/nvim-lspconfig",
     opts = {
       servers = {
-        angularls = {},
+        angularls = {
+          -- Make sure it works for Nx workspaces as well
+          root_dir = function(fname)
+            return require("lspconfig.util").root_pattern("nx.json", "angular.json")(fname)
+                or vim.loop.cwd()
+          end,
+        },
       },
       setup = {
         angularls = function()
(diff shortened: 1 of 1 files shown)


lua/lazyvim/plugins/extras/lang/angular.lua
```diff
@@ -31,7 +31,12 @@ return {
     "neovim/nvim-lspconfig",
     opts = {
       servers = {
-        angularls = {},
+        angularls = {
+          -- Make sure it works for Nx workspaces as well
+          root_dir = function(fname)
+            return require("lspconfig.util").root_pattern("nx.json", "angular.json")(fname) or vim.loop.cwd()
```
dpetka2001, line 37: Why is `vim.loop.cwd` needed here? Shouldn't the `root_dir` just be resolved based on the patterns? What will happen if it doesn't find a pattern? Will the LSP server still attach based on the `cwd`?

fegyi001: removed it, without it also works the root folder detection

dpetka2001: I don't do any webdev and like I've told you before, I'm just a simple user like you. So, this was more of a suggestion for you to test it out, since I can't test it.

About the review, I don't feel confident enough to officially approve it but it looks good to me. 

The maintainer will do the final review and will be the one to approve it. Good luck :smile:

folke:
I don't like this change. It's not because you have an `nx` workspace that angular is being used.

PR #5615, opened 2025-02-15
fix(blink): `cmdline` is now top-level on main branch

dpetka2001:
## Description
This `blink.cmp` [commit](https://github.com/Saghen/blink.cmp/commit/93215d80346e14763a67d97785dccb1e1c3a6775) introduced a breaking change on main branch. `cmdline` is now top-level and the option is `opts.cmdline.sources = {}`.

I also added a note to remove this hack when the next blink stable release comes out, which should be greater than current 0.11.0 and just do it normally in `opts` instead.

## Related Issue(s)
None.

## Checklist


folke:
ty!

PR #5617, opened 2025-02-15
fix(blink): enable Neovim native mapping with `<Tab>`

dpetka2001:
## Description
Another blink fix unfortunately. 

I noticed that native Neovim `<Tab>` completion didn't work after the breaking commit. 

I checked the commit and noticed it had a boolean `enabled` field, so I thought maybe that would disable the blink cmdline since we already have `sources = {}`, but it didn't work. Looking through the commit, I stumbled upon this piece of [code](https://github.com/Saghen/blink.cmp/commit/93215d80346e14763a67d97785dccb1e1c3a6775#diff-f313d6f0270f6fed848f1cb7e62c82d3cfaf9b6c7e0072266908eeaf202f11d5R66-R68) and it seems that it's only checked for applying the blink cmdline mappings. 

So, it seems both `enabled = false` (for Neovim native cmdline mapping) and `sources = {}` are needed to have default Neovim cmdline experience.

## Related Issue(s)
None

## Checklist

### DIFF
--- a/lua/lazyvim/plugins/extras/coding/blink.lua
+++ b/lua/lazyvim/plugins/extras/coding/blink.lua
@@ -82,6 +82,7 @@ return {
       },
 
       cmdline = {
+        enabled = false,
         sources = {},
       },
 


folke:
Perfect, thanks.

I didn't yet further look into the blink cmdline. 
Last time I checked it had issues, but maybe those have been solved in the meantime.

Saghen:
That `sources = {}` won't be needed after the `v0.12.1` update goes out in just a moment fyi

folke:
@Saghen even better :)

And thank you for all the great work you do with blink. Love it!

dpetka2001:
I don't know to be honest. I just personally don't like completions in the form of pop-up menus in the cmdline like completion engines do. I much prefer the native Neovim one for this specific purpose. Whatever you decide after looking into it yourself and get a personal feeling for it will be fine by me. I'll just disable it in my personal configuration if need be.

aweis89:
I have the override currently enabled for cmdline and I think it's pretty great for both search and commands. I've found to make it work as expected, just adding the following did the trick for me (aside from enabling and specifying the default sources):
```
        completion = {
          list = {
            selection = {
              preselect = function(ctx)
                return ctx.mode ~= "cmdline" and not require("blink.cmp").snippet_active({ direction = 1 })
              end,
            },
          },
```

PR #5620, opened 2025-02-15
fix(blink): remove unnecessary `sources` from `cmdline`

dpetka2001:
## Description
Like Saghen already mentioned in the other PR, this [commit](https://github.com/Saghen/blink.cmp/commit/19f60a675eaaf4b160bd6458bfd72fc005da5b3f) does a check based on the `enabled` field before adding sources, so it's no longer needed to define it.

## Related Issue(s)
None

## Checklist


folke:
ty

PR #5652, opened 2025-02-19
fix(sql): fix autocomplete when `omni` is used as a `blink.cmp` source

igorgue:
## Description

I was trying the `omni` source with `blink.cmp` and I had an error which you can read more about in this `nvim-compe` issue: https://github.com/hrsh7th/nvim-compe/issues/286 but the message keeps showing up after setting that to `syntax` so I mark the default plugin as loaded, and it works as expected, I think there are more items returned but I cannot confirm that :).

## Checklist

### DIFF
--- a/lua/lazyvim/plugins/extras/lang/sql.lua
+++ b/lua/lazyvim/plugins/extras/lang/sql.lua
@@ -28,6 +28,11 @@ end
 
 local sql_ft = { "sql", "mysql", "plsql" }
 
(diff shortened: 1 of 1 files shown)


igorgue:
This is not really a problem if you don't ever add the `omni` source in blink.cmp.

igorgue:
I don't think this is really working, I got an error while using dadbod completions.

PR #5654, opened 2025-02-20
feat(extras): add lua-console.nvim

YaroSpace:
## Description

[Lua-console.nvim](https://github.com/YaroSpace/lua-console.nvim) handy scratch pad / Lua REPL / debug console for Lua and Neovim  development, exploration and configuration.   A modernized and enhanced successor to `SnipRun / Luapad / Luadev`.

## Checklist

### DIFF
--- a/lua/lazyvim/plugins/extras/coding/lua-console.lua
+++ b/lua/lazyvim/plugins/extras/coding/lua-console.lua
@@ -1,0 +1,9 @@
+return {
+  "yarospace/lua-console.nvim",
+  lazy = true,
+  keys = {
+    { "`", desc = "Lua-console - toggle" },
+    { "<Leader>`", desc = "Lua-console - attach to buffer" },
+  },
+  opts = {},
+}


folke:
LazyVim already comes with a lua scratch pad from snacks.

And we also only accept new extras if they require complex setup, which is not the case here,

You did not read the contributing guidelines https://github.com/LazyVim/LazyVim/blob/main/CONTRIBUTING.md#contributing-an-extra-plugin

YaroSpace:
As a snacks user,  I am surely aware of the scratchpad. However, lua-console is more of a REPL and code runner, while scratchpad is a side feature, and is more an alternative to SnipRun than to snacks. 

As of guidelines, I did read them and find them ambiguous, since there are few of extras with little or no setup in coding and utils.

folke:
yes, popular plugins are sometimes added.

*  The plugin should be well-known and require significant configuration.
  - *this is not the case here*
*  Simple specs containing just the plugin with some options will not be accepted.
  - *you're not even setting any options*

For these types of plugins, users can simply add the spec from the readme to their install. No need to be an extra.

folke:
And the snacks scratchpad is also a code runner

YaroSpace:
There are few options that are set by default and can be configured, but I was under the impression that lazy spec should be as simple as possible and extended by the user if he wants to change the defaults. So, i deliberately kept them out. 

I know that snacks scratchpad can run lua). Lua-console, however, has extra features for debugging, output and inspection of the results, pretty prints function details and stack traces, follows their truncated paths, attaches to other buffers and runs other languages.

PR #5707, opened 2025-02-28
fix(copilot-chat): switch from deprecated picker integrations

deathbeam:
- Instead of using picker integrations (that are now deprecated), switch to using CopilotChat.select_prompt() that is using `vim.ui.select` which brings the selector integration responsibility to pickers instead of plugin
- Use vim.ui.input instead of vim.fn.input for consistency

Reason why this was deprecated in CopilotChat.nvim (by me) and will be removed in future is because these integrations never made any sense anyway when `vim.ui.select` overrides exist.

## Screenshots

Using fzf-lua in my config, outside of LazyNvim but same thing applies:

image

## Checklist

### DIFF
--- a/lua/lazyvim/plugins/extras/ai/copilot-chat.lua
+++ b/lua/lazyvim/plugins/extras/ai/copilot-chat.lua
@@ -1,28 +1,3 @@
-local M = {}
-
----@param kind string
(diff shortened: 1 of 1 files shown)


folke:
Makes sense. Thanks!

PR #5669, opened 2025-02-22
feat(mini-snippets): add support for package.json

matadaniel:
## Related Issue(s)

echasnovski/mini.nvim#1571

## Checklist

- [x] I've read the [CONTRIBUTING](https://github.com/LazyVim/LazyVim/blob/main/CONTRIBUTING.md) guidelines.

PR #5710, opened 2025-02-28
feat(mini-snippets): add leaner support for package.json

matadaniel:
## Related Issue(s)

- echasnovski/mini.nvim#1571
- #5669

## Same functionality now with 60% less LOC!

There is no need for any of the previous `json` validation when the paths are provided explicitly.
This also adds compatibility with `from_lang` so the user is not required to create a `package.json`.
The extra should be loaded before `lang` extras in case they add to `lang_additions`. It works without setting a priority since `coding` comes first alphabetically and `lang.typescript` does not add any options, but I wanted to be more explicit.

## Checklist


folke:
Sorry, but I'm just not willing to accept additional features implemented for another plugin.

This makes no sense. This should just be part of that plugin, and if they don't want to add it then that's how it is.

This just adds more to the maintenance burden, especially since it's something I don't even use myself.

PR #5754, opened 2025-03-08
fix(extras): disable blink path source in CopilotChat (#5666)

S1M0N38:
Fix #5666 with the answer given in the discussion.

### DIFF
--- a/lua/lazyvim/plugins/extras/ai/copilot-chat.lua
+++ b/lua/lazyvim/plugins/extras/ai/copilot-chat.lua
@@ -85,4 +85,24 @@ return {
       })
     end,
   },
+
+  -- Blink integration
+  {
+    "saghen/blink.cmp",
+    optional = true,
+    ---@module 'blink.cmp'
+    ---@type blink.cmp.Config
+    opts = {
+      sources = {
+        providers = {
+          path = {
+            -- Path sources triggered by "/" interfere with CopilotChat commands
+            enabled = function()
+              return vim.bo.filetype ~= "copilot-chat"
+            end,
+          },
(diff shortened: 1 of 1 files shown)


azeveco:
It fixes the problem here. Thanks @S1M0N38 

image

folke:
ty

LucasMoskun:
Just like to mention the blink and `/` is very convenient for `#file` command for pulling files into the context buffer. 

I don't have a clear solution (just overrode this change in my local as it's something i do often) but wanted to point this out  in case someone has a clear idea.

I tried replacing it with this that ignores blink for `/` as a first character which then allowed me to still do `#file./" and leverage blink, but i don't want to PR it because i think they were still fighting in some cases... 

```lua
-- Blink integration
{
  "saghen/blink.cmp",
  optional = true,
  ---@module 'blink.cmp'
  ---@type blink.cmp.Config
  opts = {
    sources = {
      providers = {
        path = {
          -- Only suppress path completion for CopilotChat *slash commands* at BOL.
          enabled = function()
            if vim.bo.filetype ~= "copilot-chat" then
              return true
            end
            local line = vim.api.nvim_get_current_line()
            local col  = vim.api.nvim_win_get_cursor(0)[2] + 1 -- 1-based
            local before = line:sub(1, col)

            -- If we're currently typing a slash command like "/he|"
            -- at the start of the line, disable path completion.
            if before:match("^/%w*$") then
              return false
            end

            -- Otherwise (e.g. "#file ./src/…"), keep path completion ON.
            return true
          end,
        },
      },
    },
  },
}

```

PR #5763, opened 2025-03-10
fix(neo-tree): import neo-tree Extra before edgy Extra

### LINKED ISSUE #5762: bug: edgy's neo-tree spec gets overridden by neo-tree's spec

### Neovim version (nvim -v)

v0.11.0-dev-1860+g877f3b7288

### Operating system/version

archlinux 6.13.2-zen1-1-zenA

### Describe the bug

hi, neo-tree maintainer here.

related: https://github.com/nvim-neo-tree/neo-tree.nvim/issues/1703 and https://github.com/LazyVim/LazyVim/discussions/5583

As of recently, [edgy.nvim's neo-tree spec](https://github.com/LazyVim/LazyVim/blob/ec5981dfb1222c3bf246d9bcaa713d5cfa486fbd/lua/lazyvim/plugins/extras/ui/edgy.lua#L147-L155) now gets overridden by the neo-tree spec instead of the other way around, so the final filetype ignore list received in setup doesn't include `edgy`. This leads to the splitting behavior mentioned in the above discussions.

I would make a PR but not sure what the best way to implement a fix would be. is there a way to ensure some specs get their opts set before others? If there's anything you'd like me to do to make this option more flexible (e.g. allow the ignored types option to accept entries like`["filetype"] = boolean`, lmk)

### Steps To Reproduce

1. Enable both neo-tree and edgy through `:LazyExtras`
2. Open neo-tree with `<leader>e`
3. Open a file
4. Observe the slim split

### Expected Behavior

No slim split should form - it should takeover the dashboard window instead

### Repro

```lua

require("lazy.minit").repro({
  spec = {
    -- add any other plugins here
  },
})
```

dpetka2001:
## Description
Seems like since the move of `neo-tree` into Extras, its plugin spec is loaded after edgy's and that causes problems. Give `neo-tree` Extra higher prio than edgy Extra.

## Related Issue(s)
Fixes #5762 

## Checklist


pynappo (approved):

hmnd (approved):

hmnd:
Can confirm this fixes the weird splits I was experiencing!

folke:
ty!

PR #5796, opened 2025-03-17
fix(lspconfig): inlay_hints toggle for Lua_ls

windowsrefund:
## Description

This makes the `<leader>uh` toggle work when editing a lua file.

## Checklist


folke:
The toggle already works as intended.
What you are proposing is showing more inlay hints by default, but I'm not a fan of that.
You can of course do this in your own config.

PR #5841, opened 2025-03-26
fix(editor): disable cmdline window flash

EasonMo:
## Description

When executing `<leader>ghs`, the command line window will flicker. Therefore, it should be run silently.

## Related Issue(s)
None

## Checklist


folke:
ty

PR #5936, opened 2025-04-11
feat(ai.coding.yanky): Use default highlight group for yanked region

xavierchen0:
Set yanky.nvim's highlight groups, YankyPut and YankyYanked, to use vim.hl.on_yank()'s default highlight group for yanked region, IncSearch.

## Description

Without yanky.nvim extra installed, yanking will use the default highlight group, IncSearch.

Reference from LazyVim's Documentation,

[link](https://www.lazyvim.org/configuration/general)

Autocommand:
```lua
-- Highlight on yank
vim.api.nvim_create_autocmd("TextYankPost", {
  group = augroup("highlight_yank"),
  callback = function()
    (vim.hl or vim.highlight).on_yank()
  end,
})
```

Reference from NeoVim's Documentation,

[link](https://neovim.io/doc/user/lua.html#vim.hl.on_yank())

documentation states that vim.hl.on_yank() uses the default highlight group for yanked region as IncSearch.

> {opts} (table?) Optional parameters
> higroup highlight group for yanked region (default "IncSearch")

However, when you install yank.nvim extra, yank.nvim's two highlight groups, YankyPut and YankyYanked, is linked to Search highlight group by default ([link](https://github.com/gbprod/yanky.nvim?tab=readme-ov-file#-colors)).

This can be puzzling for users, at least for me, because I previously had colors when I yank, but after installing yank.nvim, the colors are gone.

This PR links yank.nvim's two highlight groups, YankyPut and YankyYanked, to the default highlight group, IncSearch, used by vim.hl.on_yanked().

As a result, after installing yank.nvim extra, user should still see colors when they yank, just as expected. 

## Checklist

### DIFF
--- a/lua/lazyvim/plugins/extras/coding/yanky.lua
+++ b/lua/lazyvim/plugins/extras/coding/yanky.lua
@@ -7,6 +7,10 @@ return {
   opts = {
     highlight = { timer = 150 },
   },
+  init = function(_)
+    vim.api.nvim_set_hl(0, "YankyYanked", { link = "IncSearch" })
+    vim.api.nvim_set_hl(0, "YankyPut", { link = "IncSearch" })
+  end,
   keys = {
     {
       "<leader>p",


dpetka2001:
That is dependent on the colorscheme that you use and the `yanky.nvim` plugin itself. Tokyonight already does this as can be seen [here](https://github.com/folke/tokyonight.nvim/blob/057ef5d260c1931f1dffd0f052c685dcd14100a3/lua/tokyonight/groups/yanky.lua#L9-L10). 

So, this should be changed in the user's personal configuration in the colorscheme that he uses or in the `yanky.nvim` plugin itself in my personal opinion as a simple user. Also the link from the Neovim docs only references the default hi-group for the `yank` operation not the `put` one.

Still, I'm not making any decisions and maintainer has the final say.

xavierchen0:
> That is dependent on the colorscheme that you use and the `yanky.nvim` plugin itself. Tokyonight already does this as can be seen [here](https://github.com/folke/tokyonight.nvim/blob/057ef5d260c1931f1dffd0f052c685dcd14100a3/lua/tokyonight/groups/yanky.lua#L9-L10). 
> 
> 
> 
> So, this should be changed in the user's personal configuration in the colorscheme that he uses or in the `yanky.nvim` plugin itself in my personal opinion as a simple user. Also the link from the Neovim docs only references the default hi-group for the `yank` operation not the `put` one.
> 
> 
> 
> Still, I'm not making any decisions and maintainer has the final say.

Thank you for your feedback. I do agree with you that a user could implement this feature in two distinct ways:

1. Do what I did in this PR. Change the link to a highlight group of one's preference.
2. Configure the colorscheme plugin options (if it exists), or directly in the colorscheme plugin itself.

However, I created this PR with a new LazyVim user in mind. Without installing yanky.nvim, if colors had been set for the IncSearch highlight group by the colorscheme plugin or directly, one should see colors when they yank. This is because of vim.hl in the TextYankPost autocommand.

When the new LazyVim user installs yanky.nvim, he might expect to see colors when he yanks. But this is not the case if his colorscheme did not configure YankyYanked highlight group. 

A universal and good default for the yanky.nvim LazyExtra, would be to link YankyYanked highlight group to the IncSearch highlight group because that is the default highlight group for vim.hl in the LazyVim default TextYankPost autocommand. It is fair to assume that most colorscheme plugins would have configured the IncSearch highlight group.

Otherwise, the new LazyVim user would have to spend considerable time troubleshooting why the yank no longer has colors, and would eventually arrive at the same conclusion as me, and link it to IncSearch highlight group for simplicity in configuration.

Note: The above serves to add further clarification on the PR. Hope it helps!

Regarding the YankyPut highlight group, this is what I would have liked configured. I am fine to remove it since this is not LazyVim default feature.

dpetka2001:
I understand what you're saying as far as a new user is concerned, but I respectfully think that either the colorscheme or the plugin itself is responsible for setting appropriate highlight groups and it's up to the user to overwrite anything he wants in his personal configuration. 

I'd personally prefer if users asked questions about this and be guided as to how to solve this problem (they will also understand why this happens which is a helpful experience). 

Again this is just my personal opinion and no hard feelings (we're both free to have different opinions and that's perfectly normal). I have tried locally your PR and since you set the highlight groups in `init` function, they are correctly overwritten by LazyVim's colorscheme which is loaded later (in case colorschemes actually have these hi-groups defined), so there are no conflicts in that regard.

Cheers and good luck :smile:

xavierchen0:
> I understand what you're saying as far as a new user is concerned, but I respectfully think that either the colorscheme or the plugin itself is responsible for setting appropriate highlight groups and it's up to the user to overwrite anything he wants in his personal configuration.
> 
> I'd personally prefer if users asked questions about this and be guided as to how to solve this problem (they will also understand why this happens which is a helpful experience).
> 
> Again this is just my personal opinion and no hard feelings (we're both free to have different opinions and that's perfectly normal). I have tried locally your PR and since you set the highlight groups in `init` function, they are correctly overwritten by LazyVim's colorscheme which is loaded later (in case colorschemes actually have these hi-groups defined), so there are no conflicts in that regard.
> 
> Cheers and good luck 😄

Thank you for checking my PR! I understand where you are coming from. Cheers!

folke:
Won't merge this. Yanky already sets this as well btw https://github.com/gbprod/yanky.nvim/blob/04775cc6e10ef038c397c407bc17f00a2f52b378/lua/yanky/highlight.lua#L12

PR #5958, opened 2025-04-16
feat(lang): add neotest config for PHP tests

jradtilbrook:
## Description

Similar to some other languages (this closely follows the Go lang extra), this adds neotest runners for the major PHP test runners: PHPUnit and Pest.

Both work together, trying to use Pest if available and falling back to PHPUnit which is a decent approach for the PHP ecosystem.

I read CONTRIBUTING.md which differs from the way the Go lang extra has implemented a neotest adapters and opted to follow that instead. Ie. adapters
have been added as dependencies instead of a separate spec with `lazy=true`. Let me know if you want that changed.

## Checklist

### DIFF
--- a/lua/lazyvim/plugins/extras/lang/php.lua
+++ b/lua/lazyvim/plugins/extras/lang/php.lua
@@ -84,4 +84,21 @@ return {
       },
     },
   },
+
+  {
+    "nvim-neotest/neotest",
+    optional = true,
+    dependencies = {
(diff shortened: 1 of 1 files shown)


jradtilbrook:
😬 Will this auto close after some time being stale? I'd still love to get this merged 🙏

jradtilbrook:
Plz not stale. This is super handy, would love to get it merged in

folke:
ty

PR #5969, opened 2025-04-17
fix(lsp): use `supports_method` method instead of field on Neovim >=0.11

### LINKED ISSUE #5861: bug: ESLint's codeActionOnSave stopped working

### Neovim version (nvim -v)

0.11.0

### Operating system/version

15.3.2

### Describe the bug

I noticed that formatting on save with ESLint stopped working.

Image

Since I started using LazyVim, I followed the configuration in https://www.lazyvim.org/configuration/recipes#use-eslint-for-fix-on-save-and-prettier-for-formatting and had the 2 extras enabled.

I use a plugin called `eslint-plugin-simple-import-sort` (https://github.com/lydell/eslint-plugin-simple-import-sort) to sort my imports on save. Things have been working just fine. After this issue happened, I could still sort my imports using `eslint --fix`.

Upon checking with `LspInfo`, I also noticed that ESLint LSP's `codeActionOnSave.enable` was set to `false`.

Image

### Steps To Reproduce

1. Set up ESLint using flat config for a TypeScript project.
2. Enable the 2 extras as mentioned in https://www.lazyvim.org/configuration/recipes#use-eslint-for-fix-on-save-and-prettier-for-formatting .
3. Set up `eslint-plugin-simple-import-sort`
4. Mixed up imports in a file.
5. Save the file.

### Expected Behavior

With the steps listed in `Steps to Reproduce`, imports should be auto-sorted after the file is saved.

### Repro

```lua

require("lazy.minit").repro({
  spec = {
    -- add any other plugins here
  },
})
```

dpetka2001:
## Description
From Neovim 0.11 onwards `client.supports_method` was changed into a method `client:supports_method`. Hope I didn't forget any occurences where this has to be changed.

## Related Issue(s)
Fixes #5861 hopefully

## Checklist

### DIFF
--- a/lua/lazyvim/plugins/extras/editor/navic.lua
+++ b/lua/lazyvim/plugins/extras/editor/navic.lua
@@ -8,7 +8,10 @@ return {
     init = function()
       vim.g.navic_silence = true
       LazyVim.lsp.on_attach(function(client, buffer)
-        if client.supports_method("textDocument/documentSymbol") then
+        if
+          vim.fn.has("nvim-0.11") == 0 and client.supports_method("textDocument/documentSymbol")
+          or client:supports_method("textDocument/documentSymbol")
+        then
           require("nvim-navic").attach(client, buffer)
         end
       end)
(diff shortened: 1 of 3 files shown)


dpetka2001:
Actually this should not be needed yet. The fields that got changed into methods should still be handled internally by Neovim, as I can see for `supports_method` that it still checks to see if `bufnr` is a table and makes the appropriate adjustment. They will be removed in Neovim 0.13, so this shouldn't have any effect currently. 

Probably a larger refactor will be made once Neovim 0.13 is released, but this shouldn't cause any problem for the Eslint issue like I thought initially. Closing.

PR #5991, opened 2025-04-22
Improve comment in `keymaps.lua` on how to add keymap in lazyvim style

B1LLP4RK:
## Description
This is the file that users are directed to in `lua/config/keymaps.lua`.
```
-- Keymaps are automatically loaded on the VeryLazy event
-- Default keymaps that are always set:
https://github.com/LazyVim/LazyVim/blob/main/lua/lazyvim/config/keymaps.lua
-- Add any additional keymaps here
```
So it would be much better to explicitly guide them to use the which-key API.

Besides, the [documentation](https://www.lazyvim.org/configuration/keymaps) on keymaps might also be improved by instructing how to use which-key to add keymaps. As anyone who uses Lazyvim would like to add their keymaps with proper description to share the joy of which-key.

### DIFF
--- a/lua/lazyvim/config/keymaps.lua
+++ b/lua/lazyvim/config/keymaps.lua
@@ -3,7 +3,7 @@
 -- DO NOT USE `LazyVim.safe_keymap_set` IN YOUR OWN CONFIG!!
 -- use `vim.keymap.set` instead
 -- or use local `wk = require("which-key")` and `wk.add`, to utilise the popup from the which-key plugin.
--- for more information, go to https://github.com/folke/which-key.nvim
+-- For more information, go to https://github.com/folke/which-key.nvim
 local map = LazyVim.safe_keymap_set
 
 -- better up/down
--- a/lua/lazyvim/config/keymaps.lua
+++ b/lua/lazyvim/config/keymaps.lua
@@ -2,6 +2,8 @@
 
 -- DO NOT USE `LazyVim.safe_keymap_set` IN YOUR OWN CONFIG!!
 -- use `vim.keymap.set` instead
+-- or use local `wk = require("which-key")` and `wk.add`, to utilise the popup from the which-key plugin.
+-- for more information, go to https://github.com/folke/which-key.nvim
 local map = LazyVim.safe_keymap_set
 
 -- better up/down


dpetka2001:
`which-key` will pick up whatever you add with `vim.keymap.set`, so I don't really see the point in this? Also `wk.add` should not be encouraged to be used, because users might try to contribute PRs that use `wk.add` and that should definetely be a no-no, since it makes it difficult to remove.

dpetka2001:
It's `vim.keymap.set` not `vim.key.set`. 

To add description you can do something like `vim.keymap.set("n", "<leader>xo", function() something end, { desc = "toggle diagnostics" })`. 

See `:h vim.keymap.set()` for more info. It's explained in Neovim help. You should get more familiar about reading Neovim help files. It might be a bit hard at first, but you should keep doing it and get more familiar with the process. Most of what you want to ask can be found in Neovim help.

B1LLP4RK:
I did try to find relevant parts in help files.
```
vim.keymap.set({mode}, {lhs}, {rhs}, {opts})                *vim.keymap.set()*
    Defines a |mapping| of |keycodes| to a function or keycodes.

    Examples: >lua
        -- Map "x" to a Lua function:
        vim.keymap.set('n', 'x', function() print("real lua function") end)
        -- Map "<leader>x" to multiple modes for the current buffer:
        vim.keymap.set({'n', 'v'}, '<leader>x', vim.lsp.buf.references, { buffer = true })
        -- Map <Tab> to an expression (|:map-<expr>|):
        vim.keymap.set('i', '<Tab>', function()
          return vim.fn.pumvisible() == 1 and "<C-n>" or "<Tab>"
        end, { expr = true })
        -- Map "[%%" to a <Plug> mapping:
        vim.keymap.set('n', '[%%', '<Plug>(MatchitNormalMultiBackward)')
<

    Parameters: ~
      • {mode}  (`string|string[]`) Mode "short-name" (see
                |nvim_set_keymap()|), or a list thereof.
      • {lhs}   (`string`) Left-hand side |{lhs}| of the mapping.
      • {rhs}   (`string|function`) Right-hand side |{rhs}| of the mapping,
                can be a Lua function.
      • {opts}  (`table?`) Table of |:map-arguments|. Same as
                |nvim_set_keymap()| {opts}, except:
                • {replace_keycodes} defaults to `true` if "expr" is `true`.

                Also accepts:
                • {buffer}? (`integer|boolean`) Creates buffer-local mapping,
                  `0` or `true` for current buffer.
                • {remap}? (`boolean`, default: `false`) Make the mapping
                  recursive. Inverse of {noremap}.

    See also: ~
      • |nvim_set_keymap()|
      • |maparg()|
      • |mapcheck()|
      • |mapset()|
```

So I went to follow the :map-arguments help file.
Which was
```
1.2 SPECIAL ARGUMENTS					*:map-arguments*

"<buffer>", "<nowait>", "<silent>", "<script>", "<expr>" and
"<unique>" can be used in any order.  They must appear right after the
command, before any other arguments.
```

So there was no information in help file that I can use the desc keyword.
And it makes sense information on `desc` is not in the help file because isn't it something introduced by which-key?

I am glad that `vim.keymap.set` can be used instead of requiring `wk` and using `wk.add`. It is simpler and faster.
However I think it is problematic that no relevant is available anywhere. I looked at the `lazyvim`, `lazy.nvim`, `which-key` docs and the help files before making this pull request. Nowhere could I find the method of `vim.keymap.set("n", "<leader>xo", function() something end, { desc = "toggle diagnostics" })`

Jacobus-afk:
>       • {opts}  (`table?`) Table of |:map-arguments|. Same as
>                |nvim_set_keymap()| {opts}, except:
>                • {replace_keycodes} defaults to `true` if "expr" is `true`.

if you followed the link for `nvim_set_keymap()` you'd see the following

image

like @dpetka2001 said, reading the help is tricky and will take some time to get used to

Lazyvim is also another abstraction on top of nvim, so it might be better to first start with [nvim-kickstart](https://github.com/nvim-lua/kickstart.nvim/tree/master) and then once you've got an idea how everything fits together, you can move to a distribution such as Lazyvim

also, could you please close this pull request? :smiley:

B1LLP4RK:
I see. I appreciate your help in finding the parts I missed. Then I think it is a fault in the help files since I don't think it is natural to go see the help file of `nvim_set_keymap()` unless the help file for `vim.keymap.set` was
```
      • {opts}  (`table?`) Table of |:map-arguments| with **_extra keywords_** same as
                |nvim_set_keymap()| {opts}, except:
                • {replace_keycodes} defaults to `true` if "expr" is `true`.
```

I understand that same or overlapping parts in help files better not be in multiple parts of the help file, but it should have been more explicit.

PR #5829, opened 2025-03-22
Fix: Enable venv-selector even when telescope is not available

rabinadk1:
## Description

Enable the `venv-selector` plugin even when the telescope plugin is not available. Now the `venv-selector` plugin supports `fzf-lua` and native selection and by default, it chooses the previous, if available.

More on this PR: https://github.com/linux-cultist/venv-selector.nvim/pull/188

## Related Issue(s)

None

## Screenshots

N/A

## Checklist

- [x] I've read the [CONTRIBUTING](https://github.com/LazyVim/LazyVim/blob/main/CONTRIBUTING.md) guidelines.

PR #6024, opened 2025-05-02
feat(lang.python): supprot loading python path configured in neoconf

qizidog:
## Description

Currently, `nvim-dap-python` resolve python interpreter through [debugpy](https://github.com/microsoft/debugpy), while `lspconfig` support `neoconf`. In most case, it's not necessary to setup two different interpreter for a simple project.

`linux-cultist/venv-selector.nvim` is disabled after `telescope` removed by default, this commit add a support for loading python interpreter path configured in neoconf.

## Related Issue(s)

  - related #5829

## Screenshots

...

## Checklist


folke:
not interested in adding this. neoconf is also something I would not recommend to use.

PR #6044, opened 2025-05-07
fix(blink-cmp): luasnip + blink-cmp snippet expand

jackielii:
## Description

When using Luasnip with blink-cmp, the snippet expansion is using built-in snippet expansion wrongly. This PR fixes the issue

To reproduce the issue, make a mapping like so:

```lua
vim.keymap.set({ "i", "s" }, "<C-l>", function()
	vim.print("menu visible: " .. tostring(require("blink-cmp").is_menu_visible()))
	vim.print("snippet active: " .. tostring(require("blink-cmp").snippet_active({ direction = 1 })))
	vim.print("luasnip jumpable: " .. tostring(require("luasnip").jumpable(1)))
	vim.print("vim.snippet_active: " .. tostring(vim.snippet.active({ direction = 1 })))
end)
```

Insert a lsp snippet like so:

```lua
  vim.keymap.set(|mode, lhs, rhs, opts?)
```

When at the `|` position, press `<C-l>` to trigger the mapping. The output will be:

```
menu visible: false
snippet active:false
luasnip jumpable: false
vim.snippet_active: true
```

This is wrong because vim.snippet_active should be false, and blink-cmp snippet active should be true.

## Checklist


dpetka2001:
The `preset = "luasnip"` is a shorthand by blink to not define all the previous fields (like `expand` in your case). Shouldn't this be fixed in blink instead if the `preset = "luasnip"` doesn't resolve the snippet engine correctly?

jackielii:
> The `preset = "luasnip"` is a shorthand by blink to not define all the previous fields (like `expand` in your case). Shouldn't this be fixed in blink instead if the `preset = "luasnip"` doesn't resolve the snippet engine correctly?

It doesn't, the https://github.com/LazyVim/LazyVim/blob/main/lua/lazyvim/plugins/extras/coding/blink.lua#L39-L43 overwrites blink's default `luasnip` preset, therefore, when we include `luasnip` extra, we need to override this override :)

dpetka2001:
Oh, I see. You're right. Totally missed that override. Sorry about that :smile:

PR #6045, opened 2025-05-07
fix(config): update to mason v2

Conarius:
## Description

- Updates all references to mason.nvim and mason-lspconfig.nvim
- Updates all config files to use the suggested vim APIs instead

## Related Issue(s)

- Fixes #6039

## Checklist

- [x] I've read the [CONTRIBUTING](https://github.com/LazyVim/LazyVim/blob/main/CONTRIBUTING.md) guidelines.

PR #6041, opened 2025-05-07
fix(mason): update org, use public mappings API

minusfive:
## Description

`mason.nvim` and `mason-lspconfig.nvim` have moved GitHub orgs, causing LazyVim to break. This PR updates all references with the new org, and uses the `mason-lspconfig` public `get_mappings()` API [as suggested](https://github.com/LazyVim/LazyVim/issues/6039#issuecomment-2856502153) by @williamboman himself.

## Related Issue(s)

- https://github.com/mason-org/mason.nvim/pull/1925
- Fixes #6039

## Checklist

- [x] I've read the [CONTRIBUTING](https://github.com/LazyVim/LazyVim/blob/main/CONTRIBUTING.md) guidelines.

PR #6046, opened 2025-05-07
fix(config): all_mslp_servers fix

aauutthh:
## Description

when enter buffer, an error triggered by requiring "mason-lspconfig.mappings.server" , which  is no longer exists.

## Checklist


dpetka2001:
There's already #6041 and #6045 about this

aauutthh:
> There's already #6041 and #6045 about this↳

Yes, you're right.

PR #6053, opened 2025-05-08
fix(mason): conform to v2 changes

dpetka2001:
## Description
`mason-lspconfig` v2 made changes to conform with latest Neovim `vim.lsp.config`/`vim.lsp.enable` API. 

- Lock `mason` and `mason-lspconfig` to v1 versions for Neovim <0.11 and latest HEAD for Neovim >=0.11.
- Change `:get_install_path()` method in LazyVim to use `LazyVim.get_pkg_path()` and also update name references to new ones.
- Change `nvim-lspconfig` from `LazyFile` to `{ "BufReadPre", "BufNewFile", "BufWritePre" }` for Neovim >=0.11, since on `BufReadPost` the LSP would not attach on the first buffer. Previously the `setup.handlers()` function from `mason-lspconfig` would take care of that, but that has been removed on v2. I fail to think if there's a better way to handle this.

Most of the changes in `/lsp/init.lua` are thanks to @williamboman comment [here](https://github.com/LazyVim/LazyVim/pull/6041#issuecomment-2857266471).

I've also seen that both `mason-lspconfig` and `rustaceanvim` now have as minimum requirements Neovim >=0.11.
As such I would suggest, instead of all the conditional checks for Neovim versions to create a stable release where we pin `mason` and `mason-lspconfig` to v1 versions and then only incorporate William's suggestions in a new release and mention in the README that users who are on Neovim <0.11 should use the stable release where the plugin versions are pinned to v1. Not sure how you want to tackle this, so this is merely just a suggestion from my part.

Both #6041 and #6045 tackle only with the new name references and the offending line that was causing the error, but don't take into consideration the LSP servers configurations. `opts.setup` would not execute on both, since that was handled by `setup_handlers()` function in the past. I checked with Typescript lang and with those PRs there are no Javascript settings in `vtsls` (by checking with `:=LazyVim.opts("nvim-lspconfig").servers.vtsls`). 
#6045 also tried to change the method `:get_install_pkg()` by using `vim.fn.exepath`, but to my understanding that only returns the binary path not the package installation path and that is later concatenated to find other paths, which I believe is not correct.

PS: There are 2 commits, because my LazyVim fork was not up to date and updated it and then merged the branch I had into a new branch. Sorry if that causes any inconvenience.

## Related Issue(s)
Fixes #6039

## Checklist

- [x] I've read the [CONTRIBUTING](https://github.com/LazyVim/LazyVim/blob/main/CONTRIBUTING.md) guidelines.

PR #6056, opened 2025-05-09
Chore: fixed lsp config

mmj2023:
The lsp config has been changed to use new api to get the table for all mason lsp servers.

## Checklist

### DIFF
--- a/lua/lazyvim/plugins/lsp/init.lua
+++ b/lua/lazyvim/plugins/lsp/init.lua
@@ -212,7 +212,7 @@ return {
       local have_mason, mlsp = pcall(require, "mason-lspconfig")
       local all_mslp_servers = {}
       if have_mason then
-        all_mslp_servers = vim.tbl_keys(require("mason-lspconfig.mappings.server").lspconfig_to_package)
+        all_mslp_servers = vim.tbl_keys(require("mason-lspconfig").get_available_servers())
       end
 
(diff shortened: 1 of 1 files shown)


dpetka2001:
This will not work. Try #6053 instead. Looking for feedback. Or pin the plugin's versions to v1.

mmj2023:
Why could elaborate please?

dpetka2001:
This PR only makes the error go away. It doesn't take into consideration the LSP servers configuration.

fivemtebex:
Impressive project structure! The architecture really demonstrates solid design principles.

mmj2023:
@fivemhostingserver I guess you could fix the integration with Lazyvim and tests, if there are any issues, I use a custom config that's heavily inspired by Lazyvim, but I am not a Lazyvim user. So, pardon me if any other area of the code breaks. They have a lot of abstractions that I don't think I need, I think, @dpetka2001 has a better pr than me, see #6053 . I just updated my plug-ins and lspconfig broke, so I fixed it, and when I looked into Lazyvim's repo, it was broken here as well, so I fixed it and made a pr. This is my 1st pr so, I was actually hoping someone would either make a better pr than me or tell me what to do in order to fix mine. And I hadn't looked at the issues yet, so I didn't notice that the issue had already been created and fixes were already given.

PR #6060, opened 2025-05-10
fix(clangd): rewrite the root_dir function

ChillyWall:
## Description

Nvim 0.11 had some breaking changes in lsp related parts. And one of them is that if the root_dir field of lsp config is a function, the bufnr and a function would be passed instead of the file name in nvim 0.10. As a result, the configurations of extra.lang.clangd broke.

I rewrite the root_dir function and make it work both on 0.11 and 0.10.

It would check whether the first argument is a string to decide whether to adapt the new interface.

## Checklist

### DIFF
--- a/lua/lazyvim/plugins/extras/lang/clangd.lua
+++ b/lua/lazyvim/plugins/extras/lang/clangd.lua
@@ -60,18 +60,27 @@ return {
           keys = {
             { "<leader>ch", "<cmd>ClangdSwitchSourceHeader<cr>", desc = "Switch Source/Header (C/C++)" },
           },
-          root_dir = function(fname)
-            return require("lspconfig.util").root_pattern(
-              "Makefile",
-              "configure.ac",
-              "configure.in",
-              "config.h.in",
-              "meson.build",
-              "meson_options.txt",
-              "build.ninja"
-            )(fname) or require("lspconfig.util").root_pattern("compile_commands.json", "compile_flags.txt")(
-              fname
-            ) or require("lspconfig.util").find_git_ancestor(fname)
+          root_dir = function(bufnr, ondir)
+            local root_directory = function(fname)
+              return require("lspconfig.util").root_pattern(
+                "Makefile",
+                "configure.ac",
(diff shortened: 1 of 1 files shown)


dpetka2001:
I would try to keep this PR separate from #6053 for the moment. When maintainer comes back, maybe he will come up with an even better way to resolve the problem and my PR won't even end up being merged. 

My suggestion would be to keep your PR simple and separate. If #6053 ends up being merged, you can just rebase when the time is appropriate.

ChillyWall:
> I would try to keep this PR separate from #6053 for the moment. When maintainer comes back, maybe he will come up with an even better way to resolve the problem and my PR won't even end up being merged.
> 
> My suggestion would be to keep your PR simple and separate. If #6053 ends up being merged, you can just rebase when the time is appropriate.

I don't know the PR would automatically update when I push new commits.😅

dpetka2001:
I just noticed that you wrote 
> However, before the breaking changes of mason and mason-lspconfig, the configurations of clangd still worked. So maybe what really broke the configurations is mason-lspconfig, which uses a new way to enable lsp in v2.

Are you able to test #6053 to see if things still work without this PR, so that you can determine if this PR is actually needed or not?

ChillyWall:
> Are you able to test #6053 to see if things still work without this PR, so that you can determine if this PR is actually needed or not?

In my test, with #6053, mason-lspconfig v2 could start clangd correctly without this PR on nvim 0.10.3 but not on nvim0.11.1+. 

mason-lspconfig v1 could also start clangd without this PR no matter on nvim0.10 or 0.11. I guess it use other way rather than vim.lsp.start to start lsp.

I have read the codes of maosn-lspconfig v2 and found that it use vim.lsp.start to auto start lsp and this function has changed since nvim0.11. And I guess this may be the problem.

So I write the new function, it could work on all cases above.

AtifChy:
@folke merge this, please. with latest lsp changes clangd doesn't work anymore.

folke:
ty!

PR #6074, opened 2025-05-14
feat(editor.refactoring): Limit `editor.refactoring` extra's filetypes

TheDSCPL:
## Description

The `editor.refactoring` currently has the annoying consequence of polluting which-key with a bunch of refactor keymaps that don't do anything in the projects I work on, namely Rust.

For this reason, I've added `ft` to both the keymaps and the `ThePrimeagen/refactoring.nvim` spec, containing all the supported [filetypes of refactoring.nvim](https://github.com/ThePrimeagen/refactoring.nvim/blob/2be7ea3f10b7e59658f5abf6dffc50b5d61964d6/lua/refactoring/config/init.lua#L57).

## Checklist

### DIFF
--- a/lua/lazyvim/plugins/extras/editor/refactoring.lua
+++ b/lua/lazyvim/plugins/extras/editor/refactoring.lua
@@ -1,3 +1,26 @@
+-- Source: https://github.com/ThePrimeagen/refactoring.nvim/blob/2be7ea3f10b7e59658f5abf6dffc50b5d61964d6/lua/refactoring/config/init.lua#L57
+local refactoring_supported_ft = {
+  "ts",
+  "js",
+  "typescriptreact",
+  "javascriptreact",
+  "vue",
+  "typescript",
+  "javascript",
+  "java",
+  "lua",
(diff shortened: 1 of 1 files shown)


TheDSCPL:
I've had this in my config for some time:

```lua
-- TODO: make a PR to LazyVim so this extra only loads in the supported filetypes
local function get_refactoring()
  local refactoring_supported_ft = { "go", "java", "cpp", "c", "h", "hpp", "cxx" }

  local extra_spec = require("lazyvim.plugins.extras.editor.refactoring")
  local spec = extra_spec[1]

  local keys = spec.keys or {}
  -- add ft to the keymaps
  for i, key in ipairs(keys) do
    if type(key) == "table" then
      key.ft = refactoring_supported_ft
    end
  end

  spec.ft = refactoring_supported_ft
  spec.keys = keys

  return spec
end
```

and I wanted to clean my nvim config, so I'd love for this to be merged!

folke:
that;'s not the way to do it and there's also conflicts

PR #6075, opened 2025-05-14
feat(php): add `php-debug-adapter` to mason

TheDSCPL:
## Description

While reviewing and contributing to #6053, I noticed that the PHP language extra doesn't include `php-debug-adapter` in mason's `ensure_installed`, even though it requires it.

## Checklist

### DIFF
--- a/lua/lazyvim/plugins/extras/lang/php.lua
+++ b/lua/lazyvim/plugins/extras/lang/php.lua
@@ -46,6 +46,16 @@ return {
   {
     "mfussenegger/nvim-dap",
     optional = true,
+    specs = {
+      {
+        "mason-org/mason.nvim",
+        opts = {
+          ensure_installed = {
+            "php-debug-adapter",
+          },
+        },
+      },
+    },
     opts = function()
       local dap = require("dap")
       local path = require("mason-registry").get_package("php-debug-adapter"):get_install_path()


dpetka2001:
@TheDSCPL Did you try setting a breakpoint with `<leader>db` to load `nvim-dap`? It should automatically install the debug adapter if I'm not mistaken. Is that not the case?

dpetka2001:
Here's a screencast of a fresh LazyVim installation from scratch and enabling `dap.core` and `lang.php` Extras. When you press `<leader>db` to toggle a breakpoint, you can see that `php-debug-adapter` gets automatically installed because that's when `nvim-dap` and subsequently `mason-nvim-dap` which is a dependency actually load.

So, I don't believe this is needed??


https://github.com/user-attachments/assets/e0b31e80-697f-4cee-8eb2-44b58ef9516a

dpetka2001:
You usually define in mason's `ensure_installed` formatters and linters and maybe some other tools in my experience. 

You don't need to define LSP servers because `opts.servers` from nvim-lspconfig gets automatically merged and added to mason's `ensure_installed`. 

Same with dap adapters, because mason-nvim-dap (which is the equivalent to mason-lspconfig for DAP) has a setting that is set to true about auto-installing dap adapters for which a dap configuration exists (dap.configurations.php-debug-adapter in this case) but the dap adapter is not installed yet. So, it's actually mason-nvim-dap that does the installing based on the mappings from dap configurations to dap adapters I believe (same as mason-lspconfig does for LSP servers).

PR #6090, opened 2025-05-19
Fix screen-based up/down mappings in vscode (`j/k` -> `gj/gk`)

danielwe:
## Description

Currently, LazyVim's screen-based up/down mappings (`j/k` -> `gj/gk` et cetera) don't take effect in vscode. The reason is that the `vscode-neovim` extension maps `gj/gk` to a special command. The `j/k` -> `gj/gk` mappings need `remap = true` to pick up on this.

(As far as I understand, the reason `vscode-neovim` maps `gj/gk` to a custom command is that the embedded neovim instance doesn't know the viewport width, so soft-wrapping happens on the vscode side. Hence, screen-based line movements require a callback into vscode. See https://github.com/vscode-neovim/vscode-neovim/blob/d53b769447adfdcda931129b7ccafec3c63fc7da/runtime/vscode/overrides/vscode-motion.vim#L14-L16)

## Checklist


theol0403:
Sorry, don't merge this. Mapping these has some side effects and race conditions. Cursor updates going from vscode -> neovim are debounced wheras neovim -> vscode are instant. gj is mapped to a vscode movement command

If we could natively support this, we would.

danielwe:
Can you explain a bit further what kind of race conditions you're worried about here? Do you have steps to reproduce a problem caused by this? Just trying to understand

theol0403:
https://github.com/vscode-neovim/vscode-neovim/blob/master/CONTRIBUTING.md#cursormanager

When using gj, this calls vscode's `cursorMove`, which initiates a movement on the vscode side. Then, that movement is detected, debounced, and sent to nvim:

https://github.com/vscode-neovim/vscode-neovim/blob/284df247b4ff9843c12b57ea517495b0d83c2b18/src/cursor_manager.ts#L312

The debouncing is required for two-way sync. Cursor updates from neovim need to be sent immediately to vscode, but then the movement event triggered from vscode can't be sent back to nvim right away without causing race conditions. Instead, vscode waits until any sequence of movements is complete, then after a pause, sends back the position to nvim to confirm. 

This debounce results in jitter on slower machines when spamming gj. It also means doing something like gjdd may not execute the delete on the correct line.

PR #6091, opened 2025-05-19
fix(extras/rust): show Rust impl blocks in outline

sij1nk:
## Description

Hi!

By default, when using the Rust extra, [functions inside impl blocks](https://doc.rust-lang.org/std/keyword.impl.html#implementing-functionality-for-a-type) don't show up in [outline](https://www.lazyvim.org/extras/editor/outline), making the outline largely pointless in my experience.

The problem is that `impl` blocks are `Objects`, which are not included [lazyvim's default kind filter](https://github.com/LazyVim/LazyVim/blob/25abbf546d564dc484cf903804661ba12de45507/lua/lazyvim/config/init.lua#L102).

## Related Issue(s)

 - no issue in lazyvim repo
 - related issue from outline repo: https://github.com/hedyhli/outline.nvim/issues/81

## Screenshots

Before:

20250519_11h04m53s_grim

After:

20250519_11h05m25s_grim

## Checklist

### DIFF
--- a/lua/lazyvim/plugins/extras/lang/rust.lua
+++ b/lua/lazyvim/plugins/extras/lang/rust.lua
@@ -154,4 +154,17 @@ return {
       },
     },
   },
+
+  -- Show `impl` blocks in outline.nvim
+  {
+    "hedyhli/outline.nvim",
+    optional = true,
+    opts = {
+      symbols = {
+        filter = {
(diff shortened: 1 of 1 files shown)


LockedThread:
I like this

folke:
not willing to add this as a default

PR #6095, opened 2025-05-19
fix(chezmoi): add `hidden=false` to fzf picker

### LINKED ISSUE #6094: bug: chezmoi fzf-lua unknown flag --hidden
Thank you for your work!
Unfortunately I noticed strange behavior.

Open config from dashboard results in this:
image

After change to exclude hidden files:
image

I can't find any alias or something that would override any command.

_Originally posted by @t40mas in https://github.com/LazyVim/LazyVim/issues/5275#issuecomment-2877793503_

dpetka2001:
## Description
Since #5275, the fzf picker has changed to use the `files` picker instead of `fzf_exec` like before. 

That means it also inherits the options of `files` picker, where `hidden = true`. Just the command had been changed to use the `chezmoi` command, but it doesn't make sense to have `hidden = true`, since it just appends `--hidden` to the command which was supposed to be `fzf`, but now with `chezmoi` it's not recognized.

Add `hidden = false` to the `files` picker when the command is chezmoi.

## Related Issue(s)
Fixes #6094

## Checklist

### DIFF
--- a/lua/lazyvim/plugins/extras/util/chezmoi.lua
+++ b/lua/lazyvim/plugins/extras/util/chezmoi.lua
@@ -8,7 +8,7 @@ local pick_chezmoi = function()
         fzf_lua.actions.vimcmd_entry("ChezmoiEdit", selected, { cwd = os.getenv("HOME") })
       end,
     }
-    fzf_lua.files({ cmd = "chezmoi managed --include=files,symlinks", actions = actions })
+    fzf_lua.files({ cmd = "chezmoi managed --include=files,symlinks", actions = actions, hidden = false })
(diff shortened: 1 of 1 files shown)


zjrgov:
I have this issue as well and am working around it for now by overriding the keybinding to use my own copy of the `pick_chezmoi` function.

zjrgov:
Not stale! For posterity…
- I believe I heard @folke is away on vacation until maybe November?
- Here's a [Gist with the Chezmoi plugin config I'm using in the interim](https://gist.github.com/zjrgov/8a664a79e66e94f05edf47f11c710580).

folke:
ty

Issue #6111, opened 2025-05-23
bug: rustaceanvim shouldn't ignore some procMacro


yuusheng:
### Did you check docs and existing issues?

- [x] I have read all the LazyVim docs
- [x] I have updated the plugin to the latest version before submitting this issue
- [x] I have searched the existing issues of LazyVim
- [x] I have searched the existing issues of plugins related to this issue

### Neovim version (nvim -v)

11

### Operating system/version

MacOS 15.5

### Describe the bug

I think we should not ignore any procMacro as this might be kind of tricky.

```lua
return {
  "mrcjkb/rustaceanvim",
  ft = { "rust" },
  opts = {
    server = {
      default_settings = {
        ["rust-analyzer"] = {
          -- ...
          procMacro = {
            enable = true,
            ignored = {
              ["async-trait"] = { "async_trait" },
              ["napi-derive"] = { "napi" },
              ["async-recursion"] = { "async_recursion" },
            },
          },
        },
      },
    },
  },
  -- ...
}
```

This will cause:

*rust-analyzer: async_trait: proc-macro is explicitly disabled*

![Image](https://github.com/user-attachments/assets/bd6f0896-b028-4642-a030-804ecfa024f7)

### Steps To Reproduce

Latest LazyVim

### Expected Behavior

procMacro.ignored should be empty

### Repro

```lua
vim.env.LAZY_STDPATH = ".repro"
load(vim.fn.system("curl -s https://raw.githubusercontent.com/folke/lazy.nvim/main/bootstrap.lua"))()

require("lazy.minit").repro({
  spec = {
    { "LazyVim/LazyVim", import = "lazyvim.plugins" },
    -- add any other plugins here
  },
})
```

dpetka2001:
I'm not doing Rust, but this was made by another community member. 

You can configure it however you want in your personal configuration, thus not a bug.

And if you want to improve the default behavior, just submit a PR and maintainer will decide.

91he:
Have you disabled automatic load of rust_analyzer by mason.lspconfig?

JanKaul:
I have the same issue. The proc-macros shouldn't be ignored by default.

mirsella:
disabling the diagnostic insted is a better solution imo https://github.com/LazyVim/LazyVim/pull/6191

murar8:
Same issue for me, don't see anything wrong with `async_trait` if removed from filters, lsp works.

PR #6117, opened 2025-05-24
fix(extras): remove procMacro ignored

yuusheng:
## Description
I think we should remove rust-analyzer procMacro.ignore because it will cause some diagnostics which are not expected.

## Related Issue(s)
https://github.com/LazyVim/LazyVim/issues/6111

## Checklist

### DIFF
--- a/lua/lazyvim/plugins/extras/lang/rust.lua
+++ b/lua/lazyvim/plugins/extras/lang/rust.lua
@@ -86,11 +86,6 @@ return {
             },
             procMacro = {
               enable = true,
-              ignored = {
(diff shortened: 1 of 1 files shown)


mirsella:
not stale

mirsella:
disabling the diagnostic insted is a better solution imo https://github.com/LazyVim/LazyVim/pull/6191

yuusheng:
> disabling the diagnostic insted is a better solution imo #6191

Why? Why do we need to ignore some macros like async_trait?

folke:
Agree, it should be up to the user to ignore certain diagnostics

PR #6178, opened 2025-06-15
fix(rust): RA hanging on root scanned

xusd320:
## Description

I encounter the RA hanging problem when coding on next.js project, and found the solution to fix this at here . https://github.com/rust-lang/rust-analyzer/issues/12613#issuecomment-2096386344


ChunyuPCY:
Not long ago, I also encountered  the same problem and I was quite confused. Your PR is useful to me. Thank you.

PR #6218, opened 2025-07-02
feat(config): dynamically determine LazyVim origin spec

linw1995:
## Description

Replace hardcoded repository reference with dynamic origin detection to support different LazyVim forks automatically.

This only works with https://github.com/folke/lazy.nvim/pull/2016.

## Related Issue(s)

None

## Screenshots

image

## Checklist

### DIFF
--- a/lua/lazyvim/plugins/init.lua
+++ b/lua/lazyvim/plugins/init.lua
@@ -10,9 +10,24 @@ end
 
 require("lazyvim.config").init()
 
+local Git = require("lazy.manage.git")
+local Util = require("lazy.core.util")
+
+local function find_actual_origin_spec()
+  local current_dir = vim.fn.fnamemodify(Util.get_source(), ":h")
+  return {
+    url = Git.get_origin(Util.find_git_root(current_dir)),
+    priority = 10000,
+    lazy = false,
+    opts = {},
+    cond = true,
+    version = "*",
+  }
+end
+
 return {
   { "folke/lazy.nvim", version = "*" },
-  { "LazyVim/LazyVim", priority = 10000, lazy = false, opts = {}, cond = true, version = "*" },
+  find_actual_origin_spec(),
   {
     "folke/snacks.nvim",
     priority = 1000,


dpetka2001:
Why don't you just change the `url` in the plugin spec in your personal configuration? This isn't really needed. And with this you can also not change `version = false` if you want to follow `main`.

linw1995:
> Why don't you just change the `url` in the plugin spec in your personal configuration? This isn't really needed. And with this you can also not change `version = false` if you want to follow `main`.

Thanks for the suggestion! @dpetka2001  You're right that changing the url in the plugin spec would work. However, before implementing this solution, lazy.nvim was constantly showing alerts about the origin changes, which was quite disruptive. 

4971c18b20f0e512b267c582cc9b6b20

dpetka2001:
It should show that origin has changed, so that user is aware. That's totally normal.

linw1995:
I understand that the origin change alert is by design to keep users informed, which makes sense. However, the issue I'm facing is that it's asking me to change the spec back from linw1995/LazyVim to LazyVim/LazyVim due to the hardcoded specification in lua/lazyvim/plugins/init.lua. This creates a conflict between my fork and the expected upstream repository.

dpetka2001:
Read [the comment in this PR](https://github.com/LazyVim/LazyVim/pull/6053#issuecomment-2865604700) where I explain how to do that. What you describe is already feasible. In your screenshot it the error message also says to `run update to fix`. So you just have to press `U` (capital case `u`) to sync with the new origin.

linw1995:
image

I've read the comment in the PR and tried the suggested approach. However, the issue persists - it continues to alert about origin changes regardless of how many times I press U to update. Every time I enter the plugin management UI, it still shows the "Origin has changed" alert, even after running the update multiple times.

dpetka2001:
It works as expected for me.

Here's a new screencast with default LazyVim configuration from scratch. 

[Screencast_20250702_110425.webm](https://github.com/user-attachments/assets/ec270779-1f25-4324-8c85-0dc3e4de83ae)

Try to follow the exact steps and it should be the same for you. I don't know what you're doing in your personal configuration, but this PR doesn't offer anything that can't already be done.

linw1995:
Yes, I found out the difference between your approach and mine. My way of using my fork linw1995/LazyVim was to replace LazyVim/LazyVim directly in config/lazy.lua. Unlike your approach, which adds a new file core.lua in plugins/ that overwrites the URL in future spec merges.

I think modifying `lazy.lua` directly is the more straightforward approach for this use case. LazyVim plugins should not be configured twice, even when isolated in separate files.

dpetka2001:
In my opinion this is wrong. LazyVim is just another plugin and its plugin spec should be configured in `/lua/plugins/` directory just like other plugin specs, otherwise `lazy.nvim` is not aware of the change in the plugin spec. `lazy.lua` file should not be touched for such a purpose.

linw1995:
If a user uses a single file setup, it must be configured in lazy.lua. And this is only a surface-level fix - there must be something we can do in lazy.nvim to avoid all this disruption.

dpetka2001:
You can leave this PR open. I'm just a simple user like you expressing my opinion. Maintainer will decide in the end. Also LazyVim is not designed for single file setup (you can see that from the way it's configured). That might be more appropriate in `lazy.nvim` the package manager.

Your PR will hardcode the LazyVim spec to that function and users will not be able to change the plugin spec in their personal configuration.

Issue #5455, opened 2025-01-23
bug: Unicode (\alpha) in julia triggers cascade of errors


volkerkarle:
### Did you check docs and existing issues?

- [x] I have read all the LazyVim docs
- [x] I have updated the plugin to the latest version before submitting this issue
- [x] I have searched the existing issues of LazyVim
- [x] I have searched the existing issues of plugins related to this issue

### Neovim version (nvim -v)

v0.10.2

### Operating system/version

Archlinux 6.11.9-arch1-1

### Describe the bug

Hello there,

For some time now I'm trying to get the autocompletion running for

blink + LSP + julia-vim

with https://github.com/JuliaEditorSupport/julia-vim and https://github.com/julia-vscode/LanguageServer.jl.

Actually, I think my problem is rather related to an issue originating in julia-vim and not to LazyVim or blink, but since I need to fix it within the configuration in LazyVim, I cannot discuss my problem with them. The language server setup works now, but the only issue is that when I type, for example,

`\alpha`

I'll get the following errors: 

`                     Error detected while processing TextChangedI Autocommands for "*":
                     Error detected while processing CursorMovedI Autocommands for "*":
                       Error executing lua callback: ...are/nvim/lazy/blink.cmp/lua/blink/cmp/lib/text_edits.lua:91: index out of range
stack traceback:
	[C]: in function '_str_byteindex_enc'
	...are/nvim/lazy/blink.cmp/lua/blink/cmp/lib/text_edits.lua:91: in function 'get_line_byte_from_position'
	...are/nvim/lazy/blink.cmp/lua/blink/cmp/lib/text_edits.lua:146: in function 'to_utf_8'
	...are/nvim/lazy/blink.cmp/lua/blink/cmp/lib/text_edits.lua:130: in function 'get_from_item'
	...link.cmp/lua/blink/cmp/completion/windows/ghost_text.lua:70: in function 'draw_preview'
	...link.cmp/lua/blink/cmp/completion/windows/ghost_text.lua:33: in function <...link.cmp/lua/blink/cmp/completion/windows/ghost_text.lua:32>
Error executing lua callback: ...are/nvim/lazy/blink.cmp/lua/blink/cmp/lib/text_edits.lua:91: index out of range
stack traceback:
	[C]: in function '_str_byteindex_enc'
	...are/nvim/lazy/blink.cmp/lua/blink/cmp/lib/text_edits.lua:91: in function 'get_line_byte_from_position'
[...]

PR #6231, opened 2025-07-05
feat(lang): julia support

storopoli:
## Description

- Another take on #4436.
- Adds LSP
- Adds cmp and blink.cmp for LaTeX symbols

## Related Issue(s)

#5455 

## Screenshots

CleanShot 2025-07-05 at 08 05 01@2x

## Checklist


frankier:
Looks good!

I foresee two small problems. One is that the default config formats on save by default, but Julia formatting is quite up in the air at the moment. JuliaLanguageServer.jl is stuck on JuliaFormatter.jl 1.x, which breaks valid Julia such as using the `public` statement, while 2.x is quite unstable. There's also Runic.jl which is a bit more stable, but much less popular.

Since this is a bit of a trap, and there's no canonical option, I propose to just disable formatting by default, and users can opt-in to the formatter they want instead.

```
  {
    "stevearc/conform.nvim",
    optional = true,
    opts = {
      formatters_by_ft = {
        julia = { lsp_format = "never" },
      },
    },
  },
```

The other thing to note is that enabling https://github.com/kdheepak/cmp-latex-symbols for all file types might be a bit annoying. Ideally it would only be enabled for Julia by default since it clutters up the completions for other languages, and is actively harmful for LaTeX -- where it will replace LaTeX commands with Unicode symbols which won't work in many common LaTeX environments. Unfortunately the recommended way of doing this https://cmp.saghen.dev/recipes.html#dynamically-picking-providers-by-treesitter-node-filetype seems like it won't work well with settings inheritance. Perhaps this could be added to the blink.cmp spec? Or else something programmatic comparable to https://github.com/hrsh7th/nvim-cmp/issues/666#issuecomment-1000925581 (which is from nvim-cmp not blink.cmp -- but maybe it could be adapted). Any thoughts?

frankier:
Actually for the second one I suppose that just adding ft="julia" should do the trick.

frankier:
I'm not sure why, but whatever I do, I'm not able to blacklist the LSP formatter altogether (advice welcome). In the end I just added this to my autocmds.lua:

```julia
local set_autoformat = function(pattern, bool_val)
  vim.api.nvim_create_autocmd({ "FileType" }, {
    pattern = pattern,
    callback = function()
      vim.b.autoformat = bool_val
    end,
  })
end

set_autoformat({ "julia" }, true)
```

Cribbed from here: https://github.com/LazyVim/LazyVim/discussions/141#discussioncomment-9266704

Not sure how to blacklist the LSP formatter, or what would be best for an extra. I guess something like this could be adapted? https://github.com/LazyVim/LazyVim/discussions/6173#discussioncomment-13592350

dpetka2001:
@frankier Yes, LazyVim registers its own LSP formatter. I believe the only option to disable it is to either implement the suggestion in #6173 for a global solution or to explicitly set the format handlers in each LSP server to just empty function that don't return anything.

I don't know why maintainer made this decision, since `conform.nvim` also has the ability to fallback to LSP formatting. That is a question better forwarded to him.

I believe we could also change here https://github.com/LazyVim/LazyVim/blob/25abbf546d564dc484cf903804661ba12de45507/lua/lazyvim/util/lsp.lua#L171-L174 into 
```lua
      local ret = vim.tbl_filter(function(client)
        return (
          client.supports_method("textDocument/formatting")
          or client.supports_method("textDocument/rangeFormatting")
        ) and client.config.capabilities.textDocument.formatting ~= false
      end, clients)
```
and then people could do in their personal configuration something like 
```lua
lua_ls = {
  capabilities = {
    textDocument = {
      formatting = false,
    },
  },
}
```
to disable the LSP formatting for a specific LSP server. But I'm not confident if that causes any other side-effects, so it would be up to maintainer as to how better implement this. Code taken from [this commit](https://github.com/LazyVim/LazyVim/commit/fab6150da919a094614752f95f5fe4e7f55832f8) where such an option existed in the past before the transition to `conform.nvim` and LazyVim's explicit LSP formatter.

Edit: changed to use client capabilities corresponding fields instead of server capability `documentFormattingProvider`.

frankier:
Thanks for the explanation! I wonder too... Perhaps an oversight? I agree we need input from folke for a proper solution.

frankier:
This is what I've ended up at for the LaTeX symbols:

```lua
  {
    "https://gitlab.com/ExpandingMan/cmp-latex",
    ft = "julia",
    init = function()
      -- cmp integration
        pattern = "julia",
        callback = function()
          if LazyVim.has_extra("coding.nvim-cmp") then
            local cmp = require("cmp")

            -- global sources
            ---@param source cmp.SourceConfig
            local sources = vim.tbl_map(function(source)
              return { name = source.name }
            end, cmp.get_config().sources)

            -- add latex_symbols source
            table.insert(sources, { name = "latex_symbols" })

            -- update sources for the current buffer
            cmp.setup.buffer({ sources = sources })
          end
        end,
      })
    end,
  },

  -- cmp integration
  {
    "hrsh7th/nvim-cmp",
    optional = true,
    dependencies = { "cmp-latex" },
  },

  -- blink.cmp integration
  {
    "saghen/blink.cmp",
    optional = true,
    dependencies = { "cmp-latex", "saghen/blink.compat" },
    opts = {
      sources = {
        per_filetype = {
          julia = { inherit_defaults = true, "latex_symbols" },
        },
        providers = {
          latex_symbols = {
            name = "latex_symbols",
            module = "blink.compat.source",
            async = true,
            score_offset = 200,
          },
        },
      },
    },
  },
```

I use https://gitlab.com/ExpandingMan/cmp-latex since it works exactly like the Julia REPL. Then it turns out it to be a fit of a faff to restrict the suggestions to Julia, but this leaves my LaTeX files unicode free as is sometimes needed.

folke:
ty

PR #6250, opened 2025-07-13
feat(extras): utils: add `gh.nvim`

storopoli:
## Description

Adds [`gh.nvim`](https://github.com/ldelossa/gh.nvim) to the extras.

`gh.nvim` brings a way richer reviewer interface to NeoVim by supporting LSP and treesitter natively in review diff buffers.
I've been using successfully for a couple weeks instead of `Octo.nvim`.

## Related Issue(s)

None

## Screenshots

CleanShot 2025-07-13 at 08 02 10@2x

## Checklist


lua/lazyvim/plugins/extras/util/gh.lua
```diff
@@ -0,0 +1,64 @@
...
+    "ldelossa/gh.nvim",
+    dependencies = {
+      {
+        "ldelossa/litee.nvim",
+        config = function()
+          require("litee.lib").setup()
+        end,
+      },
+    },
+    event = "VeryLazy",
```
dpetka2001, line 17: Is this needed? You're already defining `keys`, which means that the plugin will be lazy-loaded on the keymaps that the user will press. I'm genuinely asking because I haven't used the plugin itself. Is it needed to load on `VeryLazy` instead of when the user presses one of the mappings?

dpetka2001: I tried the PR locally and it also works without `event = "VeryLazy"` from my minimal testing. I don't think there's a reason to have this when the plugin can also be loaded on `keys`.

Also thank you for bringing this plugin to my attention. Wasn't aware of this and will give it a go :smile:

lua/lazyvim/plugins/extras/util/gh.lua
```diff
@@ -0,0 +1,63 @@
...
+      {
+        "ldelossa/litee.nvim",
+        config = function()
+          require("litee.lib").setup()
+        end,
+      },
+    },
+    config = function()
+      require("litee.gh").setup()
+    end,
```
dpetka2001, line 19:
```suggestion
    config = function(_, opts)
      require("litee.gh").setup(opts)
    end,
```
This way if a user wants to change default settings, he can directly change them in `opts` in his personal configuration without needing to overwrite the `config` function.

folke:
Thanks for the PR! I've updated the keymaps to use `<leader>G` (capital G) prefix and simplified the plugin setup. Changes:

- Moved to `<leader>G*` keymaps to avoid conflicts with git keymaps
- Made litee.nvim a lazy dependency 
- Simplified config setup with opts pattern
- Removed explicit which-key configuration (now automatic with empty desc)

―  

PR #6253, opened 2025-07-14
fix(treesitter): use the treesitter main branch

OneOfOne:
## Description
Treesitter's master branch has been archived, this switches to the main branch.

## Related Issue(s)
N/A

## Checklist

### DIFF
--- a/lua/lazyvim/plugins/treesitter.lua
+++ b/lua/lazyvim/plugins/treesitter.lua
@@ -25,7 +25,6 @@ return {
       -- Luckily, the only things that those plugins need are the custom queries, which we make available
       -- during startup.
       require("lazy.core.loader").add_to_rtp(plugin)
-      require("nvim-treesitter.query_predicates")
     end,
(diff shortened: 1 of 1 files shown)


dpetka2001:
There's already a branch from when Folke made it available as an Extra before a request was made from `nvim-treesitter` maintainer to not make it available yet at that time. You can see it [here](https://github.com/LazyVim/LazyVim/compare/main...treesitter_rewrite). 

Your PR leaves some things unfixed. For example `incremental_selection` and `textobjects` are not available any more in `main` branch. The `init` function is not needed any more. Also the options `opts.indent.enable` and `opts.highlight.enable` are not taken into account (instead you always enable highlighting on `FileType` event with no apparent way for a user to disable it if he wants to).

My guess is that branch that I linked will become the default implementation with some minor changes maybe. I have been using it as a custom User Extra with no apparent problems so far.

OneOfOne:
I wasn't aware, thank you.

PR #6265, opened 2025-07-17
feat(extra): Allow users to add custom bundles through opts.init_options.bundles to jdtls

g0ne150:
Allow users to add custom bundles through opts.init_options.bundles to jdtls with lang java extra is on.


dpetka2001:
You can already do that through `opts.jdtls`, no? 

This [code here](https://github.com/LazyVim/LazyVim/blob/25abbf546d564dc484cf903804661ba12de45507/lua/lazyvim/plugins/extras/lang/java.lua#L178-L187) is supposed to allow the user to extend settings through `opts.jdtls`.

g0ne150:
> You can already do that through `opts.jdtls`, no?
> 
> This [code here](https://github.com/LazyVim/LazyVim/blob/25abbf546d564dc484cf903804661ba12de45507/lua/lazyvim/plugins/extras/lang/java.lua#L178-L187) is supposed to allow the user to extend settings through `opts.jdtls`.

Actually no, setting `opts.jdtls.init_options.bundles` would overwrite the java-debug and java-test JARs that were added above, rather than extending bundles value.

dpetka2001:
Oh I see. `bundles` is only `local` to `config` function and not exposed in `opts`. I didn't notice that.

qw457812:
Actually, I think we can, I have specs like [this](https://github.com/qw457812/dotfiles/blob/dbb7ce5e642aedd2136a245232107773acfa5bd0/dot_config/nvim/lua/plugins/lang/java.lua#L116-L138):

```lua
  {
    "mfussenegger/nvim-jdtls",
    optional = true,
    opts = {
      jdtls = function(config)
        if not vim.g.user_is_termux then
          config.cmd = vim.list_extend(vim.deepcopy(config.cmd), {
            "--jvm-arg=-Xms8g",
            "--jvm-arg=-Xmx16g",
          })
        end

        -- https://github.com/LazyVim/LazyVim/pull/5218
        config.capabilities = config.capabilities
          or LazyVim.has("blink.cmp") and require("blink.cmp").get_lsp_capabilities()
          or nil

        config.handlers = config.handlers or {}
        -- config.handlers["$/progress"] = function() end -- disable progress updates
        config.handlers["language/status"] = function() end -- mute; having progress reports is enough

        if LazyVim.has("spring-boot.nvim") then
          config.init_options = config.init_options or {}
          config.init_options.bundles =
            vim.list_extend(vim.deepcopy(config.init_options.bundles or {}), require("spring_boot").java_extensions())
        end
      end,
    },
  },

  -- https://github.com/AstroNvim/astrocommunity/blob/5f74d5fb8d8dc9b8e2904846809121068d7afaca/lua/astrocommunity/pack/spring-boot/init.lua
  {
    "JavaHello/spring-boot.nvim",
    ft = {
      "java",
      -- "yaml",
      "jproperties",
    },
    ---@type bootls.Config|{}
    opts = {},
  },
  {
    "mason-org/mason.nvim",
    opts = { ensure_installed = { "vscode-spring-boot-tools" } },
  },
```

I have confirmed with adding `print(vim.inspect(config))` above this line:
https://github.com/LazyVim/LazyVim/blob/c20c4022958780617424a090eee2360d560eb939/lua/lazyvim/plugins/extras/lang/java.lua#L190-L190 

Tips about using `opts.jdtls` function:

- When opening a java file, `opts.jdtls` is called every time, and the parameter `config` of `opts.jdtls` seems to share the same instance. So here you need to use `vim.deepcopy`. You can add `print(vim.inspect(config))` and open two java files to make sure the config is not duplicated.

dpetka2001:
> When opening a java file, opts.jdtls is called every time, and the parameter config of opts.jdtls seems to share the same instance

@qw457812 That's because the config is extended/overwritten every time the LSP server attaches. Maybe the config could be moved out of the function `attach_jdtls`, so that it's only calculated once when the plugin loads? Or you could also use `LazyVim.dedup()` to wrap your `list_extend`, so that the list doesn't end up with duplicated bundles. I guess this is also the same for every property that is a list like `cmd` as well.

I don't do Java and can't test thoroughly, but if you feel like testing it out and creating an appropriate PR feel free. 

I also tested what you said and indeed it's already possible to do what this PR tries to do, making it redundant. I'm guessing the problem that OP had about `bundles` being overwritten is because he didn't use `vim.list_extend`.

g0ne150:
> Actually, I think we can, I have specs like [this](https://github.com/qw457812/dotfiles/blob/dbb7ce5e642aedd2136a245232107773acfa5bd0/dot_config/nvim/lua/plugins/lang/java.lua#L116-L138):
> 
> ```lua
>   {
>     "mfussenegger/nvim-jdtls",
>     optional = true,
>     opts = {
>       jdtls = function(config)
>         if LazyVim.has("spring-boot.nvim") then
>           config.init_options = config.init_options or {}
>           config.init_options.bundles =
>             vim.list_extend(vim.deepcopy(config.init_options.bundles or {}), require("spring_boot").java_extensions())
>         end
>       end,
>     },
>   },
> ```
> 
> Tips about using `opts.jdtls` function:
> 
> * When opening a java file, `opts.jdtls` is called every time, and the parameter `config` of `opts.jdtls` seems to share the same instance. So here you need to use `vim.deepcopy`. You can add `print(vim.inspect(config))` and open two java files to make sure the config is not duplicated.

I think users should declare their configuration rather than hack it.

Since LazyVim's Java Extra takes care of the bundles configuration, there should be a more declarative way to extend the configuration.

I'm opening a new pr but ignore opts.jdtls.init_options.bundles, use opts.init_options.bundles only.

PR #6296, opened 2025-07-28
fix(keymap): remove select mode remaps of printable characters

akioweh:
## Description

Most of the "visual mode" mappings use "v" instead of "x", and therefore also affect select mode.
If the key is a printable character, it is now unusable in select mode. This is most prominent with `<leader>`, which is space by default, a printable character.

The most common use of select mode is when it is automatically triggered by snippets. Currently, trying to type space, g, <, >, etc. as the first character of a snippet field will trigger their bind instead of actually inserting the character.

I cannot currently think of any good reason why anyone would rely on using select mode to execute any of these mapped actions.

## Checklist

### DIFF
--- a/lua/lazyvim/plugins/editor.lua
+++ b/lua/lazyvim/plugins/editor.lua
@@ -54,7 +54,7 @@ return {
       defaults = {},
       spec = {
         {
-          mode = { "n", "v" },
+          mode = { "n", "x" },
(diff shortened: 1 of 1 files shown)


folke:
Great catch! This fixes a real UX issue with snippets. Thanks for the thorough PR!

―  

PR #6311, opened 2025-08-03
feat(go): add linting with golangci-lint

Bahaaio:
## Description

Add Go linting support using [golangci-lint](https://github.com/golangci/golangci-lint) and nvim-lint.

- Installs golangci-lint via Mason.
- Configures nvim-lint to run it on Go files.

This improves the Go development experience by running a fast, popular linter automatically

## Related Issue(s)

None

## Checklist

### DIFF
--- a/lua/lazyvim/plugins/extras/lang/go.lua
+++ b/lua/lazyvim/plugins/extras/lang/go.lua
@@ -97,6 +97,22 @@ return {
       })
     end,
   },
+  -- Add linting
+  {
+    "mfussenegger/nvim-lint",
+    optional = true,
+    dependencies = {
+      {
+        "mason-org/mason.nvim",
+        opts = { ensure_installed = { "golangci-lint" } },
+      },
+    },
+    opts = {
+      linters_by_ft = {
+        go = { "golangcilint" },
+      },
+    },
+  },
   {
     "stevearc/conform.nvim",
     optional = true,


AtifChy:
@BahaaMohamed98 Doesn't `gopls` also work as a linter?

Bahaaio:
@AtifChy Yes, `gopls` does some basic linting, but `golangci-lint` runs many more linters in parallel (`errcheck`, `ineffassign`, etc.).
full list: https://golangci-lint.run/usage/linters/

Bahaaio:
Projects like [docker/cli](https://github.com/docker/cli), [docker/compose](https://github.com/docker/compose), and [bubbletea](https://github.com/charmbracelet/bubbletea) also use golangci-lint in production — just to name a few.

Skylli202:
Hi, @Bahaaio thank you for this PR. This would probably solve one of my new issue. We started to use `golangci-lint` in our code base and LazyVim with the lang Go extra is strictier than golangci-lint.

An example of this would be the staticcheck (rule ST1000 or ST1003). They are disabled by default because they are too opinionated. This is documented [here](https://staticcheck.dev/docs/configuration/options/#checks).

Bahaaio:
@Skylli202 Yes,  `golangci-lint` is widely used, more flexible, and less opinionated by default.
You can then disable `staticcheck` by overriding LazyVim's config:

```lua
{
    "neovim/nvim-lspconfig",
    opts = {
      servers = {
        gopls = {
          settings = {
            gopls = {
              staticcheck = false,
            },
          },
        },
      },
    },
  },
```

Maybe I should add this part as a dependency so that `staticcheck` is automatically disabled when  `golangci-lint` is enabled.

Skylli202:
> @Skylli202 Yes, `golangci-lint` is widely used, more flexible, and less opinionated by default. You can then disable `staticcheck` by overriding LazyVim's config:
> 
> ```lua
> {
>     "neovim/nvim-lspconfig",
>     opts = {
>       servers = {
>         gopls = {
>           settings = {
>             gopls = {
>               staticcheck = false,
>             },
>           },
>         },
>       },
>     },
>   },
> ```
> 
> Maybe I should add this part as a dependency so that `staticcheck` is automatically disabled when `golangci-lint` is enabled.

Thank you for your reply @Bahaaio however, I believe this would completely disable `staticcheck` while default golangci-lint only has a few rules disabled, not all of them.

folke:
Thanks for adding golangci-lint support! This is a widely used linter in the Go community and adds value beyond gopls diagnostics. LGTM!

―  

PR #6384, opened 2025-09-02
fix(venv-selector): update venv-selector.nvim branch to 'main'

you-n-g:
## Related Issue(s)
image

## Checklist

### DIFF
--- a/lua/lazyvim/plugins/extras/lang/python.lua
+++ b/lua/lazyvim/plugins/extras/lang/python.lua
@@ -114,7 +114,7 @@ return {
 
   {
     "linux-cultist/venv-selector.nvim",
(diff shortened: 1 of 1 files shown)


toondaey:
Should this use the main branch directly? I'm not sure if lazyvim allows using commits, but I think that would be more logical because anything can be merged to the main branch at anytime and this could break LazyVim IMO.

ckane:
> Should this use the main branch directly? I'm not sure if lazyvim allows using commits, but I think that would be more logical because anything can be merged to the main branch at anytime and this could break LazyVim IMO.

I vote for using the `main` branch...which also becomes the behavior if you just remove the `branch = ` field altogether. It's what the upstream package author says to do and it's how most of the rest of the packages in these files are handled.

PR #6386, opened 2025-09-02
fix(typescript): DAP configuration for Typescript

dqnk:
## Description
DAP configuration for Typescript did not work right with imports as described in #6382 . Setting the `runtimeExecutable` to `ts-node` in the `Launch File` config fixes the issue. Additionally also set `sourceMaps`, because else the debugger complains that it can't read them. 

The `skipFiles` field is optional but I feel like it makes sense if I am already updating the debugger (it skips lib files). 

Added settings in `Attach` config might not be necessary (I have never used `Attach), but I'm assuming this makes sense?

## Related ~~Issue(s)~~ Discussion

https://github.com/LazyVim/LazyVim/discussions/6382#discussion-8826409

## Checklist

### DIFF
--- a/lua/lazyvim/plugins/extras/lang/typescript.lua
+++ b/lua/lazyvim/plugins/extras/lang/typescript.lua
@@ -247,6 +247,10 @@ return {
               name = "Launch file",
               program = "${file}",
               cwd = "${workspaceFolder}",
+              sourceMaps = true,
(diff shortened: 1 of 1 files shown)


dqnk:
Alternatively we could just point people to configure their `.vscode/launch.json`, but since this is Typescript specific, I thought that using `ts-node` would be a sane default.

folke:
I refactored it a bit. Do let me know if it would not work for you.

PR #6395, opened 2025-09-04
fix(snacks_picker): merge user config with call options

cesconix:
## Description

User configuration for snacks picker may not be applied when called via `LazyVim.pick()`. The picker.open() function currently only passes call-specific options to Snacks.picker.pick(), without explicitly merging user configuration from Snacks.config.picker[source].

For example, this configuration might be ignored:

```lua
return {
  "folke/snacks.nvim",
  opts = {
    picker = {
      grep = {
        regex = false,
      },
    },
  },
}
```

## Checklist

### DIFF
--- a/lua/lazyvim/plugins/extras/editor/snacks_picker.lua
+++ b/lua/lazyvim/plugins/extras/editor/snacks_picker.lua
@@ -18,6 +18,7 @@ local picker = {
   ---@param source string
   ---@param opts? snacks.picker.Config
   open = function(source, opts)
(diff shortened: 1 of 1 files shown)


dpetka2001:
The configuration should actually be
```lua
return {
	"folke/snacks.nvim",
	opts = {
		picker = {
			sources = {
				grep = {
					regex = false,
				},
			},
		},
	},
}
```
With this it works as expected for me.

cesconix:
Thanks @dpetka2001, you’re right — using picker.sources.grep works correctly on my side too.

PR #6419, opened 2025-09-15
fix(lang.haskell): prevent Haskell extras from installing telescope.nvim

ChristophSchmidpeter:
## Description
Installing the Haskell extras currently installs telescope.nvim as well,
since the dependency is not marked optional. This marks it optional so
telescope.nvim is no longer pulled in automatically.

## Checklist


folke:
ty

PR #6425, opened 2025-09-15
feat(extras): add workflow.hardtime and workflow.precognition

phucisstupid:
## Checklist

### DIFF
--- a/lua/lazyvim/plugins/extras/workflow/hardtime.lua
+++ b/lua/lazyvim/plugins/extras/workflow/hardtime.lua
@@ -1,0 +1,14 @@
+return {
+  "m4xshen/hardtime.nvim",
+  lazy = false,
+  dependencies = { "MunifTanjim/nui.nvim" },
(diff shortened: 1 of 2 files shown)


dpetka2001:
These are plugins that each user can just add to his personal configuration. They don't actually provide an improvement to LazyVim's default configuration. 

I don't think LazyVim should be a place to make other plugins' discoverability easier. This can create an unwanted precedent for other plugins as well. There are already other sites and even Neovim plugins as of recently that provide some kind of store for users to discover plugins.

Of course I'm just stating my personal opinion as a user and it's up to the maintainer to decide.

yriveiro:
Maybe we can create a category for beginner extras?

It's easy to say that everyone can do it in their personal configuration, but the reality is that if you are joining Vim for the first time and using Neovim + Lazy, a beginner mode could improve adoption.

When you start doing something, it's not always easy to find/choose the best plugins to help you to master the tool.

PR #6427, opened 2025-09-15
fix(json,yaml): use `before_init` instead of `on_new_config`

AtifChy:
## Description

`vim.lsp.config` doesn't provide `on_new_config`, which is causing `b0o/SchemaStore.nvim` fail to load.
This PR replaces `on_new_config` with `before_init` fixing the issue.

## Checklist


folke:
ty!

PR #6428, opened 2025-09-15
fix(ocaml): use `root_dir` instead of `root_markers` for globs support

dpetka2001:
## Description
As per [comment](https://github.com/LazyVim/LazyVim/commit/23b9cdeb3471b655532e9884fa2dd36ee83062d5#r165830889), use `root_dir` with `util.root_pattern` instead of `root_markers` for globs support.

I don't actually do Ocaml to thoroughly test it, but it should work in theory :stuck_out_tongue: 

## Related Issue(s)
None

## Checklist


folke:
ty!

PR #6430, opened 2025-09-16
fix(treesitter-main): set vim.bo.indentexpr in FileType autocmd

qw457812:
## Description

Ref: https://github.com/nvim-treesitter/nvim-treesitter/blob/7aa24acae3a288e442e06928171f360bbdf75ba4/scripts/minimal_init.lua#L19-L24

## Checklist


folke:
Makes sense, thank you!

PR #6431, opened 2025-09-16
feat(lang.nix): Use `alejandra` formatter instead of `nixfmt`

phucisstupid:
## Description
This pr change `conform.nvim` formatter to [Alejandra]( https://github.com/kamadorueda/alejandra) because It's faster and prettier

## Checklist


amaanq:
I'd think keeping the default in LazyVim as nixfmt would be better, and you can override this in your config to use alejandra

folke:
I agree with @amaanq. Thank you anyway :)

PR #6483, opened 2025-09-20
fix(mini.animate): schedule the toggle mapping to correctly take effect

dpetka2001:
## Description
`mini.animate`'s toggle mapping didn't take effect, presumably because `keymaps.lua` is the last one to execute on `VeryLazy`. Schedule it to correctly overwrite the default from `keymaps.lua`.

I would also be interested if you don't mind to explain to me how can I find out the order of the events execution if there are multiple executions on the same event (i.e on `VeryLazy` what is the order of the execution? Is it arbitrary?)

## Related Issue(s)
None. Saw a discussion on `snacks.nvim` and OP wanted to change the mapping and when I tested with `mini.animate` Extra I noticed that its mapping wasn't overriding the default one from `keymaps.lua`.

## Checklist

### DIFF
--- a/lua/lazyvim/plugins/extras/ui/mini-animate.lua
+++ b/lua/lazyvim/plugins/extras/ui/mini-animate.lua
@@ -32,15 +32,19 @@ return {
         end,
       })
 
-      Snacks.toggle({
-        name = "Mini Animate",
-        get = function()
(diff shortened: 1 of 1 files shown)


folke:
ty

PR #6498, opened 2025-09-22
fix(jdtls): allow mason jdtls installation

dareni:
Bug fix jdtls config for compatibility with last lsp refactor.

### DIFF
--- a/lua/lazyvim/plugins/extras/lang/java.lua
+++ b/lua/lazyvim/plugins/extras/lang/java.lua
@@ -71,11 +71,6 @@ return {
       servers = {
         jdtls = {},
       },
-      setup = {
-        jdtls = function()
-          return true -- avoid duplicate servers
-        end,
-      },
     },
   },
 


folke:
ty!

agriffis:
Hello @dareni 

Could you explain a bit more what this fixes? I wrote these lines originally but I haven't followed everything to understand why they should be removed. Thanks!

dareni:
Hi @agriffis

When the config for nvim-lspconfig is processed the configured lsp(jdtls in this case) is installed by mason only when 'setup' does not exist(for nvim-lspconfig).  Removing setup() allows mason to do the install and the processor of the config ensures 'duplicate servers' will be avoided.

qw457812:
@dareni Hi, I think this PR has some problems:

Before this PR, we are using [nvim-jdtls](https://github.com/mfussenegger/nvim-jdtls).

After this PR, I think we are using jdtls from nvim-lspconfig.
- It has [less features](https://github.com/neovim/nvim-lspconfig/blob/c023a6832c4301923a20e2c1d65be3d50ff42efd/lsp/jdtls.lua#L7-L9) than nvim-jdtls.
- The `opts.jdtls` of nvim-jdtls that LazyVim [provided](https://github.com/LazyVim/LazyVim/blob/c4237659621e406089f4d13d32192954d52c4924/lua/lazyvim/plugins/extras/lang/java.lua#L158-L168) to override config not working anymore. [My opts.jdtls](https://github.com/qw457812/dotfiles/blob/00673493fe35ba6c6e1d09d4e240a3f7c6a60585/dot_config/nvim/lua/plugins/lang/java.lua#L116-L133) for example.

qw457812:
To fix jdtls auto-install via Mason, I now see two approaches, but I'm not sure about either of them:

1. Remove the [nvim-lspconfig spec](https://github.com/LazyVim/LazyVim/blob/c4237659621e406089f4d13d32192954d52c4924/lua/lazyvim/plugins/extras/lang/java.lua#L65-L75) from extras.lang.java and add:
```lua
  {
    "mason-org/mason.nvim",
    opts = { ensure_installed = { "jdtls" } },
  },
```
However, I'm not sure about the original purpose of using nvim-lspconfig instead of using nvim-jdtls directly. If Mason auto-install is the only concern, then I think we can do it this way.

2. Revert this PR and update [these lines](https://github.com/LazyVim/LazyVim/blob/a90b56518f4f2291985952d79472d4f935c0bad0/lua/lazyvim/plugins/lsp/init.lua#L191-L222) in `plugins/lsp/init.lua`.
I think jdtls auto-install was broken after the <https://github.com/LazyVim/LazyVim/commit/773f28b4912d5505a21da9686d56ab7cf41a9be7> commit (I haven't tested to confirm that this specific commit broke jdtls Mason auto-install). After a quick look, I think all `opts.servers` from nvim-lspconfig were previously added to `ensure_installed` of mason-lspconfig before the migration. We can investigate whether we can do the same thing now.

dpetka2001:
Yes, these [lines here](https://github.com/LazyVim/LazyVim/blob/21518ae8a32cc56cfd8eefbbe33b8fe801d5bb3a/lua/lazyvim/plugins/lsp/init.lua#L217-L219) are responsible for filtering out from `ensure_installed` every server that is also present in the `exclude` table. I think those lines should be
```lua
ensure_installed = vim.list_extend(servers, LazyVim.opts("mason-lspconfig.nvim").ensure_installed or {})
```
@folke What do you think?

PS: actually no, then it installs `ts_ls` server. Something else must be done.

dpetka2001:
I think it should be
```lua
ensure_installed = vim.tbl_filter(function(server)
  return opts.servers[server].enabled ~= false
end, vim.list_extend(servers, LazyVim.opts("mason-lspconfig.nvim").ensure_installed or {})),
```
Because a server should not be installed only when it has `enabled = false`, no?

PS: So, yes I believe this PR should be reverted and the logic for `ensure_installed` changed.

dpetka2001:
I submitted #6504 which should fix the automatic installation for servers. I don't do Java however and am reluctant to submit a PR that reverts the changes by this PR. 

@qw457812 @agriffis @dareni Could you also test the behavior with #6504 and reverting the changes made by this PR? If all is good with just #6504 maybe someone could submit a PR that reverts the changes made in this PR?

qw457812:
@dpetka2001 Thanks. I tested #6504 , jdtls did auto-installed and works as expected, but there is a warning for metals (scala extra).

Add this spec to revert this PR in my config:
```lua
  {
    "neovim/nvim-lspconfig",
    opts = {
      setup = {
        jdtls = function()
          return true
        end,
      },
    },
  },
```

Uninstall jdtls from mason.

Use #6504 and open nvim, wait till mason auto-install finished, notification history:
 ```
 18:17 󰋽 INFO mason.nvim jdtls was successfully installed.
 18:17 󰋽 INFO mason-lspconfig.nvim [mason-lspconfig.nvim] jdtls was successfully installed
 18:15 󰋽 INFO mason-lspconfig.nvim [mason-lspconfig.nvim] installing jdtls
 18:15  WARN mason-lspconfig.nvim [mason-lspconfig.nvim] Server "metals" is not a valid entry in ensure_installed. Make sure to only provide lspconfig server names.
 ```

folke:
I've just reverted this change.
I don't use java, so didn't test...

dpetka2001:
@qw457812 I guess the Scala Extra needs `mason = false`? 

Mason doesn't seem to provide the option to install the language server. See also https://github.com/mason-org/mason.nvim/issues/369. 

Seems that users using `nvim-metals` plugin should just use `:MetalsInstall` command as the recommended way.

PS: I guess with Folke's recent changes that won't be needed, since `metals` won't be listed in `mason_all` and that will result to `use_mason` being `false`, which should not try to install the server via Mason?

folke:
Yep, indeed. It roughly all works again as before the refactor

qw457812:
Yes, all good now, thank you guys.

dareni:
Hi @qw457812,

A lot of this code is new to me, will take a look at why your config no longer works and will get back to you.

qw457812:
> Hi @qw457812,
> 
> A lot of this code is new to me, will take a look at why your config no longer works and will get back to you.

@dareni Hi, it's working now. Folke has fixed my config problems by revert this PR (commit 67781bd), and then he re-fix the mason jdtls installation via commit 37b1ec4. So no need to worry about my config now.

PR #6499, opened 2025-09-22
fix(jdtls): bundle configuration bug fix

dareni:
The jar for the debug adapter was included but the operation to append the java-test jar files failed.


folke:
ty!

PR #6504, opened 2025-09-23
fix(mason): only don't install servers with `enabled=false`

dpetka2001:
## Description
I don't believe we should filter out servers from `exclude` table from installing via Mason (which is the current behavior).

I believe servers should not be installed only when `enabled = false`. 

You might also want to take a look at the comments in #6498 and might revert that commit as it seems to break the Java Extra. See [comment here](https://github.com/LazyVim/LazyVim/pull/6498#issuecomment-3322117658). I don't do Java to thoroughly test.

## Related Issue(s)
None. Just the comment I linked above for the Java Extra PR, which prompted me to test again the current code.

## Checklist


folke:
The logic was indeed not correct. I've had to make some additional changes to fix all cases.

PR #6543, opened 2025-09-28
fix(copilot-chat): use up to date config for chat headers

deathbeam:
The config format changed a while ago due to function calling support


folke:
ty!

PR #6544, opened 2025-09-28
refactor(lang/dotnet): rename omnisharp.lua to dotnet.lua

ChristophSchmidpeter:
## Checklist


sebastianstudniczek:
Seems like a good idea to include dotnet in the name but IMO it would be better to preserve `omnisharp` to avoid ambiguity as I would expect more people migrating to roslyn, therefore in the near future such extra probably will be created

folke:
not a good idea. In it's current state it would also break peoples setups when they have it enabled

PR #6547, opened 2025-09-29
Feat(vimtex): add keymaps with whichkey

ChillyWall:
## Description

Since vimtex is written in vimscript, the key bindings don't display properly with which-key.nvim, and you have to remap them in your own config. There's already a solution provided on [vimtex's wiki](https://github.com/lervag/vimtex/wiki/which%E2%80%90key.nvim-support) and I just copy it here.

> Note: pressing `ds` is pressing `d` and `s` within an interval shorter than the value of `timeout_len` you set (default to 300), or only flash or leap would be triggered instead. `tse`, `tsc` would not be displayed by `ts` for it's also a built-in command of vim, you have to press them as whole ones. 

## Screenshots

Press `<localleader>`
Screenshot from 2025-09-29 13-56-13

Press `<localleader>l`
Screenshot from 2025-09-29 13-56-24

Press `ds`
Screenshot from 2025-09-29 13-56-51

Press `cs`
Screenshot from 2025-09-29 13-57-02

Press `v`
Screenshot from 2025-09-29 13-57-39

Press `v` and `a`
Screenshot from 2025-09-29 13-57-49

Press `v` and `i`
Screenshot from 2025-09-29 13-58-00

Press `[`
Screenshot from 2025-09-29 13-58-12

Press `]`
Screenshot from 2025-09-29 13-58-31

## Checklist


folke:
Thanks for the contribution! However, we prefer to keep LazyVim minimal and let users configure which-key integrations in their own configs. The vimtex wiki already provides a good solution for this, so we'd rather not include it in the base distribution.

―  

PR #6582, opened 2025-10-03
fix(lualine): increase placeholder width for better rendering

mattr-:
## Description

In the pretty path lualine helper, the ellipsis is being cut off. In order to fix that, I've added an extra space of padding in order for things to visually line up correctly again.

I'm using Iosevka as my font and am seeing this in both Wezterm and Ghostty. It could potentially be a font problem, but in case it's not, I decided to submit a potential fix.

## Screenshots

Before:
Screenshot showing the problem

Notice the ellipsis being cut off in the statusline.

After:

Screenshot with the problem fixed

No more cut off!

## Checklist

### DIFF
--- a/lua/lazyvim/util/lualine.lua
+++ b/lua/lazyvim/util/lualine.lua
@@ -124,7 +124,7 @@ function M.pretty_path(opts)
     if opts.length == 0 then
       parts = parts
     elseif #parts > opts.length then
-      parts = { parts[1], "…", unpack(parts, #parts - opts.length + 2, #parts) }
+      parts = { parts[1], "… ", unpack(parts, #parts - opts.length + 2, #parts) }
     end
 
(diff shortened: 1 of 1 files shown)


dpetka2001:
To me it seems like it's either a font problem or a combination of specific font with your terminal font size?

This is how it looks for me with Kitty terminal and Fira Code font



The upper half of the image is with your PR checked out and the bottom half with default LazyVim config. You can see the extra padding that you added also is visible in the upper half image.

mattr-:
I found https://github.com/be5invis/Iosevka/issues/2301, so it is indeed it is a font bug. Looks like Iosevka has U+2026 (the ellipsis) taking two character widths instead of just one. 

Thanks for the double check @dpetka2001! :heart:

Closing this and finding a new font, I guess. :joy: :sob:

PR #6588, opened 2025-10-05
fix(xtras): load `copilot-native` before `blink`

dpetka2001:
## Description
`blink` Extra loads earlier than `copilot-native` Extra and that causes `ghost_text.enabled = vim.g.ai_cmp` to wrongly evaluate to `true` even though `copilot-native` explicitly sets `vim.g.ai_cmp = false`.

I also changed the order of `blink` and `typescript` so that `blink` comes logically right after `copilot-native`, hope that is not a problem. If it is, feel free to revert the order and just keep the change about `copilot-native`.

## Related Issue(s)
None, I was just trying to create a toggle mapping to disable/enable both NES and inline_completion and even  though they were disabled, I could see ghost_text from blink.

## Checklist


folke:
ty!

PR #6657, opened 2025-10-20
fix(avante): unset default keymaps

bryanniwa:
## Description

The extras spec includes its own keymaps for the most common commands now so these are redundant and causing duplicate entries in which-key.

## Checklist


folke:
ty!

jmarcelomb:
Hello, I noticed that it removed the keybinds in my config. I enabled again.

PR #6156, opened 2025-06-06
this fix #6155

CaueAnjos:
## Description
This PR updates the CSharpier integration in LazyVim to reflect changes in how Mason installs the formatter.

Previously, Mason installed CSharpier as the `dotnet-csharpier` binary, which allowed running it without additional arguments. Now, Mason installs it simply as `csharpier`, requiring the explicit format argument for it to work correctly.

This change updates the formatter command and arguments accordingly to ensure compatibility with the latest Mason installation and CSharpier CLI behavior.

## Changes
Replaced `dotnet-csharpier` with `csharpier` as the executable command.

Added the required `format` argument to the command invocation.

## Related Issue(s)
Fixes #6155

## Checklist
 I have read the [CONTRIBUTING](https://github.com/LazyVim/LazyVim/blob/main/CONTRIBUTING.md) guidelines.

PR #6665, opened 2025-10-21
fix(dart): add file name to formatter

jordyslagter:
Fix dart_format failing to format because there wasn't a file name included.

## Description

I noticed the dart formatter wasn't working when installing the lang.dart extra without adding any of my own configuration. I tested this in a Docker container with a new install of LazyVim and lang.dart as well and the problem happens there as well.
The problem originates from the $FILENAME not being included in the dart format command, this PR adds that and thus fixes the formatting for dart.

## Related Issue(s)

-

## Screenshots

-

## Checklist

### DIFF
--- a/lua/lazyvim/plugins/extras/lang/dart.lua
+++ b/lua/lazyvim/plugins/extras/lang/dart.lua
@@ -22,7 +22,7 @@ return {
     opts = {
       formatters = {
         dart_format = {
-          args = { "format", "--line-length", "120" },
+          args = { "format", "--line-length", "120", "$FILENAME" },
         },
       },
       formatters_by_ft = {


dpetka2001:
My suggestion would be the same as #6156. 

Just remove lines 23-27. These settings get inherited from `conform.nvim` upstream and preferably users should change these settings in their personal configuration if they want something different.

folke:
Agreed, that's just not needed. Remove those lines

dpetka2001:
Along with this PR, I would also suggest to make a change to Solidity Extra and remove these lines as well on the same precipice 

https://github.com/LazyVim/LazyVim/blob/e8a1d8b6286f82d75e5830de5ea93bd7622cec8c/lua/lazyvim/plugins/extras/lang/solidity.lua#L35-L40

In general it would be a good guideline to avoid overwriting `cmd` and `args` about formatters, as they tend to lead to unforeseeable breaking changes that LazyVim has to account for if they do get overwritten in some Extra.

PR #6690, opened 2025-10-24
fix(extras.lang): update nvimMetals dependencies

linux-root:
## Description
- Removed  Plenary  as nvim-metals's dependency due to https://github.com/scalameta/nvim-metals/pull/753
- explicitly added telescope as a nvim-metals' dependency to avoid issue if LazyVim users don't have telescope lazyvim extra
## Checklist


dpetka2001:
I don't think Telescope should explicitly be a dependency. It's only being used for a specific mapping, so guard that mapping with `LazyVim.has("telescope.nvim")` and it will only be present for users that have Telescope installed.

dpetka2001:
While you're at it with this PR I would also suggest to change all the mappings to use `<localleader>`, see also latest addition in [Contributing](https://github.com/LazyVim/LazyVim/blob/main/CONTRIBUTING.md#language-specific-keymaps). 

You can see latest Haskell PR #6230 for an example.

linux-root:
> I don't think Telescope should explicitly be a dependency. It's only being used for a specific mapping, so guard that mapping with `LazyVim.has("telescope.nvim")` and it will only be present for users that have Telescope installed.

@dpetka2001  That's a good point! I do have a small concern, though - from my experience, one of the nice things about LazyVim is having solid “good defaults” right out of the box. I actually didn’t even know what Telescope was until I started using LazyVim for coding in Scala

dpetka2001:
A good default is that mappings should only be available when a certain plugin is also available (if the mapping requires an explicit plugin to work). Since Telescope is an Extra and optional for the users to enable, it doesn't really make sense to make it a hard dependency of an Extra.

PS: Btw I'm just a simple user like you expressing my personal opinion and the maintainer will make the final decision. So, the above is just a suggestion of mine of what I understand as a good practice from reading [Contributing.md](https://github.com/LazyVim/LazyVim/blob/main/CONTRIBUTING.md) and my personal experience as a LazyVim user thus far.

linux-root:
> A good default is that mappings should only be available when a certain plugin is also available (if the mapping requires an explicit plugin to work). Since Telescope is an Extra and optional for the users to enable, it doesn't really make sense to make it a hard dependency of an Extra.

"it doesn't really make sense to make it a hard dependency of an Extra." -> I got your point now. Is it possible to make the Telescope extras install automatically when a user enables the Scala lang extras?

dpetka2001:
> Is it possible to make the Telescope extras install automatically when a user enables the Scala lang extras?

I would not suggest that, because that will also change the fuzzy finder that LazyVim uses. The default is Snacks.picker, but users can also enable fzf-lua or Telescope depending on their preferences from the Extras. But if you automatically enable the Telescope Extra for the Scala Extra that will change the default fuzzy finder for users and I don't think that's a good solution.

So, the best solution from my perspective would be something like in [this comment](https://github.com/LazyVim/LazyVim/pull/6230#discussion_r2187087798) and the mapping will only be available when users have Telescope installed in their personal configuration.

folke:
Just remove plenary. Shouldn;'t have been there in the first place.

For telescope, just add a telescope spec, mark it optional and add the keys on that one.

That way you don't need to use `LazyVim.has` and much cleaner.

dpetka2001:
@folke Out of curiosity, why do you prefer an additional optional plugin spec over `LazyVim.has`? When should it be used?

folke:
It's just cleaner?

It has extra configuration for telescope, but it will only be used if the extra is enabled and telescope exists as a plugin.

PR #6708, opened 2025-10-27
fix(lsp): use correct mason-lspconfig API for getting mappings

prabhat0206:
## Description

Fixes the error: `module 'mason-lspconfig.mappings' not found`

The mason-lspconfig plugin changed its API. The module `mason-lspconfig.mappings` no longer has a `get_mason_map()` function. 

## Changes

Updated line 225 in `lua/lazyvim/plugins/lsp/init.lua` to use the correct API:

**Before:**
```lua
require("mason-lspconfig.mappings").get_mason_map().lspconfig_to_package
```

**After:**
```lua
require("mason-lspconfig").get_mappings().lspconfig_to_mason
```

The correct API is documented in mason-lspconfig at:
https://github.com/williamboman/mason-lspconfig.nvim/blob/main/lua/mason-lspconfig/init.lua#L163-L169

## Testing

- [x] Tested locally and confirmed the error is resolved
- [x] No other breaking changes introduced

## Related Issues

This appears to be related to the refactoring in commit 16637dd6 which fixed this issue, but it was inadvertently reintroduced in a later commit.

### DIFF
--- a/lua/lazyvim/plugins/lsp/init.lua
+++ b/lua/lazyvim/plugins/lsp/init.lua
@@ -222,7 +222,7 @@ return {
       -- get all the servers that are available through mason-lspconfig
       local have_mason = LazyVim.has("mason-lspconfig.nvim")
       local mason_all = have_mason
-          and vim.tbl_keys(require("mason-lspconfig.mappings").get_mason_map().lspconfig_to_package)
+          and vim.tbl_keys(require("mason-lspconfig").get_mappings().lspconfig_to_package)
         or {} --[[ @as string[] ]]
       local mason_exclude = {} ---@type string[]
 


prabhat0206:
Updated the fix to use the correct API key: `lspconfig_to_package` instead of `lspconfig_to_mason`.

Verified that:
- The `get_mappings()` function returns a table with `lspconfig_to_package` key (not `lspconfig_to_mason`)
- Neovim starts successfully without errors
- The fix works correctly with the latest mason-lspconfig.nvim

dpetka2001:
> The mason-lspconfig plugin changed its API. The module mason-lspconfig.mappings no longer has a get_mason_map() function.

Where exactly did you see this? Here is the file you mention with the function still there https://github.com/mason-org/mason-lspconfig.nvim/blob/1d77bd86e1c3eaa777010c95dad455b83823f247/lua/mason-lspconfig/mappings.lua#L11-L25

folke:
Your "fix" is incorrect. You're probably still pinned to v1 for mason-lspconfig.

In v2, you need the new API (which is the one you believe is incorrect)

PR #6756, opened 2025-11-06
feat(extras.editor)!: update Leap configuration

ggandor:
## Description

Leap's API has undergone significant changes in the meantime, and flit.nvim has been deprecated recently. The one in the PR is my suggested starter configuration now.

- `opts`: `preview` filter function, some more `equivalence_classes`
- replaced flit.nvim with simple `leap()` wrapper
- added mappings for treesitter node selection (`R` is used by flash's treesitter-search otherwise)
- added mappings for remote operations (`gs`)
- updated default mappings (bidirectional `s`) (breaking)

I'd be fine with keeping the current Sneak-like mappings for the default motions (`s`/`S`/`gs`), my only problem is that it's hard to find good mappings for remote then.

Feel free to discuss if you think you'd rather cherry-pick, but in any case, `add_default_mappings()` is deprecated, and I don't intend to maintain flit.nvim as a separate plugin anymore, at least those parts should be updated. Adding the treesitter mapping is also a no-brainer I guess.

## Checklist

### DIFF
--- a/lua/lazyvim/plugins/extras/editor/leap.lua
+++ b/lua/lazyvim/plugins/extras/editor/leap.lua
@@ -1,37 +1,94 @@
+-- A helper function making it easier to set "clever-f" behavior
+-- (use f/F or t/T instead of ;/, - see the plugin clever-f.vim).
+local function clever(...)
+  return require("leap.user").with_traversal_keys(...)
+end
+
+local function ft(key_specific_args)
+  require("leap").leap(vim.tbl_deep_extend("keep", key_specific_args, {
+    inputlen = 1,
+    inclusive = true,
+    opts = {
+      labels = "",
+      safe_labels = vim.fn.mode(true):match("no?") and "" or nil,
+    },
+  }))
+end
+
 return {
   -- disable flash
   { "folke/flash.nvim", enabled = false, optional = true },
 
-  -- easily jump to any location and enhanced f/t motions for Leap
+  -- Leap provides improved vim-sneak-style navigation with live label
+  -- preview and multi-window scope, and also extra features like remote
+  -- operations and incremental treesitter node selection.
   {
     url = "https://codeberg.org/andyg/leap.nvim.git",
-    enabled = true,
-    keys = function()
-      ---@type LazyKeysSpec[]
-      local ret = {}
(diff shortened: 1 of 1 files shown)


folke:
I trust your judgement, since you have quite some experience with leap :)

So feel free to structure this any way you want.

Thank you!

ggandor:
Okay, thanks, let's go with this then :)

I added some comments for the extra features for people browsing the Lazy Extras documentation or wanting to try out the plugin, to understand right away what the heck the intended use is, without having to deep dive into the README.

dpetka2001:
Sorry, but I don't understand this. Why not define the mappings directly in `keys`? That way users can overwrite them without having to overwrite the whole `config` function. 

Also the second spec overwrites the first one. `s` key not working. Second spec also needs `local leap = require "leap"`, otherwise it throws an error.

Please also try with a default LazyVim installation and enabling this Extra to make sure things work as expected.

PS: I don't think you can use 2 specs with different `name` so that you can have 2 different `config` functions. @folke you might want to take a closer look here.

ggandor:
The url should also be updated, as we're moving to [Codeberg](https://codeberg.org/andyg/leap.nvim), but I'm not sure what to do to satisfy the failing test.

ggandor:
I think we are good to go.

k-soomro:
I don't know if this was the right way to do it, but I copied this file manually into `nvim/lua/plugins/leap.lua` and now pressing `s` actives the plugin but it still appears to be broken. The hints are completely off and highlight the wrong character. Even if you select a hint shown, the cursor often doesn't jump to it.

b0lle:
> I don't know if this was the right way to do it, but I copied this file manually into `nvim/lua/plugins/leap.lua` and now pressing `s` actives the plugin but it still appears to be broken. The hints are completely off and highlight the wrong character. Even if you select a hint shown, the cursor often doesn't jump to it.

I did the same because after updating to lazyvim 15.15.0 leap was broken for me. With the leap.lua file from this PR its working again 🤞🏽

ggandor:
@k-soomro Isn't the preview filter confusing you? See `:help leap.opts.preview`.

PR #6765, opened 2025-11-08
feat(extras.lang): add java dependencies sidebar to java extra

josemiguelo:
## Description

This PR adds support for displaying a list of java dependencies in a sidebar through the [java-deps.nvim](https://github.com/g0ne150/java-deps.nvim) plugin. 

## Related Issue(s)

If the inclusion of this plugin should be avoided for whatever reason (keep the extra simple with the current plugins only for example), we should figure out a way to configure the list of bundles that we pass to jdtls, so that the user of this extra handles this scenario locally.
Other than that, I believe this is a good addition since it enhances the java experience significantly.

## Screenshots

image
image
image

## Checklist

### DIFF
--- a/lua/lazyvim/plugins/extras/lang/java.lua
+++ b/lua/lazyvim/plugins/extras/lang/java.lua
@@ -79,6 +79,20 @@ return {
     },
   },
 
+  -- Opens up a sidebar with all the project's java dependencies
+  -- similar functionality to vscode's: https://github.com/microsoft/vscode-java-dependency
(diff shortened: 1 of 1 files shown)


josemiguelo:
I'll close the issue since it's possible to inject custom bundles into jdtls:
https://github.com/LazyVim/LazyVim/pull/6265#issuecomment-3203831504
I do think adding a plugin that handles java dependencies comes in really handy, tough.

PR #6770, opened 2025-11-10
fix(terminal): partially revert previous mappings

dpetka2001:
## Description
Commit 4efd0e2bea2a387d6fcdea8cf36a62049e8bafed introduced some changes to terminal mappings. I tested in a new LazyVim installation and `sidekick.nvim` is not affected by `<C-/>` any more.

The mappings `hide_underscore` and `hide_slash` don't let user to create new terminals with `2<C-/>`, `3<C-/>` and so on. Instead they hide the terminal and user can't open split terminals next to each other any more.

Therefore, I believe it would be a better default to revert these 2 mappings explicitly.

## Related Issue(s)
None

## Checklist


folke:
Ty!

stefanHoefler:
Seams like with this change using `<C-/>` in a terminal no longer hides it but creates a new split terminal. Is that intended? And how to close it now? Using `<Esc><Esc>:q` is the only way if found.

dpetka2001:
@stefanHoefler It was by accident, because of `cwd` being evaluated when you press the keymap from the terminal as well. #6774 will fix that. It just hasn't been merged yet.

PR #6774, opened 2025-11-11
fix(terminal): terminal mappings again

### LINKED ISSUE #6573: feature: avoid closing sidekick AI CLI pannel with <C-/>
### Did you check the docs?

When I'm in the Sidekick AI CLI panel and use <C-/>, the panel closes. However, I think it makes more sense to treat the Sidekick panel differently from a regular terminal panel and avoid closing it with this shortcut.

Sidekick defines its own set of keymaps for closing/focusing/opening/etc., so it might be better to keep them separate from the keymaps that control terminal buffers.

This is just my personal preference, but others may expect the same behavior. Feel free to close this issue without explanation if it doesn’t align with your tastes.

Disable <C-/> for sidekick_terminal ft.

Provide examples in the docs how to disable <C-/> for sidekick_terminal ft.

### Additional context

_No response_

dpetka2001:
## Description
Previous change results in a weird behavior where you can indeed open multiple terminals vertically, but then whey you press `<C-/>` it always toggles the last terminal. After some discussion in #6773 as well, I guess this is a better compromise.

In my personal opinion, since all this stemmed from https://github.com/LazyVim/LazyVim/commit/4efd0e2bea2a387d6fcdea8cf36a62049e8bafed in order to resolve #6573, I believe it would be best to reinstate the former mappings (which undoubtedly provide a more intuitive overall behavior regarding terminals) and just create a buffer local mapping for Sidekick either in an autocommand as suggested also in #6573 or maybe Sidekick can handle this internally.

This is just my personal opinion though and you of course have the final decision. This PR still leaves some unintuitive behavior when in normal mode with the cursor focused in a terminal window, but it tries to operate based on the assumption that toggling off the terminal makes more sense when in terminal mode rather than normal mode. Of course there might be a better way I can't think of, in which case I'll gladly accept whatever else you come up with over everything I've mentioned above.

## Checklist

### DIFF
--- a/lua/lazyvim/config/keymaps.lua
+++ b/lua/lazyvim/config/keymaps.lua
@@ -189,8 +189,12 @@ map("n", "<leader>L", function() LazyVim.news.changelog() end, { desc = "LazyVim
 -- floating terminal
 map("n", "<leader>fT", function() Snacks.terminal() end, { desc = "Terminal (cwd)" })
 map("n", "<leader>ft", function() Snacks.terminal(nil, { cwd = LazyVim.root() }) end, { desc = "Terminal (Root Dir)" })
-map({"n","t"}, "<c-/>",function() Snacks.terminal(nil, { cwd = LazyVim.root() }) end, { desc = "Terminal (Root Dir)" })
-map({"n","t"}, "<c-_>",function() Snacks.terminal(nil, { cwd = LazyVim.root() }) end, { desc = "which_key_ignore" })
+map("n", "<c-/>",      function() Snacks.terminal(nil, { cwd = LazyVim.root() }) end, { desc = "Terminal (Root Dir)" })
+map("n", "<c-_>",      function() Snacks.terminal(nil, { cwd = LazyVim.root() }) end, { desc = "which_key_ignore" })
+
+-- Terminal Mappings
+map("t", "<C-/>", "<cmd>close<cr>", { desc = "Hide Terminal" })
+map("t", "<c-_>", "<cmd>close<cr>", { desc = "which_key_ignore" })
 
 -- windows
 map("n", "<leader>-", "<C-W>s", { desc = "Split Window Below", remap = true })
--- a/lua/lazyvim/plugins/extras/ai/sidekick.lua
+++ b/lua/lazyvim/plugins/extras/ai/sidekick.lua
@@ -74,6 +74,18 @@ return {
           require("sidekick.nes").enable(state)
         end,
       }):map("<leader>uN")
+      local opts = {
+        cli = {
+          win = {
+            --stylua: ignore
+            keys = {
+              terminal_slash = { "<C-/>", function() Snacks.terminal(nil, { cwd = LazyVim.root() }) end, desc = "Terminal (Root Dir)", mode = "t" },
+              terminal_underscore = { "<C-_>", function() Snacks.terminal(nil, { cwd = LazyVim.root() }) end, desc = "which_key_ignore", mode = "t" },
+            },
+          },
+        },
+      }
+      return opts
     end,
     -- stylua: ignore
     keys = {


dpetka2001:
I also tried the following diff locally
```diff
diff --git a/lua/lazyvim/config/keymaps.lua b/lua/lazyvim/config/keymaps.lua
index 05d07a21..6f4e0575 100644
--- a/lua/lazyvim/config/keymaps.lua
+++ b/lua/lazyvim/config/keymaps.lua
@@ -189,8 +189,12 @@ map("n", "<leader>L", function() LazyVim.news.changelog() end, { desc = "LazyVim
 -- floating terminal
 map("n", "<leader>fT", function() Snacks.terminal() end, { desc = "Terminal (cwd)" })
 map("n", "<leader>ft", function() Snacks.terminal(nil, { cwd = LazyVim.root() }) end, { desc = "Terminal (Root Dir)" })
-map({"n","t"}, "<c-/>",function() Snacks.terminal(nil, { cwd = LazyVim.root() }) end, { desc = "Terminal (Root Dir)" })
-map({"n","t"}, "<c-_>",function() Snacks.terminal(nil, { cwd = LazyVim.root() }) end, { desc = "which_key_ignore" })
+map("n", "<c-/>",      function() Snacks.terminal(nil, { cwd = LazyVim.root() }) end, { desc = "Terminal (Root Dir)" })
+map("n", "<c-_>",      function() Snacks.terminal(nil, { cwd = LazyVim.root() }) end, { desc = "which_key_ignore" })
+
+-- Terminal Mappings
+map("t", "<C-/>", "<cmd>close<cr>", { desc = "Hide Terminal" })
+map("t", "<c-_>", "<cmd>close<cr>", { desc = "which_key_ignore" })

 -- windows
 map("n", "<leader>-", "<C-W>s", { desc = "Split Window Below", remap = true })
diff --git a/lua/lazyvim/plugins/extras/ai/sidekick.lua b/lua/lazyvim/plugins/extras/ai/sidekick.lua
index d9c0c2e2..ef5b03af 100644
--- a/lua/lazyvim/plugins/extras/ai/sidekick.lua
+++ b/lua/lazyvim/plugins/extras/ai/sidekick.lua
@@ -74,6 +74,18 @@ return {
           require("sidekick.nes").enable(state)
         end,
       }):map("<leader>uN")
+      local opts = {
+        cli = {
+          win = {
+            --stylua: ignore
+            keys = {
+              terminal_slash = { "<C-/>", function() Snacks.terminal(nil, { cwd = LazyVim.root() }) end, desc = "Terminal (Root Dir)", mode = "t" },
+              terminal_underscore = { "<C-_>", function() Snacks.terminal(nil, { cwd = LazyVim.root() }) end, desc = "which_key_ignore", mode = "t" },
+            },
+          },
+        },
+      }
+      return opts
     end,
     -- stylua: ignore
     keys = {
diff --git a/lua/lazyvim/plugins/util.lua b/lua/lazyvim/plugins/util.lua
index eede7bcb..8d8aa220 100644
--- a/lua/lazyvim/plugins/util.lua
+++ b/lua/lazyvim/plugins/util.lua
@@ -23,8 +23,6 @@ return {
             nav_j = { "<C-j>", term_nav("j"), desc = "Go to Lower Window", expr = true, mode = "t" },
             nav_k = { "<C-k>", term_nav("k"), desc = "Go to Upper Window", expr = true, mode = "t" },
             nav_l = { "<C-l>", term_nav("l"), desc = "Go to Right Window", expr = true, mode = "t" },
-            hide_slash = { "<C-/>", "hide", desc = "Hide Terminal Slash", mode = "t" },
-            hide_underscore = { "<C-_>", "hide", desc = "which_key_ignore", mode = "t" },
           },
         },
       },
```
and to me seems like the best solution. Would it be acceptable to modify the current state of this PR to the one of the mentioned diff? Let me know what you think.

PS: It's exactly the old behavior for terminal mappings with the addition of terminal mappings to Sidekick cli, so that it doesn't close when pressing `<C-/>`.

qw457812:
For `<c-/>` -> `<esc><esc>` -> `2<c-/>` -> `<esc><esc>` -> `<c-/>`, the last `<c-/>` can close the second term if use something like this:

```lua
map({ "n", "t" }, "<c-/>", function()
  if
    vim.fn.mode() == "n" -- terminal mode handled by `hide_slash` below
    and vim.bo.filetype == "snacks_terminal"
    and vim.v.count == 0 -- instead of toggling the [1](https://github.com/folke/snacks.nvim/blob/8c501965beff9a741b29eea53c7f876b039bddea/lua/snacks/terminal.lua#L182), we want to close current term
  then
    vim.cmd("close")
    return
  end
  Snacks.terminal(nil, {
    cwd = LazyVim.root(),
    win = { keys = { hide_slash = { "<c-/>", "hide", desc = "Hide Terminal", mode = "t" } } }, -- remove this line if it already mapped to all snacks terminal
  })
end, { desc = "Terminal (Root Dir)" })
```

dpetka2001:
Yes, but that's not ideal. I believe the diff patch that I mentioned in my previous comment provides the most intuitive behavior. 

I will wait for Folke's input regarding this.

dpetka2001:
Went ahead and merged the diff patch from my comment earlier. If this is not something you'd like, feel free to disregard.

folke:
This has been resolved differently. Here's a full analysis of all the related issues and how the new approach addresses them:

## The underlying problems

**1. `<C-/>` closes unrelated terminals like Sidekick (#6573)**
The old `Snacks.terminal()` calls `toggle()`, which hides any matching terminal regardless of whether you're looking at it. Pressing `<C-/>` while in Sidekick's terminal would close it.

**2. `hide_slash`/`hide_underscore` win keys prevent `2<C-/>` splits (#6770)**
When `<C-/>` was mapped as a win key to `"hide"` in both `t` and `n` modes, you couldn't `<Esc><Esc>` then `2<C-/>` to open a second terminal in a vertical split — it would just hide the current one.

**3. `<C-/>` in normal mode toggles the wrong terminal (#6773)**
With two terminals open, `<C-/>` in normal mode always toggled the terminal matching the computed root/count ID, not necessarily the one the cursor is in. If you were in terminal A (normal mode) and the ID resolved to terminal B, it would toggle B.

**4. `LazyVim.root()` differs in terminal vs file buffers (#7048)**
A terminal buffer has no LSP attached, so `LazyVim.root()` falls back to cwd instead of the LSP-detected root. Pressing `<C-/>` inside a terminal would compute a different root, creating a new terminal instead of toggling the current one.

## The fix

Two changes across both repos:

**snacks.nvim** ([12128b9](https://github.com/folke/snacks.nvim/commit/12128b9)): New `Snacks.terminal.focus()` method — if the current buffer IS the target terminal, hide it. Otherwise, show and focus the target terminal (never toggle something you can't see).

**LazyVim** ([96f4f18d](https://github.com/LazyVim/LazyVim/commit/96f4f18d)): `<C-/>` now uses `Snacks.terminal.focus()` instead of `Snacks.terminal()`. The `hide_slash`/`hide_underscore` win keys are restored in terminal mode only (`mode = "t"`).

## Behavior in all scenarios

| Scenario | Old behavior | New behavior |
|---|---|---|
| `<C-/>` from file buffer | Toggle root terminal (could hide unseen terminal) | Show/focus root terminal |
| `<C-/>` in terminal mode (in the terminal) | Win key hides it | Win key hides it (same) |
| `<C-/>` in normal mode, cursor in the right terminal | Toggle (hide) | Focus sees `buf == terminal.buf`, hides it |
| `<C-/>` in normal mode, cursor in a different terminal | Toggle the computed terminal (wrong one) | Focus the correct root terminal (jumps to it) |
| `<C-/>` inside terminal with different root (#7048) | Creates new terminal (different ID) | Hides current terminal (win key fires in `t` mode) |
| `<Esc><Esc>` then `2<C-/>` for splits | Blocked by `n` mode win key | Works — win key is `t` mode only, `focus()` creates/shows terminal 2 |
| `<C-/>` in Sidekick (#6573) | Closes Sidekick | Sidekick is not a `Snacks.terminal`, unaffected |

Closing this PR as the new approach supersedes it. Thanks @dpetka2001 for the thorough analysis throughout these issues — it helped clarify all the edge cases.

folke:
The new approach with focus/toggle also make a lot more sense to me.
So if you have a terminal window open and you're in another window, you can press `ctrl+/` to focus it and then `ctrl+/` again to close it.

So same behavior as the changes I did yesterday for sidekick and `ctrl+.`

PR #6823, opened 2025-11-27
fix(leap): "default mappings are deprecated" warning from leap

awilkins:
## Description

The first use of Leap anywhere generates a popup warning that
the default mappings are deprecated.

PR replaces the call to `leap.add_default_mappings` with the
part of the default mappings that wasn't immediately
deleted afterwards.


ggandor:
Related: https://github.com/LazyVim/LazyVim/pull/6756

PR #5737, opened 2025-03-06
fix(util.project): different mapping on dashboard than `snacks.projects`

dpetka2001:
## Description
Different mapping and description to not be confused with `snacks.projects`.

## Related Issue(s)
Fixes #5736 

## Checklist

- [x] I've read the [CONTRIBUTING](https://github.com/LazyVim/LazyVim/blob/main/CONTRIBUTING.md) guidelines.

Issue #5736, opened 2025-03-06
bug: duplicate 'project' item on dashboard when `extra.util.project` is enabled


winter233:
### Did you check docs and existing issues?

- [x] I have read all the LazyVim docs
- [x] I have updated the plugin to the latest version before submitting this issue
- [x] I have searched the existing issues of LazyVim
- [x] I have searched the existing issues of plugins related to this issue

### Neovim version (nvim -v)

NVIM v0.10.4 Build type: Release LuaJIT 2.1.1713484068

### Operating system/version

MacOS 11.7.10

### Describe the bug

As described in the title
<img width="612" alt="Image" src="https://github.com/user-attachments/assets/c7f391f0-e9b6-43ff-872c-bf475530d3b0" />

### Steps To Reproduce

1. enable `lazyvim.plugins.extras.util.project`
2. open vim

### Expected Behavior

should only be one project using `lazyvim.plugins.extras.util.project`

### Repro

```lua
vim.env.LAZY_STDPATH = ".repro"
load(vim.fn.system("curl -s https://raw.githubusercontent.com/folke/lazy.nvim/main/bootstrap.lua"))()

require("lazy.minit").repro({
  spec = {
    { "LazyVim/LazyVim", import = "lazyvim.plugins" },
    { import = "lazyvim.plugins.extras.util.project" },
    -- add any other plugins here
  },
})
```

PR #6832, opened 2025-12-01
fix(util.project): Change project key from 'P' to 'p' in dashboard

GeoDaoyu:
## Description

Change project key from 'P' to 'p' in dashboard.

I agree to make a distinction between snack/project and utils/project, but we should not modify the shortcut keys.

It is almost impossible to use both of them simultaneously.

## Related Issue(s)

https://github.com/LazyVim/LazyVim/pull/5737

## Screenshots

image

## Checklist

### DIFF
--- a/lua/lazyvim/plugins/extras/util/project.lua
+++ b/lua/lazyvim/plugins/extras/util/project.lua
@@ -126,7 +126,7 @@ return {
     "goolord/alpha-nvim",
     optional = true,
     opts = function(_, dashboard)
-      local button = dashboard.button("P", " " .. " Projects (util.project)", pick)
+      local button = dashboard.button("p", " " .. " Projects (util.project)", pick)
       button.opts.hl = "AlphaButtons"
       button.opts.hl_shortcut = "AlphaShortcut"
       table.insert(dashboard.section.buttons.val, 4, button)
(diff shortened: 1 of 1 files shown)


dpetka2001:
#5736

GeoDaoyu:
I don't understand why the default shortcut key needs to be modified.🤔

When util.projects is not enabled, snacks.projects(p) should be used.
When starting util.projects, should disable snacks.projects(p) and use util.projects(p) instead.

dpetka2001:
Snacks.projects is just another picker of Snacks.picker (just like files, grep etc). So, when Snacks.picker is enabled (which is the default), Snacks.projects will always be enabled. Snacks.projects is not a different plugin like `util.project` to disable it separately. It's just part of the Snacks suite.

GeoDaoyu:
So is there no way to restore the shortcut key of `p`? Can it be configured through personal Settings?

dpetka2001:
Yes, you can do it in personal configuration.

For example with FzfLua Extra and util.project enabled, the following should change the key to `p` for Snacks.dashboard
```lua
return {
  {
    "folke/snacks.nvim",
    opts = {
      dashboard = {
        config = function(opts, _)
          opts.preset.keys[3].key = "p"
        end,
      },
    },
  },
}
```
Same should be able to achieve for other dashboards as well. Read the configuration and change the corresponding component.

GeoDaoyu:
@dpetka2001 Thank you very much. It's worked.
https://github.com/GeoDaoyu/dotfiles/commit/bc2f0c980f339b20c4b2daea1c4abd8b0f14c56a
and I will close this pr.

PR #6903, opened 2025-12-27
fix(lsp): sort `opts.servers` before setting up keymaps

dpetka2001:
## Description
I noticed with Typescript Extra that some times `gD` was mapped to `vim.lsp.buf.declaration` and other times to `Go to Source Definition`. 

There was a race condition because `pairs(opts.servers)` doesn't guarantee the order in which servers will get evaluated for setting up keymaps. Some times `"*"` would get evaluated before `vtsls` and `Go to Source Definition` would take precedence and other times `"*"` would get evaluated after `vtsls` and thus resulting in `Goto Declaration` to take precedence. 

Solution: Put the keys of `opts.servers` into a list, sort the list and evaluate the servers based on the sorted list to setup keymaps. This way `"*"` server will always get evaluated first and other servers will be able to correctly overwrite keymaps that already exist.

## Related Issue(s)
None

## Checklist

### DIFF
--- a/lua/lazyvim/plugins/lsp/init.lua
+++ b/lua/lazyvim/plugins/lsp/init.lua
@@ -155,7 +155,14 @@ return {
       LazyVim.format.register(LazyVim.lsp.formatter())
 
       -- setup keymaps
-      for server, server_opts in pairs(opts.servers) do
+      local keys = {}
+      for server in pairs(opts.servers) do
+        table.insert(keys, server)
+      end
+      table.sort(keys)
+
+      for _, server in ipairs(keys) do
+        local server_opts = opts.servers[server]
         if type(server_opts) == "table" and server_opts.keys then
           require("lazyvim.plugins.lsp.keymaps").set({ name = server ~= "*" and server or nil }, server_opts.keys)
         end


dpetka2001:
I don't think that's correct. It does indeed fix the issue for Typescript, but omnisharp `gd` mapping still gets overwritten by snacks-picker `gd`. There must be something wrong with `Snacks.keymap.set` or `Snacks.lsp.on` I believe.

PS: Actually it seems it doesn't even for Typescript. Seems I just got lucky among a few restarts.

dpetka2001:
The new way of defining LSP keymaps does not guarantee a deterministic order. I don't know how this could be handled in `Snacks.keymap.set` or `Snacks.lsp.on`.

However, I managed to make it work as expected by mostly mimicking the previous behavior with following patch in addition to the change that this PR includes
```diff
diff --git a/lua/lazyvim/plugins/lsp/init.lua b/lua/lazyvim/plugins/lsp/init.lua
index b7272283..44465b63 100644
--- a/lua/lazyvim/plugins/lsp/init.lua
+++ b/lua/lazyvim/plugins/lsp/init.lua
@@ -164,7 +164,21 @@ return {
       for _, server in ipairs(keys) do
         local server_opts = opts.servers[server]
         if type(server_opts) == "table" and server_opts.keys then
-          require("lazyvim.plugins.lsp.keymaps").set({ name = server ~= "*" and server or nil }, server_opts.keys)
+          server_opts.on_attach = function(_, bufnr)
+            local map = function(lhs, rhs, desc, mode)
+              mode = mode or "n"
+              vim.keymap.set(mode, lhs, rhs, { buffer = bufnr, desc = desc, silent = true })
+            end
+            local spec = vim.tbl_extend("force", {}, opts.servers["*"].keys)
+            local clients = vim.lsp.get_clients({ bufnr = bufnr })
+            for _, client in ipairs(clients) do
+              local maps = opts.servers[client.name].keys or {}
+              vim.list_extend(spec, maps)
+            end
+            for _, key in ipairs(spec) do
+              map(key[1], key[2], key.desc, key.mode)
+            end
+          end
         end
       end
```
This handles both Typescript and Omnisharp cases correctly. Of course further changes would be required to accommodate for cases of `has`, `enabled` etc. This is just a proof of concept that when defining LSP keymaps in `on_attach` works and that there should be some kind of change in Snacks aforementioned functions to handle the mappings with some sort of order.

dpetka2001:
Ok, I tried to delve a bit more into `Snacks.util.lsp` code. I did a lot of printing to grasp in what order the handlers were being evaluated. 

There was no deterministic order. So, this is what I've tried
- Use a separate key for generic handlers with empty filter `{}` and only evaluate them once per buffer by keeping track of the buffer only instead of `client_id:buf` for the rest of the handlers that do have a filter (I saw that from commit history where you added the code for `Snacks.util.lsp.on()` should trigger for each client per buffer, but thought that for the generic LSP `"*"` which has an empty `filter.name` that should not be the case and should be executed only once per buffer).
- Sort the `handlers` table, so that it first evaluates generic handlers and also evaluates handlers that have `filter = { name = "something" }` last

The following patch seems to be working
```diff
diff --git a/lua/snacks/util/lsp.lua b/lua/snacks/util/lsp.lua
index ddac576c..d14796e6 100644
--- a/lua/snacks/util/lsp.lua
+++ b/lua/snacks/util/lsp.lua
@@ -29,13 +29,33 @@ local function _handle(filter)
     return
   end
 
+  table.sort(handlers, function(a, b)
+    local a_empty = next(a.filter) == nil
+    local b_empty = next(b.filter) == nil
+    local a_has_name = a.filter.name ~= nil
+    local b_has_name = b.filter.name ~= nil
+
+    -- stylua: ignore start
+    -- Generic filters first
+    if a_empty and not b_empty then return true end
+    if b_empty and not a_empty then return false end
+
+    -- Name-based filters last (so they override method-based filters)
+    if a_has_name and not b_has_name then return false end
+    if b_has_name and not a_has_name then return true end
+    -- stylua: ignore end
+
+    return false
+  end)
+
   for _, state in ipairs(handlers) do
     local f = vim.deepcopy(state.filter)
     f = vim.tbl_extend("force", f, filter)
     local clients = vim.lsp.get_clients(f)
     for _, client in ipairs(clients) do
       for buf in pairs(client.attached_buffers) do
-        local key = ("%d:%d"):format(client.id, buf)
+        local is_generic = next(state.filter) == nil
+        local key = is_generic and ("generic:%d"):format(buf) or ("%d:%d"):format(client.id, buf)
         if not state.done[key] then
           state.done[key] = true
           local ok, err = pcall(state.cb, buf, client)
@@ -78,8 +98,9 @@ local function setup()
   vim.api.nvim_create_autocmd("LspDetach", {
     group = group,
     callback = function(ev)
-      local key = ("%d:%d"):format(ev.data.client_id, ev.buf)
       for _, state in ipairs(_handlers) do
+        local is_generic = next(state.filter) == nil
+        local key = is_generic and ("generic:%d"):format(ev.buf) or ("%d:%d"):format(ev.data.client_id, ev.buf)
         state.done[key] = nil
       end
     end,
```
This seems to solve the Typescript problem. For the dotnet Extra we have to add `has = "definition"` because again from printing I observed that even with the above patch diff the key was at first correctly set by handler with `filter = { name = "omnisharp" }` but was later on overwritten by handler with `filter = { method = textDocument/definition }` which I assume is being evaluated by `vim.lsp.handlers["client/registerCapability"]` for dynamic capability registration and that happens after `LspAttach`??

I'm not at all confident if the above diff patch does not have any other side effects, so I'm refraining from creating a PR at Snacks repo. Just wanted to share my observations and hopefully fill you in with as much detail as possible.

dpetka2001:
I also have a question about the function `handle`. 

In the `LspAttach` event we call `_handle({ id = ev.data.client_id, buffer = ev.buf })`. When we call this how does the actual filtering from `vim.tbl_filter(..., _handlers)` work? 

It seems to me that all handlers will pass because the table `_handlers` that gets generated from `M.on()` function only creates handlers with either have an empty filter or a `{ name = "some" }` filter or a `{ method = "bla" }` filter or a combination filter of the 2 latter. 

So, when we call `_handle({ id = ev.data.client_id, buffer = ev.buf })` it will do a comparison for keys `id` and `buffer` which will be `nil` for every handler in the `_handlers` table, thus `return true` will be the outcome for every handler from `_handlers` table. You only merge the 2 filters later on with `vim.tbl_extend` which will get passed onto `vim.lsp.get_clients`. Only the call of `handle(filter)` inside `M.on()` function actually does filtering of the `_handlers` table based on the `filter` that is passed to it.

PS: I would really appreciate it if you could shed some more info with regards to the workflow. Obviously, my knowledge is limited when concerning LSP stuff, so I would really like to understand better. This is the first time I've looked into `Snacks.util.lsp` code since you implemented it.

dpetka2001:
This PR is wrong btw, just leaving it open for you to look into the underlying issue.

folke:
This has been properly fixed in two places:

- **snacks.nvim** ([809f979](https://github.com/folke/snacks.nvim/commit/809f979)): LSP keymaps now use a debounced full re-evaluation system that respects registration order. When multiple keymaps share the same `mode:lhs`, the last registered one wins deterministically.
- **LazyVim** ([6d0da34](https://github.com/LazyVim/LazyVim/commit/6d0da34d)): Server names are sorted before setting keymaps, ensuring `*` (wildcard) keymaps are always registered before named servers.

Thanks for the investigation @dpetka2001, your analysis of the non-deterministic ordering was spot on.

PR #6904, opened 2025-12-27
fix(util.lsp): `LazyVim.lsp.execute` only request single client

### LINKED ISSUE #6900: bug: missing `lsp.execute()` error handling

### Neovim version (nvim -v)

NVIM v0.12.0-dev-1879+gb95848a8f1

### Operating system/version

Windows 11

### Describe the bug

the few usages of `LazyVim.lsp.execute()` (e.g. in `extras.lang.typst`'s keymap to "pin main file") broadcasts an lsp-specific command to all attached lsps. This causes the irrelevant lsps to throw an error which is not handled and ends up as an ERROR through `vim.notify`.

In my case with typst (Tinymist lsp), I also have copilot attached, so triggering the "Pin main file" keybind causes an error notification:

```
copilot: -32603: Request workspace/executeCommand failed with message: Unknown command: tinymist.pinMain
```

the other uses of `lsp.execute()` are in `typescript` and `elixir` language estras... while I don't use them, they look like they'd have the same issue.

### Steps To Reproduce

Case with Tinymist and Copilot lsp:

1. enable both extras
2. open `.typ` file
3. press `<leader>cP` to trigger "Pin main"
4. profit

### Expected Behavior

no extraneous errors from copilot

dpetka2001:
## Description
`LazyVim.lsp.execute` uses `vim.lsp.buf_request` which sends the request to all attached clients. 

`LazyVim.lsp.execute` seems to me to be the equivalent of [vim.lsp.buf.execute_command](https://github.com/neovim/neovim/blob/922816877febf397fe854f01d8013a510d73f1d2/runtime/lua/vim/lsp/buf.lua#L1366-L1376), but adds the option to also open in Trouble. Otherwise the request is being made the same way. This command will be deprecated in 0.12 and it's advised to use `client:exec_cmd` instead.

Since LazyVim supports Neovim >=0.11.2, it makes sense to refactor the corresponding codeblock and use `client:cmd_exec` instead (it's already available in Neovim 0.11).

I only changed the necessary parts where `vim.lsp.buf_request` was being used and not Trouble (when `opts.open = true`), since I'm not familiar with its codebase. That's also why I went for extending `params` in the `else` branch instead of adding `title` into original `params`, since I didn't know how Trouble will react to this change. Not sure if maybe there also needs to be some change there.

## Related Issue(s)
Fixes #6900

## Checklist

### DIFF
--- a/lua/lazyvim/plugins/extras/lang/elixir.lua
+++ b/lua/lazyvim/plugins/extras/lang/elixir.lua
@@ -16,6 +16,8 @@ return {
               function()
                 local params = vim.lsp.util.make_position_params()
                 LazyVim.lsp.execute({
+                  title = "toPipe",
+                  filter = "elixirls",
                   command = "manipulatePipes:serverid",
                   arguments = { "toPipe", params.textDocument.uri, params.position.line, params.position.character },
                 })
@@ -27,6 +29,8 @@ return {
               function()
                 local params = vim.lsp.util.make_position_params()
                 LazyVim.lsp.execute({
+                  title = "fromPipe",
+                  filter = "elixirls",
                   command = "manipulatePipes:serverid",
                   arguments = { "fromPipe", params.textDocument.uri, params.position.line, params.position.character },
                 })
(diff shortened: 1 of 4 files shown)


folke:
Thanks!

PR #6923, opened 2026-01-08
fix(alpha): pick config files

madjxatw:
## Description

The `alpha` dashboard fails to open a picker for config files selection.

## Related Issue(s)

None

## Screenshots

None

## Checklist


folke:
Closing this — changing `"` to `[[]]` is a no-op here. There are no special characters in the string that would require double-bracket quoting, and other buttons on the same dashboard (Lazy Extras, Lazy, Quit) use regular quotes without any issues.

If you're experiencing a bug with the Config picker, please open an issue with reproduction steps so we can investigate the actual root cause.

Issue #5854, opened 2025-03-28
bug: Potential incorrect import of lazyvim extras in examples


michaelbarton:
### Did you check docs and existing issues?

- [x] I have read all the LazyVim docs
- [x] I have updated the plugin to the latest version before submitting this issue
- [x] I have searched the existing issues of LazyVim
- [x] I have searched the existing issues of plugins related to this issue

### Neovim version (nvim -v)

v0.10.4

### Operating system/version

OSX 15.3

### Describe the bug

I believe the documentation may be incorrect on the [examples page.](https://www.lazyvim.org/configuration/examples) Specifically the line:

```lua
  { import = "lazyvim.plugins.extras.lang.typescript" },
```

I am new to lazyvim, and I was following the examples provided and I tried importing the extras as outlined on this page. However I got a dialog box with a warning about the imports being in the wrong order and that the lazyvim extras should be imported before other plugins. I am relatively inexperienced with nvim, so my guess at the cause of the issue is:

- files in `plugins/*.lua` get loaded after the all the lazyvim plugins
- these files contain an import of the typescript extra lang plugin, as suggested by the documentation, which may or may not come after 3rd party plugins
- this raises the warning about import order
- moving the import statement instead to [before this line](https://github.com/LazyVim/starter/blob/main/lua/config/lazy.lua#L22) fixes the issue. Because this comes before the import of any 3rd party libraries.

I've tried searching through the issues, but couldn't find anyone else reporting something similar

### Steps To Reproduce

1. Create a lua file in `plugins` using the starter config cloned from github
2. Add an import any 3rd party plugin
3. Following the documentation then import a lazyvim extra plugin after a 3rd party plugin

### Expected Behavior

I think the expected behaviour is the docs suggest adding the lazy vim extra imports to before [L22 of lazy.lua](https://github.com/LazyVim/starter/blob/main/lua/config/lazy.lua#L22) instead. But I am just making a guess here since I am not familiar with writing nvim plugins.

### Repro

```lua
vim.env.LAZY_STDPATH = ".repro"
load(vim.fn.system("curl -s https://raw.githubusercontent.com/folke/lazy.nvim/main/bootstrap.lua"))()

require("lazy.minit").repro({
  spec = {
    { "LazyVim/LazyVim", import = "lazyvim.plugins" },
    -- add any other plugins here
  },
})
```

dpetka2001:
[...]

PR #6928, opened 2026-01-14
feat: add a hook for user to enable extras in lua config with correct order

suimong:
## Description

This PR adds a `vim.g.lazyvim_user_extras` hook to address the pain point that user currently does not have a way to enable extras in lua config that respects LazyVim's plugin/extras loading order and predefined priorities. It is a table whose elements can be either an extras name (i.e. `lazyvim.plugins.extras.*`, using default priority of 50), or a kv pair of `["lazyvim.plugins.extras.<name>"] = <custom priority>`. This table is **merged and sorted together** with extras enabled in `lazyvim.json` and default core extras. It also detects and disallows user from overriding priority of default core extras and extras enabled via `lazyvim.json`. 

## Motivation

Currently, the user has two ways of enabling extras:
1. updating `lazyvim.json`: this can be done by either updating the file directly or via `:LazyExtras`
2. add `{ import = "lazyvim.plugins.extras.<name>" }` to a plugin spec, which usually stays under `lua/plugins` directory.

The two approaches serve different use cases and user preferences. However, the second approach has the downside that LazyVim would warn about plugins not loaded in the correct order: all extras should be loaded before user plugins, which is not possible if user "import" it in `lua/plugins`. The user could modify `lua/config/lazy.lua` to something like:
```
require("lazy").setup({
  spec = {
    -- import user-enabled extras
    {import = "extras"},
    -- import/override with your plugins
    { import = "plugins" },
  },
  ...
}
```

However, this means that all extras are loaded after those enabled via  `lazyvim.json`. 
With this PR, user now has the ability to precisely control the order in which extras are enabled, whilst keeping the declarations in lua land.

## Related Issue(s)

Addresses https://github.com/LazyVim/LazyVim/issues/5854 in an alternative way.

## Checklist

### DIFF
--- a/lua/lazyvim/plugins/xtras.lua
+++ b/lua/lazyvim/plugins/xtras.lua
@@ -59,9 +59,9 @@ for name, extra in pairs(defaults) do
   end
 end
 
--- Add user defined extras via `vim.g.lazyvim_user_extras`
-local user_defined_extras = vim.g.lazyvim_user_extras or {}
-for k, v in pairs(user_defined_extras) do
+-- Add user enabled extras via `vim.g.lazyvim_user_extras`
+local user_enabled_extras = vim.g.lazyvim_user_extras or {}
+for k, v in pairs(user_enabled_extras) do
   local is_atomic_value = type(k) == "number"
   local name = is_atomic_value and v or k
   local priority = is_atomic_value and default_prio or v


dpetka2001:
Doesn't this existing code https://github.com/LazyVim/LazyVim/blob/c64a61734fc9d45470a72603395c02137802bc6f/lua/lazyvim/plugins/xtras.lua#L21-L23 already serve the same purpose?

suimong:
> already serve the same purpose?

So at first I thought `vim.g.xtra_prios` would solve my problem. Then I realized it is different: `xtra_prios` allows the user to explicitly specify priorities for some extras, but those extras are still gonna be enabled via `lazyvim.json`. looking at the implementation in `xtras.lua`, `vim.g.xtras_prios` is merged with `prios` table. `prios` table only stores priority information keyed by xtras name. the actual extras to be enabled is stored in `extras` table, which is populated by reading `lazyvim.json` plus defaults (`LazyVim.config.get_defaults()`). Then the `table.sort` functions sorts `extras` table with information in the `prios` table. So whatever is in `xtra_prios` won't show up automatically in `extras`.

My PR would allow enabling extras and (optionally) specifying priorities at the same place, and in lua file.

dpetka2001:
Yes, that is deliberate because maintainer wanted users to use LazyExtras instead of relying on manual imports which caused problems in the past. So, users should add their Extras to `/lua/plugins/extras/` directory and then enable them via LazyExtras and configure their priority via `vim.g.xtras_prios` if they wanted. 

Manual `import` should be avoided unless the users know what they're doing. You can also disable the check warning message by setting `vim.g.lazyvim_check_order = false` in your `options.lua` file.

suimong:
Yes, manual import causes problems precisely because they have a de facto priority lower than everything enabled in LazyExtras, which is a false assumption. I have actually tripped by this because i tried manually importing extras.lang.typescript, which has a high priority of 5 defined in xtras.lua.

That's why this PR does not rely on lazy.nvim's import directly. User are free to set vim.g.lazyvim_user_extras however they want, as long as it is in the specified shape and set early enough for xtras.lua to consume it. Thus it is a safe mechanism because a) it participates in the same priority sorting process as LazyExtras, b) user are not allowed to override priority values defined internally by LazyVim or by those set via vim.g.xtra_prios, and c) this PR is backward compatible. 

Admittedly this is a niche feature, but I feel like it complements LazyExtras for users who want more flexibility. In particular, this enables a usage pattern that allows users to enable extras written in multiple files (collected dynamically into vim.g.lazyvim_user_extras), whereas with LazyExtras everything must be written in one place, the lazyvim.json file.

dpetka2001:
> In particular, this enables a usage pattern that allows users to enable extras written in multiple files (collected dynamically into vim.g.lazyvim_user_extras), whereas with LazyExtras everything must be written in one place, the lazyvim.json file.

That is the reason why LazyExtras were created, whereas previously users would just define their own `import`s. It caused problems and maintainer wanted to discourage users from opening unnecessary issues because they didn't know what they were doing.

So, the consensus was to use LazyExtras instead of imports for that reason. And users who wanted to use imports could do so in their personal configuration but they would be responsible for taking care of the correct order.

In my personal opinion, this has the potential to cause unnecessary confusion to users, thus I would suggest against incorporating it into LazyVim. But maintainer has the final say and will make final decision.

suimong:
Thank you for the discussion. I should probably add some of the arguments in our discussion to the PR description.

PR #7004, opened 2026-02-13
fix(keymaps): terminal-toggle <C-/> should hide instead of re-evaluat…

0xKev:
## Description

The terminal-mode `<C-/>` mapping currently calls `Snacks.terminal(nil, { cwd = LazyVim.root() })`, which re-evaluates the project root from within the terminal buffer. Since terminal buffers have no LSP and no file path, `LazyVim.root()` falls back to `cwd`, which can resolve to a different directory than the code buffer that originally opened the terminal.

This causes Snacks to compute a different terminal ID (since the ID is based on `cwd`), spawning a second terminal instead of toggling the first one closed.

The fix splits the `{"n","t"}` mapping so that:
- Normal mode still calls `Snacks.terminal(nil, { cwd = LazyVim.root() })` to open/toggle
- Terminal mode uses `<cmd>hide<cr>` to simply hide the window

This is consistent with Snacks' own built-in `q = "hide"` key for terminal windows (https://github.com/folke/snacks.nvim/blob/main/docs/terminal.md)

## Steps to reproduce

1. Open a project with nested directories (e.g. `~/projects/myapp/src/module`)
2. Press `<C-/>` from a code buffer, terminal opens correctly
3. Press `<C-/>` from inside the terminal, a second terminal spawns alongside the first

## Screenshots

Two terminal buffers spawned side by side after pressing `<C-/>` twice. 
Note the different working directories and PIDs:
terminal-spawn-ids

## Checklist

### DIFF
--- a/lua/lazyvim/config/keymaps.lua
+++ b/lua/lazyvim/config/keymaps.lua
@@ -189,8 +189,10 @@ map("n", "<leader>L", function() LazyVim.news.changelog() end, { desc = "LazyVim
 -- floating terminal
 map("n", "<leader>fT", function() Snacks.terminal() end, { desc = "Terminal (cwd)" })
 map("n", "<leader>ft", function() Snacks.terminal(nil, { cwd = LazyVim.root() }) end, { desc = "Terminal (Root Dir)" })
-map({"n","t"}, "<c-/>",function() Snacks.terminal(nil, { cwd = LazyVim.root() }) end, { desc = "Terminal (Root Dir)" })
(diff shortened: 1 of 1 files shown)


dpetka2001:
This will be fixed by #6774. There's also a need to address `sidekick.nvim` which with the behavior in this PR would just cause it to hide as well (but that is probably not desired as stated in #6573).

80avin:
This is already fixed in Lazyvim 15.15.0 ( https://github.com/LazyVim/LazyVim/commit/96f4f18d )

PR #7011, opened 2026-02-16
fix(vtsls): remove redundant deno logic, upstreamed in `nvim-lspconfig`

dpetka2001:
## Description
Since https://github.com/neovim/nvim-lspconfig/pull/4304 `nvim-lspconfig` has inherent deno logic to disambiguate between deno/non-deno projects. 

I believe it'd be better to remove the logic from LazyVim, since it might lead to issues because of different behavior compared to `nvim-lspconfig` defaults (whose logic doesn't only rely on the existence of `deno/json`, but also looks for lock files and takes into account which is the largest path with regards to the found pattern files). It'd be better for users to change `root_dir` themselves if they don't agree with the inherent logic.

## Related Issue(s)
None

## Checklist

### DIFF
--- a/lua/lazyvim/plugins/extras/lang/typescript.lua
+++ b/lua/lazyvim/plugins/extras/lang/typescript.lua
@@ -132,28 +132,6 @@ return {
           return true
         end,
         vtsls = function(_, opts)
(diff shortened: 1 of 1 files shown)


folke:
ty!

PR #7020, opened 2026-02-21
fix(news): use $VIMRUNTIME for news.txt path

### LINKED ISSUE #7019: bug: news resolves to a plugin's `doc/news.txt` instead of Neovim's

### Neovim version (nvim -v)

NVIM v0.11.6

### Operating system/version

macOS 15.7.3

### Describe the bug

When `news.neovim` is enabled, LazyVim uses `nvim_get_runtime_file("doc/news.txt", false)` to locate Neovim's changelog. Because this returns only the **first** match on the runtime path, any plugin that ships its own `doc/news.txt` (e.g., [`rainbow-delimiters.nvim`](https://github.com/HiPhish/rainbow-delimiters.nvim)) can shadow Neovim's file.

**Root cause:** [news.lua#L54](https://github.com/samyakbardiya/LazyVim/blob/main/lua/lazyvim/util/news.lua#L54) — the call passes `false` (first match only) and does not verify the result actually belongs to the Neovim runtime.

I'm happy to open a PR to fix this if it's unintentional.

Related: https://github.com/HiPhish/rainbow-delimiters.nvim/issues/183

### Steps To Reproduce

1. Open Neovim with the repro config: `nvim -u repro.lua`
2. Run:
   `:lua print(vim.inspect(vim.api.nvim_get_runtime_file("doc/news.txt", false)[1]))`
3. Observe the returned path points to a plugin, not Neovim:
   `/private/tmp/LazyVim/.repro/data/nvim/lazy/rainbow-delimiters.nvim/doc/news.txt`

### Expected Behavior

The resolved path should be Neovim's own news file:
`/opt/homebrew/Cellar/neovim/0.11.6/share/nvim/runtime/doc/news.txt`

For reference, fetching _all_ matches shows both files:
```lua
print(vim.inspect(vim.api.nvim_get_runtime_file("doc/news.txt", true)))
{
  "/private/tmp/LazyVim/.repro/data/nvim/lazy/rainbow-delimiters.nvim/doc/news.txt",
  "/opt/homebrew/Cellar/neovim/0.11.6/share/nvim/runtime/doc/news.txt"
}
```

### Repro

```lua

require("lazy.minit").repro({
  spec = {
    {
      "LazyVim/LazyVim",
      import = "lazyvim.plugins",
      opts = {
        news = {
          neovim = true,
        },
      },
    },
    { "HiPhish/rainbow-delimiters.nvim" },
  },
})
```

samyakbardiya:
## Description

Use `vim.env.VIMRUNTIME` to resolve `news.txt` instead of `nvim_get_runtime_file()`, which can return a plugin's `news.txt` (e.g., rainbow-delimiters.nvim) instead of Neovim's bundled one.

I have tested this in macOS & Arch Linux, seems to be working in both.  
Let me know if this approach is not good, I have gone through `:h $VIMRUNTIME`, and it should work according to the doc. 

## Related Issue(s)

- Fixes #7019 

## Checklist

### DIFF
--- a/lua/lazyvim/util/news.lua
+++ b/lua/lazyvim/util/news.lua
@@ -51,7 +51,7 @@ function M.open(file, opts)
     end
     file = plugin.dir .. "/" .. file
   elseif opts.rtp then
-    file = vim.api.nvim_get_runtime_file(file, false)[1]
+    file = vim.fs.joinpath(vim.env.VIMRUNTIME, "doc", "news.txt")
   end
 
   if not file then


folke:
Thanks for the report and the PR! I've fixed this in a slightly different way in 60fea623 — instead of hardcoding `doc/news.txt`, the fix first checks `$VIMRUNTIME` for the file and falls back to `nvim_get_runtime_file()`. This keeps the function generic (it receives `file` as a parameter and also handles other rtp files) while still prioritizing Neovim's bundled file over plugin shadows.

samyakbardiya:
Hello @folke, your approach looks much better!

Thanks for the fix :)

Issue #4695, opened 2024-11-03
bug: Adding prepend_args to markdowncli2 causes markdownlint-cli2 to not load


shauryagoel:
### Did you check docs and existing issues?

- [x] I have read all the LazyVim docs
- [x] I have updated the plugin to the latest version before submitting this issue
- [x] I have searched the existing issues of LazyVim
- [x] I have searched the existing issues of plugins related to this issue

### Neovim version (nvim -v)

v0.10.2

### Operating system/version

MacOS 11.6.1

### Describe the bug

I want to disable long line length warning in markdown files. To solve this, I updated this config-
```
{
    "mfussenegger/nvim-lint",
    optional = true,
    opts = {
      linters_by_ft = {
        markdown = { "markdownlint-cli2" },
      },
      linters = {
        ["markdownlint-cli2"] = {
          prepend_args = {
            -- Disable long line length warning
            "--config",
            "~/.config/nvim/.markdownlint-cli2.yaml",
          },
        },
      },
    },
  }
```

.markdownlint-cli2.yaml-
```
---
config:
  line-length: false  # Disable MD013 lint for line length

```
However, when I open a markdown file, no other diagnostics are being shown. On using `:ConformInfo`, I see `markdownlint-cli2 unavailable: Condition failed`. However, if I use markdownlint-cli2 in the terminal and pass in the yaml config file, everything works as expected. 

### Steps To Reproduce

1. Use the above neovim config and markdownlint-cli2 config yaml in `~/.config/nvim/.markdownlint-cli2.yaml`. 
2. Start a markdown file and type `:ConformInfo`. 

### Expected Behavior

Other diagnostics besides the line length are displayed in neovim markdown buffer. 
Also, `:ConformInfo` shows that `markdownlint-cli2` is found.

dpetka2001:
I have the following and it seems to be working correctly
```lua
  {
    "mfussenegger/nvim-lint",
    opts = {
      linters = {
        ["markdownlint-cli2"] = {
          prepend_args = { "--config", "/home/jrn23/.config/LazyVimDev/.markdownlint-cli2.yaml", "--" },
        },
      },
    },
  },
```

shauryagoel:
The config by @dpetka2001 works as expected. I think there might be some issue with my config. Sorry for creating noise.

PR #7038, opened 2026-03-02
fix(markdown): enable config file discovery for markdownlint-cli2

carlesandres:
## Summary

This PR fixes markdownlint-cli2 config file discovery in nvim-lint by disabling stdin mode.

## Problem

PR #3843 switched LazyVim from `markdownlint-cli` to `markdownlint-cli2`, citing "more flexible configuration through files in the whole directory tree" as a key benefit. However, nvim-lint's default `markdownlint-cli2` linter uses `stdin = true` with `args = { "-" }`, which pipes content via stdin.

**The issue**: When using stdin, markdownlint-cli2 doesn't know the file path and cannot traverse directories to find `.markdownlint.json`, `.markdownlint-cli2.jsonc`, or other config files. This defeats the primary reason for choosing markdownlint-cli2.

**Symptoms**:
- Local `.markdownlint.json` configs are ignored in Neovim
- CLI works fine: `markdownlint-cli2 file.md` respects configs
- Multiple users have reported this: #4695, discussions #4094, #2268, #2367

## Solution

Override nvim-lint's defaults to:
1. Set `stdin = false` so markdownlint-cli2 receives the actual file path
2. Set `args = {}` to remove the `-` (stdin indicator)
3. Update the error format parser to handle filename output instead of "stdin:"

## Testing

Verified locally that `.markdownlint.json` configs are now respected when linting markdown files in Neovim.

## Related

- Closes workaround from #4695
- Implements the config discovery promised by #3843

### DIFF
--- a/lua/lazyvim/plugins/extras/lang/markdown.lua
+++ b/lua/lazyvim/plugins/extras/lang/markdown.lua
@@ -56,11 +56,23 @@ return {
   {
     "mfussenegger/nvim-lint",
     optional = true,
-    opts = {
-      linters_by_ft = {
-        markdown = { "markdownlint-cli2" },
-      },
-    },
+    opts = function(_, opts)
+      opts.linters_by_ft = opts.linters_by_ft or {}
+      opts.linters_by_ft.markdown = { "markdownlint-cli2" }
+
+      opts.linters = opts.linters or {}
+      opts.linters["markdownlint-cli2"] = {
+        -- Disable stdin to enable config file discovery.
+        -- markdownlint-cli2 traverses directories to find .markdownlint.json,
+        -- .markdownlint-cli2.jsonc, etc., but this only works with file paths.
+        stdin = false,
+        args = {},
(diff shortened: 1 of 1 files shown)


dpetka2001:
I would advise against changing default settings by `nvim-lint` linters, as that may introduce unnecessary conflict with future changes and implies that the configuration introduced by LazyVim needs to be monitored against `nvim-lint` default configuration at any time in the future.

Simply using 
```lua
 {
    "mfussenegger/nvim-lint",
    opts = {
      linters = {
        ["markdownlint-cli2"] = {
          prepend_args = { "--config", os.getenv("HOME") .. "/.config/nvim/.markdownlint.yaml", "--" },
        },
      },
    },
  },
```
works as expected from my limited experience with markdown files (maybe I'm missing something else, but as I said I don't work much with markdown except in some local forked repos). If users would like to change the default configuration for any linter, they should do so in their personal configuration.

Just a simple user here and not making any decisions. Final say is entirely up to the maintainer.

carlesandres:
I think you are right @dpetka2001. Thanks for the config suggestion.

PR #7105, opened 2026-04-07
fix(nvim-lint): fix `prepend_args` logic

### LINKED ISSUE #7104: bug: Fix linting `prepend_args` logic

### Neovim version (nvim -v)

v0.11.6 Release

### Operating system/version

MacOS 26.4

## Description

The `prepend_args` logic was first added in this PR: https://github.com/LazyVim/LazyVim/issues/1887 to allow for similar argument customisation as [conform.nvim](https://github.com/stevearc/conform.nvim). The original logic has a flaw where it *appends* the extra args, rather than actually *prepending* them:

```lua
vim.list_extend(lint.linters[name].args, linter.prepend_args)
```

Thus this causes issues with linters that have `-` as the last default argument:

Ex: `selene` args

* Before: `"--display-style", "json", "-"`
* After: `"--display-style", "json", "-", "--config", "/some/path/selene.toml"`

> The linter silently crashes, and doesn't report anything

One could circumvent the issue by completely overriding the arguments, but it defeats the purpose of this feature.

> [!NOTE]
> This issue might be related: https://github.com/LazyVim/LazyVim/issues/4695

## Proposed Solution

To allow this feature to work as one would expect, a simple amendment needs to be made:

```lua
local function list_prepend(dst, src)
  for i = 1, #src do
    table.insert(dst, i, src[i])
  end
end
...
list_prepend(lint.linters[name].args, linter.prepend_args)
```

### Steps To Reproduce

1. Configure a linter (e.g. `selene`/`luacheck`/etc.)
2. Customise `prepend_args` in some way
3. Inspect the state of the linter configuration after start-up

### Expected Behavior

The extra args should be *prepended*, rather than *appended*, since it's called `prepend_args`.

### Repro

```lua

require("lazy.minit").repro({
  spec = {
    -- add any other plugins here
  },
})
```

ivanisakov99:
## Description

`prepend_args` logic doesn't work as expected, as it appends the extra args rather than prepends them. This works for some linter configurations, but breaks other ones that use `-` as the last argument in the default configuration.

## Related Issue(s)

- Fixes #7104

## Screenshots

N/A

## Testing

Manual testing done with some impacted linters (`selene`, `luacheck`):

* Args before:
  * selene: `{ "--display-style", "json", "-" }`
  * luacheck: `{ "--formatter", "plain", "--codes", "--ranges", "-" }`
* Args after:
  * selene: `{ "--config", "/some/path/selene.toml", "--display-style", "json", "-" }`
  * luacheck: `{ "--default-config", "/some/path/.luacheckrc", "--formatter", "plain", "--codes", "--ranges", "-" }`

## Checklist

### DIFF
--- a/lua/lazyvim/plugins/linting.lua
+++ b/lua/lazyvim/plugins/linting.lua
@@ -32,13 +32,19 @@ return {
     config = function(_, opts)
       local M = {}
 
+      local function list_prepend(dst, src)
+        for i = 1, #src do
+          table.insert(dst, i, src[i])
+        end
+      end
+
       local lint = require("lint")
       for name, linter in pairs(opts.linters) do
         if type(linter) == "table" and type(lint.linters[name]) == "table" then
           lint.linters[name] = vim.tbl_deep_extend("force", lint.linters[name], linter)
           if type(linter.prepend_args) == "table" then
             lint.linters[name].args = lint.linters[name].args or {}
-            vim.list_extend(lint.linters[name].args, linter.prepend_args)
+            list_prepend(lint.linters[name].args, linter.prepend_args)
           end
         else
           lint.linters[name] = linter


dpetka2001:
Why not just 
```lua
vim.list_extend(linter.prepend_args, lint.linters[name].args)
lint.linters[name].args = linter.prepend_args
```
Just a suggestion, as I'm just a simple user. Feel free to keep whatever you feel like. Maintainer will do the actual review.

ivanisakov99:
@dpetka2001

Yep that's a valid solution too, but there are some other issues that might arise with it, since now the `prepend_args` table is being mutated and the original args table being overwritten.

* TL;DR I've seen some complex configs that, for instance, define a local function within the linter table and reference it in the default args list, so copying the just the args to a new list will invalidate that reference. (ex: https://github.com/mfussenegger/nvim-lint/blob/4b03656c09c1561f89b6aa0665c15d292ba9499d/lua/lint/linters/checkstyle.lua#L3)

So I'm just keeping with the convention of modifying the original table, in case of other weird edge cases like that

dpetka2001:
Oh, I see what you mean and you're right of course.

dpetka2001:
Actually are you sure about that? Because `config` is a reference to a function won't it be copied by reference? I think it should not matter and still be valid when we reassign the default `args` to `prepend_args`. And `config` function itself is a closure over the `M` table itself.

I'm just asking for educational purposes this time, not to convince that you should change into what I proposed. Please do not misunderstand.

ivanisakov99:
No worries, I don't have an example handy atm (and maybe that config is a bad example to begin with) but essentially that particular issue doesn't directly stem from the manipulation of args alone.

Just before `prepend_args` are handled, the default and user configurations are merged together via:

```lua
lint.linters[name] = vim.tbl_deep_extend("force", lint.linters[name], linter)
```

This is what messes up the references, since it creates a new copy.

In the end, you are right that your way will keep the references intact, it just mutates more than the original table. So I don't mind changing it, if that's more desirable.

dpetka2001:
No, don't change it. I believe your function delivers more to the point what is supposed to be done. Let the maintainer decide when he'll become available.

I just asked to understand better, that's all. Thanks for your input.

ivanisakov99:
Sounds good, no worries!

folke:
ty!

PR #7113, opened 2026-04-14
feat(keymaps): improvements for gitsigns, refactoring, lsp

cub-uanic:
## Checklist

### DIFF
--- a/lua/lazyvim/plugins/editor.lua
+++ b/lua/lazyvim/plugins/editor.lua
@@ -72,7 +72,6 @@ return {
           { "<leader>f", group = "file/find" },
           { "<leader>g", group = "git" },
           { "<leader>gh", group = "hunks" },
-          { "<leader>gt", group = "toggle" },
           { "<leader>q", group = "quit/session" },
           { "<leader>s", group = "search" },
           { "<leader>u", group = "ui" },
@@ -184,11 +183,10 @@ return {
         map("n", "<leader>ghB", function() gs.blame() end, "Blame Buffer")
         map("n", "<leader>ghd", gs.diffthis, "Diff This")
         map("n", "<leader>ghD", function() gs.diffthis("~") end, "Diff This ~")
-        map("n", "<leader>ghq", gs.setqflist, "Show hunks as Quickfix")
-        map("n", "<leader>ghQ", function() gs.setqflist("all") end, "Show hunks as Quickfix (all)")
-        map("n", "<leader>gtb", gs.toggle_current_line_blame, "Toggle current line blame")
-        map("n", "<leader>gtw", gs.toggle_word_diff, "Toggle word diff")
         map({ "o", "x" }, "ih", ":<C-U>Gitsigns select_hunk<CR>", "GitSigns Select Hunk")
+
+        map("n", "<leader>xh", gs.setqflist, "Git hunks (file)")
+        map("n", "<leader>xH", function() gs.setqflist("all") end, "Git hunks (all)")
       end,
     },
   },
@@ -204,6 +202,38 @@ return {
           require("gitsigns").toggle_signs(state)
         end,
       }):map("<leader>uG")
+      Snacks.toggle({
+        name = "Git Line Blame",
+        get = function()
+          return require("gitsigns.config").config.current_line_blame
+        end,
+        set = function(state)
+          require("gitsigns").toggle_current_line_blame(state)
+        end,
+      }):map("<leader>uB")
+      Snacks.toggle({
+        name = "Git Word Diff",
+        get = function()
+          return require("gitsigns.config").config.word_diff
+        end,
+        set = function(state)
+          require("gitsigns").toggle_word_diff(state)
+        end,
+      }):map("<leader>uW")
+    end,
+  },
+  {
+    "neovim/nvim-lspconfig",
+    opts = function()
+      Snacks.toggle({
+        name = "LSP Codelens",
+        get = function()
+          return vim.lsp.codelens.is_enabled()
+        end,
+        set = function(state)
+          vim.lsp.codelens.enable(state)
+        end,
+      }):map("<leader>uH")
     end,
   },
 
--- a/lua/lazyvim/plugins/editor.lua
+++ b/lua/lazyvim/plugins/editor.lua
@@ -72,6 +72,7 @@ return {
           { "<leader>f", group = "file/find" },
           { "<leader>g", group = "git" },
           { "<leader>gh", group = "hunks" },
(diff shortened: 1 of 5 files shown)


dpetka2001:
If you follow the example [here](https://github.com/LazyVim/LazyVim/blob/83d90f339defdb109a6ede333865a66ffc7ef6aa/lua/lazyvim/plugins/editor.lua#L193-L201) then the toggle will provide visual information about whether it's enabled or not. 

All toggles in LazyVim live under `<leader>u`. I would suggest to just use `<leader>uB` and `<leader>uW` respectively. Another alternative would be to have a new group under `<leader>u` for Gitsigns toggles specifically (maybe `<leader>uG` that is already being used and then under it mappings like `<leader>uGb` or something similar).

Just a simple user myself though and these are just suggestions based on current LazyVim structure. Maintainer will do the final review that really counts.

cub-uanic:
@dpetka2001 reworked as you suggested, to `<leader>uB` and `<leader>uW`
Thanks for the idea!

lua/lazyvim/plugins/editor.lua
```diff
@@ -204,6 +202,38 @@ return {
...
+    opts = function()
+      Snacks.toggle({
+        name = "LSP Codelens",
+        get = function()
+          return vim.lsp.codelens.is_enabled()
+        end,
+        set = function(state)
+          vim.lsp.codelens.enable(state)
+        end,
+      }):map("<leader>uH")
```
dpetka2001: Better to put this block of code [here](https://github.com/LazyVim/LazyVim/blob/main/lua/lazyvim/plugins/lsp/init.lua#L202-L210). That way the mapping will only be available when a user has explicitly enabled `opts.codelens.enabled` and also when a server actually supports `textDocument/codeLens` capability. So, this block of code should be put inside `Snacks.util.lsp.on` function.

cub-uanic: I tried this approach, but it didn’t work well for me.

With my setup, `opts.codelens.enabled` is initially not set at all (and so it's `false`), which I believe is perfectly valid.
I can still run CodeLens manually with `<leader>cC`, whenever I want.
But in that case, the code inside `Snacks.util.lsp.on` never executes, even after CodeLens is started manually.

After trying several options, I decided to implement it the same way as the other toggles.
I’m not claiming this is the best solution, but at least it is consistent with the current code structure and conventions.
As a result, it now depends on folke/snacks.nvim#2831.

dpetka2001: `vim.lsp.codelens.is_enabled()` is only present in Neovim 0.12+, not previous versions. So, this Snacks toggle mapping will only work on Neovim 0.12+. On Neovim 0.11.x it will throw error about `is_enabled (a nil value)`. LazyVim is supposed to support version 0.11.2 onwards. 

Something like the following should support both
```diff
diff --git a/lua/lazyvim/plugins/lsp/init.lua b/lua/lazyvim/plugins/lsp/init.lua
index 490bec3c..68a6d8cd 100644
--- a/lua/lazyvim/plugins/lsp/init.lua
+++ b/lua/lazyvim/plugins/lsp/init.lua
@@ -199,14 +199,26 @@ return {
       end
 
       -- code lens
-      if opts.codelens.enabled and vim.lsp.codelens then
-        Snacks.util.lsp.on({ method = "textDocument/codeLens" }, function(buffer)
-          vim.lsp.codelens.refresh()
-          vim.api.nvim_create_autocmd({ "BufEnter", "CursorHold", "InsertLeave" }, {
-            buffer = buffer,
-            callback = vim.lsp.codelens.refresh,
-          })
-        end)
+      if vim.fn.has("nvim-0.12") == 0 then
+        if opts.codelens.enabled and vim.lsp.codelens then
+          Snacks.util.lsp.on({ method = "textDocument/codeLens" }, function(buffer)
+            vim.lsp.codelens.refresh()
+            vim.api.nvim_create_autocmd({ "BufEnter", "CursorHold", "InsertLeave" }, {
+              buffer = buffer,
+              callback = vim.lsp.codelens.refresh,
+            })
+          end)
+        end
+      else
+        if opts.codelens.enabled then
+           Snacks.notify.warn(
+            "`opts.codelens` will only take effect on Neovim <0.12.\nRemove option from your config and use the toggle mapping instead",
+            { title = "LSP Codelens" }
+          )
+        end
+        Snacks.util.lsp.on({ method = "textDocument/codeLens" }, function()
+          Snacks.toggle({
+            name = "LSP Codelens",
+            get = function()
+              return vim.lsp.codelens.is_enabled()
+            end,
+            set = function(state)
+              vim.lsp.codelens.enable(state)
+            end,
+          }):map("<leader>uH")
+        end)
       end
```
Even your PR at Snacks repo should take into consideration which Neovim versions it supports.

PS: I don't think maintainer will be welcoming dropping support for Neovim <0.12 given the `nvim-treesitter` backlash that went on, but I might be wrong.

cub-uanic: I'm not sure why you propose calling `vim.lsp.codelens.refresh()` only in one branch - but that's fine, as I'm not very familiar with the internals of NVim and LazyVim. However, I'm unclear why the key mapping is created only in one branch.

I'm also not sure that Vim version checking should be part of my PR, since I don't have enough experience with NVim and LazyVim to handle that properly. The main goal of my proposal was simply to add a few missing shortcuts that I would like to have in the LazyVim repository rather than in my custom configuration. I believe that version checks and other compatibility concerns should be addressed in a separate commit. My PR is "feat:(keymaps)" and intended to focus solely on adding key mappings.

dpetka2001: Because your keymap won't work in Neovim <0.12 and will throw an error about `is_enabled (a nil value)`. So, your PR at its current form will cause breakage for users that still use Neovim 0.11.7 for example.

Anyway I left the diff patch in my previous comment and hopefully maintainer will read it and is able to make changes to your PR when the time comes for him to review it. `vim.lsp.codelens.refresh()` in 0.12 is deprecated and it essentially just calls `vim.lsp.codelens.enable()` under the hood for the time being until it is hard deprecated in some later Neovim version. The whole `codelens` code has been reworked in Neovim 0.12+ and is entirely different than the code in Neovim 0.11.7 for example. That's why I only call `refresh` in versions <0.12 in my diff patch.

I'm not suggesting that you incorporate my suggestions. Since you don't feel comfortable doing it, that's perfectly acceptable. Let's just leave the final review up to the maintainer himself.

cub-uanic: Totally agreed, and thank you for the reviews and for your time! 🫡

PR #7139, opened 2026-04-25
fix(blink): change `build` for V2 main

dpetka2001:
## Description
V2 main branch has changed the required `build` command. Conform to that change.

## Related Issue(s)
None

## Checklist

### DIFF
--- a/lua/lazyvim/plugins/extras/coding/blink.lua
+++ b/lua/lazyvim/plugins/extras/coding/blink.lua
@@ -15,13 +15,18 @@ return {
   {
     "saghen/blink.cmp",
     version = not vim.g.lazyvim_blink_main and "*",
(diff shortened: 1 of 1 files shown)


jyuan0:
I'm not sure if it's necessary, but [blink.lib](https://github.com/saghen/blink.lib) is listed as a dependency in the [V2 install docs](https://main.cmp.saghen.dev/installation.html). Maybe it should be added as well?

```lua
    dependencies = {
      vim.g.lazyvim_blink_main and { "saghen/blink.lib" },
    },
 ```

Either way, thanks for the PR.

dpetka2001:
@jyuan0 Nice catch. I forgot about it because I had it installed in my personal configuration and I only made this PR mainly because of the warning blink gave due to the new `build` function.

Just added that as well. Tnx a lot :smile:

PR #7147, opened 2026-04-30
fix(haskell): prevent mason-lspconfig from starting hls

### LINKED ISSUE #7146: bug: Haskell extra starts duplicate HLS clients with mason-lspconfig automatic_enable

### Neovim version (nvim -v)

0.12.2

### Operating system/version

macOS 15.7.2

### Describe the bug

The Haskell extra can start duplicate HLS clients when `haskell-language-server` is installed by Mason.

The extra already prevents `nvim-lspconfig` from starting `hls`:

```lua
setup = {
  hls = function()
    return true
  end,
}
````

However, with recent `mason-lspconfig.nvim` / Neovim builtin LSP behavior, `hls` can still be automatically enabled through Mason / `vim.lsp.enable("hls")`.

This causes two clients to attach to the same Haskell buffer:

```text
haskell-tools.nvim
hls
```

and two HLS processes to be started:

```text
haskell-language-server-9.12.2 --lsp --logfile ...
haskell-language-server-9.12.2 --lsp
```

The fix that worked locally was to mark `hls` as not Mason-managed:

```lua
servers = {
  hls = {
    mason = false,
  },
}
```

### Steps To Reproduce

1. Use LazyVim with the Haskell extra enabled.
2. Open a Haskell file.

### Expected Behavior

The Haskell extra should only start one HLS client.

Since the extra uses `haskell-tools.nvim`, HLS should be managed by `haskell-tools.nvim` only. Neither `nvim-lspconfig` nor `mason-lspconfig` automatic enable should start a second `hls` client.

s7tya:
## Description

The Haskell extra uses `haskell-tools.nvim`, which manages its own HLS client. The extra already defines a `setup.hls` handler that returns `true`, but `hls` was not present in `opts.servers`, so LazyVim never applied that handler. With `mason-lspconfig` automatic enable, this can cause a second `hls` client to start alongside `haskell-tools.nvim`.

This PR adds `hls = {}` to `opts.servers` so the existing `setup.hls` handler is applied and `hls` is excluded from the normal LSP/Mason automatic-enable path.

Since `hls` is now registered as an LSP server, `haskell-language-server` no longer needs to be explicitly listed in `mason.nvim`’s `ensure_installed`.

This also adds `hlint` to Mason when `nvim-lint` is enabled, since the extra configures `hlint` for Haskell files.

## Related Issue(s)

- Fixes #7146

## Checklist

### DIFF
--- a/lua/lazyvim/plugins/extras/lang/haskell.lua
+++ b/lua/lazyvim/plugins/extras/lang/haskell.lua
@@ -51,11 +51,6 @@ return {
   },
 
   {
-    "mason-org/mason.nvim",
-    opts = { ensure_installed = { "haskell-language-server" } },
-  },
-
-  {
     "mfussenegger/nvim-dap",
     optional = true,
     dependencies = {
@@ -132,13 +127,22 @@ return {
         haskell = { "hlint" },
       },
     },
+    dependencies = {
+      {
+        "mason-org/mason.nvim",
+        opts = { ensure_installed = { "hlint" } },
+      },
+    },
   },
 
-  -- Make sure lspconfig doesn't start hls,
+  -- Make sure neither lspconfig nor mason-lspconfig starts hls,
   -- as it conflicts with haskell-tools
   {
     "neovim/nvim-lspconfig",
     opts = {
+      servers = {
+        hls = {},
+      },
       setup = {
         hls = function()
           return true


lua/lazyvim/plugins/extras/lang/haskell.lua
```diff
@@ -134,11 +134,16 @@ return {
...
-  -- Make sure lspconfig doesn't start hls,
+  -- Make sure neither lspconfig nor mason-lspconfig starts hls,
   -- as it conflicts with haskell-tools
   {
     "neovim/nvim-lspconfig",
     opts = {
+      servers = {
+        hls = {
+          mason = false,
+        },
```
dpetka2001:
```suggestion
        hls = {},
```
It doesn't need `mason = false`. It just needs the LSP server name to exist in `opts.servers`, so that `configure` function gets applied to the server.

s7tya: Thank you for the review. That worked, and I've updated the PR. So it turns out the issue wasn't caused by the mason-lspconfig v2 update, but was actually due to missing settings that should have been there all along?

dpetka2001: Yes, the setting should have been there. Because later in the code we do https://github.com/LazyVim/LazyVim/blob/7c1301b89563198d1b2a295e673f6a7205adc076/lua/lazyvim/plugins/lsp/init.lua#L270 This only checks the `configure` function against the keys of `opts.servers` (which are the LSP servers names). 

There was a major refactor of `mason-lspconfig` last year and LazyVim adjusted as well (so the new config is not the same as the old was before that change), but it seems there must have been some slippages like with the Haskell Extra.

s7tya: That makes sense. I appreciate the explanation.

dpetka2001: Please do note I'm just a simple user like you that just tries to help to the best of my knowledge. You'll have to wait for the maintainer for the actual review whenever he becomes available.

s7tya: No problem. I just started using LazyVim yesterday and finally got it working. I wasn't very confident about the PR I created, so this was a huge help.

lua/lazyvim/plugins/extras/lang/haskell.lua
dpetka2001:
```suggestion
    opts = { ensure_installed = { "hlint" } },
```
I believe this should be changed as well. `hlint` is not installed by Mason otherwise, but it's configured for `nvim-lint`. And we can also remove `haskell-language-server` as the server will automatically get installed by Mason by using `opts.servers.hls` (unless `opts.servers.hls.mason = false` explicitly).

s7tya: Hmm, while I agree with removing the unnecessary `haskell-language-server`, I don't see the need to add `hlint` in this PR.

dpetka2001: `nvim-lint` is enabled by default in LazyVim. `optional` only means that if the user decides to disable it, then it will not be enabled by the Extra as well. 

I agree that my previous recommendation is not correct. The correct one would be to do the same as here with `nvim-dap` https://github.com/LazyVim/LazyVim/blob/7c1301b89563198d1b2a295e673f6a7205adc076/lua/lazyvim/plugins/extras/lang/haskell.lua#L58-L67 So just add in `nvim-lint` spec a `mason.nvim` dependency that installs `hlint`.

So, just remove lines 53-56 and add the dependency for `mason.nvim` directly under `nvim-lint` spec.

s7tya: That looks better. It might be out of scope for this PR, but I've updated it accordingly.

s7tya:
I'm looking forward to your review. If there are any issues I should fix in this PR, please feel free to let me know.

PR #7151, opened 2026-05-03
feat(lsp): wildcard disable mason

alexandre-abrioux:
This PR adds the option to opt out all LSP servers from `mason-lspconfig` automatic installation. This is handy when the user wants to have total control over the `ensure_installed` list. To do so:

```lua
servers = {
  ["*"] = {
    mason = false,
  },
}
```

### DIFF
--- a/lua/lazyvim/plugins/lsp/init.lua
+++ b/lua/lazyvim/plugins/lsp/init.lua
@@ -254,7 +254,9 @@ return {
           return
         end
 
-        local use_mason = sopts.mason ~= false and vim.tbl_contains(mason_all, server)
+        local use_mason = sopts.mason ~= false
+          and opts.servers["*"].mason ~= false
+          and vim.tbl_contains(mason_all, server)
         local setup = opts.setup[server] or opts.setup["*"]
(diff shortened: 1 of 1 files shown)


lua/lazyvim/plugins/lsp/init.lua
dpetka2001, line 246:
```suggestion
        if server == "*" or opts.servers["*"].mason == false then
```
Better here to return early? Just a suggestion from a simple user.

alexandre-abrioux: Hi, this PR is about ensuring that Mason doesn't download the server binary, but we shouldn't disable the server itself. We still need to set up the server with `vim.lsp.config` and enable the server with `vim.lsp.enable`. It should have the same behavior as setting `mason = false` for all LSP servers.

dpetka2001: Oh sorry, you're right. Didn't think this through. My bad.

alexandre-abrioux: No worries, thanks for the review 🙏

PR #7156, opened 2026-05-10
feat(lsp): update configs for codelens

PFiS1737:
## Description

The legacy `vim.lsp.codelens.refresh` has been deprecated and will be removed in Neovim 0.13. 

This PR switches CodeLens using the new `vim.lsp.codelens.enable` interface. For configuration, it follows `inlay_hint` by introducing an `exclude` option to filter specific filetypes. 

This PR also adds a toggler for CodeLens.

## Checklist

### DIFF
--- a/lua/lazyvim/config/keymaps.lua
+++ b/lua/lazyvim/config/keymaps.lua
@@ -161,6 +161,24 @@ if vim.lsp.inlay_hint then
   Snacks.toggle.inlay_hints():map("<leader>uh")
 end
 
+if vim.lsp.codelens.enable then
+  if Snacks.toggle.codelens then
+    Snacks.toggle.codelens():map("<leader>cC")
+  else
+    -- TODO: remove this after codelens toggler is added to Snacks
+    Snacks.toggle.new({
+      id = "codelens",
+      name = "Codelens",
+      get = function()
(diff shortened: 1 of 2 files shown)


lua/lazyvim/config/keymaps.lua
```diff
@@ -161,6 +161,28 @@ if vim.lsp.inlay_hint then
...
+      end,
+      set = function(state)
+        vim.lsp.codelens.enable(state, { bufnr = 0 })
+      end,
+    }):map("<leader>cC")
+  end
+else
+  require("lazyvim.plugins.lsp.keymaps").set({}, {
+    { "<leader>cC", vim.lsp.codelens.refresh, desc = "Refresh & Display Codelens", mode = { "n" }, has = "codeLens" },
+  })
```
dpetka2001, line 183: You could change this to just 
```suggestion
Snacks.keymap.set("n", "<leader>cC", vim.lsp.codelens.refresh, { desc = "Refresh & Display Codelens", lsp = { method = "textDocument/codeLens" } })
```
That's what essentially `require("lazyvim.plugins.lsp.keymaps").set()` also does. 

But I'm just a simple user and this is only a suggestion. Don't feel obligated to change it if you don't want to. I just think it's cleaner this way, but it's just a personal preference in the end.

PR #7172, opened 2026-05-19
fix(go): use semanticTokens if setup in gopls config init_options

davidmdm:
## Description

Since upgrading to the go lsp (gopls) to version 0.22.0 we receive the following error on load of any Go file:

```
...nvim/lazy/LazyVim/lua/lazyvim/plugins/extras/lang/go.lua:64: attempt to index local 'semantic' (a nil value)
```

This PR adds a check to avoid this error.

If this causes degraded functionality, or a different fix around semantic tokens is needed, I do not know.

## Screenshots

image

## Checklist

### DIFF
--- a/lua/lazyvim/plugins/extras/lang/go.lua
+++ b/lua/lazyvim/plugins/extras/lang/go.lua
@@ -14,6 +14,9 @@ return {
     opts = {
       servers = {
         gopls = {
+          init_options = {
+            semanticTokens = true,
+          },
           settings = {
             gopls = {
               gofumpt = true,
@@ -46,7 +49,6 @@ return {
               completeUnimported = true,
               staticcheck = true,
               directoryFilters = { "-.git", "-.vscode", "-.idea", "-.vscode-test", "-node_modules" },
-              semanticTokens = true,
             },
           },
         },
@@ -56,7 +58,12 @@ return {
           -- workaround for gopls not supporting semanticTokensProvider
           -- https://github.com/golang/go/issues/54531#issuecomment-1464982242
           Snacks.util.lsp.on({ name = "gopls" }, function(_, client)
-            if not client.server_capabilities.semanticTokensProvider then
+            if
+              client.config
+              and client.config.init_options
+              and client.config.init_options.semanticTokens
+              and not client.server_capabilities.semanticTokensProvider
+            then
               local semantic = client.config.capabilities.textDocument.semanticTokens
               client.server_capabilities.semanticTokensProvider = {
                 full = true,


odaven:
Hello. 

Your solution was still giving the issue for me. I solved it by adding the nil guard for `textDocument` which is what the error is complaining about.

```lua
return {
  {
    "neovim/nvim-lspconfig",
    opts = {
      setup = {
        gopls = function(_, opts)
          Snacks.util.lsp.on({ name = "gopls" }, function(_, client)
            if not client.server_capabilities.semanticTokensProvider then
              local capabilities = client.config and client.config.capabilities
              local textDocument = capabilities and capabilities.textDocument
              local semantic = textDocument and textDocument.semanticTokens
              if semantic then
                client.server_capabilities.semanticTokensProvider = {
                  full = true,
                  legend = {
                    tokenTypes = semantic.tokenTypes,
                    tokenModifiers = semantic.tokenModifiers,
                  },
                  range = true,
                }
              end
            end
          end)
        end,
      },
    },
  },
}
```

davidmdm:
@odaven updated with your solution but defaulting to lsp.protocol capabilities if nil.

odaven:
This is with your solution.
```
vim.schedule callback: ....12.2/share/nvim/runtime/lua/vim/lsp/semantic_tokens.lua:437: ....12.2/share/nvim/runtime/lua/vim/lsp/semantic_tokens.lua:66: table index is nil
stack traceback:
	[builtin#36]: at 0x0101465f4c
	....12.2/share/nvim/runtime/lua/vim/lsp/semantic_tokens.lua:437: in function 'handler'
	.../neovim/0.12.2/share/nvim/runtime/lua/vim/lsp/client.lua:742: in function ''
	vim/_core/editor.lua: in function <vim/_core/editor.lua:0>
```



I removed 
```lua
or vim.lsp.protocol.make_client_capabilities().textDocument.semanticTokens
```
and the error disappeared.

The original comment says something about a workaround so I'm not sure if this is the final/best solution but the error disappear and looks the LSP is still working.

davidmdm:
Yeah, I am just taking a stab at it, but I'm unable to test, as I am not sure how best to load my local lazyvim instead of the remote git clone.

Could you perchance fill me in on how you do it?

davidmdm:
I mainly wanted to bring attention to this issue, happy if it's somebody else's contribution that fixes it. But would love to learn.

odaven:
In your file `lazy.lua`

```lua
 { "LazyVim/LazyVim", dev = true, dir = "~/Dev/LazyVim/", import = "lazyvim.plugins" },
```

When adding `dev = true`, it will use local project pointed by `dir`.

Opening Lazy you will be able to see the dir.

Screenshot 2026-05-19 at 23 35 44

dpetka2001:
Please track this issue at gopls repo https://github.com/golang/go/issues/79524. Seems like a regression to the server. If that's the case, then this PR should not be needed and users should downgrade gopls until issue is fixed upstream. Not sure if that's the case or not as I don't use gopls myself.

davidmdm:
@odaven thanks! I was able to test it out locally, and basically reverted to your exact solution. Looks great.

@dpetka2001 sounds good. There could be an argument to make that since this is technically a nillable property, the nil-check wouldn't hurt to have. But regardless, waiting for an upstream fix should be fine, too.

We can leave this PR open for others to use if they want, and close when the upstream is fixed. Or merge if you (or Folke) feel the nil-check stands on its own merits.

dpetka2001:
I'm just a simple user like you. Maintainer has final say and responsible for any merging. There's no harm at all about the nil check from my point of view. 

I linked the issue because as things stand you wouldn't have semantic tokens with latest gopls version either way. Try installing previous gopls version and check for yourself (from my brief understanding of the linked issue).

dpetka2001:
So, this PR only solves the error, but semantic tokens are still disabled. According to comments from the linked issue, `gopls` explicitly has semantic tokens disabled since version 0.22 and users now have to explicitly enable it in initialization options instead.

The following diff patch solves both issues
```diff
diff --git a/lua/lazyvim/plugins/extras/lang/go.lua b/lua/lazyvim/plugins/extras/lang/go.lua
index 7098a055..7b12977d 100644
--- a/lua/lazyvim/plugins/extras/lang/go.lua
+++ b/lua/lazyvim/plugins/extras/lang/go.lua
@@ -14,6 +14,9 @@ return {
     opts = {
       servers = {
         gopls = {
+          init_options = {
+            semanticTokens = true,
+          },
           settings = {
             gopls = {
               gofumpt = true,
@@ -46,7 +49,6 @@ return {
               completeUnimported = true,
               staticcheck = true,
               directoryFilters = { "-.git", "-.vscode", "-.idea", "-.vscode-test", "-node_modules" },
-              semanticTokens = true,
             },
           },
         },
@@ -56,16 +58,18 @@ return {
           -- workaround for gopls not supporting semanticTokensProvider
           -- https://github.com/golang/go/issues/54531#issuecomment-1464982242
           Snacks.util.lsp.on({ name = "gopls" }, function(_, client)
-            if not client.server_capabilities.semanticTokensProvider then
-              local semantic = client.config.capabilities.textDocument.semanticTokens
-              client.server_capabilities.semanticTokensProvider = {
-                full = true,
-                legend = {
-                  tokenTypes = semantic.tokenTypes,
-                  tokenModifiers = semantic.tokenModifiers,
-                },
-                range = true,
-              }
+            if client.config and client.config.init_options and client.config.init_options.semanticTokens then
+              if not client.server_capabilities.semanticTokensProvider then
+                local semantic = client.config.capabilities.textDocument.semanticTokens
+                client.server_capabilities.semanticTokensProvider = {
+                  full = true,
+                  legend = {
+                    tokenTypes = semantic.tokenTypes,
+                    tokenModifiers = semantic.tokenModifiers,
+                  },
+                  range = true,
+                }
+              end
             end
           end)
           -- end workaround
```
@davidmdm @odaven Please verify. If you go with your cursor on whatever field token and then press `<leader>ui` you will see that it doesn't mention `Semantic tokens`, only highlights from `Treesitter` and `Extmarks`. With the diff patch above, you will still get highlights from `Semantic tokens` and users can still disable it in their personal configuration if they wish to (`init_options.semanticTokens = false`).

davidmdm:
Thanks @dpetka2001 !

dpetka2001:
The nil check for `semantic`, although it doesn't hurt, is not necessary any more. `semantic` will always be `true` when the server is initialized with `semanticTokens = true`. And the nil check about `init_options.semanticTokens` already takes into account that the rest of the block will not be executed unless it's explicitly set to `true`.

davidmdm:
@dpetka2001 removed nil-check against `semantic`!

dpetka2001:
@davidmdm No need to report to me. As I mentioned before, I'm just a simple user like you. These are just personal suggestions. Whether you would like to incorporate them or not, is entirely up to you.

I also trust that you test these changes, because I don't do Go and can't test that thoroughly. I just did a basic test with some random AI generated Go file. I hope I did not miss and misunderstand anything from the linked issue. If I missed anything please feel free to point it out. I'm definitely not a know-it-all kind of person.

davidmdm:
I have rebased and validated that it works:
image

dpetka2001:
This will need to be reverted after release of gopls 0.22.1. See also `nvim-lspconfig#4438` and `nvim-lspconfig#4446`.

PR #7129, opened 2026-04-23
refactor(keymaps): remove hardcoded diagnostic jump float option

nowaylifer:
## Description

Currently, the diagnostic jump keymap overrides the default float behavior defined in `vim.diagnostic.config`. This change ensures the keybinding respects the user's diagnostic configuration.

## Related Issue(s)
None

## Checklist

- [X] I've read the [CONTRIBUTING](https://github.com/LazyVim/LazyVim/blob/main/CONTRIBUTING.md) guidelines.

PR #7228, opened 2026-07-31
fix(lsp): replace deprecated `float` option with `on_jump` in `vim.diagnostic.jump()`

vandalt:
## Description

The `float` option for `vim.diagnostic.jump()` is deprecated in Neovim 0.13 and will be removed in Neovim 0.14.

Here is the full message from `:checkhealth  vim.deprecated`

```text
- ⚠️ WARNING opts.float is deprecated. Feature will be removed in Nvim 0.14
  - ADVICE:
    - use opts.on_jump instead.
    - stack traceback:
        /usr/share/nvim/runtime/lua/vim/diagnostic/_jump.lua:277
        /home/vandal/repos/nvim-local/opt/LazyVim/lua/lazyvim/config/keymaps.lua:127
```

This PR fixes the deprecation by using the recommended `on_jump` instead. I based the implementation on what upstream currently does when users specify `float = true` (https://github.com/neovim/neovim/blob/02a25caa1b5211e349c41ac2e0642de905ae7be9/runtime/lua/vim/diagnostic/_jump.lua#L156).

The PR might have some conflicts with #7129 so I wanted to mention it just in case.

Thank you!

## Checklist


dpetka2001:
You could do the same as the other PR in the same position, but instead use the following
```lua
jump = vim.fn.has("nvim-0.13") == 1 and {
            on_jump = function(_, bufnr)
              vim.diagnostic.open_float({
                bufnr = bufnr,
                scope = "cursor",
                focus = false,
              })
            end,
          } or { float = true },
```
So, not in the `keymaps.lua` file in `diagnostic_go_to` function but instead in `lsp/init.lua` file under `diagnostics` table like the other PR does.

This way it will also support up to Neovim 0.11.2, which is the minimum version that LazyVim supports. 

Maybe you should make a suggestion to the other PR about it? I believe it has precedence over yours. But I'm just a simple user and not the one to make the final decision.

Also maybe the check for Neovim version could also be `vim.fn.has("nvim-0.12") == 1` since it still shows the message in `checkhealth`, it just doesn't show the deprecation notification pop up because of the version check (it only shows it in the version where it's hard deprecated just before the removal). So, only 0.11.x should use `float = true`.

vandalt:
Thanks @dpetka2001, I just rebased on #7129 (`nowaylifer/main`). This was the easiest way to test moving the change to config while preserving @nowaylifer's commit authorship.

I'm perfectly OK with #7129 merging changes from here instead or with @nowaylifer rewriting the changes themselves (I don't really mind showing up in the commit history or not, these changes are pretty small).

Cheers!

dpetka2001:
Like i said, i'm just a simple user like you. Not making any decisions. Maintainer will make final decision. I also left a comment in the other PR, so that the author would take notice, but no response so far.

PR #7233, opened 2026-08-12
fix(typescript): transition from `tsgo` to `tsc`

dpetka2001:
## Description
I also added `mason = false`, cuz `mason.nvim` does not provide a package yet. Users would have to install `typescript` manually with the package manager of their choice.

## Related Issue(s)
None

## Checklist


eriklindebratt:
Thanks for this @dpetka2001!

Given that https://github.com/neovim/nvim-lspconfig/pull/4493 is merged now it seems to me that latest nvim-lspconfig + LazyVim will break unless this PR is merged?

Any chance we can get it merged? 🙂

dpetka2001:
@eriklindebratt We have to wait for maintainer to review and merge it. No idea when he's available. I'm just a simple user like you. You can always just copy in your `~/.config/nvim/lua/plugins/lsp.lua` (or whatever file under `/lua/plugins/` directory) the following contents
```lua
{
  "neovim/nvim-lspconfig",
  opts = {
    servers = {
      tsc = { mason = false },
      vtsls = { enabled = false },
    },
  },
}
```
and disable the LazyVim Extra `lang.tsgo` until this gets merged. Then you should just install `typescript` whichever way you like and I believe it should work. Just try it out. Make sure to also uninstall `vtsls` from Mason UI because it doesn't get automatically uninstalled and it will get attached to the buffer otherwise.

The breakage with latest `nvim-lspconfig` is likely because they made the `tsgo` config redirect to the `tsc` one I'm assuming.

I would either copy the above code in my personal configuration and temporarily disable `lang.tsgo` Extra or pin back `nvim-lspconfig` to the commit before the `tsc` config introduction so that it doesn't cause any breakage.

PS: I'm seeing in the new `tsc.lua` of `nvim-lspconfig` that they also take into account for the `tsgo` binary. So, maybe it's because of some incompatibility due to the different Typescript version that users use in their projects?

lua/lazyvim/plugins/extras/lang/typescript/tsc.lua
```diff
@@ -20,7 +20,8 @@ return {
       -- make sure mason installs the server
       servers = {
         ---@type lspconfig.settings.tsgo
```
aitorciki: The type annotation should likely also be updated?

dpetka2001: `nvim-lspconfig` doesn't have type annotations for `tsc` yet. Only for `ts_ls`. Still I renamed it to what it will probably will be whenever they add support for it and also moved the annotation before `settings` where it should've been initially. Tnx for catching this :smile:

aitorciki:
> I also added `mason = false`, cuz `mason.nvim` does not provide a package yet.

This should be fixed after https://github.com/mason-org/mason-registry/pull/16462

dpetka2001:
@aitorciki No access to PC at the moment. Will remove `mason = false` when I get to it again in the next few days. I also saw that `nvim-lspconfig` changed `settings.typescript` into `settings["js/ts"]` in their latest commit. If someone could also confirm about which one should be the correct key under the `settings` field I'd appreciate it.

dpetka2001:
I found the following in [typescript-go repo](https://github.com/microsoft/typescript-go/blob/1bcfa18d79a3be41772223d5c05dfe4480e614ff/_extension/src/configurationMiddleware.ts#L6-L25). So, I guess I should also update to `settings["js/ts"]` next time I push again.

folke:
Got back this week from my vacation. I use ts myself, so will check fully next week. ty

pedroresende:
Hi, any news about this ?

folke:
Made some small additional changes and merged. Thank you!!

